Issue #315, opened 2022-03-30
Add explicit gen-p2p command


corverroos:
## Problem to be solved

Charon currently auto-creates a p2p key (and enr) if no p2p key is found. The UX of this isn't great, since charon will still error on startup when a p2p key was generated since the manifest cannot match that p2p key.

## Proposed solution

Add an explicit `gen-p2pkey` command that creates a p2p key and outputs the enr.
Do not auto generate p2pkey on startup, rather error.

PR #316, opened 2022-03-30
cmd: add gen-p2pkey command

corverroos:
Makes loading vs creating p2pkeys explicit both when running charon and as commands.

category: feature
ticket: #315

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
         tags: "${{ steps.get-version.outputs.version }}"
 
     - name: Generate cli reference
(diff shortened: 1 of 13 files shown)


cmd/enr.go
```diff
@@ -46,20 +46,14 @@ func newEnrCmd(runFunc func(io.Writer, p2p.Config, string) error) *cobra.Command
 	return cmd
 }
 
-// Function for printing status of ENR for this instance.
+// runNewENR loads the p2pkey from disk and prints the ENR for the provided config.
 func runNewENR(w io.Writer, config p2p.Config, dataDir string) error {
-	identityKey, loaded, err := p2p.LoadOrCreatePrivKey(dataDir)
```
xenowits, line 51: https://github.com/ObolNetwork/charon/issues/893 was fixed here

Issue #413, opened 2022-04-12
Add support for libp2p circuit relay


corverroos:
## Problem to be solved

Charon nodes need to have exposed ports in order to connect directly to each other via libp2p. This increases the barrier to entry to get a charon cluster up and running. This applies even more to dkg once it has been added to charon.

## Proposed solution

Add support for libp2p circuit relay ([1](https://docs.libp2p.io/concepts/circuit-relay/),[2](https://github.com/libp2p/go-libp2p-circuit)) via charon bootnodes.

- Add an optional libp2p host to the charon bootnode.
- Add an optional flag to charon for fallback p2p addresses via bootnode circuit relay.
- libp2p priorises direct connections over circuit relay, so it will only use circuit relay if if cannot connect directly to a peer.

PR #421, opened 2022-04-13
p2p: support circuit relay via bootnodes

corverroos:
Adds support for libp2p circuit relay via bootnodes. Allows clusters with non-publicly accessible charon nodes.

category: feature
ticket: #413

### DIFF
--- a/core/leadercast/transport.go
+++ b/core/leadercast/transport.go
@@ -124,7 +124,8 @@ func (t *p2pTransport) AwaitNext(ctx context.Context) (int, core.Duty, core.Unsi
 }
 
 func sendData(ctx context.Context, t *p2pTransport, p peer.ID, b []byte) error {
-	s, err := t.tcpNode.NewStream(ctx, p, protocol)
+	// Circuit relay connections are transient
+	s, err := t.tcpNode.NewStream(network.WithUseTransient(ctx, "leadercast"), p, protocol)
 	if err != nil {
 		return errors.Wrap(err, "tcpNode stream")
 	}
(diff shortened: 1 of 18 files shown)


p2p/p2p.go
```diff
@@ -116,18 +116,58 @@ func adaptDiscRouting(udpNode *discover.UDPv5, peers map[peer.ID]enode.Node) pee
...
+			if relay.Enode.TCP() == 0 {
+				continue
+			}
+
+			relayAddr, err := multiAddrViaRelay(relay, peerID)
+			if err != nil {
+				return peer.AddrInfo{}, err
+			}
+			mAddrs = append(mAddrs, relayAddr)
+		}
```
corverroos, line 141: here we basically tell libp2p that it should do circuit relay if the above resolved address isn't reachable.

p2p/p2p.go
```diff
@@ -67,17 +67,19 @@ func NewTCPNode(cfg Config, key *ecdsa.PrivateKey, connGater ConnGater,
 		// Limit connections to DV peers.
 		libp2p.ConnectionGater(connGater),
 
+		libp2p.AddrsFactory(factory),
```
corverroos, line 70: So turns out libp2p also does a form of peer discovery. We should turns this off since we use discv5 for peer discovery. See two factories above.

core/leadercast/transport.go
```diff
@@ -124,7 +124,8 @@ func (t *p2pTransport) AwaitNext(ctx context.Context) (int, core.Duty, core.Unsi
 }
 
 func sendData(ctx context.Context, t *p2pTransport, p peer.ID, b []byte) error {
-	s, err := t.tcpNode.NewStream(ctx, p, protocol)
+	// Circuit relay connections are transient
+	s, err := t.tcpNode.NewStream(network.WithUseTransient(ctx, "leadercast"), p, protocol)
```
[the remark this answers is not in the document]
dB2510: maybe add a TODO for more investigation

dB2510 (approved):
LGTM

PR #426, opened 2022-04-14
core: wire getduty in validatorapi

dB2510:
Wires GetDuty in validatorapi alongwith other stuff required for block proposer and improve architecture docs.

category: feature
ticket: none

### DIFF
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -28,7 +28,7 @@ type Scheduler interface {
 	Subscribe(func(context.Context, Duty, FetchArgSet) error)
 
 	// GetDuty returns the argSet for a duty if resolved already.
-	GetDuty(ctx context.Context, duty Duty) (FetchArgSet, error)
+	GetDuty(context.Context, Duty) (FetchArgSet, error)
 }
 
 // Fetcher fetches proposed unsigned duty data.
@@ -86,7 +86,7 @@ type ValidatorAPI interface {
 	RegisterPubKeyByAttestation(func(ctx context.Context, slot, commIdx, valCommIdx int64) (PubKey, error))
 
 	// RegisterGetDutyFunc registers a function to query duty data.
-	RegisterGetDutyFunc(func(ctx context.Context, duty Duty) (FetchArgSet, error))
+	RegisterGetDutyFunc(func(context.Context, Duty) (FetchArgSet, error))
 
 	// RegisterParSigDB registers a function to store partially signed data sets.
 	RegisterParSigDB(func(context.Context, Duty, ParSignedDataSet) error)
@@ -147,7 +147,7 @@ type Broadcaster interface {
 // instead as interfaces, since functions are easier to wrap than interfaces.
 type wireFuncs struct {
 	SchedulerSubscribe              func(func(context.Context, Duty, FetchArgSet) error)
-	SchedulerGetDuty                func(ctx context.Context, duty Duty) (FetchArgSet, error)
+	SchedulerGetDuty                func(context.Context, Duty) (FetchArgSet, error)
 	FetcherFetch                    func(context.Context, Duty, FetchArgSet) error
 	FetcherSubscribe                func(func(context.Context, Duty, UnsignedDataSet) error)
 	FetcherRegisterAggSigDB         func(func(context.Context, Duty, PubKey) (AggSignedData, error))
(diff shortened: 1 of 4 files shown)


core/interfaces.go
```diff
@@ -26,6 +26,9 @@ import (
 type Scheduler interface {
 	// Subscribe registers a callback for fetching a duty.
 	Subscribe(func(context.Context, Duty, FetchArgSet) error)
+
+	// GetDuty returns the argSet for a duty if resolved already.
+	GetDuty(ctx context.Context, duty Duty) (FetchArgSet, error)
```
corverroos, line 31: can remove parameter names in interface if it isn't confusing (which it isn't in this case), see `Subscribe` above

core/interfaces.go
```diff
@@ -82,6 +85,9 @@ type ValidatorAPI interface {
 	// RegisterPubKeyByAttestation registers a function to query validator by attestation.
 	RegisterPubKeyByAttestation(func(ctx context.Context, slot, commIdx, valCommIdx int64) (PubKey, error))
 
+	// RegisterGetDutyFunc registers a function to query duty data.
+	RegisterGetDutyFunc(func(ctx context.Context, duty Duty) (FetchArgSet, error))
```
corverroos, line 89: same here, remove param names. `RegisterPubKeyByAttestation` has names, since there are `3x int64` which is not clear.

core/interfaces.go
```diff
@@ -141,14 +147,19 @@ type Broadcaster interface {
 // instead as interfaces, since functions are easier to wrap than interfaces.
 type wireFuncs struct {
 	SchedulerSubscribe              func(func(context.Context, Duty, FetchArgSet) error)
+	SchedulerGetDuty                func(ctx context.Context, duty Duty) (FetchArgSet, error)
```
corverroos, line 150: same here and everywhere else

core/validatorapi/validatorapi.go
```diff
@@ -167,11 +167,9 @@ func (c *Component) RegisterPubKeyByAttestation(fn func(ctx context.Context, slo
 	c.pubKeyByAttFunc = fn
 }
 
-// RegisterAwaitProposer registers a function to query proposer PubKey by slot.
-// It supports a single function, since it is an input of the component.
-// TODO(corver): Best place to get this is probably Scheduler.
-func (c *Component) RegisterAwaitProposer(fn func(ctx context.Context, slot int64) (core.PubKey, error)) {
-	c.awaitProposerFunc = fn
+// RegisterGetDutyFunc registers a function to query duty data by duty.
+func (c *Component) RegisterGetDutyFunc(fn func(ctx context.Context, duty core.Duty) (core.FetchArgSet, error)) {
```
corverroos, line 171: should probably keep:
```
// It supports a single function, since it is an input of the component.
```

core/validatorapi/validatorapi.go
```diff
@@ -265,18 +263,24 @@ func (c Component) SubmitAttestations(ctx context.Context, attestations []*eth2p
 // BeaconBlockProposal submits the randao for aggregation and inclusion in DutyProposer and then queries the dutyDB for an unsigned beacon block.
 func (c Component) BeaconBlockProposal(ctx context.Context, slot eth2p0.Slot, randao eth2p0.BLSSignature, _ []byte) (*spec.VersionedBeaconBlock, error) {
 	// Get proposer pubkey (this is a blocking query).
-	pubKey, err := c.awaitProposerFunc(ctx, int64(slot))
+	dutySet, err := c.getDutyFunc(ctx, core.Duty{Slot: int64(slot), Type: core.DutyProposer})
 	if err != nil {
 		return nil, err
 	}
```
corverroos, line 269: maybe add:
```
else if len(dutySet) != 1 {
  return errors.New("unexpected amount of proposer duties")
}
```

core/validatorapi/validatorapi.go
```diff
@@ -309,11 +313,18 @@ func (c Component) SubmitBeaconBlock(ctx context.Context, block *spec.VersionedS
...
+	dutySet, err := c.getDutyFunc(ctx, core.Duty{Slot: int64(slot), Type: core.DutyProposer})
 	if err != nil {
 		return err
 	}
 
+	// There should be single duty proposer for the slot
+	var pubkey core.PubKey
+	for pk := range dutySet {
+		pubkey = pk
+	}
```
corverroos, line 326: could maybe extract this a reuable method (it is used in two places and quite verbose). Also, it isn't blocking anymore

```
pubkey, err := c.getProposerPubkey(ctx, slot)
```

corverroos (approved):
LGTM

core/validatorapi/validatorapi.go
dB2510: done

core/validatorapi/validatorapi.go
dB2510: done

core/validatorapi/validatorapi.go
dB2510: done

core/interfaces.go
dB2510: done

core/interfaces.go
dB2510: done

core/interfaces.go
dB2510: done

Issue #337, opened 2022-04-01
Re-enable complexity linters


corverroos:
## Problem to be solved

We disabled the complexity linters since we had lots of //nolint:cyclop,revive. This is probably not a good long term solution.  

## Proposed solution

Enable one of the two (suggest cyclop).
Fix the lint issues by splitting large functions into smaller ones.

corverroos:
It seems like we opt for longer functions of over bigger objects. There are no lints for large complex objects. Closing this for now.

PR #429, opened 2022-04-19
app/lifecycle: reduce the cyclomatic complexity of running hooks

leolara:
Reduce the cyclomatic complexity of running hooks in lifecylce component and reorganise some files.

category: refactor
ticket: #337


app/lifecycle/hook.go
```diff
@@ -0,0 +1,201 @@
...
+
+	stopAllHooks(stopCtx, stopHooks, cancel, cacheErr)
+
+	cacheErr(nil) // Ensure there is something in firstErr.
+
+	return <-firstErr
+}
+
+// startHook starts a hook, block until it returns.
+func startHook(ctx context.Context, hook hook, cancel context.CancelFunc, cacheErr func(err error)) {
```
corverroos, line 141: I think we can try to do normal function that returns an error, then you can do context and error handling in the parent function.

app/lifecycle/hook.go
```diff
@@ -0,0 +1,201 @@
...
+	err := hook.Func.Call(ctx)
+	if err != nil && !errors.Is(err, context.Canceled) {
+		cacheErr(errors.Wrap(err, "start hook", z.Str("hook", hook.Label)))
+		cancel()
+	}
+}
+
+// startAllHooks starts all hooks from hook collection.
+func startAllHooks(
```
corverroos, line 150: same here, rather stick to normal pattern of returning an error and do fancy context stuff and error in parent

app/lifecycle/hook.go
leolara: I don't understand, are you talking about cacheErr?

app/lifecycle/hook.go
leolara: so if you mean to do not call `cacheErr()` and `cancel()` and from inside here, and return an error when something happen and the caller exectute `cacheErr()` and `cancel()` that would change the behaviour, as an error in one of the hooks, does not stop the hooks loop.

Also AsyncBackground and AsyncAppCtx hooks would not be able to return the error

app/lifecycle/hook.go
corverroos: ok yeah, I see that is not possible. But the problem is that even though you broke up the big function into smaller functions, you haven't achieved any form of decoupling. The inner functions references all the state they referenced previously. 

So technically decreased the lines of code per function, but I think this is harder to read and understand.

I had a look and do not see an easy way to simplify and decouple the code into proper smaller decoupled functions.

app/lifecycle/hook.go
corverroos: yeah, passing two contexts and and a cancel function doesn't feel right. I think we should either do a proper refactor that allows for smaller decoupled functions, or just keep it the way it was.

leolara: we cannot keep it the way it was, the functions are so long that it doesn't even pass linting

app/lifecycle/hook.go
leolara: I don't that because there are no additional advantages to these beyond the splitting the functions into smaller functions we should not take this advantage. This code is incrementally better that what we have.

dB2510 (approved):
LGTM
I think we cannot refactor hook.go beyond this at this point. This looks fine to me.

dB2510:
Since it's just one file, we can leave this as we have bigger problems to solve. We can come back to this if it needs more refactor.
I don't think it is a tech debt at the moment.

PR #430, opened 2022-04-19
testutil/validatormock: add block proposal to vmock

dB2510:
Adds block proposal support to validatormock and a propose block unit test.

category: feature
ticket: none


core/scheduler/testdata/TestSchedulerDuties_grouped.golden
```diff
@@ -2,8 +2,6 @@
  {
   "Duty": "0/proposer",
   "DutyArgSet": {
-   "0x8dae41352b69f2b3a1c0b05330c1bf65f03730c520273028864b11fcb94d8ce8f26d64f979a0ee3025467f45fd2241ea": "{\"pubkey\":\"0x8dae41352b69f2b3a1c0b05330c1bf65f03730c520273028864b11fcb94d8ce8f26d64f979a0ee3025467f45fd2241ea\",\"slot\":\"0\",\"validator_index\":\"2\"}",
```
dB2510, line 5: this is removed because this meant there are multiple block proposer for a single slot

testutil/validatormock/validatormock_test.go
```diff
@@ -93,3 +93,33 @@ func TestAttest(t *testing.T) {
 		})
 	}
 }
+
+func TestProposeBlock(t *testing.T) {
```
dB2510, line 97: Added a basic unit test for proposeBlock

testutil/validatormock/validatormock.go
```diff
@@ -164,7 +188,12 @@ func ProposeBlock(ctx context.Context, eth2Cl Eth2Provider, signFunc SignFunc, s
 		}
 
 		// Get Unsigned beacon block with given randao and slot
-		block, err = eth2Cl.BeaconBlockProposal(ctx, slot, randao, nil)
+		if addr == "" {
+			// For testing purposes assuming beaconmock is used in place of validatorapi
+			block, err = eth2Cl.BeaconBlockProposal(ctx, slot, randao, nil)
```
corverroos, line 193: not sure this is required, why support eth2client, it will just error? I think just using the workaround is fine.

dB2510: in this case it is not eth2client, it is our beaconmock which implements ethclient's interface and will return our desired block. used this in validatormock_test

dB2510: here i am assuming that if we got empty `addr` parameter then we would assume that we are using beaconmock for testing and not validatorapi.

leolara: I find also difficult to understand what is happening here, it seems to assume which is the implementation of eth2Cl, if we pass an interface we cannot assume that it is an implemention or another

testutil/validatormock/validatormock.go
```diff
@@ -197,16 +226,19 @@ func ProposeBlock(ctx context.Context, eth2Cl Eth2Provider, signFunc SignFunc, s
...
 
 	// create signed beacon block
-	var signedBlock *spec.VersionedSignedBeaconBlock
+	signedBlock := new(spec.VersionedSignedBeaconBlock)
 	signedBlock.Version = block.Version
 	switch block.Version {
 	case spec.DataVersionPhase0:
+		signedBlock.Phase0 = new(eth2p0.SignedBeaconBlock)
 		signedBlock.Phase0.Message = block.Phase0
 		signedBlock.Phase0.Signature = sig
```
corverroos, line 235: suggest standard struct literals:
```
signedBlock.Phase0 := &eth2p0.SignedBeaconBlock{
  Message: block.Phase0,
  Signature: sig,
  }
  
```

testutil/validatormock/validatormock.go
```diff
@@ -257,3 +289,106 @@ func getSecret(secrets []*bls_sig.SecretKey, pubkey eth2p0.BLSPubKey) (*bls_sig.
...
+
+type altairBeaconBlockProposalJSON struct {
+	Data *altair.BeaconBlock `json:"data"`
+}
+
+type bellatrixBeaconBlockProposalJSON struct {
+	Data *bellatrix.BeaconBlock `json:"data"`
+}
+
+func beaconBlockProposal(_ context.Context, slot eth2p0.Slot, randaoReveal eth2p0.BLSSignature, graffiti []byte, addr string) (*spec.VersionedBeaconBlock, error) {
```
corverroos, line 310: add a comment why you had to do this

dB2510: done

testutil/validatormock/validatormock.go
dB2510: done

corverroos (approved):
LGTM

testutil/validatormock/validatormock.go
```diff
@@ -122,15 +132,29 @@ func Attest(ctx context.Context, eth2Cl Eth2Provider, signFunc SignFunc,
 }
 
 // ProposeBlock proposes block for the given slot.
-func ProposeBlock(ctx context.Context, eth2Cl Eth2Provider, signFunc SignFunc, slot eth2p0.Slot) error {
+
```
corverroos, line 135: remove empty line

PR #431, opened 2022-04-19
validatorapi: update validatorapi for proposer

dB2510:
Improves router and validatorapi for block proposer flow.

category: feature
ticket: none


core/validatorapi/router.go
```diff
@@ -367,7 +367,7 @@ func proposeBlock(p eth2client.BeaconBlockProposalProvider) handlerFunc {
 			}
 
 			return proposeBlockResponsePhase0{
-				Version: block.Version,
+				Version: "PHASE0",
```
dB2510, line 370: @corverroos i didn't added marshalling thing because we do marshalling of response as part of writeResponse.

Also didn't added any specific function to enforce this constant because anyways these fields will be exported so why having a seperate function for that when we can do that directly.

core/validatorapi/validatorapi.go
```diff
@@ -150,8 +153,22 @@ type Component struct {
 	parSigDBFuncs   []func(context.Context, core.Duty, core.ParSignedDataSet) error
 }
 
-func (*Component) ProposerDuties(context.Context, eth2p0.Epoch, []eth2p0.ValidatorIndex) ([]*eth2v1.ProposerDuty, error) {
-	return []*eth2v1.ProposerDuty{}, nil // No proposer duties for now.
+func (c *Component) ProposerDuties(ctx context.Context, epoch eth2p0.Epoch, validatorIndices []eth2p0.ValidatorIndex) ([]*eth2v1.ProposerDuty, error) {
```
corverroos, line 156: need to add a test for this

core/validatorapi/router.go
corverroos: fair enough

corverroos (approved):
Looks good, just add a test please

core/validatorapi/validatorapi.go
dB2510: done

PR #436, opened 2022-04-23
dkg: implement keycast

corverroos:
Implements a stub non-production key generation ceremony called keycast. The leader generates the keys and shares and broadcasts them to the participants. This allows us to wire and configure the dkg without an actual implementation. Unblocking multiple streams of works.

category: feature 
ticket: none

### DIFF
--- a/dkg/keycastjoin.go
+++ b/dkg/keycastjoin.go
@@ -33,8 +33,10 @@ import (
 // joinKeyCast returns the output for a keycast participant identified by the libp2p peer ID.
 //nolint:deadcode // Will be tested and wired in subsequent PRs.
 func joinKeyCast(ctx context.Context, tcpNode host.Host) (output, error) {
-	outCh := make(chan output, 1)
-	errCh := make(chan error, 1)
+	var (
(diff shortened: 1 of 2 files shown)


leolara (approved):
Perhaps we are putting too many things in the main directory. Perhaps we should think about the overall structure and where to put things that are not in core or app

corverroos:
So these files are under the `dkg` package, they are the only files so far.

dkg/keycastjoin.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return output{}, errors.Wrap(err, "verifier hex")
+		}
+
+		commitments = append(commitments, c)
+	}
+
+	secretShare := new(bls_sig.SecretKeyShare)
+	if err := secretShare.UnmarshalBinary(msg.Share); err != nil {
+		return output{}, errors.Wrap(err, "unmarshal pubkey")
```
xenowits, line 92:
```suggestion
		return output{}, errors.Wrap(err, "unmarshal secretkey")
```

PR #438, opened 2022-04-25
docs: spec remote signer architecture

corverroos:
Documents the remote signer architecture and how it fits into the current core workflow.

category: docs
ticket: none

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -52,8 +52,8 @@ Charon core business logic is modelled as a workflow, with a duty being performe
 ```
 Core Workflow
 
-      Phases │ Components
-─────────────┴───────────────────────────
+      Phases │ Components                              │ External
+─────────────┴─────────────────────────────────────────┴──────────────────────
 
              │                ┌─────────┐
   *Schedule* │          ┌─────►Scheduler│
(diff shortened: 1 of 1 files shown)


docs/architecture.md
```diff
@@ -391,6 +391,34 @@ type ValidatorAPI interface {
...
+> 🏗️ TODO: Figure out if signer should query DutyDB for slashing, or if DutyDB should push to signer.
+
+```go
+// Signer signs unsigned duty data sets via one or more remote signer instances.
+type Signer interface {
+    // Sign signs the unsigned duty data set.
+    Sign(context.Context, Duty, UnsignedDataSet) error
+
+    // RegisterParSigDB registers a function to store partially signed data sets.
+    RegisterParSigDB(func(context.Context, Duty, ParSignedDataSet) error))
```
leolara, line 418: Mandatory dependencies should be in the constructors. Not in setters. In most Go code, including Prysm we use interfaces instead of function types for dependencies

docs/architecture.md
```diff
@@ -66,15 +66,15 @@ Core Workflow
...
-             │  |       |     │DutyDB│
-             │  |       |     └──▲───┘
-                |       |        │           │
-      *Sign* │  |       |     ┌──┴─┐         │
-        duty │  |       └----─┤VAPI◄─────────│── VC
-        data │  |             └──┬─┘         │ Query, sign, submit
-                |                │           │
+        sign │  |       |        ├────────────┐
+             │  |       |     ┌──▼───┐     ┌──▼───┐    │
+             │  |       |     │DutyDB│     │Signer├────│─► RS
```
dB2510, line 71: will there be any other responsibilities of `Signer` component apart from interaction with RS?

docs/architecture.md
```diff
@@ -391,6 +391,34 @@ type ValidatorAPI interface {
...
+which authorises charon to request adhoc signatures from a [remote signer instance](https://lighthouse-book.sigmaprime.io/validator-web3signer.html).
+In the _middleware_ architecture charon cannot initiate signatures itself and has to
+wait for the VC to submit signatures.
+
+Duties originating in the `scheduler` (`DutyAttester`, `DutyProposer`) are not significantly affected by this change in architecture.
+Instead of waiting for the `validatorapi` to submit signatures, these duties directly request
+signatures from the remote signer instance. The flow is otherwise unaffected.
+
+Duties originating in the `validatorapi` (`DutyRandao`, `DutyAggregator`) has to refactored to
+originate in the `scheduler`, since charon is in full control of the duties in this architecture.
```
dB2510, line 405: how would that look like? (`DutyRandao` and `DutyAggregator` originating from scheduler rather than vapi)

dB2510 (approved):
LGTM

docs/architecture.md
corverroos: So this is the architecture we are following. We wire components via input and output functions.

docs/architecture.md
corverroos: I do not think so.

docs/architecture.md
corverroos: Following the same pattern as DutyAttester and DutyProposer, Scheduler will just trigger the duty, fetcher will do nothing, since those duties just sign slot and epochs, then consensus, then the rest of the flow. So nothing too complex.

Issue #439, opened 2022-04-25
Implement stub dkg


corverroos:
## Problem to be solved

We are starting on dkg implementation, but the algorithm to be used is still unknown. The exact interfaces and manifest files are also not 100% clear. We can however start integration of a stub dkg to test the overall structure and the how it will interact with the rest of charon code base.

## Proposed solution

- Create a `charon dkg lead` command that sets up networking and then calls `dkg.leadKeyCast` and writes the output files to disk.
- Create a `charon dkg join` command that sets up networking and then calls `dkg.joinKeyCast` and writes the output files to disk.

### Out of Scope

- Use stub dkg algorithm
- Use existing manifest format, and write new manifest 
- Use existing keystore files and password.
-

PR #440, opened 2022-04-25
cmd: wire dkg stub

corverroos:
Wires a stub dkg into the charon cli, starts lbp2p tcp nodes, and runs stub algorithms, and writes output to disk.

This is a starting point for DKG integration, something to build on, extend, and improve.

category: feature
ticket: #439

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/obolnetwork/charon/app"
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/dkg"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


dB2510:
suggest: keep keycast in seperate package like `dkg/keycast/join.go` or `dkg/keycast/lead.go

corverroos:
I'm avoiding pre-mature abstraction. The interfaces and dependencies of the dkg protocol vs transport vs command isn't clear yet. Also keycast will be removed. Single packages keep things simple as well.

dB2510:
Got it!
I didn't meant pre-mature abstraction, i was thinking along the lines of what we have in other packages for ex: for leadercast and parsigex we have p2p one and in-memory one. But yeah there we were clear with the interface part of it.

Also I liked the idea of extending keycast or building on top of it as per the algorithm. I more aligned with keeping one simple implementation with no fancy stuff like keycast to have a working flow with other charon components that would be used with dkg, alongwith other algorithms that we will use.

dB2510 (approved):
LGTM

xenowits (approved):
Looks good

Issue #381, opened 2022-04-07
Feature flag design proposal


corverroos:
## Problem to be solved

Once we have released versions of charon running in the wild (public testnets), we need to be much more careful and controlled about how we rollout new features. Some features might be high risk, and we might want phase then in gradually over time (e.g. simnet only, then devnet, then testnet, and then mainnet). 

Some considerations:
 - We do not really have classic SaaS "environments" like dev, staging and prod.
 - We also do not really have a single linear set of "networks"/"chains" like devnet, testnet, prod.
 - There are a myriad of different "networks"/"chains" that charon can be deployed to: multiple testnets, gnosis chain, etc.
 - Some users might want to run "beta" features on mainnet.
 - Features should have a default "state", but this must be configurable via flags/config.
 - Be careful with flags, since they cannot be removed easily, see [prysm](https://github.com/prysmaticlabs/prysm/blob/develop/.github/ISSUE_TEMPLATE/feature_flag.md). 

## Proposed solution

- Introduce a new package/concept called `rollout`. 
- Features are "rolled out" in three [phases](https://en.wikipedia.org/wiki/Software_release_life_cycle): `alpha`, `beta`, `gold`.
- Features in `alpha` are unstable and only to be used/tested by the core team.
- Features in `beta` are more stable can be used/tested by the community (opt in), but still lacks confidence for general availability.
- Features in `gold` are stable and generally available.
- Package `rollout` enumerates a set of `Feature` strings that are each hardcoded to a `Phase`.
- Charon `run` command has a flag `--rollout` which defaults to `gold` but can be set to `alpha` or `beta`.
- This flag controls which features are active.
- Features are enabled/disabled in the code by simple `if` statements: `if rollout.Enabled(rollout.FeatureFoo) {`
- Features are rolled out by updating their hardcoded phase.
- The feature and check should be removed and permanently integrated/enabled after a full release in gold.
- Individual features can be enabled/disabled via a unstructured flag `--rollout-enable=csv,list,of,features` / `--rollout-disable`
- This flag just warns if features do not exist (or have been removed).

corverroos:
Do we really need to split 100 lines of code into separate files? Feels a bit forced.

leolara:
In general in go we separate by component or subcomponent subject not by number of lines

PR #446, opened 2022-04-26
app/featureset: implement feature flags

corverroos:
Implement `featureset` package as per design proposal.

category:  feature 
ticket: #381


app/featureset/featureset.go
```diff
@@ -0,0 +1,57 @@
...
+	// statusSentinel is an internal tail-end placeholder.
+	statusSentinel // Must always be last
+)
+
+// Feature is a feature being rolled out.
+type Feature string
+
+const (
+	// QBFTConsensus introduces qbft consensus, see https://github.com/ObolNetwork/charon/issues/445.
+	QBFTConsensus Feature = "qbft_consensus"
```
dB2510, line 40: i think we can also have proposer feature here

app/featureset/featureset.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+var (
+	// state defines the current rollout status of each feature.
+	state = map[Feature]status{
+		QBFTConsensus: statusAlpha,
+		// Add all features and there status here.
```
dB2510, line 47: here too proposer feature with `statusAlpha`

app/featureset/config.go
```diff
@@ -0,0 +1,117 @@
...
+
+// DefaultConfig returns the default config enabling only stable features.
+func DefaultConfig() Config {
+	return Config{
+		MinStatus: statusStable.String(),
+	}
+}
+
+// Init initialises the global feature set state.
+func Init(ctx context.Context, config Config) error {
```
dB2510, line 52: so basically minStatus is very important and enabled and disabled are not that important it seems

cmd/cmd_internal_test.go
```diff
@@ -73,6 +74,11 @@ func TestCmdFlags(t *testing.T) {
 					Denylist:  "",
 					DBPath:    "",
 				},
+				Feature: featureset.Config{
```
dB2510, line 77: nit: add cmd test including enabled and disabled features too so that reader of this code would get more clarity with the use of these flags.
Because from the name of these flags it seems that they are boolean flags (like enabled=true/false).

dB2510 (approved):
LGTM

app/featureset/featureset.go
corverroos: Can add that in next PR

app/featureset/config.go
corverroos: yup, `minStatus` enables/disables different sets of features. While `Enable/Disable` is to override individual features

cmd/cmd_internal_test.go
corverroos: I'll see what I can do

PR #451, opened 2022-04-27
dutydb: add shutdown and clean api

corverroos:
Add Shutdown to DutyDB so ValidatorAPI doesn't timeout on shutdown due to blocked DutyDB.Await* requests.

Also remove PubKey from AwaitBlockProposer as per previous TODO. 

Also reorder stop hooks in order of dependency tree.

category: refactor 
ticket: none
feature_set: stable

### DIFF
--- a/core/dutydb/memory.go
+++ b/core/dutydb/memory.go
@@ -18,7 +18,6 @@ package dutydb
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"sync"
 
 	"github.com/attestantio/go-eth2-client/spec"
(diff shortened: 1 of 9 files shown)


core/dutydb/memory.go
```diff
@@ -56,8 +63,11 @@ func (db *MemDB) Store(_ context.Context, duty core.Duty, unsignedSet core.Unsig
 
 	switch duty.Type {
 	case core.DutyProposer:
-		for pubkey, unsignedData := range unsignedSet {
-			err := db.storeBeaconBlockUnsafe(pubkey, unsignedData)
+		if len(unsignedSet) > 1 {
```
dB2510, line 66: add a comment on what does this if condition mean

dB2510 (approved):
LGTM

core/dutydb/memory.go
corverroos:
```suggestion
		// Sanity check max one proposer per slot 
		if len(unsignedSet) > 1 {
```

PR #453, opened 2022-04-28
cluster: add definition and lock config types

corverroos:
Adds the `cluster` package and the two config types: `Definition` and `Lock`. Also do basic json serialisation and hash based validation. Also implement EIP712 signatures and verification.

category: feature
ticket: none
feature_set: alpha


cluster/cluster.go
```diff
@@ -0,0 +1,65 @@
...
+	"io"
+)
+
+const (
+	paramsVersion = "v1.0.0"
+	dkgAlgo       = "default"
+)
+
+// NewParams returns a new params with populated version and UUID.
+func NewParams(name string, numVals int, threshold int,
```
leolara, line 39: @corverroos parameters to a function should be either all in one line or each in its own line. Doesn't linter detect this?

cluster/types.go
```diff
@@ -0,0 +1,93 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cluster
+
+// Params defines an intended charon cluster configuration.
+type Params struct {
```
leolara, line 19: @corverroos the public fields make this structure mutable, unfortunatelly, making it less functional. But I understand that to do the json encoding would require a bit of boilerplate code if we made them private, so perhaps it is acceptable?

cluster/types.go
```diff
@@ -0,0 +1,93 @@
+// Copyright © 2022 Obol Labs Inc.
```
leolara, line 1: @corverroos it is cleaner to have a params.go operator.go lock.go  and distvalidator.go and then the methods of each in that file

cluster/encode.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"strings"
+
+	"github.com/obolnetwork/charon/app/errors"
+)
+
+func (p Params) MarshalJSON() ([]byte, error) {
+	type fieldsOnly Params // Marshal fields-only version of params
```
leolara, line 27: @corverroos what does this renaming of Params?

cluster/types.go
leolara: considering we are already writing a MarshallJSON and UnmarshallJSOn for Params, there is no need for these fields to be public?

leolara (changes requested):

cluster/types.go
corverroos: So this is a DTO, ie. a pure value, fields should be public.

cluster/types.go
corverroos: I chose the option to colocate logic, almost like how eth2client does things.

cluster/encode.go
corverroos: I've refactored this to use json formatter type, same as `Lock`

cluster/cluster.go
corverroos: sure, will make it flat.

leolara:
I don't think we should add everything as a new package in the root. I am not sure about this one, but perhaps we need to think about a package hierarchy and where to put certain packages that do not fit on app or core.

corverroos:
It is just 4 types, I do not think creating more packages is required.

cluster/encode.go
leolara: ok, I didn't understood this version

cluster/types.go
leolara: in eth2client they subsecuently split things of a component into subjects (as they do in prysm) for the same component, but I don't see they do exactly what you do here of putting every component together and then put the same subject of all the components in the same file.

leolara:
I am not talking about creating more packages, folders can be empty, my argument is about not putting all new packages in the root.

cluster/types.go
corverroos: ok, grouping functions by type.

cluster/types.go
corverroos: ok, grouping files by type. So now a single file for each type

corverroos:
`cluster` package is in root because it is a top level thing. The cluster configuration. It will be used by `dkg` and `app` packages. So I do not think it should be in `app/cluster`, neither `cmd/cluster` or `dkg/cluster`.

cluster/types.go
leolara: Thanks

leolara:
Approved

PR #460, opened 2022-04-29
github/workflows: add GITHUB_TOKEN to build

dB2510:
Replaces CONTAINER_REGISTRY_SECRET with GITHUB_TOKEN. It seems that @OisinKyne has configured his Personal Access Token for CONTAINER_REGISTRY_SECRET which is not a good practice for security.

category: fixbuild
ticket: none


corverroos (approved):
LGTM

leolara (approved):

Issue #427, opened 2022-04-14
Error: mkdir ./charon: not a directory


OisinKyne:
## Problem to be solved
I'm not sure how you guys run commands on charon, but what I do is:
```sh
go build .
./charon help
```

So when I choose:
```sh
./charon gen-p2pkey
```

I get:
```sh
./charon gen-p2pkey 
Error: mkdir: mkdir ./charon: not a directory
Usage:
  charon gen-p2pkey [flags]
```

## Proposed solution

Change `./charon` as the default folder to `./.charon/`.

PR #467, opened 2022-04-30
app: fix error mkdir ./charon: not a directory

xenowits:
When running commands like `./charon gen-p2pkey` or `./charon create-cluster`, users including me got an error:
`Error: mkdir ./charon: not a directory`

This could happen as the default directory name starts with the prefix `./charon`, for example, `./charon/data` for `create-cluster`. Removing the prefix solves this. Directories are always created off of `charon` root.

category: bug
ticket: #427

### DIFF
--- a/cmd/cmd_internal_test.go
+++ b/cmd/cmd_internal_test.go
@@ -79,7 +79,7 @@ func TestCmdFlags(t *testing.T) {
 					Enabled:   nil,
 					Disabled:  nil,
 				},
-				ManifestFile:     "./.charon/manifest.json",
+				ManifestFile:     ".charon/manifest.json",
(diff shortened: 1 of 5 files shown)


cmd/cmd_internal_test.go
```diff
@@ -79,7 +79,7 @@ func TestCmdFlags(t *testing.T) {
 					Enabled:   nil,
 					Disabled:  nil,
 				},
-				ManifestFile:     "./charon/manifest.json",
+				ManifestFile:     "./.charon/manifest.json",
```
corverroos, line 82: can remove the `./` prefix, so just `.charon/...`

corverroos (approved):
LGTM

PR #468, opened 2022-05-02
app/log: align warn and error APIs

corverroos:
Aligns `log.Warn` and `log.Error` APIs. Since warn and error are mostly related to logging errors, just at different level, the API should be the same. Using `z.Err(err)` also resulted in inconsistent logging, which is now mitigated by aligned APIs. Also support `nil` errors for both `log.Warn` and `log.Error` that just does logging.
 
category: refactor
ticket: none
feature_set: stable

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -496,7 +496,7 @@ func wireValidatorMock(conf Config, pubshares []eth2p0.BLSPubKey, sched core.Sch
 				eth2http.WithTimeout(time.Second*10), // Allow sufficient time to block while fetching duties.
 			)
 			if err != nil {
-				log.Warn(ctx, "Cannot connect to validatorapi", z.Err(err))
+				log.Warn(ctx, "Cannot connect to validatorapi", err)
 				return
 			}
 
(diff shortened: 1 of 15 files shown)


app/app.go
```diff
@@ -496,7 +496,7 @@ func wireValidatorMock(conf Config, pubshares []eth2p0.BLSPubKey, sched core.Sch
 				eth2http.WithTimeout(time.Second*10), // Allow sufficient time to block while fetching duties.
 			)
 			if err != nil {
-				log.Warn(ctx, "Cannot connect to validatorapi", z.Err(err))
+				log.Warn(ctx, "Cannot connect to validatorapi", err)
 				return
```
corverroos, line 500: most `log.Warn` had `z.Err(err)`, which caused inconsistent logging and resulted in inconsistent API

app/app.go
```diff
@@ -515,18 +515,18 @@ func callValidatorMock(ctx context.Context, duty core.Duty, cl eth2client.Servic
...
 		err := validatormock.ProposeBlock(ctx, cl.(*eth2http.Service), signer, eth2p0.Slot(duty.Slot), addr, pubshares...)
 		if err != nil {
-			log.Warn(ctx, "Failed to propose block", z.Err(err))
+			log.Warn(ctx, "Failed to propose block", err)
 		} else {
 			log.Info(ctx, "Block proposed successfully", z.I64("slot", duty.Slot))
 		}
 	default:
-		log.Warn(ctx, "Invalid duty type")
+		log.Warn(ctx, "Invalid duty type", nil)
```
corverroos, line 530: `nil` error is supported, resulting in normal `warn` logs

dB2510 (approved):
LGTM

PR #469, opened 2022-05-02
dkg: add blackbox test

corverroos:
Adds a blackbox dkg test (that should work for any DKG algorithm).
Also adds `cluster.NewForT` which is a port of `app.NewClusterForT` using new cluster types.
This code isn't integrated yet.

category: test
ticket: none
feature_set: alpha

### DIFF
--- a/cluster/test_cluster.go
+++ b/cluster/test_cluster.go
@@ -13,11 +13,12 @@
 // You should have received a copy of the GNU General Public License along with
 // this program.  If not, see <http://www.gnu.org/licenses/>.
 
-package app
+package cluster
 
 import (
 	"crypto/ecdsa"
 	crand "crypto/rand"
+	"fmt"
 	"io"
 	"math/rand"
 	"net"
@@ -30,22 +31,22 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/obolnetwork/charon/p2p"
-	crypto2 "github.com/obolnetwork/charon/tbls"
+	"github.com/obolnetwork/charon/tbls"
 	"github.com/obolnetwork/charon/testutil"
 )
 
-// NewClusterForT returns a new cluster manifest with dv number of distributed validators, k threshold and n peers.
+// NewForT returns a new cluster lock with dv number of distributed validators, k threshold and n peers.
 // It also returns the peer p2p keys and BLS secret shares. If the seed is zero a random cluster on available loopback
 // ports is generated, else a deterministic cluster is generated.
-// Note this is not defined in testutil since it is tightly coupled with the app package.
-func NewClusterForT(t *testing.T, dv, k, n, seed int) (Manifest, []*ecdsa.PrivateKey, [][]*bls_sig.SecretKeyShare) {
+// Note this is not defined in testutil since it is tightly coupled with the cluster package.
+func NewForT(t *testing.T, dv, k, n, seed int) (Lock, []*ecdsa.PrivateKey, [][]*bls_sig.SecretKeyShare) {
 	t.Helper()
 
 	var (
-		dvs      []crypto2.TSS
-		dvShares [][]*bls_sig.SecretKeyShare
+		vals     []DistValidator
 		p2pKeys  []*ecdsa.PrivateKey
-		peers    []p2p.Peer
+		ops      []Operator
+		dvShares [][]*bls_sig.SecretKeyShare
 	)
 
 	addrFunc := getAddrFunc(seed)
@@ -53,13 +54,26 @@ func NewClusterForT(t *testing.T, dv, k, n, seed int) (Manifest, []*ecdsa.Privat
 	random := io.Reader(rand.New(rand.NewSource(int64(seed)))) //nolint:gosec // Explicit use of weak random generator for determinism.
 	if seed == 0 {
 		random = crand.Reader
+	} else {
+		rand.Seed(int64(seed))
 	}
 
 	for i := 0; i < dv; i++ {
-		tss, shares, err := crypto2.GenerateTSS(k, n, random)
+		tss, shares, err := tbls.GenerateTSS(k, n, random)
 		require.NoError(t, err)
 
-		dvs = append(dvs, tss)
+		pk, err := tss.PublicKey().MarshalBinary()
+		require.NoError(t, err)
+
+		var verifiers [][]byte
+		for _, commitment := range tss.Verifier().Commitments {
+			verifiers = append(verifiers, commitment.ToAffineCompressed())
+		}
+
+		vals = append(vals, DistValidator{
+			PubKey:    fmt.Sprintf("%#x", pk),
+			Verifiers: verifiers,
+		})
 		dvShares = append(dvShares, shares)
 	}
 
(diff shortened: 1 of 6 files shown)


cluster/test_cluster.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/obolnetwork/charon/p2p"
+	"github.com/obolnetwork/charon/tbls"
+	"github.com/obolnetwork/charon/testutil"
+)
+
+// NewForT returns a new cluster lock with dv number of distributed validators, k threshold and n peers.
+// It also returns the peer p2p keys and BLS secret shares. If the seed is zero a random cluster on available loopback
+// ports is generated, else a deterministic cluster is generated.
+// Note this is not defined in testutil since it is tightly coupled with the cluster package.
+func NewForT(t *testing.T, dv, k, n, seed int) (Lock, []*ecdsa.PrivateKey, [][]*bls_sig.SecretKeyShare) {
```
corverroos, line 42: mostly copied from `app.NewClusterForT`

dkg/dkg.go
```diff
@@ -124,7 +124,12 @@ func setupP2P(datadir string, p2pConf p2p.Config, peers []p2p.Peer) (host.Host,
 		return nil, nil, errors.Wrap(err, "failed to open enode")
 	}
 
-	udpNode, err := p2p.NewUDPNode(p2pConf, localEnode, key, nil)
+	bootnodes, err := p2p.NewUDPBootnodes(ctx, p2pConf, peers, localEnode.ID())
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "new bootnodes")
+	}
+
```
corverroos, line 131: missing bootnode support

dkg/keycast_internal_test.go
```diff
@@ -60,12 +60,14 @@ func TestKeyCastNoNetwork(t *testing.T) {
 	for i := 0; i < nodes; i++ {
 		i := i // Copy loop variable.
 		eg.Go(func() error {
-			defer cancel()
```
corverroos, line 63: fixing flapping test. Only cancel on error, otherwise let all DKG complete by themselves.

cluster/test_cluster.go
dB2510, line 42: shouldn't we name it something like NewCLockForT or something specific?

cluster/test_cluster.go
```diff
@@ -0,0 +1,136 @@
...
+
+		ops = append(ops, Operator{
+			Address: crypto.PubkeyToAddress(p2pKey.PublicKey).String(),
+			ENR:     enrStr,
+		})
+
+		p2pKeys = append(p2pKeys, p2pKey)
+	}
+
+	// TODO(corver): Add signatures.
```
dB2510, line 108: to test secret shares?

dB2510 (approved):
LGTM

dkg/dkg_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// Wait until complete
+
+	runChan := make(chan error, 1)
+	go func() {
+		runChan <- eg.Wait()
+	}()
+
+	select {
```
leolara, line 98: @corverroos if I understand correctly, there will be an  error sent via errChan and as many as `nodes` sent via runChan. Why then we wait for only one of them?

dkg/dkg_test.go
```diff
@@ -0,0 +1,199 @@
...
+		testutil.SkipIfBindErr(t, err)
+		require.NoError(t, err)
+	case err := <-runChan:
+		cancel()
+		testutil.SkipIfBindErr(t, err)
+		require.NoError(t, err)
+	}
+
+	// Read generated lock and keystores from disk
+	var (
```
leolara, line 110: @corverroos I dont see using var or const with brackets often in Go.

cluster/test_cluster.go
leolara: For me the problem is not not saying Cluster, as if this is the only thing this package contains we will use it as cluster.New. Foir me the problem is using T for Test, as it is not a common abbreviation

dkg/dkg_test.go
```diff
@@ -0,0 +1,199 @@
...
+		require.NoError(t, os.MkdirAll(conf.DataDir, 0o755))
+		err := crypto.SaveECDSA(p2p.KeyPath(conf.DataDir), keys[i])
+		require.NoError(t, err)
+
+		eg.Go(func() error {
+			return dkg.Run(ctx, conf)
+		})
+		if i == 0 {
+			// Allow node0 some time to startup, this just mitigates startup races and backoffs but isn't required.
+			time.Sleep(time.Millisecond * 100)
```
leolara, line 87: @corverroos is there a way to check that a node has started? If not I think we should add it as part of another task

dkg/dkg_test.go
dB2510: It is recommended to group variables declared in multiple lines for better readability

leolara: Where is it recommended? I have seen the opposite in Go codebases:

https://github.com/prysmaticlabs/prysm/blob/ff40a682150cbe12f044e8d636d682b68fe6bffb/beacon-chain/state/stategen/errors.go#L5-L7

https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go#L112-L113

cluster/test_cluster.go
corverroos: it might not be common, but it is clear, `NewForT(t testing.T, ...)` is very obvious. Don't be scared to innovate 😉

cluster/test_cluster.go
corverroos: So the cluster definition contains signatures that "validate/authorise" it. We should add those to the test cluster definition at some point.

dkg/dkg_test.go
corverroos: No `runChan` and `errChan` only returns a single error. In general `runChan` will return something and then we shutdown the bootnode.

cluster/test_cluster.go
dB2510: cool

dkg/dkg_test.go
corverroos: There is no way at this point to do that. So this is best effort that allows for fast test to mitigate startup race and backoff.

dkg/dkg_test.go
dB2510: You can find it in following:
* https://github.com/ethereum/go-ethereum/blob/master/core/beacon/errors.go
* https://github.com/prysmaticlabs/prysm/blob/ff40a682150cbe12f044e8d636d682b68fe6bffb/cmd/beacon-chain/flags/base.go#L10
* https://github.com/prysmaticlabs/prysm/blob/9615484fe9a0312d76ac68a4a4086e64f0bdfb71/io/logs/stream.go#L17
* https://github.com/prysmaticlabs/prysm/blob/9615484fe9a0312d76ac68a4a4086e64f0bdfb71/async/event/feed_test.go#L120

dkg/dkg_test.go
leolara: I didn't mean to do it now

PR #475, opened 2022-05-03
dkg: implement frost dkg

corverroos:
Implements a parallel Frost DKG using kryptology library and following example in `github.com/coinbase/kryptology/test/frost_dkg/bls/main.go`.

category: feature 
ticket: none
feature_set: alpha

### DIFF
--- a/dkg/frost.go
+++ b/dkg/frost.go
@@ -195,6 +195,7 @@ func round2(
 
 // makeShares returns a slice of shares (one for each validator) from the DKG participants and round 2 results.
 func makeShares(validators map[uint32]*frost.DkgParticipant, r2Result map[msgKey]frost.Round2Bcast) ([]share, error) {
+	// Get our ID from any validator (they all have our ID). 
 	targetID := validators[0].Id
 
 	// Get set of public shares for each validator.
--- a/dkg/frost.go
+++ b/dkg/frost.go
@@ -17,9 +17,6 @@ package dkg
 
 import (
 	"context"
-	"encoding/binary"
-	"io"
-	"strconv"
 
 	"github.com/coinbase/kryptology/pkg/core/curves"
 	"github.com/coinbase/kryptology/pkg/dkg/frost"
@@ -58,8 +55,8 @@ type fTransport interface {
 // runFrostParallel runs numValidators Frost DKG processes in parallel (sharing transport rounds)
 // and returns a list of shares (one for each distributed validator).
 //nolint:deadcode // Will be tested and integrated in subsequent PRs.
-func runFrostParallel(ctx context.Context, tp fTransport, numValidators, numNodes, threshold, shareIdx uint32, random io.Reader) ([]share, error) {
-	validators, err := newFrostParticipants(numValidators, numNodes, threshold, shareIdx, random)
+func runFrostParallel(ctx context.Context, tp fTransport, numValidators, numNodes, threshold, shareIdx uint32, dgkCtx string) ([]share, error) {
+	validators, err := newFrostParticipants(numValidators, numNodes, threshold, shareIdx, dgkCtx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


dkg/frost.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/coinbase/kryptology/pkg/sharing"
+	"github.com/coinbase/kryptology/pkg/signatures/bls/bls_sig"
+
+	"github.com/obolnetwork/charon/app/errors"
+)
+
+// msgKey identifies the source and target nodes and validator index the message belongs to.
+type msgKey struct {
+	// ValIdx identifies the distributed validator (Ith parallel participant) the message belongs to.
+	ValIdx uint32
```
leolara, line 32: @corverroos if `msgKey` is only used on this package perhaps the fields should be private

dkg/frost.go
```diff
@@ -0,0 +1,275 @@
...
+	// It is 1-indexed and equivalent to `cluster.NodeIdx.ShareIdx`.
+	// The zero value indicates outgoing broadcast messages.
+	TargetID uint32
+}
+
+// fTransport abstracts the transport of frost DKG messages.
+type fTransport interface {
+	// Round1 returns results of all round 1 communication; the received round 1 broadcasts from all other nodes
+	// and the round 1 P2P sends to this node.
+	Round1(context.Context, map[msgKey]frost.Round1Bcast, map[msgKey]sharing.ShamirShare) (
```
leolara, line 48: @corverroos is this an interface that another package is going to implement and inject? or all the implementations are in this package?

Either:
 + Round1, Round2 should be private
 + msgKey should be public

dkg/frost.go
```diff
@@ -0,0 +1,275 @@
...
+	castR2Result, err := tp.Round2(ctx, castR2)
+	if err != nil {
+		return nil, errors.Wrap(err, "transport round 2")
+	}
+
+	return makeShares(validators, castR2Result)
+}
+
+// newFrostParticipant returns multiple frost dkg participants (one for each parallel validator).
+func newFrostParticipants(numValidators, numNodes, threshold, shareIdx uint32, dgkCtx string) (map[uint32]*frost.DkgParticipant, error) {
```
leolara, line 88: @corverroos in my opinion it is more readable if all the parameters have the type explicit, but I have seen both styles

dkg/frost.go
```diff
@@ -0,0 +1,275 @@
...
+	// The zero value indicates outgoing broadcast messages.
+	TargetID uint32
+}
+
+// fTransport abstracts the transport of frost DKG messages.
+type fTransport interface {
+	// Round1 returns results of all round 1 communication; the received round 1 broadcasts from all other nodes
+	// and the round 1 P2P sends to this node.
+	Round1(context.Context, map[msgKey]frost.Round1Bcast, map[msgKey]sharing.ShamirShare) (
+		map[msgKey]frost.Round1Bcast, map[msgKey]sharing.ShamirShare, error)
```
leolara, line 49: @corverroos I have not seen the new line done in this place before, if the opening bracket is in a different line I guess the closign one should be also in a different one?

leolara: I am not sure what is the best option for this? Perhaps having each input paramater in a different line? like in:

https://github.com/ObolNetwork/charon/pull/475/files#diff-97a44ec75e8f5c1ad277c0f745acc3e68bcc8ecc5e42cc6d9568abab6d04f272R148-R152

dkg/frost.go
```diff
@@ -0,0 +1,275 @@
...
+			TargetID: 0, // Broadcast
+		}] = *castR2
+	}
+
+	return castResults, nil
+}
+
+// makeShares returns a slice of shares (one for each validator) from the DKG participants and round 2 results.
+func makeShares(validators map[uint32]*frost.DkgParticipant, r2Result map[msgKey]frost.Round2Bcast) ([]share, error) {
+	targetID := validators[0].Id
```
xenowits, line 198: can u pls explain this line? why the 0th validator's id?

xenowits (approved):

dkg/frost.go
corverroos: it is a DTO, so fields are public.

dkg/frost.go
corverroos: It is a package internal interface. since the fTransport itself is not exported. Having unexported methods isn't required, nor does it makes things clearer.

dkg/frost.go
corverroos: I'm trying to keep the function signature on a single line.

dkg/frost.go
corverroos: Since this is an interface method, having each on a new line VERY verbose and weird looking, this is a good compromise I think.

dkg/frost.go
corverroos:
```suggestion
	// Get our ID from any validator (they all have our ID). 
	targetID := validators[0].Id
```

dkg/frost.go
leolara: it might? if it is not going to be marshalled and not used outside, it should be private

dkg/frost.go
leolara: private is the default, if something are not going to be exported it should not be made public

PR #476, opened 2022-05-03
core: add logs for bcast and scheduler

dB2510:
Adds logs for successful and failed broadcast in bcast component and improves logs in scheduler from debug to info

category: refactor
ticket: #000
featureset: stable

### DIFF
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -66,9 +66,7 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 		}
 
 		err = b.eth2Cl.SubmitAttestations(ctx, []*eth2p0.Attestation{att})
-		if err != nil {
-			log.Error(ctx, "Failed to submit attestation to beacon node", err, z.Int("slot", int(duty.Slot)))
-		} else {
+		if err == nil {
 			log.Info(ctx, "Attestation submitted successfully to beacon node", z.Int("slot", int(duty.Slot)))
 		}
 
@@ -80,13 +78,11 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 		}
 
 		err = b.eth2Cl.SubmitBeaconBlock(ctx, block)
-		if err != nil {
-			log.Error(ctx, "Failed to submit block to beacon node", err, z.Int("slot", int(duty.Slot)))
-		} else {
+		if err == nil {
 			log.Info(ctx, "Block submitted successfully to beacon node", z.Int("slot", int(duty.Slot)))
 		}
 
-		return b.eth2Cl.SubmitBeaconBlock(ctx, block)
+		return err
 	case core.DutyRandao:
 		// Randao is an internal duty, not broadcasted to beacon chain
 		return nil
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -24,6 +24,8 @@ import (
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
 	"github.com/obolnetwork/charon/core"
 )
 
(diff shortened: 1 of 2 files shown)


core/bcast/bcast.go
```diff
@@ -63,13 +65,27 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
...
+		return err
 	case core.DutyProposer:
 		block, err := core.DecodeBlockAggSignedData(aggData)
 		if err != nil {
 			return err
 		}
 
+		err = b.eth2Cl.SubmitBeaconBlock(ctx, block)
+		if err != nil {
+			log.Info(ctx, "Block submitted successfully to beacon node", z.Int("slot", int(duty.Slot)))
```
xenowits, line 84: Lines 84 and 86 are interchanged

core/bcast/bcast.go
```diff
@@ -63,13 +65,27 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 			return err
 		}
 
-		return b.eth2Cl.SubmitAttestations(ctx, []*eth2p0.Attestation{att})
+		err = b.eth2Cl.SubmitAttestations(ctx, []*eth2p0.Attestation{att})
+		if err == nil {
```
xenowits, line 69: use either `err != nil` first or `err == nil` consistently, though i'd prefer the first one

xenowits (approved):

core/bcast/bcast.go
dB2510: done

core/bcast/bcast.go
dB2510: corrected

corverroos: should be `if err == nil {`

core/bcast/bcast.go
```diff
@@ -63,13 +65,27 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 			return err
 		}
 
-		return b.eth2Cl.SubmitAttestations(ctx, []*eth2p0.Attestation{att})
+		err = b.eth2Cl.SubmitAttestations(ctx, []*eth2p0.Attestation{att})
+		if err != nil {
+			log.Error(ctx, "Failed to submit attestation to beacon node", err, z.Int("slot", int(duty.Slot)))
```
corverroos, line 70: as per go guidelines, do not log AND return errors

core/scheduler/scheduler.go
```diff
@@ -114,7 +114,7 @@ func (s *Scheduler) Run() error {
 			return nil
 		case slot := <-slotTicker:
 			slotCtx := log.WithCtx(ctx, z.I64("slot", slot.Slot))
-			log.Debug(slotCtx, "Slot ticked")
+			log.Info(slotCtx, "Slot ticked")
```
corverroos, line 117: mmmm, not sure about this, lets leave as debug for now.

core/scheduler/scheduler.go
```diff
@@ -446,7 +444,7 @@ func resolveActiveValidators(ctx context.Context, eth2Cl eth2Provider,
 	var resp []validator
 	for index, val := range vals {
 		if !val.Status.IsActive() {
-			log.Debug(ctx, "skipping inactive validator", z.U64("index", uint64(index)))
+			log.Info(ctx, "skipping inactive validator", z.U64("index", uint64(index)))
```
corverroos, line 447: This will spam each time it fetches validators

corverroos (approved):
LGTM

core/bcast/bcast.go
dB2510: yeah you are right, we shouldn't log and return errors. but i added this to log so that it will be visible in logs at the time running charon-docker-compose or individual node.

Although, since we are returning error it will get propagated between core components

core/scheduler/scheduler.go
dB2510: sure

core/scheduler/scheduler.go
dB2510: only if validator is inactive. i think i should remove this

corverroos (approved):
LGTM

PR #477, opened 2022-05-04
core/validatorapi: ignore proposer duties error

dB2510:
Ignore the error for proposer duties as they are the rare ones. It is spamming in logs since it is being checked in every epoch.

category: bug
ticket: #000


corverroos (changes requested):
This is unsafe. If something is broken with `getPubShare`, we will NEVER know about it. Swallowing errors is a HUGE source of hidden bugs.

corverroos (approved):
LGTM

Issue #478, opened 2022-05-04
Implement Frost DKG


corverroos:
## Problem to be solved

We need a V1 DKG algorithm; something simple and compatible with our tbls library, preferably using kryptology. 

## Proposed solution

Implement Frost DKG from kryptology `github.com/coinbase/kryptology/pkg/dkg/frost`. See example `github.com/coinbase/kryptology/test/frost_dkg/bls`.

- Integrate into existing `dkg.Run` function

PR #483, opened 2022-05-05
cmd: wire dkg command

corverroos:
Wires the dkg command to the cli

category: feature
ticket: #478 
feature_set: stable

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/obolnetwork/charon/app"
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/dkg"
 )
 
 const (
@@ -46,13 +47,14 @@ func New() *cobra.Command {
 		newRunCmd(app.Run),
 		newBootnodeCmd(RunBootnode),
 		newCreateClusterCmd(runCreateCluster),
+		newDKGCmd(dkg.Run),
 	)
 }
 
 func newRootCmd(cmds ...*cobra.Command) *cobra.Command {
 	root := &cobra.Command{
 		Use:   "charon",
-		Short: "Charon - The Ethereum DVT middleware client",
+		Short: "Charon - Proof of Stake Ethereum Distributed Validator Client",
 		Long:  `Charon enables the operation of Ethereum validators in a fault tolerant manner by splitting the validating keys across a group of trusted parties using threshold cryptography.`,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			return initializeConfig(cmd)
--- a/cmd/dkg.go
+++ b/cmd/dkg.go
@@ -1,0 +1,52 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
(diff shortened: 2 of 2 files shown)


cmd/cmd.go
```diff
@@ -46,13 +47,14 @@ func New() *cobra.Command {
...
 		newCreateClusterCmd(runCreateCluster),
+		newDKGCmd(dkg.Run),
 	)
 }
 
 func newRootCmd(cmds ...*cobra.Command) *cobra.Command {
 	root := &cobra.Command{
 		Use:   "charon",
-		Short: "Charon - The Ethereum DVT middleware client",
+		Short: "Charon - Proof of Stake Ethereum Distributed Validator Client",
```
corverroos, line 57: aligning the latest github repo description

cmd/dkg.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/obolnetwork/charon/dkg"
+)
+
+func newDKGCmd(runFunc func(context.Context, dkg.Config) error) *cobra.Command {
```
leolara, line 27: I don't understand the goal of passing runFunc instead of calling directly dkg.Run from here. Are we going to use this parameter with something different at some point?

leolara (approved):

cmd/dkg.go
```diff
@@ -0,0 +1,52 @@
...
+	var config dkg.Config
+
+	cmd := &cobra.Command{
+		Use:   "dkg",
+		Short: "Create distributed validator key shares by participating in a DKG ceremony",
+		Long: `Participate in a distributed key generation ceremony for a specific cluster definition that creates
+distributed validator key shares and a final cluster lock configuration. Note that all other cluster operators should run
+this command at the same time.`,
```
xenowits, line 35:
```suggestion
this command at the same time in order to participate in the ceremony.`,
```

cmd/dkg.go
```diff
@@ -0,0 +1,52 @@
...
+	bindDataDirFlag(cmd.Flags(), &config.DataDir)
+	bindDefDirFlag(cmd.Flags(), &config.DefFile)
+	bindP2PFlags(cmd.Flags(), &config.P2P)
+	bindLogFlags(cmd.Flags(), &config.Log)
+
+	return cmd
+}
+
+func bindDefDirFlag(flags *pflag.FlagSet, dataDir *string) {
+	flags.StringVar(dataDir, "definition-file", ".charon/cluster_definition.json", "The path to the cluster definition configuration file.")
```
xenowits, line 51:
```suggestion
	flags.StringVar(dataDir, "definition-file", ".charon/cluster_definition.json", "The path to the cluster definition file.")
```

xenowits (approved):

cmd/dkg.go
corverroos: yeah, this allows testing the configuration flags

cmd/dkg.go
corverroos: this feels unnecessary

xenowits: i reason i say so is because when i read `... at the same time` for the first time, it looked to me as if everyone should be perfectly synced to the clock to join a ceremony which i suppose is not the case.

corverroos: sure, but your suggestion doesn't clarify that.

corverroos: And they do need to run it at the same time, they just do not need to start it as the same time 😉

xenowits: yeah i mean this: they can start at different times

cmd/dkg.go
corverroos: cool

cmd/dkg.go
leolara: ok

Issue #442, opened 2022-04-25
Refactor to use a create command with subcommands


OisinKyne:
## Problem to be solved

To clean up the charon [command] output, we want to nest a number of subcommands under `charon create`. 

## Proposed solution

Add the following commands under `charon create`
- `enr`: Create an Ethereum Node Record private key to identify charon
- `dkg`: Create a charon DKG configuration file for `charon dkg`
- `cluster`: Create the configuration and private keys for a local cluster

### Out of Scope

The functions don't need to work for this ticket. They should just exist, have appropriate flags, perform appropriate validation, and error with meaningful errors.

OisinKyne:
Let me know if you want to jam on the feature flags for these commands if you want to take an opportunity to simplify them @dB2510

dB2510:
@OisinKyne sure let's get on call to discuss

PR #484, opened 2022-05-05
cmd: add create dkg command

corverroos:
Adds the `charon create dkg` command that generates a cluster definition file.

category: feature
ticket: #442

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -48,6 +48,9 @@ func New() *cobra.Command {
 		newBootnodeCmd(RunBootnode),
 		newCreateClusterCmd(runCreateCluster),
 		newDKGCmd(dkg.Run),
+		newCreateCmd(
+			newCreateDKGCmd(runCreateDKG),
+		),
 	)
 }
 
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -1,0 +1,31 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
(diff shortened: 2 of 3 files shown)


cmd/cmd.go
```diff
@@ -48,6 +48,9 @@ func New() *cobra.Command {
 		newBootnodeCmd(RunBootnode),
 		newCreateClusterCmd(runCreateCluster),
 		newDKGCmd(dkg.Run),
+		newCreateCmd(
+			newCreateDKGCmd(runCreateDKG),
```
xenowits, line 52: `charon create dkg` just doesn't feel right to me

xenowits: i mean you can `do` a dkg but `creating` a dkg doesn't make sense to me

corverroos: I pushed for:
 - `charon define dkg` -> `cluster_definition.json`
 - `charon seal` -> `cluster_lock.json`
 - `charon run`
 
 Or for local cluster flow:
  - `charon define local`
  - `charon seal`
  - `charon run`
 
 But this this was rejected for:
  - `charon create dkg` -> `cluster_definition.json`
  - `charon dkg` -> `cluster_lock.json`
  - `charon run`

xenowits: is building a `cluster_definition` even part of a dkg?

xenowits: it's an input to a dkg, if i'm not wrong

dB2510: or we can do something like:
`charon create dkg-cluster`
or
`charon create dkg-config`

xenowits (approved):
Naming to be discussed later. Approving this for now!

cmd/cmd.go
leolara: Perhaps this should have been discussed as part of the ticket grooming. I don't think we should stop the merge of this because we do not agree on the command name. Let's create an issue and discuss it there the definete command name

PR #485, opened 2022-05-05
dkg: wait for peers to connect

corverroos:
Waits for all peers to connect before starting Frost DKG. This isn't required for keycast since it has retries built-in.

Also:
 - Improve logging
 - Make configuration files read-only
 - Add `DKGAlgo` to `charon create dkg` command.
 - Don't error when making folders (since this doesn't work in docker)

category: feature
ticket: #478 
feature_set: stable


xenowits (approved):
Looks good!

PR #486, opened 2022-05-05
app/log: remove caller from console logs

corverroos:
Removes the `{"caller": "dkg.go:121"}` field from console logs.

Reasoning:
 - The `caller` field present in ALL console logs adds more noise than signal.
 - Removing it reduces noise and focusses the logging on the message.
 - Console logs already include a "topic", so the general caller package/scope is already known.
 - Since we do structured logging, finding the source/caller is as easy as grepping the code base for the message.
 - Proposer production deployments should use JSON log format in any case.

category: refactor
ticket: none
feature_set: stable

### DIFF
--- a/app/log/config.go
+++ b/app/log/config.go
@@ -120,8 +120,6 @@ func buildConsoleLogger(level zapcore.Level, ws zapcore.WriteSyncer, opts ...fun
 			encoder, ws,
 			zap.NewAtomicLevelAt(level),
 		),
-		zap.WithCaller(true),
-		zap.AddCallerSkip(1),
 	)
 }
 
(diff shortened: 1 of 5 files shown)


corverroos:
Example result
image

dB2510 (approved):
I think we should keep the caller for DEBUG logs but remove it for INFO logs.
It gives you direct source of errors at the time of debugging even while you are writing a unit test and it fails. or a test fails in a github action. 

The caller line number instantly grabs attention to where the hell was the origin of that error.

corverroos:
Source of error is still present in the error stack trace.

corverroos:
Debug logs are for developers, Info logs are for customers.

Customers only care THAT stuff works. They do not care about HOW it works. 

As a customer, I really do not like spamming logs, I will never read them. If the logs are sparse, I will look at them and glean the important high level information (info logging). The sparser the logs, the better for customers.

PR #488, opened 2022-05-05
app/log: fix logfmt bug

corverroos:
This fixes a panic when using `logfmt` formatter and logging slices (which is doesn't support) by convertng any non-standard values to string first.

category: bug
ticket: none
feature_set: stable


corverroos:
Example logfmt output:
image

Issue #481, opened 2022-05-05
dkg: generate deposit_data


xenowits:
## Problem to be solved

Generate ETH2 `deposit_data` as we have a concrete dkg implementation (frost) ready which generates the keys.

Sample `deposit_data` file:
```json
  {
    "pubkey": "b41734eb148a763517988120bff29bfde8aa6a81a44810ba3184421d323b32893697ee4b02bf9c1ba7bfb7e6a21380c7",
    "withdrawal_credentials": "010000000000000000000000dfeb5c3093eeabb44e799b8da9d0b71d2eb37ccd",
    "amount": 32000000000,
    "signature": "b56d8c29817ee92ae663a36e5ab4da8d65a7362c4036120d26a42d69d267eb34bc4a360bdc945e273ab275a343de73cf15a9949d6ee511479ae2487e4b45daa39f426bf738ee97be9c54b66d05bd428789bef34c6ae32e812e59d96328fa6852",
    "deposit_message_root": "7704b2e98ace61d42a903ac68b0df923282a6f91508489f8e6394819f9a22f4b",
    "deposit_data_root": "10650f133403e8fe23ad24a38f90667a682d4904b8d2c19dcab603b8c0f8874b",
    "fork_version": "00000064",
    "eth2_network_name": "gnosis",
    "deposit_cli_version": "1.2.0"
  },
```

## Proposed solution

1. `pubkey` is the group pubkey and is an output of the dkg.
2. `withdrawal_credentials` exists in the cluster definition, `cluster.Definition.WithdrawalAddress`.
3. `signature` is a BLS signature of the serialized `deposit_data`
4. The `root`s can be calculated using https://benjaminion.xyz/eth2-annotated-spec/phase0/beacon-chain/#compute_signing_root

Explore prysm code for more details on how to generate deposit_data. See https://github.com/prysmaticlabs/prysm/blob/7af89a82c96a824bb09f6b9b46e4e47d153aa406/tools/genesis-state-gen/main.go

Add code to `testutil` folder for now.

### DoD
Types for `deposit-data` and `deposit-message` are implemented .
Implement `func MessageRoot(pubkey, withdrawalAddr) eth2p0.Root` and `func MarshalDepositData(pubkey, withdrawalAddr, sig, network) []byte`.

PR #491, opened 2022-05-05
dkg: generate deposit data

xenowits:
Generate ETH2 `deposit_data` file. See ticket for more description.

category: feature
ticket: #481 
feature_set: alpha


testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
+package eth2
```
leolara, line 1: @xenowits why is this in testutil? isn't it used in the normal DKG process?

corverroos: I asked him to put it here for now, next to `keystore`, we need to move them both and some other "eth2" related utils somewhere else but I'm not sure where. Waiting for a pattern to emerge.

@xenowits please rename package to `depositdata`

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.5.0
 	github.com/prometheus/client_golang v1.12.1
 	github.com/prysmaticlabs/go-bitfield v0.0.0-20210809151128-385d8c5e3fb7
+	github.com/prysmaticlabs/prysm/v2 v2.0.1
```
corverroos, line 25: I would avoid importing prysm, it is a HUGE dependency. And we do not need it actually.

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+func GenerateDepositData(def cluster.Definition, pubkey string) (DepositData, error) {
+	amount := uint64(32000000000) // 32 Ether in Gwei
+	depositMessage := DepositMessage{
+		PubKey:                pubkey,
+		Amount:                amount,
+		WithdrawalCredentials: def.WithdrawalAddress,
+	}
+
+	domainType := [4]byte{1, 2, 3, 4}
+	domain, err := signing.ComputeDomain(domainType, nil, nil)
```
corverroos, line 39: we already have this logic, see `validatorapi.GetDomain()`

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+// DepositMessage contains the basic information necessary to activate a validator. The fields are
+// hashed to get the DepositMessageRoot. This root is signed and the signature is added to DepositData.
+type DepositMessage struct {
+	PubKey                string `json:"pubkey"`
+	Amount                uint64 `json:"amount"`
+	WithdrawalCredentials string `json:"withdrawal_credentials"`
+}
+
+// GenerateDepositData generates a deposit data object by populating the required fields.
+func GenerateDepositData(def cluster.Definition, pubkey string) (DepositData, error) {
```
corverroos, line 30: pass in only what you need. You only use withdrawal address and forkversion from the definition, rather just pass them in directly.

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+	case "00001020":
+		return "prater"
+	case "60000069":
+		return "kintsugi"
+	case "70000069":
+		return "kiln"
+	case "00000064":
+		return "gnosis"
+	default:
+		return "mainnet"
```
corverroos, line 77: not sure defaulting to mainnet is correct. Rather default to "unknown"

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+	amount := uint64(32000000000) // 32 Ether in Gwei
+	depositMessage := DepositMessage{
+		PubKey:                pubkey,
+		Amount:                amount,
+		WithdrawalCredentials: def.WithdrawalAddress,
+	}
+
+	domainType := [4]byte{1, 2, 3, 4}
+	domain, err := signing.ComputeDomain(domainType, nil, nil)
+	depositMessageRoot, err := signing.ComputeSigningRoot(depositMessage, domain)
```
corverroos, line 40: we also alreadyhave this, see `validatorapi.prepSigningData`. You can just copy it.

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+type DepositMessage struct {
+	PubKey                string `json:"pubkey"`
+	Amount                uint64 `json:"amount"`
+	WithdrawalCredentials string `json:"withdrawal_credentials"`
+}
+
+// GenerateDepositData generates a deposit data object by populating the required fields.
+func GenerateDepositData(def cluster.Definition, pubkey string) (DepositData, error) {
+	amount := uint64(32000000000) // 32 Ether in Gwei
+	depositMessage := DepositMessage{
```
corverroos, line 32: suggest making DepositMessage unexported, and adding a `HashTreeRoot()` function to it. See prysm, `cluster.Lock` and `cluster.Defintion` how to implement `HashTreeRoot`

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	domainType := [4]byte{1, 2, 3, 4}
+	domain, err := signing.ComputeDomain(domainType, nil, nil)
+	depositMessageRoot, err := signing.ComputeSigningRoot(depositMessage, domain)
+	if err != nil {
+		return DepositData{}, errors.Wrap(err, "compute domain")
+	}
+
+	depositDataRoot := ""
```
corverroos, line 45: same here depositDataRoot needs to be calcualted by a method you need to add to `DepositData` called `HashTreeRoot`, see prsym which fields and what order to hash them.

corverroos (commented):
commented

testutil/eth2/deposit_data.go
corverroos: you might beed to dig into how `eth2client` does it and copy inner logic since genesis and forkversion is nil in this case.

testutil/eth2/deposit_data.go
corverroos, line 30: always use proper types, use eth2p0.PublicKey not string

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+	signature := "" // sign the depositMsgRoot with the bls key
+
+	deposit_data := DepositData{
+		PubKey:                pubkey,
+		Amount:                amount,
+		WithdrawalCredentials: def.WithdrawalAddress,
```
corverroos, line 51: note that withdrawal credentials isn't an address, is looks different to me

testutil/eth2/deposit_data.go
```diff
@@ -0,0 +1,79 @@
...
+
+	domainType := [4]byte{1, 2, 3, 4}
+	domain, err := signing.ComputeDomain(domainType, nil, nil)
+	depositMessageRoot, err := signing.ComputeSigningRoot(depositMessage, domain)
+	if err != nil {
+		return DepositData{}, errors.Wrap(err, "compute domain")
+	}
+
+	depositDataRoot := ""
+	signature := "" // sign the depositMsgRoot with the bls key
```
corverroos, line 46: signing in a distributed validator is not trivial, so we need to restructure this logic, you cannot sign "inside a function".

Suggest splitting into two functions:
 - `deposits.GetMessageRoot(ctx, eth2Cl, pubkey, withdrawalPubKey) eth2p0.Root, error`
 - We will then threshold sign this sig root, and aggregate it to get a the signature, and only then make deposit data.
 - `deposits.GetDepositData(pubkey, withdrawalPubKey), forkversion, messageRoot, messageSig) DepositData, error`

testutil/eth2/deposit_message.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	return b, nil
+}
+
+func (d DepositMessage) HashTreeRootWith(hh *ssz.Hasher) error {
+	idx := hh.Index()
+
+	// Field 0 'PubKey`
+	hh.PutBytes([]byte(d.PubKey))
```
corverroos, line 30: this is going to fail, since the type of pubkey is a string, so it is really unclear what format should be. Rather make it `[]byte` then we know it is a standard public key bytes format

testutil/eth2/deposit_message.go
```diff
@@ -0,0 +1,41 @@
...
+func (d DepositMessage) HashTreeRoot() ([32]byte, error) {
+	b, err := ssz.HashWithDefaultHasher(d)
+	if err != nil {
+		return [32]byte{}, errors.Wrap(err, "hash deposit message")
+	}
+
+	return b, nil
+}
+
+func (d DepositMessage) HashTreeRootWith(hh *ssz.Hasher) error {
```
corverroos, line 26: please add tests for this, try to find reference data in prysm. Ie. given a message with values X,Y,Z, we know the hash should be something

testutil/eth2/deposit_message.go
```diff
@@ -0,0 +1,41 @@
...
+	idx := hh.Index()
+
+	// Field 0 'PubKey`
+	hh.PutBytes([]byte(d.PubKey))
+
+	// Field 1 'Amount'
+	hh.PutUint64(d.Amount)
+
+	// Field 2 'WithdrawalCredentials'
+	hh.PutBytes([]byte(d.WithdrawalCredentials))
```
corverroos, line 36: same here, this should be bytes, also please reference the format spec.

testutil/eth2/deposit_data.go
xenowits: i'm not sure either but sure will default to `unknown`

testutil/eth2/deposit_data.go
leolara: @corverroos don't you think we should avoid the eth2 term as it has become unofficial? Perhaps instead of using eth2 we could use ECL?

testutil/eth2/deposit_data.go
OisinKyne: No no! Jim McDonald and others pushed to change defaults to mainnet instead of testnet, to reduce the chance of an invalid deposit on mainnet. An unknown/non-mainnet default could blow up lots of money. Sending a mainnet deposit data to testnet blows up monopoly money. (I once blew up 120 testnet ether when Jim made this change. Before then ethdo pulled fork_version from connected beacon node, now it defaults to mainnet and you have to explicitly tell it to do something else.

testutil/eth2/deposit_data.go
corverroos: as discussed, we are sticking with eth2 for now

testutil/eth2/deposit_data.go
leolara: please, don't copy/paste code if it is possible

eth2util/deposit/deposit_message.go
```diff
@@ -0,0 +1,106 @@
...
+
+	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/ethereum/go-ethereum/common"
+	ssz "github.com/ferranbt/fastssz"
+
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+const ETH1_ADDRESS_WITHDRAWAL_PREFIX = "01"
```
corverroos, line 29: do not export types/interfaces/variables/constants by default, always start unexported

eth2util/deposit/deposit_message.go
```diff
@@ -0,0 +1,106 @@
...
+// withdrawalCredentialsFromAddr returns the WithdrawalCredentials corresponding to a '0x01' Ethereum withdrawal address.
+func withdrawalCredentialsFromAddr(addr string) (WithdrawalCredentials, error) {
+	// Check for validity of address.
+	if !common.IsHexAddress(addr) {
+		return WithdrawalCredentials{}, errors.New("invalid withdrawal address", z.Str("address", addr))
+	}
+
+	withdrawalCreds := make([]byte, 0, 32)
+
+	prefix, err := hex.DecodeString(ETH1_ADDRESS_WITHDRAWAL_PREFIX)
```
corverroos, line 54: rather make the constant a []byte, then no need to convert

eth2util/deposit/deposit_message.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// withdrawalCredentialsFromAddr returns the WithdrawalCredentials corresponding to a '0x01' Ethereum withdrawal address.
+func withdrawalCredentialsFromAddr(addr string) (WithdrawalCredentials, error) {
+	// Check for validity of address.
+	if !common.IsHexAddress(addr) {
+		return WithdrawalCredentials{}, errors.New("invalid withdrawal address", z.Str("address", addr))
+	}
+
+	withdrawalCreds := make([]byte, 0, 32)
```
corverroos, line 52: introduces variables just before they are used

canonical slice instantiation: `var creds []byte`

corverroos: avoid micro optimisations (adding length to slice). It draws attention ... to nothing in this case. 

Only do micro optimisations if really required, like in some hot path.

eth2util/deposit/deposit_message.go
```diff
@@ -0,0 +1,106 @@
...
+
+	prefix, err := hex.DecodeString(ETH1_ADDRESS_WITHDRAWAL_PREFIX)
+	if err != nil {
+		return WithdrawalCredentials{}, errors.Wrap(err, "decode prefix")
+	}
+	// Append the single byte ETH1_ADDRESS_WITHDRAWAL_PREFIX as prefix.
+	withdrawalCreds = append(withdrawalCreds, prefix...)
+
+	zeroes, err := hex.DecodeString("0000000000000000000000")
```
corverroos, line 61: same here, make this a global `var elevenZeros = []byte{0x00,0x00, ...}```

eth2util/deposit/deposit_message.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return WithdrawalCredentials{}, errors.Wrap(err, "decode zero string")
+	}
+	// Append 11 bytes of 0.
+	withdrawalCreds = append(withdrawalCreds, zeroes...)
+
+	// Finally, append 20 bytes of ethereum address.
+	withdrawalCreds = append(withdrawalCreds, addr...)
```
corverroos, line 69: you are appending a hex string to a byte slice...

eth2util/deposit/deposit_message.go
xenowits: ok. made it a single byte: `byte(1)`

eth2util/deposit/deposit_data.go
```diff
@@ -0,0 +1,229 @@
...
+		amount:                amount,
+		withdrawalCredentials: withdrawalCreds,
+	}
+
+	// calculate the hash tree root of depositMessage
+	depositMessageRoot, err := depositMessage.HashTreeRoot()
+	if err != nil {
+		return DepositData{}, err
+	}
+
```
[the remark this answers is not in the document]
corverroos: verify sig using: `pubkey == crypto.SigToPub(depositMessageRoot[:], sig)`

eth2util/deposit/deposit_data.go
```diff
@@ -0,0 +1,229 @@
...
+	// Signature is constructed from DepositMessageRoot combined with DOMAIN_DEPOSIT.
+	Signature eth2p0.BLSSignature
+
+	// ForkVersion identifies the network/chainID.
+	ForkVersion eth2p0.Version
+}
+
+// GenerateDepositData generates a deposit data object by populating the required fields.
+func GenerateDepositData(addr string, forkVersion string, pubkey eth2p0.BLSPubKey) (DepositData, error) {
```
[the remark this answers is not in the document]
xenowits: the idea to skip the structs obviously seems very simple. let's try this way then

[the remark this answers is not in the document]
xenowits: we can get rid of the unmarshal functions though since we only need to marshal and save to disk

eth2util/deposit/deposit_data.go
```diff
@@ -0,0 +1,155 @@
...
+	// Field 5 'ForkVersion'
+	hh.PutBytes(d.ForkVersion[:])
+
+	hh.Merkleize(idx)
+
+	return nil
+}
+
+// MarshalDepositData returns the json serialised deposit data bytes to be written to disk.
+func MarshalDepositData(pubkey eth2p0.BLSPubKey, msgRoot eth2p0.Root, sig eth2p0.BLSSignature, withdrawalAddr, forkVersion string) ([]byte, error) {
```
corverroos, line 87: maybe rather remove msgRoot, since you can calculate it

eth2util/deposit/deposit_data.go
```diff
@@ -0,0 +1,155 @@
...
+	idx := hh.Index()
+
+	// Field 0 'PubKey`
+	hh.PutBytes(d.PubKey[:])
+
+	// Field 1 'Amount'
+	hh.PutUint64(uint64(d.Amount))
+
+	// Field 2 'Eth1WithdrawalAddress'
+	hh.PutBytes([]byte(d.Eth1WithdrawalAddress))
```
corverroos, line 70: this should probably be credentials, not address

corverroos (approved):

eth2util/deposit/deposit_data.go
xenowits: i'm thinking along the lines of if we have a `sig` (that's calculated on `msgRoot`), we already have a `msgRoot`. So, we could have that as an input parameter.

eth2util/deposit/deposit_data.go
xenowits: correct, also changed the ordering of fields according to spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#depositdata

eth2util/deposit/deposit_message.go
```diff
@@ -0,0 +1,105 @@
...
+
+	var withdrawalCreds []byte
+
+	// Append the single byte ETH1_ADDRESS_WITHDRAWAL_PREFIX as prefix.
+	withdrawalCreds = append(withdrawalCreds, eth1AddressWithdrawalPrefix)
+
+	// Append 11 bytes of 0.
+	withdrawalCreds = append(withdrawalCreds, elevenZeroes...)
+
+	addrBytes, err := hex.DecodeString(addr)
```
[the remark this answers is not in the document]
OisinKyne: Is this appending or prepending the zeros? Because it should be at the start of the string.

E.g. 0x01 + 00000000000 + %s

Issue #490, opened 2022-05-05
Help users figure out their ENR


OisinKyne:
## Problem to be solved

The first step on the journey to running a multi operator cluster is getting your ENR. This is a key hurdle to overcome. 

Right now if you type `charon enr` on a fresh install this is what you get

```
./charon enr 
Error: load key: open .charon/data/p2pkey: no such file or directory
Usage:
  charon enr [flags]

Flags:
      --data-dir string                The directory where charon will store all its internal data (default ".charon/data")
  -h, --help                           Help for enr
      --p2p-allowlist string           Comma-separated list of CIDR subnets for allowing only certain peer connections. Example: 192.168.0.0/16 would permit connections to peers on your local network only. The default is to accept all connections.
      --p2p-bootmanifest               Enables using manifest ENRs as discv5 bootnodes. Allows skipping explicit bootnodes if key generation ceremony included correct IPs.
      --p2p-bootnode-relay             Enables using bootnodes as libp2p circuit relays. Useful if some charon nodes are not have publicly accessible.
      --p2p-bootnodes strings          Comma-separated list of discv5 bootnode URLs or ENRs. Example: enode://<hex node id>@10.3.58.6:30303?discport=30301.
      --p2p-denylist string            Comma-separated list of CIDR subnets for disallowing certain peer connections. Example: 192.168.0.0/16 would disallow connections to peers on your local network. The default is to accept all connections.
      --p2p-external-hostname string   The DNS hostname advertised by libp2p. This may be used to advertise an external DNS.
      --p2p-external-ip string         The IP address advertised by libp2p. This may be used to advertise an external IP.
      --p2p-peerdb string              Path to store a discv5 peer database. Empty default results in in-memory database.
      --p2p-tcp-address strings        Comma-separated list of listening TCP addresses (ip and port) for libP2P traffic. (default [127.0.0.1:16003])
      --p2p-udp-address string         Listening UDP address (ip and port) for discv5 discovery. (default "127.0.0.1:16004")

Error: load key: open .charon/data/p2pkey: no such file or directory
```

## Proposed solution

Handle this error more gracefully and probably less verbosely, maybe prompt to run `charon create enr` in the error message. 
Maybe an error like:
```
Error: No ENR private key found in ${data_dir}. If this is your first time running this client, create one with  `charon create enr`.

```

### Out of Scope

If there is anything to highlight as out of scope for this issue, please outline it here.

PR #493, opened 2022-05-05
refactor: modify enr error message

xenowits:
Assist users running `charon enr` in creating their ENRs. See ticket for more details. 
Probably may also need to do more than just a helpful message. Thoughts are welcome!

category: refactor

ticket: #490 
feature_set: alpha

### DIFF
--- a/cmd/enr.go
+++ b/cmd/enr.go
@@ -50,8 +50,9 @@ func newEnrCmd(runFunc func(io.Writer, p2p.Config, string) error) *cobra.Command
 // runNewENR loads the p2pkey from disk and prints the ENR for the provided config.
 func runNewENR(w io.Writer, config p2p.Config, dataDir string) error {
 	key, err := p2p.LoadPrivKey(dataDir)
+	helpMsg := fmt.Sprintf("No ENR private key found in %s. If this is your first time running this client, create one with `charon create enr`.", p2p.KeyPath(dataDir))
 	if err != nil {
-		return errors.New(fmt.Sprintf("No ENR private key found in %s. If this is your first time running this client, create one with `charon create enr`.", p2p.KeyPath(dataDir)))
+		return errors.New(helpMsg)
 	}
 
 	localEnode, db, err := p2p.NewLocalEnode(config, key)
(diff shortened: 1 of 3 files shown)


OisinKyne (approved):

cmd/enr.go
```diff
@@ -50,8 +50,9 @@ func newEnrCmd(runFunc func(io.Writer, p2p.Config, string) error) *cobra.Command
 // runNewENR loads the p2pkey from disk and prints the ENR for the provided config.
 func runNewENR(w io.Writer, config p2p.Config, dataDir string) error {
 	key, err := p2p.LoadPrivKey(dataDir)
+	helpMsg := fmt.Sprintf("No ENR private key found in %s. If this is your first time running this client, create one with `charon create enr`.", p2p.KeyPath(dataDir))
 	if err != nil {
-		return err
+		return errors.New(helpMsg)
```
corverroos, line 55: You are assuming that this is the only error that can occur, which is not true. Also please use structured errors always.

Suggest
```
key, err := p2p.LoadP2PKey(dir)
if os.IsNotExist(err) {
  return errors.New("ENR private key not found. Maybe create one with `charon create enr`", z.Str("path", p2p.KeyPath(dataDir))
} else if err != nil {
  return err
} 
return errors.Wrap(err, ""
```
Also please add a test for this!

corverroos (changes requested):
Please add test

cmd/enr.go
xenowits: For `os.IsNotExist(err)` to work, the error coming from calling `p2p.LoadPrivKey` should not be wrapped. wrapping changes the error type from `*fs.PathError` to `structured`.

But, when i attempt to simply return err, like, 
```go

func LoadPrivKey(dataDir string) (*ecdsa.PrivateKey, error) {
	key, err := crypto.LoadECDSA(KeyPath(dataDir))
	if err != nil {
		// return nil, errors.Wrap(err, "load key")
                // But, this throws an error by golangci-lint's wrapcheck linter
                	return nil, err
	}

	return key, nil
}
```

corverroos (approved):
LGTM

PR #494, opened 2022-05-06
cmd: make commands backwards compatible

dB2510:
Make commands backwards compatible until charon-docker-compose and docs are updated.

category: refactor
ticket: #000
feature_set: stable

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -43,11 +43,11 @@ func New() *cobra.Command {
 	return newRootCmd(
 		newVersionCmd(runVersionCmd),
 		newEnrCmd(runNewENR),
-
+		// TODO(dhruv): remove genp2pkey command once charon-docker-compose and docs are updated
 		newGenP2PKeyCmd(runGenP2PKey),
 		newRunCmd(app.Run),
 		newBootnodeCmd(RunBootnode),
-
+		// TODO(dhruv): replace newCreateClusterCmdNew with newCreateClusterCmd once charon-docker-compose and docs are updated
 		newCreateClusterCmd(runCreateCluster),
 		newDKGCmd(dkg.Run),
 		newCreateCmd(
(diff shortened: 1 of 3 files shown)


cmd/cmd.go
```diff
@@ -43,13 +43,17 @@ func New() *cobra.Command {
 	return newRootCmd(
 		newVersionCmd(runVersionCmd),
 		newEnrCmd(runNewENR),
+
```
xenowits, line 46: nit: why this additional line?

cmd/createcluster.go
```diff
@@ -98,6 +98,25 @@ type clusterConfig struct {
...
+			return runFunc(cmd.OutOrStdout(), conf)
+		},
+	}
+
+	bindClusterFlags(cmd.Flags(), &conf)
+
+	return cmd
+}
+
+// TODO(dhruv): replace newCreateClusterCmd with newCreateClusterCmdNew once charon-docker-compose are updated.
```
xenowits, line 116: can also add a comment before `newCreateClusterCmd` saying that it will be deprecated out

xenowits (approved):
Looks good!

cmd/createcluster.go
dB2510: actually `newCreateClusterCmdNew` will be renamed to `newCreateClusterCmd` and existing `newCreateClusterCmd` will be removed

leolara: I agree @xenowits I agree the legacy commands should have a big "DEPRECATED" in they Short and Long fields. Independently of the name of the `new...Cmd` function

leolara (changes requested):
Show DEPRECATED in short and long docs for commands that are going to be removed

cmd/createcluster.go
dB2510: Done!

dB2510:
> Show DEPRECATED in short and long docs for commands that are going to be removed

Done!

leolara (approved):

PR #495, opened 2022-05-06
eth2util: refactor project structure

corverroos:
Introduces `eth2util` package that contains eth2 common functionality. With introduction of DKG, we need keystores and depositdatas in production code, not just test code. We also need to sign stuff in production code, not just test code. So move signing and keystore logic to eth2util package.

category: refactor
ticket: none
feature_set: stable


leolara (changes requested):
@corverroos

 + I wouldn't use eth2 as it is a depracted name, I would use ECL or some other suggestion.
 + util and tools and common is usually a code smell that something does not have a particular goal. I don't think this is the case but I wouldn't use util in the name
  + As we are working on the ECL protocol exclusively, perhaps this package should be called protocol?

Issue #445, opened 2022-04-26
QBFT opt-in beta consensus algo


corverroos:
## Problem to be solved

We have a QBFT PoC but it isn't integrated and so isn't really tested in the field. Add QBFT as opt-in Beta feature consensus algorithm. After sufficient testing and verification, then we can bump to Alpha and Stable.

## Proposed solution

Add QBFT as opt-in beta consensus algorithm.

### Out of Scope

Related but out of scope is adding the `featureset` package to control feature rollout.

PR #500, opened 2022-05-07
core/qbft: refactor to explicit justifications

corverroos:
Refactors QBFT to explicit justifications.

Also:
 - Do not exit when decided, keep running until context cancelled
 - Send Qcommit to others after decided
 - Jump ahead when receiving justified messages
 - Trim old round messages from buffer
 - Add constant period tests
 - Make instance type bytes (to support "any" instance type)

category: refactor 
ticket: #445 
feature_set: alpha

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -36,6 +36,9 @@ type Transport struct {
 	// processes in the system (including this process).
 	Broadcast func(Msg)
 
+	// SendQCommit sends the commit messages to a specific process.
+	SendQCommit func(target int64, qCommit []Msg)
+
 	// Receive returns a stream of messages received
 	// from other processes in the system (including this process).
 	Receive <-chan Msg
@@ -45,13 +48,13 @@ type Transport struct {
 // This remains constant across multiple instances of consensus (calls to Run).
 type Definition struct {
 	// IsLeader is a deterministic leader election function.
-	IsLeader func(instance, round, process int64) bool
+	IsLeader func(instance []byte, round, process int64) bool
 	// NewTimer returns a new timer channel and stop function for the round.
 	NewTimer func(round int64) (<-chan time.Time, func())
-	// IsValid validates messages.
-	IsValid func(instance int64, msg Msg) bool
+	// Decide is called when consensus has been reached on a value.
+	Decide func(instance []byte, value []byte, qcommit []Msg)
 	// LogUponRule allows debug logging of triggered upon rules on message receipt.
-	LogUponRule func(instance, process, round int64, msg Msg, uponRule string)
+	LogUponRule func(instance []byte, process, round int64, msg Msg, uponRule string)
 	// Quorum is the quorum count for the system.
 	Quorum int
 	// Faulty is the maximum faulty process count for the system.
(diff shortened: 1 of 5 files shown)


dB2510 (approved):

xenowits (approved):

core/qbft/qbft.go
```diff
@@ -87,18 +99,21 @@ type Msg struct {
...
+	uponQuorumCommits
+	uponUnjustRoundChange
+	uponFPlus1RoundChanges
+	uponQuorumRoundChanges
 )
 
-// Run returns the consensus decided value (Qcommit) or a context closed error.
-func Run(ctx context.Context, d Definition, t Transport, instance, process int64, inputValue []byte) (value []byte, err error) {
+// Run executes the consensus algorithm until the context closed.
+//nolint:gocognit // If is indeed a complex algorithm.
```
dB2510, line 113:
```suggestion
//nolint:gocognit // It is indeed a complex algorithm.
```

core/qbft/qbft.go
```diff
@@ -109,167 +124,264 @@ func Run(ctx context.Context, d Definition, t Transport, instance, process int64
...
+	// === State ===
+
+	var (
+		round           int64 = 1
+		preparedRound   int64
+		preparedValue   []byte
+		preparedJustify []Msg
+		qCommit         []Msg
+		buffer          []Msg
+		dedupIn         = make(map[dedupKey]bool)
```
dB2510, line 136: What is dedup?

core/qbft/qbft.go
```diff
@@ -87,18 +99,21 @@ type Msg struct {
...
+	uponUnjustRoundChange
+	uponFPlus1RoundChanges
+	uponQuorumRoundChanges
 )
 
-// Run returns the consensus decided value (Qcommit) or a context closed error.
-func Run(ctx context.Context, d Definition, t Transport, instance, process int64, inputValue []byte) (value []byte, err error) {
+// Run executes the consensus algorithm until the context closed.
+//nolint:gocognit // If is indeed a complex algorithm.
+func Run(ctx context.Context, d Definition, t Transport, instance []byte, process int64, inputValue []byte) (err error) {
```
dB2510, line 114: Suggest: parameter names can be `def Defination, trn Transport`

core/qbft/qbft.go
```diff
@@ -72,13 +75,22 @@ const (
 
 // Msg defines the inter process messages.
```
dB2510, line 73: suggest: replace word `process` with `nodes` as it is confusing since we deal with Charon **nodes** not processes

corverroos: the algorithm paper talks about processes (so sticking with its terminology)

core/qbft/qbft.go
corverroos: sure, but `d` and `t` also works no?

core/qbft/qbft.go
corverroos: A very common term for "deduplication"

core/qbft/qbft.go
corverroos: will fix in generics PR 👍

core/qbft/qbft.go
dB2510: def sounds more of defination and t confuses with t of testing

core/qbft/qbft.go
dB2510: yeah sure, paper talks about processes but we are dealing with nodes not processes so we can/should modify terminologies as per our usecases

PR #501, opened 2022-05-09
core/qbft: refactor to use generics

corverroos:
Refactors qbft to use generics. This simplified the integration significantly since an external buffer isn't required and casting and conversion and (errors) isn't required.

category: refactor 
ticket: #445 
feature_set: alpha

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
         exclude: ".pb.go"
 
   - repo: https://github.com/corverroos/fiximports
-    rev: 496e5392530966aa8f12e88a289c07782075cf91
+    rev: 886071dd81a69ab030db85d65803fb2048aa0722
     hooks:
       - id: go-fiximports # format imports for go source files
 
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -26,8 +26,8 @@ import (
 	"github.com/obolnetwork/charon/app/errors"
 )
 
-// Value defines the generic value type constraint
-// that provides an equality method.
+// Value defines the constraints of the generic value type.
+// The only constraint is an equality method.
 type Value[V any] interface {
 	Equal(V) bool
 }
@@ -37,9 +37,9 @@ type Value[V any] interface {
 // Note that broadcasting doesn't return an error. Since this algorithm is idempotent
 // it is suggested to just retry broadcasting indefinitely until it succeeds or times out.
 type Transport[I any, V Value[V]] struct {
-	// Broadcast sends the message to all other
+	// Broadcast sends a message with the provided fields to all other
 	// processes in the system (including this process).
-	Broadcast func(Msg[I, V])
+	Broadcast func(typ MsgType, instance I, source int64, round int64, value V, pr int64, pv V, justify []Msg[I, V])
 
 	// SendQCommit sends the commit messages to a specific process.
 	SendQCommit func(target int64, qCommit []Msg[I, V])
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
         exclude: ".pb.go"
 
   - repo: https://github.com/corverroos/fiximports
-    rev: 496e5392530966aa8f12e88a289c07782075cf91
```
corverroos, line 26: had to bump `fiximports` to support go1.18 generics

core/qbft/qbft.go
```diff
@@ -66,31 +72,38 @@ type Definition struct {
...
+	msgSentinel    MsgType = 5
 )
 
+func (i MsgType) Valid() bool {
+	return i > MsgUnknown && i < msgSentinel
+}
+
 // Msg defines the inter process messages.
-type Msg struct {
+type Msg[I any, V Value[V]] interface {
```
corverroos, line 91: Changing message to an interface simplifies explicit justifications, since external buffer of signed messages isn't required.

core/qbft/qbft.go
```diff
@@ -27,34 +26,42 @@ import (
 	"github.com/obolnetwork/charon/app/errors"
 )
 
+// Value defines the generic value type constraint
+// that provides an equality method.
+type Value[V any] interface {
```
leolara, line 31: @corverroos I don't think the name of this interface represents the methods it countais. Something like Equaler is better

core/qbft/qbft.go
```diff
@@ -27,34 +26,42 @@ import (
...
+	Equal(V) bool
+}
+
 // Transport abstracts the transport layer between processes in the consensus system.
 //
 // Note that broadcasting doesn't return an error. Since this algorithm is idempotent
 // it is suggested to just retry broadcasting indefinitely until it succeeds or times out.
-type Transport struct {
+type Transport[I any, V Value[V]] struct {
```
leolara, line 39: @corverroos what is I and what is V ?

core/qbft/qbft.go
leolara: Also, here it is not clear to me what is I and what is V

leolara:
I like that messages changes from public fields to public accessors

core/qbft/qbft.go
leolara: So, I guess V is the type of the value reaching a consensus on? But I?

core/qbft/qbft.go
OisinKyne: `Instance` and `Value` https://github.com/ObolNetwork/charon/pull/501/files#diff-675cdeeca6e8a606721a8f590fe58fbae35d8ba37a64956b018f047ace94b541R62

core/qbft/qbft.go
corverroos: `I` is the instance of consensus as per the paper

core/qbft/qbft.go
corverroos: Yeah, but using `V Value` for the "value" makes a lot more sense than `Equaler` for "value". In this case, this interface defines what constraints the "value" type. Which is only that it has an equal method.

core/qbft/qbft.go
corverroos: Yup:
```
// Instance identifies the consensus instance.
Instance() I

// Value being proposed.
Value() V
```

OisinKyne (approved):

core/qbft/qbft.go
leolara: I see that this is the interface for the value we are reaching consensus for, it makes sense to have an interface called Value. I was confused because it didn't have other methods, I guess you will need to add it some other method to be able to use it, like something that returns V? In order to use the Value[v] returned by the consensus?

core/qbft/qbft.go
leolara: Ok, then

PR #503, opened 2022-05-09
core/qbft: calculate q and f

corverroos:
Calculates `Quorum` and `Faulty` from `Nodes` as per Roberto suggestion. Also using his paper for the formulas.

Also rename `justify` to `justification`

category: refactor
ticket: #445 
feature_set: alpha

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -60,10 +60,20 @@ type Definition[I any, V Value[V]] struct {
 	Decide func(instance I, value V, qcommit []Msg[I, V])
 	// LogUponRule allows debug logging of triggered upon rules on message receipt.
 	LogUponRule func(instance I, process, round int64, msg Msg[I, V], uponRule string)
-	// Quorum is the quorum count for the system.
-	Quorum int
-	// Faulty is the maximum faulty process count for the system.
-	Faulty int
+	// Nodes is the total number of nodes/processes participating in consensus.
+	Nodes int
+}
+
+// Quorum returns the quorum count for the system.
+// See IBFT 2.0 paper for correct formula: https://arxiv.org/pdf/1909.10194.pdf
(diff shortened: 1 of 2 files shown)


core/qbft/qbft.go
```diff
@@ -39,7 +39,7 @@ type Value[V any] interface {
 type Transport[I any, V Value[V]] struct {
 	// Broadcast sends a message with the provided fields to all other
 	// processes in the system (including this process).
-	Broadcast func(typ MsgType, instance I, source int64, round int64, value V, pr int64, pv V, justify []Msg[I, V])
+	Broadcast func(typ MsgType, instance I, source int64, round int64, value V, pr int64, pv V, justification []Msg[I, V])
```
dB2510, line 42: Maybe
```suggestion
	Broadcast func(typ MsgType, instance I, source int64, round int64, value V, pr int64, pv V, justifications []Msg[I, V])
```

core/qbft/qbft.go
```diff
@@ -60,10 +60,20 @@ type Definition[I any, V Value[V]] struct {
 	Decide func(instance I, value V, qcommit []Msg[I, V])
 	// LogUponRule allows debug logging of triggered upon rules on message receipt.
 	LogUponRule func(instance I, process, round int64, msg Msg[I, V], uponRule string)
-	// Quorum is the quorum count for the system.
-	Quorum int
-	// Faulty is the maximum faulty process count for the system.
-	Faulty int
+	// Nodes is the total number of nodes/processes participating in consensus.
+	Nodes int
```
dB2510, line 64: 😏

dB2510 (approved):

OisinKyne (approved):

core/qbft/qbft.go
corverroos: technically it is a single justification containing multiple messages...

PR #504, opened 2022-05-09
core/qbft: add drop tests and fix bugs

corverroos:
Add the 10% dropped message test, which found some bugs:

- Condition J2 doesn't require quorum identical `preparedValue`, only a single RC with that value (confirmed this with GoQuorum `quorum/consensus/istanbul/qbft/core/justification.go:79`).
- Typo on line 652.
- Trimming older rounds removes required PREPARE message. Disable for now.
- Adds a fail fast if quorum round changes are not justified.

category: bug
ticket: #445 
feature_set: alpha

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -199,21 +199,22 @@ func Run[I any, V Value[V]](ctx context.Context, d Definition[I, V], t Transport
 
 	// trimBuffer drops all older round's buffered messages.
 	trimBuffer := func() {
-		var selected []Msg[I, V]
-		for _, msg := range buffer {
-			if msg.Round() >= round {
-				selected = append(selected, msg)
-			}
-		}
-		buffer = selected
-
-		dedup := make(map[dedupKey]bool)
-		for k := range dedupIn {
-			if k.Round >= round {
-				dedup[k] = true
-			}
-		}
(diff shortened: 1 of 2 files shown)


OisinKyne (approved):
I like the spacing changes to align everything 😅

core/qbft/qbft.go
```diff
@@ -199,21 +199,22 @@ func Run[I any, V Value[V]](ctx context.Context, d Definition[I, V], t Transport
...
+		// }
+		// buffer = selected
+		//
+		// dedup := make(map[dedupKey]bool)
+		// for k := range dedupIn {
+		//	if k.Round >= round {
+		//		dedup[k] = true
+		//	}
+		// }
+		// dedupIn = dedup
```
dB2510, line 217: what is wrong with this approach? I assume there would be cases where we would need message of rounds before `round-1`

dB2510 (approved):

core/qbft/qbft.go
corverroos: We sometimes loose some PREPAREs from previous rounds, then we cannot trigger UnponQuorumRoundChange singe Qrc isn't justified. I have a plan to fix this in PR.

PR #507, opened 2022-05-10
core/qbft: fix trimming issue

corverroos:
Fixes buffer trimming issue:
- There is only a few uses for Justification messages: 
  - a) Check if a received PRE-PREPARE is justified (only use justifications of that single message)
  - b) Check if a received ROUND-CHANGE is justified  (only use justifications of that single message)
  - c) Generate justification of a quorum ROUND-CHANGEs (select from all buffered messages and their justifications)
- We only need previously received justifications for (c), and we can get that by `flattening` the buffer.
- Since we do not add old justifications to the buffer anymore
- We can trim all old messages again.

Also make code a bit more explicit and readable.
Also ensure that send-to-self in the test is always first and never drops or jitters.

category: bug
ticket: #445

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -354,7 +354,7 @@ func classify[I any, V Value[V]](d Definition[I, V], instance I, round, process 
 		}
 
 	case MsgRoundChange:
-		if !IsJustifiedRoundChange(d, msg) {
+		if !isJustifiedRoundChange(d, msg) {
 			return uponUnjustRoundChange, nil
 		}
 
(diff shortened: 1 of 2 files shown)


core/qbft/qbft.go
```diff
@@ -428,25 +421,27 @@ func highestPrepared[I any, V Value[V]](qrc []Msg[I, V]) (int64, V) {
 
 // nextMinRound implements algorithm 3:6 and returns the next minimum round
 // from received round change messages.
-func nextMinRound[I any, V Value[V]](d Definition[I, V], msgs []Msg[I, V], round int64) int64 {
+func nextMinRound[I any, V Value[V]](d Definition[I, V], frc []Msg[I, V], round int64) int64 {
 	// Get all RoundChange messages with round (rj) higher than current round (ri)
-	frc, ok := getFPlus1RoundChanges(d, msgs, round)
```
corverroos, line 433: no need to filter again, the input is already Frc

dB2510 (approved):

xenowits (approved):
LGTM!

PR #508, opened 2022-05-10
core/qbft: send qcommit as decided message

corverroos:
Broadcasting individual COMMIT messages after Decided can result in some being dropped. This can postpone finality in lossy networks. 

Rather batch `Qcommit` as justification of a new message type `MsgDecided`. This ensures that if a valid justified `MsgDecided` is received, that any process will decide. This also makes sending Qcommit explicit and aligns the API with that other messages, mitigating requirement of special case handling.

Also ensure that when filtering/counting messages, that only unique sources/processes are included. Generalise this via `uniq := uniqSource` helper function.

category: refactor
ticket: #445

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -41,9 +41,6 @@ type Transport[I any, V Value[V]] struct {
 	// processes in the system (including this process).
 	Broadcast func(typ MsgType, instance I, source int64, round int64, value V, pr int64, pv V, justification []Msg[I, V])
 
-	// SendQCommit sends the commit messages to a specific process.
-	SendQCommit func(target int64, qCommit []Msg[I, V])
-
 	// Receive returns a stream of messages received
 	// from other processes in the system (including this process).
 	Receive <-chan Msg[I, V]
@@ -88,7 +85,8 @@ const (
 	MsgPrepare     MsgType = 2
 	MsgCommit      MsgType = 3
 	MsgRoundChange MsgType = 4
-	msgSentinel    MsgType = 5
+	MsgDecided     MsgType = 5
+	msgSentinel    MsgType = 6
 )
 
 func (i MsgType) Valid() bool {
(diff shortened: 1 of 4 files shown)


core/qbft/qbft.go
```diff
@@ -468,7 +469,11 @@ func isJustifiedRoundChange[I any, V Value[V]](d Definition[I, V], msg Msg[I, V]
 		return false
 	}
 
+	uniq := uniqSource[I, V]()
 	for _, prepare := range prepares {
+		if !uniq(prepare) {
```
corverroos, line 474: when counting messages, ensure that only one message per source/process is counted.

core/qbft/qbft.go
```diff
@@ -695,7 +717,7 @@ func filterRoundChange[I any, V Value[V]](msgs []Msg[I, V], round int64) []Msg[I
 func filterMsgs[I any, V Value[V]](msgs []Msg[I, V], typ MsgType, round int64, value *V, pr *int64, pv *V) []Msg[I, V] {
 	var (
 		resp []Msg[I, V]
-		dups = make(map[dedupKey]bool)
```
corverroos, line 698: we already did this in `filterMsgs` but since we do not always use this function, generalise the `uniqSource[I, V]()` helper function

core/qbft/qbft_test.go
```diff
@@ -142,6 +142,20 @@ func TestQBFT(t *testing.T) {
 			RandomRound: true,
 		})
 	})
+
+	t.Run("drop 30% const", func(t *testing.T) {
```
corverroos, line 146: this issue was uncovered when running multiple 30% lossy tests

core/qbft/qbft_test.go
```diff
@@ -204,22 +221,24 @@ func testQBFT(t *testing.T, test test) {
 			Broadcast: func(typ qbft.MsgType, instance int64, source int64, round int64, value value,
 				pr int64, pv value, justify []qbft.Msg[int64, value],
 			) {
+				if round > maxRound {
+					cancel()
+					return
+				}
+				t.Logf("%s %v => %v@%d", clock.NowStr(), source, typ, round)
```
corverroos, line 228: Add more test logs

xenowits (approved):

PR #509, opened 2022-05-10
core/qbft: refactor value to generic comparable

corverroos:
Simplify and standardise qbft generic `Value` to built-in `comparable`.

This aligns the implementation with `GoQuorum` that uses hashes of the actual values in the algo. This makes the algo simpler and probably more performant. Actual values will need to "hidden" in the `Msg` implementation.

category: refactor
ticket: #445

### DIFF
--- a/core/qbft/qbft_test.go
+++ b/core/qbft/qbft_test.go
@@ -28,9 +28,6 @@ import (
 	"github.com/obolnetwork/charon/core/qbft"
 )
 
-// Suggest running tests continuously until cancelled with Ctrl-C.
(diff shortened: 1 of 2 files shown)


core/qbft/qbft_test.go
```diff
@@ -28,9 +28,6 @@ import (
 	"github.com/obolnetwork/charon/core/qbft"
 )
 
-// Suggest running tests continuously until cancelled with Ctrl-C.
-//go:generate while go test . -count=1 -timeout=5s; do; done
```
corverroos, line 32: `while` in generate doesn't work

leolara (approved):

PR #512, opened 2022-05-10
core/qbft: add ctx and error to broadcast interface

corverroos:
Adds `ctx` and `err` to the `Transport.Broadcast` function. This allows for tracing and shutdown on fatal errors.

category: refactor
ticket: #445 
feature_flag: qbft_consensus


OisinKyne (approved):
`continue`s to `break`s is a nice tweak.

PR #514, opened 2022-05-10
core/consensus: implement qbft component

corverroos:
Implements a qbft consensus component that integrates qbft into p2p and the core workflow.

Note this isn't wired yet. Tests will also be added in subsequent PR.

category: feature
ticket: #445 
feature_flag: qbft_consensus


testutil/random.go
```diff
@@ -59,7 +60,8 @@ func RandomCorePubKey(t *testing.T) core.PubKey {
 // RandomEth2PubKey returns a random eth2 phase0 bls pubkey.
 func RandomEth2PubKey(t *testing.T) eth2p0.BLSPubKey {
 	t.Helper()
-	pubkey, _, err := tbls.Keygen()
+	random := rand.New(rand.NewSource(rand.Int63()))
```
dB2510, line 63: is this deterministic?

core/consensus/component.go
```diff
@@ -0,0 +1,217 @@
...
+
+// NewComponent returns a new consensus QBFT component.
+func NewComponent(tcpNode host.Host, peers []p2p.Peer,
+	peerIdx int64, p2pKey *ecdsa.PrivateKey,
+) *Component {
+	c := &Component{
+		tcpNode:     tcpNode,
+		peers:       peers,
+		p2pKey:      p2pKey,
+		recvBuffers: make(map[core.Duty]chan msg),
```
dB2510, line 51: why do we need receive buffers for each duty with size 100?
and why `100`?

dB2510 (approved):

testutil/random.go
corverroos: yup, since it uses math.Rand for the seed

core/consensus/transport.go
```diff
@@ -0,0 +1,188 @@
...
+			return
+		case msg := <-outerBuffer:
+			if err := validateMsg(msg); err != nil {
+				log.Warn(ctx, "Dropping invalid message", err)
+				continue
+			}
+			t.setValues(msg)
+
+			select {
+			case <-ctx.Done():
```
OisinKyne, line 123: This is in many places so I take it its intentional, but what's with a `select { }` block not counting for an extra indentation, the  `case`s aren't tabbed, convention I guess but seems weird to me.

core/consensus/transport.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+// send sends the protobuf message to the peer.
+func send(ctx context.Context, tcpNode host.Host, id peer.ID, pb proto.Message) error {
+	s, err := tcpNode.NewStream(ctx, id, protocolID)
+	if err != nil {
+		return errors.Wrap(err, "new stream")
+	}
+	defer s.Close()
```
OisinKyne, line 175: There's no scenario where s can be non-nil when err is non-nil, right? Otherwise we could have a memory leak from returning before deferring the close.

OisinKyne (commented):
Apologies I didn't get this finished yesterday. Busy day. Not all of these questions are the most insightful comments and are partly for my benefit. 🙃

Issue #515, opened 2022-05-11
Scheduler supports triggering duties at offset in slot


corverroos:
## Problem to be solved

Scheduler triggers duty at the start of each slot. Certain duties however need to triggered at offsets within the slot. We need to add support for this.

This is causing validatormock not to work with real VCs since it is triggered by scheduler and attests much earlier than other VCs, resulting in timeouts. 

## Proposed solution

Add support for triggering duties at slot offsets.

### Out of Scope

Adding support for 1/3 slot offset for DutyAttester is enough, can add other duties later.

PR #516, opened 2022-05-11
core/scheduler: trigger duty at slot offset

corverroos:
Trigger duties at an offset within each slot.

category: feature
ticket: #515

### DIFF
--- a/core/scheduler/scheduler_test.go
+++ b/core/scheduler/scheduler_test.go
@@ -20,6 +20,8 @@ import (
 	"encoding/json"
 	"flag"
 	"os"
+	"sort"
+	"sync"
 	"testing"
 	"time"
 
@@ -159,7 +161,7 @@ func TestSchedulerWait(t *testing.T) {
 				}, err
 			}
 
-			sched := scheduler.NewForT(t, clock, nil, eth2Cl)
+			sched := scheduler.NewForT(t, clock, new(delayer).Delay, nil, eth2Cl)
 			sched.Stop() // Just run wait functions, then quit.
 			require.NoError(t, sched.Run())
 			require.EqualValues(t, test.WaitSecs, clock.Since(t0).Seconds())
@@ -175,22 +177,26 @@ func TestSchedulerDuties(t *testing.T) {
 		Name     string
 		Factor   int // Determines how duties are spread per epoch
 		PropErrs int
+		Results  int
 	}{
 		{
 			// All duties grouped in first slot of epoch
-			Name:   "grouped",
-			Factor: 0,
+			Name:    "grouped",
+			Factor:  0,
+			Results: 2,
 		},
 		{
 			// All duties spread in first N slots of epoch (N is number of validators)
-			Name:   "spread",
-			Factor: 1,
+			Name:    "spread",
+			Factor:  1,
+			Results: 6,
 		},
 		{
 			// All duties spread in first N slots of epoch (except first proposer errors)
 			Name:     "spread_errors",
 			Factor:   1,
 			PropErrs: 1,
+			Results:  5,
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


core/scheduler/scheduler_test.go
```diff
@@ -312,6 +347,34 @@ func TestScheduler_GetDuty(t *testing.T) {
 	require.NoError(t, sched.Run())
 }
 
+// delayer implements scheduler.delayFunc and records the deadline and returns it immediately.
+type delayer struct {
+	mu        sync.Mutex
+	deadlines map[core.Duty]time.Time
+}
+
+func (d *delayer) Get() map[core.Duty]time.Time {
```
leolara, line 355: Is this exported method used or aimed to used outside the package?

core/scheduler/scheduler_test.go
```diff
@@ -312,6 +346,34 @@ func TestScheduler_GetDuty(t *testing.T) {
...
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	return d.deadlines
+}
+
+// Delay implements scheduler.delayFunc and records the deadline and returns it immediately.
+func (d *delayer) Delay(duty core.Duty, deadline time.Time) <-chan time.Time {
+	d.mu.Lock()
```
leolara, line 364: Is this method used or aimed to be used outside the package?

core/scheduler/scheduler_test.go
corverroos: Also since this is a `_test` package, it is **impossible** to use it from outside the package.

core/scheduler/scheduler_test.go
corverroos: see comment above

leolara (changes requested):
Please, make methods that are not for comsumtion outside the package private. If there is a reason why it is necessary to make `delay` and `get` public, let me know

core/scheduler/scheduler_test.go
leolara: When you make a method public you are documenting that the method is to be comsumed by other packages

corverroos: 1. It is impossible to import anything from a `_test.go` or `_internal_test.go` file in another package in Golang. So your second point is not valid.
2. You general approach of making _everything_ private, both types **and** fields/methods is a subjective opinion which I have already mentioned that I do not agree with. I feel making types private is sufficient to ensure encapsulation.

corverroos: > When you make a method public you are documenting that the method is to be comsumed by other packages

I do not agree with this. Only the **type** indicates whether it is used outside the package, not the methods or fields.

leolara: > It is impossible to import anything from a _test.go or _internal_test.go file in another package in Golang. So your second point is not valid.

_internal_test.go is to test private methods, you told me so yourself

leolara: > You general approach of making everything private, both types and fields/methods is a subjective opinion which I have already mentioned that I do not agree with. I feel making types private is sufficient to ensure encapsulation. Public fields/method then indicate that they are used by other things in the same package as opposed to only for internal use in the type itself.

public methods in private structs does not provide encapsulation in Go as shown in the code https://github.com/leolara/privateExperiment

leolara: here there is an example of code: https://willhaley.com/blog/private-and-public-visibility-with-go-packages/

a private struct with public method or private method

corverroos: I've documented my approach here: https://github.com/corverroos/go-visibilty

core/scheduler/scheduler_test.go
corverroos: ok made it unexported.

core/scheduler/scheduler_test.go
corverroos: ok made it unexported.

leolara (approved):

PR #517, opened 2022-05-11
core/consensus: add message signing

corverroos:
Adds message signing and verification helper functions for qbft consensus,

category: feature 
ticket: #445 
feature_flag: qbft_consensus


core/consensus/msg.go
```diff
@@ -131,4 +134,41 @@ func hashProto(msg proto.Message) ([32]byte, error) {
 	return hash, nil
 }
 
+// verifyMsgSig returns true if the message with signed by the pubkey.
```
xenowits, line 137:
```suggestion
// verifyMsgSig returns true if the message was signed by the same priv key.
```

PR #518, opened 2022-05-12
dkg: add aggregate lock hash signing skeleton

corverroos:
Adds the general steps to create an aggregate signature of the lock hash by all validators.

category: feature 
ticket: #478

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -197,11 +197,13 @@ func setupP2P(ctx context.Context, datadir string, p2pConf p2p.Config, peers []p
 func aggSignLockHash(ctx context.Context, tcpNode host.Host, peerIdx int,
 	peers []peer.ID, lock cluster.Lock, shares []share,
 ) (*bls_sig.MultiSignature, error) {
+	// Create partial signatures of lock hash
 	signedSet, err := signLockHash(lock, shares)
 	if err != nil {
 		return nil, err
 	}
 
+	// Wire some core workflow components.
 	sigChan := make(chan *bls_sig.Signature, len(lock.Validators))
 	db := parsigdb.NewMemDB(lock.Threshold)
 	exchange := parsigex.NewParSigEx(tcpNode, peerIdx, peers)
@@ -209,11 +211,13 @@ func aggSignLockHash(ctx context.Context, tcpNode host.Host, peerIdx int,
 	db.SubscribeThreshold(makeSigAgg(sigChan))
 	exchange.Subscribe(db.StoreExternal)
 
+	// Start the process by inserting the partial signatures
 	err = db.StoreInternal(ctx, core.Duty{}, signedSet)
 	if err != nil {
 		return nil, err
 	}
 
+	// Wait until all group signatures pop out.
 	var sigs []*bls_sig.Signature
 	for {
 		select {
@@ -227,6 +231,7 @@ func aggSignLockHash(ctx context.Context, tcpNode host.Host, peerIdx int,
 		}
 	}
 
+	// Aggregate all group signatures.
 	b, err := tbls.Scheme().AggregateSignatures(sigs...)
 	if err != nil {
 		return nil, errors.Wrap(err, "aggregate signature")
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/coinbase/kryptology/pkg/signatures/bls/bls_sig"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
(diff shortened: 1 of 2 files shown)


dkg/dkg.go
```diff
@@ -181,8 +192,60 @@ func setupP2P(ctx context.Context, datadir string, p2pConf p2p.Config, peers []p
...
+	if err != nil {
+		return nil, err
+	}
+
+	sigChan := make(chan *bls_sig.Signature, len(lock.Validators))
+	db := parsigdb.NewMemDB(lock.Threshold)
+	exchange := parsigex.NewParSigEx(tcpNode, peerIdx, peers)
+	db.SubscribeInternal(exchange.Broadcast)
+	db.SubscribeThreshold(makeSigAgg(sigChan))
+	exchange.Subscribe(db.StoreExternal)
```
corverroos, line 210: using core workflow components for all the heavy lifting.

Note that the generic way these components have been implemented, allows them to be used in a different use-case here in DKG.

dkg/dkg.go
```diff
@@ -181,8 +192,60 @@ func setupP2P(ctx context.Context, datadir string, p2pConf p2p.Config, peers []p
 	}, nil
 }
 
-func aggSignLockHash(_ context.Context, _ kcTransport, _ cluster.Lock) ([]byte, error) {
-	// TODO(corver): Implement lock hash signing by all DVs and aggregation.
+// aggSignLockHash returns the aggregated multi signature of the lock hash
+// signed by all the distributed validator group private keys.
+func aggSignLockHash(ctx context.Context, tcpNode host.Host, peerIdx int,
```
leolara, line 197: @corverroos please, either one line or each parameter on its own line. I think that is how I always saw it. Isn't a linter for that?

dkg/dkg.go
corverroos: that is the select function below

dkg/dkg.go
corverroos: at this point this isn't linted or specified anywhere

dkg/dkg.go
leolara: I don't understand, which line?

corverroos: these methods only wires stuff, the output is below at:
```
		    case sig := <-sigChan:
			sigs = append(sigs, sig)
		}
		if len(sigs) == len(lock.Validators) {
			break
		}
```

leolara: yes, my question if it the stuff is unwired or if they keep wired, is because this func is only called once?

corverroos: those components are pure stateless functions, so there is nothing to shutdown.

corverroos: There are no goroutines inside those components.

leolara: is this like a mini-DI container?

leolara: I am seeing that there is this events subscriptions, and I don't see the unsubscription, so I guess it is not too weird too ask. If this function is going to be run in the background in a goroutine as long charon is running I guess there is no need for unsubscription. Is that right?

dkg/dkg.go
```diff
@@ -181,8 +192,65 @@ func setupP2P(ctx context.Context, datadir string, p2pConf p2p.Config, peers []p
...
+		return nil, err
+	}
+
+	// Wire some core workflow components.
+	sigChan := make(chan *bls_sig.Signature, len(lock.Validators))
+	sigdb := parsigdb.NewMemDB(lock.Threshold)
+	exchange := parsigex.NewParSigEx(tcpNode, peerIdx, peers)
+	sigdb.SubscribeInternal(exchange.Broadcast)
+	sigdb.SubscribeThreshold(makeSigAgg(sigChan))
+	exchange.Subscribe(sigdb.StoreExternal)
```
dB2510, line 212: nit: add comments for these lines

dB2510 (approved):
LGTM

dkg/dkg.go
leolara: @dB2510 so you know the answer to this?

dB2510: yeah that's right, there's no need to unsubscribe. So basically in charon we follow functional programming paradigm where we keep our functions/methods to be stateless and only introduce statefulness whenever necessary. We don't need unsubscription because we aren't persisting anything. So when that function goes out of scope, we are done. Go's GC will take care of that.
On the other hand, we introduce statefulness for functions mainly by returning functions i.e, closures.

Am I right, @corverroos ?

corverroos: exactly right! 🚀

dkg/dkg.go
corverroos: sure will do in next PR

dkg/dkg.go
dB2510: > Is this intended for some reason? or just a temporary shortcut that will be refactored in the future?
yes, this is temporary. You can refer here for next steps: #522

leolara: ok 👍

Issue #525, opened 2022-05-13
QBFT: Do not buffer unjust messages


corverroos:
## Problem to be solved

The current QBFT implementation is buffering ALL messages.
A valid message may however be unjust.
This is either due to a bug in the code, or due to Byzantine behaviour
To err on the side of caution, it should be assumed the be Byzantine, and those messages should not be buffered.

## Proposed solution

Move buffering of messages after classification.
Note that classification assumes the message is already in the buffer, create a local `var all []Msg` slice that includes the latest message (without modifying the buffer itself).

PR #535, opened 2022-05-14
core/qbft: do not buffer unjust messages

leolara:
The current QBFT implementation is buffering ALL messages.
A valid message may however be unjust.
This is either due to a bug in the code, or due to Byzantine behaviour
To err on the side of caution, it should be assumed the be Byzantine, and those messages should not be buffered.

category: refactor
ticket: #525

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -226,8 +226,7 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 				break
 			}
 
-			justified := isJustified(d, instance, msg)
-			if !justified {
+			if !isJustified(d, instance, msg) {
 				break
 			}
 
@@ -456,7 +455,7 @@ func isJustified[I any, V comparable](d Definition[I, V], instance I, msg Msg[I,
 	case MsgDecided:
 		return isJustifiedDecided(d, msg)
 	default:
-		return false
+		panic("invalid message type")
 	}
 }
 
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -114,15 +114,12 @@ type uponRule int64
 const (
 	uponNothing uponRule = iota
 	uponJustifiedPrePrepare
-	uponUnjustPrePrepare
 	uponQuorumPrepares
 	uponQuorumCommits
-	uponUnjustRoundChange
 	uponUnjustQuorumRoundChanges
 	uponFPlus1RoundChanges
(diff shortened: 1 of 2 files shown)


core/qbft/qbft.go
```diff
@@ -229,6 +226,11 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 				break
 			}
 
+			justified := isJustified(d, instance, msg)
+			if !justified {
```
corverroos, line 230: nit: can inline `if !isJustified {`

Also, we should probably still debug log this fact:
```
if !isJustified(d, instance, msg){
  d.LogUponRule(instance, process, round, msg, "unjust")
  break
}
```

core/qbft/qbft.go
```diff
@@ -451,6 +442,24 @@ func nextMinRound[I any, V comparable](d Definition[I, V], frc []Msg[I, V], roun
...
+	case MsgPrepare:
+		return true
+	case MsgCommit:
+		return true
+	case MsgRoundChange:
+		return isJustifiedRoundChange(d, msg)
+	case MsgDecided:
+		return isJustifiedDecided(d, msg)
+	default:
+		return false
```
corverroos, line 459: So I have been using panics as sanity checks in the algo, since this case is unexpected, I would panic here: `panic("bug: invalid message type")`

corverroos (approved):

core/qbft/qbft.go
leolara: @corverroos I could add a panic, but I would consider it tech debt, we should handle all errors.

core/qbft/qbft.go
leolara: yes, I thought I was going to pass it to clasify but in the end is not necessary

core/qbft/qbft.go
corverroos: Yeah I know people avoid panics, so I am breaking the "rules". But I opted for panics for sanity checks to simplify the code and to distinguish between normal errors (like when doing networking) and sanity checks (things that fatal bugs that should never happen). It also makes the code more explicit, stating what is assumptions about the state of the world.

core/qbft/qbft.go
leolara: Done

core/qbft/qbft.go
leolara: Done

Issue #523, opened 2022-05-13
QBFT: Only trigger each upon rule once per round


corverroos:
## Problem to be solved

The QBFT paper states that each upon rules should only be triggered once per round (with the exception of `uponJustifiedDecided`.

The current implementation allows for rules to be triggered multiple times. 

Note this isn't such a big problem, since the algorithm is idempotent, the only affect is that more messages are sent over the wire then necessary.

## Proposed solution

- Similar to `dedupIn`, implement a `dedupRule` map to deduplicate upon rules per round. 
- Remember to trim that map in `trimBuffer`

PR #536, opened 2022-05-14
core/qbft: only trigger each upon rule once per round

leolara:
The QBFT paper states that each upon rules should only be triggered once per round (with the exception of uponJustifiedDecided.

The current implementation allows for rules to be triggered multiple times.

Note this isn't such a big problem, since the algorithm is idempotent, the only affect is that more messages are sent over the wire then necessary.

category: refactor
ticket: #523

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -152,10 +152,10 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 		preparedJustification []Msg[I, V]
 		qCommit               []Msg[I, V]
 		buffer                []Msg[I, V]
-		dedupIn               = make(map[dedupKey]bool)
+		dedupMsgs             = make(map[dedupKey]bool)
+		dedupRules            = make(map[uponRule]bool)
 		timerChan             <-chan time.Time
 		stopTimer             func()
-		dedupRules            = make(map[uponRule]bool)
 	)
 
 	// === Helpers ==
@@ -175,10 +175,10 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 	// bufferMsg returns true if the message is unique and was added to the buffer.
 	// It returns false if the message is a duplicate and should be discarded.
 	bufferMsg := func(msg Msg[I, V]) bool {
-		if dedupIn[key(msg)] {
+		if dedupMsgs[key(msg)] {
 			return false
 		}
-		dedupIn[key(msg)] = true
+		dedupMsgs[key(msg)] = true
 		buffer = append(buffer, msg)
 
 		return true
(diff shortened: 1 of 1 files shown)


core/qbft/qbft.go
```diff
@@ -202,6 +203,29 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 		dedupIn = dedup
 	}
 
+	isDuplicatedRule := func(rule uponRule) bool {
```
corverroos, line 206: please add comments to the functions

core/qbft/qbft.go
```diff
@@ -202,6 +203,29 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
...
+
+		return false
+	}
+
+	resetDedupRules := func() {
+		dedupRules = make(map[uponRule]bool)
+	}
+
+	changeRound := func(newRound int64) {
+		resetDedupRules()
```
corverroos, line 225: can probably inline `resetDedupRules` since it is just a single line function, and I would also move `trimBuffer` here:
```
// changeRound updates the round and clears state of the previous round
changeRound := func(newRound int64) {
    round = newRound
    dedupRules = make(map[uponRule]bool)
    trimBuffer()
}
```

core/qbft/qbft.go
```diff
@@ -202,6 +203,29 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
...
 
+	isDuplicatedRule := func(rule uponRule) bool {
+		if rule == uponJustifiedDecided {
+			return false
+		}
+		if dedupRules[rule] {
+			return true
+		}
+
+		// first time for this rule
```
corverroos, line 214: nit (also please capitalise comment sentences): `// First time for this rule in current round`

core/qbft/qbft.go
```diff
@@ -202,6 +203,29 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 		dedupIn = dedup
 	}
 
+	isDuplicatedRule := func(rule uponRule) bool {
+		if rule == uponJustifiedDecided {
```
corverroos, line 207: On second thought, I think it is **_sending_** of `qCommit` that may be triggered multiple times per round, not the **_receiving_** of `qCommit`.  Receiving it once is sufficient, since then we have decided and we are done. So I think we must remove this line.

core/qbft/qbft.go
```diff
@@ -155,6 +155,7 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 		dedupIn               = make(map[dedupKey]bool)
 		timerChan             <-chan time.Time
 		stopTimer             func()
+		dedupRules            = make(map[uponRule]bool)
```
corverroos, line 158: nit: I would put the two dedup maps next to each other. We could also rename `dedupIn` to `dedupMsgs`

corverroos (approved):

core/qbft/qbft.go
leolara: Done

core/qbft/qbft.go
leolara: Done

core/qbft/qbft.go
leolara: Done

core/qbft/qbft.go
leolara: Done

core/qbft/qbft.go
leolara: Done

PR #540, opened 2022-05-15
core/qbft: fix justified preprepare issue

corverroos:
Fixes two related issues with justified preprepare messages:
 - it comes down to `highestPrepared` that didn't support multiple pvs at the highest round (see test message).
 - When getting pv to send with preprepare, we cannot just get it by only looking at qrc in the justification, since it might contain different pvs at the highest round. We should get it from the prepares included (or not) in the justification.
 - The same issue also triggered `unjust preprepare`, so refactored isJustifiedPreprepare to also use prepares to get pv to check, not qrc. 

This issue was triggered by running the test suite thousands of times.

category: bug
ticket: none

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -550,7 +550,7 @@ func containsJustifiedQrc[I any, V comparable](d Definition[I, V], justification
 		if rc.PreparedRound() > pr {
 			return zeroVal[V](), false
 		}
-		// Ensure one ROUND-CHANGE with pr and pv
+		// Ensure at least one ROUND-CHANGE with pr and pv
 		if rc.PreparedRound() == pr && rc.PreparedValue() == pv {
 			found = true
 		}
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -386,29 +386,6 @@ func classify[I any, V comparable](d Definition[I, V], instance I, round, proces
 	return uponNothing, nil
 }
 
-// highestPrepared implements algorithm 4:5 and returns
-// the highest prepared round (and pv) from the set of quorum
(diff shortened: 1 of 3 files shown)


core/qbft/qbft.go
```diff
@@ -389,29 +386,6 @@ func classify[I any, V comparable](d Definition[I, V], instance I, round, proces
...
-	}
-
-	var (
-		pr int64
-		pv V
-	)
-	for _, msg := range qrc {
-		if pr < msg.PreparedRound() {
-			pr = msg.PreparedRound()
-			pv = msg.PreparedValue()
```
corverroos, line 408: there can be different pv at highest pr.

leolara (approved):

core/qbft/qbft_internal_test.go
```diff
@@ -463,20 +460,46 @@ func (m msg) PreparedValue() int64 {
...
+	const (
+		n        = 4
+		instance = 1
+	)
+
+	// Preprepare with identical Pr but different Pvs.
+	preprepare := msg{msgType: 1, instance: 1, peerIdx: 3, round: 6, value: 2, pr: 0, pv: 0, justify: []msg{
+		{msgType: 4, instance: 1, peerIdx: 2, round: 6, value: 0, pr: 2, pv: 3},
+		{msgType: 4, instance: 1, peerIdx: 3, round: 6, value: 0, pr: 2, pv: 3},
+		{msgType: 4, instance: 1, peerIdx: 1, round: 6, value: 0, pr: 2, pv: 2},
```
corverroos, line 482: two ROUND-CHANGEs with identical pr, but different pv

leolara (approved):

Issue #534, opened 2022-05-14
Rename core workflow


corverroos:
## Problem to be solved

We currently use the term "aggregate" in many names in the core workflow. This is however incorrect, since we are referring to threshold BLS signature "combination" (M-of-N multisig). "Aggregation" actually means normal BLs signature aggregation, as used in ETH2. 

We therefore need to rename all "agg" components.

We might also take this opportunity to improve naming of a few other entities as well,:
 -  `core.FetchArgs` was originally only used as input arguments to `fetcher`, so `FetchArgs` made sense. But this is now also used in `validatorapi`. Since `FetchArgs` define the "duty" to be performed, we are suggesting "DutyDefinition" as the new name.
  
**The following is a list of all OLD core workflow names, the groups they form, and contexts they are used in:**

---
 - `FetchArgs` - fetch arguments - duty definition, used to fetch duty data
 - `ParSignedData` - partial signature data - duty data containing partial bls signatures
 - `AggSignedData` - aggregated signature data - duty data containing aggregated bls signatures
 - `UnsignedData` - unsigned data - duty data to be signed
 ---
 - `FetchArgSet` - a set of fetch arguments
 - `ParSignedDataSet` - a set of partial signature data
 - `AggSignedDataSet` - a set of aggregated signature data
 - `UnsignedDataSet` - a set of unsigned data
 --- 
 - `ParSigDB` - partial signature database
 - `ParSigEx` - partial signature exchange
 - `SigAgg` - signature aggregator
 - `AggSigDB` - aggregated signature database
 ---
 Concatenated method names:
 -  `Decode*` - e.g. `DecodeAggSignedDataSet`
 -  `Encode*` - e.g. `EcnodeUnsignedDataSet`
 - `*ToProto` - e.g `ParSignedDataToProto`
 - `*FromProto` - e.g. `AggSignedDataFromProto`
 ---
 
## Proposed solution

The suggested renaming approach is:
 - Partial signatures and data signed with them get a `Share` prefix.
 - t-BLS "Combined" signatures and data signed with them get a `Group` prefix.
 - So `shares` are "combined" into a `group`.
 
**The following is a list of all NEW core workflow names, the groups they form, and contexts they are used in:**

---
 - `FetchArgs` -> `DutyDefinition` - duty definition, used to fetch duty data
 - `ParSignedData` -> `ShareSignedData`- partial signatures and data containing partial signatures 
[...]

PR #541, opened 2022-05-15
core: rename workflow

leolara:
We currently use the term "aggregate" in many names in the core workflow. This is however incorrect, since we are referring to threshold BLS signature "combination" (M-of-N multisig). "Aggregation" actually means normal BLs signature aggregation, as used in ETH2.

We therefore need to rename all "agg" components.

We might also take this opportunity to improve naming of a few other entities as well,

category: refactor
ticket: #534


dB2510 (changes requested):
I don't think we need these changes. I have explained the reasons here: https://github.com/ObolNetwork/charon/issues/534#issuecomment-1127002293

app/app.go
```diff
@@ -89,16 +89,16 @@ type TestConfig struct {
 	DisablePing bool
 	// PingCallback is called when a ping was completed to a peer.
 	PingCallback func(peer.ID)
-	// ParSigExFunc provides an in-memory partial signature exchange.
-	ParSigExFunc func() core.ParSigEx
+	// ParSigExchangeFunc provides an in-memory partial signature exchange.
```
corverroos, line 92: As per ticket, this should be `ShareSigExchange(Func)`

corverroos (commented):
Yeah, this is a massive refactor, we shouldn't fast track this, lets have more discussions and get everyones buy-in before we actually implement the code.

leolara:
@corverroos @dB2510 I think it is easier to merge it and then rename again if necessary.

the only dicussion is about combine term?

app/app.go
leolara: Done

leolara:
@OisinKyne @dB2510 I see this is much more contentious than I thought. I close the PR. I never take PRs personally

Issue #522, opened 2022-05-13
Add generic threshold aggregation functionality to DKG


corverroos:
## Problem to be solved

At this point, DKG supports distributed aggregation of lock hash partial signatures.
There are however additional aggregation signatures that should be created as part of the DKG command.

This includes:
 - Threshold aggregate signature of deposit data message roots (one for each validator).
 - Maybe other things...?
 
 The current signing logic:
  - is actually broken, since all peers should complete wiring BEFORE any peer sends a message (otherwise those messages are dropped).
  - only supports single lock hash signing and aggregation by all validators.

## Proposed solution

- Refactor DKG to support "general distributed aggregation of partial signatures".
- Do this by extracting a `aggregator` type/function.
- It should be wired and constructed before any peers send signing messages (before `switch def.DKGAlgorithm` step).
- It should provide a function `Aggregate(core.ParSignedDataSet) (map[core.PubKey]core.AggSignedData, error)`
- It should use the `aggSignLockHash` logic at its core (including parsigdb and parsigex and a custom sigAgg function).
- This `Aggregate` function will then be called after creating a core.ParSignedDataSet for the lock hash (same hash for all pubkeys)

### Out of Scope

- This `Aggregate` function will later be after creating a core.ParSignedDataSet for the deposit datas hash (different hashes for each pubkey).

PR #545, opened 2022-05-16
dkg: add minimal exchanger

dB2510:
Adds minimal exchanger struct with methods to get []core.ParSignedDataSet.

category: feature
ticket: #522


dkg/dkg.go
```diff
@@ -48,6 +48,52 @@ type Config struct {
 	TestSigning bool
 }
 
+type exchanger struct {
```
xenowits, line 51: please add comment as the name's not self explanatory

dkg/dkg.go
```diff
@@ -48,6 +48,52 @@ type Config struct {
...
+	var sets []core.ParSignedDataSet
+	sets = append(sets, set)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, err
+		case peerSet := <-e.setChan:
+			sets = append(sets, peerSet)
+		}
+		if len(sets) == cap(e.setChan) {
```
dB2510, line 83: we are done when we have ParSignedDataSet from all the peers including the current one

dkg/dkg.go
```diff
@@ -242,7 +290,7 @@ func aggSignLockHash(ctx context.Context, tcpNode host.Host, nodeIdx cluster.Nod
 	// Aggregate all group signatures.
 	b, err := tbls.Scheme().AggregateSignatures(sigs...)
 	if err != nil {
-		return nil, errors.Wrap(err, "aggregate signature")
+		return nil, errors.Wrap(err, "exchangeSigs signature")
```
xenowits, line 293: nit
```suggestion
		return nil, errors.Wrap(err, "exchange sigs signature")
```

xenowits (approved):

dkg/dkg.go
xenowits: nit: can we have the same as a comment?

dkg/dkg.go
dB2510: this got committed by mistake
reverting

dkg/dkg.go
```diff
@@ -48,6 +48,55 @@ type Config struct {
...
+		return nil, err
+	}
+
+	var sets []core.ParSignedDataSet
+	sets = append(sets, set)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, err
+		case peerSet := <-e.setChan:
```
corverroos, line 81: how are you going to distinguish between lockhash and depositdata signatures...? (core workflow uses dutyType to distinguish, maybe do the same, by adding local duty types here `type dutyLock core.DutyType = 101`)

corverroos: and you probably need to do some fanout thing, where you have a router that routes signatures to the right duty type receive channel. 

Remember, different peers can send signatures at different times, some before this node and others after this node, and some might send lockhashes when this node is listenting for depositdatas...

corverroos: timing in distributed system is hard

dkg/dkg.go
```diff
@@ -48,6 +48,55 @@ type Config struct {
...
+}
+
+func (e *exchanger) exchange(ctx context.Context, set core.ParSignedDataSet) ([]core.ParSignedDataSet, error) {
+	err := e.sigex.Broadcast(ctx, core.Duty{}, set)
+	if err != nil {
+		return nil, err
+	}
+
+	var sets []core.ParSignedDataSet
+	sets = append(sets, set)
```
corverroos, line 76: can inline: `sets := []core.ParSignedDataSet{set}`

dkg/dkg.go
```diff
@@ -48,6 +48,55 @@ type Config struct {
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, err
+		case peerSet := <-e.setChan:
+			sets = append(sets, peerSet)
+		}
+
+		// We are done when we have ParSignedDataSet from all the peers including the current one
+		if len(sets) == cap(e.setChan) {
```
corverroos, line 86: I would just compare against length of peers rather. Also remember that peers might be send duplicate messages by mistake (network retries, bugs, byzantine), so I would dedup

app/featureset/featureset.go
```diff
@@ -38,12 +38,15 @@ type Feature string
 const (
 	// QBFTConsensus introduces qbft consensus, see https://github.com/ObolNetwork/charon/issues/445.
 	QBFTConsensus Feature = "qbft_consensus"
+	// DKG introduces Frost DKG algorithm into charon.
+	DKG Feature = "dkg"
```
corverroos, line 42: if frost, then probably name `frost_dkg`. Also, where do you plan to put the feature flag to enable/disable the feature?

dkg/dkg.go
```diff
@@ -83,6 +129,8 @@ func Run(ctx context.Context, conf Config) error {
 	}
 	clusterID := fmt.Sprintf("%x", defHash[:])
 
+	_ = newExchanger(def.NumValidators, tcpNode, nodeIdx.PeerIdx, nil)
```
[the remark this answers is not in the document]
corverroos: I would just add `nolint` rather

dkg/dkg.go
```diff
@@ -48,6 +48,55 @@ type Config struct {
...
+// exchanger is responsible for exchanging partial signatures between peers on libp2p.
+type exchanger struct {
+	setChan chan core.ParSignedDataSet
+	sigex   *parsigex.ParSigEx
+}
+
+func newExchanger(tcpNode host.Host, peerIdx int, peers []peer.ID) exchanger {
+	ex := exchanger{
+		sigex:   parsigex.NewParSigEx(tcpNode, peerIdx, peers),
+		setChan: make(chan core.ParSignedDataSet, len(peers)),
```
corverroos, line 60: not sure about `len(peers)`, since:
  - peers can send duplicate messages
  - peers can send different types of messages (lock hash AND deposit datas)

dkg/dkg.go
```diff
@@ -48,6 +48,55 @@ type Config struct {
...
+	}
+
+	// Wiring core workflow components
+	ex.sigex.Subscribe(ex.getSets)
+
+	return ex
+}
+
+func (e *exchanger) exchange(ctx context.Context, set core.ParSignedDataSet) ([]core.ParSignedDataSet, error) {
```
corverroos, line 69: godoc FTW. it isn't obvious what `[]core.ParSignedDataSet` is...

corverroos:
I just added no merge since the PR was approved but I think there are some work still to be done 😄

app/featureset/featureset.go
dB2510: how feature flags marked as alpha are enabled/disabled

dkg/dkg.go
dB2510: so what should be the ideal length of the buffer?

dkg/dkg.go
dB2510: you mean to be specific about whether it is lockhash or deposit data ?

dkg/dkg.go
dB2510: yeah right

app/featureset/featureset.go
```diff
@@ -38,12 +38,15 @@ type Feature string
 const (
 	// QBFTConsensus introduces qbft consensus, see https://github.com/ObolNetwork/charon/issues/445.
 	QBFTConsensus Feature = "qbft_consensus"
+	// FrostDKG introduces Frost DKG algorithm into charon.
+	FrostDKG Feature = "frost_dkg"
```
corverroos, line 42: My comment might have gotten lost, but where are you going to put this feature flag to switch stuff on and off? Is we are not going to use a feature flag in the code, then there is no need to add something here.

dkg/dkg.go
```diff
@@ -38,6 +38,11 @@ import (
 	"github.com/obolnetwork/charon/tbls/tblsconv"
 )
 
+const (
+	DutyLock        core.DutyType = 101
```
corverroos, line 42: Add comments please

dkg/dkg.go
```diff
@@ -48,6 +53,104 @@ type Config struct {
...
+// signatures of the group according to public key of each DV.
+func (e *exchanger) exchange(ctx context.Context, duty core.DutyType, set core.ParSignedDataSet) ([]core.ParSignedDataSet, error) {
+	// Start the process by storing current peer's ParSignedDataSet
+	err := e.sigdb.StoreInternal(ctx, core.Duty{Type: duty}, set)
+	if err != nil {
+		return nil, err
+	}
+
+	// len(sets) is numPeers + 1 because of shareIdx being 1-indexed
+	sets := make([]core.ParSignedDataSet, e.numPeers+1)
```
corverroos, line 99: I would not use slices that are 1-indexed. I would rather:
 - use map: `map[shareIdx]core.ParSignedDataSet`
 - or 0-indexed slice: [peerIdx]core.ParSignedDataSet

dkg/dkg.go
```diff
@@ -48,6 +53,104 @@ type Config struct {
...
+	// remaining is set to number of DVs because we are getting response with respect to each DV
+	remaining := e.numVals
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case peerSet := <-e.sigChan:
+			switch peerSet.duty {
+			case DutyLock, DutyDepositData:
```
corverroos, line 108: avoid nesting if possible:
```
if peerSet.duty != duty {
  continue
}
```

dkg/dkg.go
corverroos: This would also allow you to count the length of this thing, as opposed to keeping a separate counter (remaining)

dkg/dkg.go
```diff
@@ -48,6 +53,104 @@ type Config struct {
...
+	numPeers int
+	numVals  int
+}
+
+func newExchanger(tcpNode host.Host, peerIdx int, peers []peer.ID, vals int) *exchanger {
+	ex := &exchanger{
+		// threshold is len(peers) to wait until we get all the partial sigs from all the peers per DV
+		sigdb:    parsigdb.NewMemDB(len(peers)),
+		sigex:    parsigex.NewParSigEx(tcpNode, peerIdx, peers),
+		sigChan:  make(chan sigData, vals),
```
corverroos, line 76: this can be a blocking channel (len==0) since the `exchange` goroutine is reading from the channel when the parsigdb threshold is reached

dkg/dkg_internal_test.go
```diff
@@ -0,0 +1,111 @@
...
+		pubkeys[i] = testutil.RandomCorePubKey(t)
+	}
+
+	// Expected data is what is desired at the end of exchange
+	expectedData := make([]core.ParSignedDataSet, nodes+1)
+	for i := 1; i <= nodes; i++ {
+		set := make(core.ParSignedDataSet)
+		for j := 0; j < dvs; j++ {
+			set[pubkeys[j]] = core.ParSignedData{
+				Data:      nil,
```
corverroos, line 53: no need to set zero values, can omit

dkg/dkg_internal_test.go
```diff
@@ -0,0 +1,111 @@
...
+			Addrs: h.Addrs(),
+		}
+		hostsInfo = append(hostsInfo, info)
+		peers = append(peers, h.ID())
+		hosts = append(hosts, h)
+	}
+
+	// Connect each host with its peers
+	for i := 0; i < nodes; i++ {
+		for k := 0; k < nodes; k++ {
```
corverroos, line 82: usually it is i and j 😉

dkg/dkg.go
```diff
@@ -48,6 +53,104 @@ type Config struct {
 	TestSigning bool
 }
 
+type sigData struct {
```
corverroos, line 56: godoc please

dkg/dkg.go
```diff
@@ -38,6 +38,11 @@ import (
 	"github.com/obolnetwork/charon/tbls/tblsconv"
 )
 
+const (
+	DutyLock        core.DutyType = 101
+	DutyDepositData core.DutyType = 202
```
corverroos, line 43: do not export this

corverroos: this should probably be `101 and 102`

Also add a comment that this cannot change since it breaks backwards compatibility.

corverroos (approved):

leolara:
@dB2510 a nice thing would be to put in a exchanger.go the exchanger component and its methods and things that are only used by exchanger and that are small, leaving the rest in dkg.go

dB2510:
> @dB2510 a nice thing would be to put in a exchanger.go the exchanger component and its methods and things that are only used by exchanger and that are small, leaving the rest in dkg.go

yeah i was also thinking the same

dkg/dkg.go
```diff
@@ -48,6 +56,86 @@ type Config struct {
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case peerSet := <-e.sigChan:
+			if duty != peerSet.duty {
+				// Do nothing if duty doesn't match
+				continue
+			}
+
+			switch peerSet.duty {
+			case dutyLock, dutyDepositData:
```
corverroos, line 112: you already checked duty above, so no need to switch again, can remove the switch completely

Issue #549, opened 2022-05-16
cmd: output deposit data in charon create cluster


corverroos:
## Problem to be solved

For Devnet 2 (or maybe 1) we need `charon create cluster` to output deposit datas so customers can activate those validators.

## Proposed solution

Create deposit data file for the newly created DVs

PR #554, opened 2022-05-17
deposit: save deposit data in charon create cluster

xenowits:
Output `deposit-data.json` files for each DV when running `charon create cluster`.

category: feature
ticket: #549


eth2util/deposit/deposit.go
```diff
@@ -105,6 +105,27 @@ func MarshalDepositData(pubkey eth2p0.BLSPubKey, withdrawalAddr string, network
 	return bytes, nil
 }
 
+// MarshalDepositDatas serializes a list of deposit data into a single file.
+func MarshalDepositDatas(depositDatas [][]byte) ([]byte, error) {
```
xenowits, line 109: is it fair to accept a list of serialized deposit datas and again deserialize the whole list here?

cmd/createcluster.go
```diff
@@ -191,6 +194,47 @@ func runCreateCluster(w io.Writer, conf clusterConfig) error {
 		}
 	}
 
+	// TODO(xenowits): add flag to specify the number of distributed validators in a cluster
```
dB2510, line 197: This TODO is already there on line 311

cmd/createcluster.go
```diff
@@ -84,6 +85,7 @@ type clusterConfig struct {
 	ClusterDir string
 	Clean      bool
 	NumNodes   int
+	NumDVs     int
```
dB2510, line 88: I think we shouldn't introduce this field now when we are not having the flag for it.
Either introduce the flag and modify everything but that would make this PR big or just hardcode this with 1 at the time of creating deposit data and take care of this seperately in another PR

cmd/createcluster.go
```diff
@@ -191,6 +194,47 @@ func runCreateCluster(w io.Writer, conf clusterConfig) error {
 		}
 	}
 
+	// TODO(xenowits): add flag to specify the number of distributed validators in a cluster
+	// Currently, we assume that we create a cluster of ONLY 1 Distributed Validator
+	var depositDatas [][]byte
+	for i := 0; i < conf.NumDVs+1; i++ { // Note that default value of conf.NumDVs is 0
```
dB2510, line 200: As suggested above remove this for loop and just create a deposit data for single DV.
And add a TODO to add deposit datas when the flag is introduced.

cmd/createcluster.go
```diff
@@ -349,6 +414,8 @@ func writeOutput(out io.Writer, conf clusterConfig) {
 	_, _ = sb.WriteString("\n")
 	_, _ = sb.WriteString(strings.TrimSuffix(conf.ClusterDir, "/") + "/\n")
 	_, _ = sb.WriteString("├─ manifest.json\tCluster manifest defines the cluster; used by all nodes\n")
+	_, _ = sb.WriteString("├─ deposit-data.json\tDeposit data file used to activate the Distributed Validator\n")
```
dB2510, line 417:
```suggestion
	_, _ = sb.WriteString("├─ deposit-data.json\tDeposit data file is used to activate the Distributed Validator on DV launchpad\n")
```

cmd/createcluster.go
xenowits: yeah, let's not have the field for now then

eth2util/deposit/deposit.go
xenowits: removed this function in https://github.com/ObolNetwork/charon/pull/557

cmd/createcluster.go
```diff
@@ -191,6 +197,45 @@ func runCreateCluster(w io.Writer, conf clusterConfig) error {
...
+		if err != nil {
+			return errors.Wrap(err, "secret to pubkey")
+		}
+
+		pk, err := tblsconv.KeyToETH2(pubkey) // Group pubkey
+		if err != nil {
+			return err
+		}
+
+		withdrawalAddr := "0xc0404ed740a69d11201f5ed297c5732f562c6e4e"
```
xenowits, line 213: these fields to be replaced by flags in https://github.com/ObolNetwork/charon/pull/556

cmd/createcluster.go
```diff
@@ -191,6 +197,45 @@ func runCreateCluster(w io.Writer, conf clusterConfig) error {
...
 
+	var depositDatas [][]byte
+	for i := 0; i < numDVs; i++ {
+		sk := secrets[i] // Group secret key for this DV
+		pubkey, err := sk.GetPublicKey()
+		if err != nil {
+			return errors.Wrap(err, "secret to pubkey")
+		}
+
+		pk, err := tblsconv.KeyToETH2(pubkey) // Group pubkey
```
OisinKyne, line 208: Nit: Lots of people use `pk` to refer to private key rather than public key.

OisinKyne: ... Now I spot sk for secret key, maybe its fine as is, not sure what we use in the rest of our codebase.

cmd/createcluster.go
```diff
@@ -264,14 +310,34 @@ func getKeys(conf clusterConfig) ([]*bls_sig.SecretKey, error) {
...
+	bytes, err := deposit.MarshalDepositDatas(b)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return []*bls_sig.SecretKey{secret}, nil
+	err = os.WriteFile(depositPath, bytes, 0o400) // read-only
+	if err != nil {
+		return errors.Wrap(err, "write deposit data")
```
OisinKyne, line 337: Nit: `Failed to write deposit data to ${depositPath}`

OisinKyne (approved):

cmd/createcluster.go
```diff
@@ -191,6 +197,45 @@ func runCreateCluster(w io.Writer, conf clusterConfig) error {
 		}
 	}
 
+	var depositDatas [][]byte
+	for i := 0; i < numDVs; i++ {
+		sk := secrets[i] // Group secret key for this DV
```
dB2510, line 202:
```suggestion
		sk := secrets[i] // Secret key for this DV
```
can be confusing with Group concepts from cryptography 😅

cmd/createcluster.go
dB2510, line 208:
```suggestion
		pk, err := tblsconv.KeyToETH2(pubkey)
```

dB2510 (approved):
LGTM

leolara (approved):
good cose style

cmd/createcluster.go
xenowits: good point, i tend to restrict to using `sk` for secret keys and `pubkey` for public keys.

cmd/createcluster.go
xenowits: abstract..

PR #557, opened 2022-05-18
deposit: marshal array of deposits

xenowits:
* Add tests for testing `MarshalDepositDatas`. 
* The deposit data created by the likes of `staking-deposit-cli` consist of an array of deposit datas, one for each validator. For us, it means one for each DV.
* It also contains an additional field, the `deposit_cli_version`.
* Adds code to generate and test a deposit data file that can be used in og launchpad.
* Modify `createcluster` to use the updated `MarshalDepositData` function.

category: feature
ticket: #554


dB2510 (approved):
LGTM

Issue #527, opened 2022-05-13
app: refactor charon run to use cluster lock instead of manifest


corverroos:
## Problem to be solved

We are moving away from the manifest file to a cluster lock file. This has been integrated in the DKG command, but not into charon run.

## Proposed solution

Delete `app.Manifest`.
Refactor the codebase to the cluster lock file.
Migrate any missing convenience functions from the manifest to the lock file.

### Out of Scope

Ignore other commands than Run for now.
This might mean we cannot delete the manifest yet.

Issue #528, opened 2022-05-13
cmd: refactor charon create cluster to generate a cluster lock file


corverroos:
## Problem to be solved

We are migrating away from the manifest file to the cluster lock file.
This has been done for the charon dkg command, but not for the charon create cluster command.

## Proposed solution

Generate a cluster lock file, instead of generating a manifest file.

PR #562, opened 2022-05-19
app: replace manifest with lock

corverroos:
Replaces the manifest with cluster lock.

category: refactor 
ticket: #527


cmd/createcluster.go
```diff
@@ -379,21 +377,8 @@ func writeDepositData(config clusterConfig, pubkeys []eth2p0.BLSPubKey, msgSigs
...
-		return errors.Wrap(err, "json marshal manifest")
-	}
-
-	manifestPath := path.Join(config.ClusterDir, "manifest.json")
-	if err = os.WriteFile(manifestPath, manifestJSON, 0o600); err != nil {
-		return errors.Wrap(err, "write manifest.json")
-	}
-
+func writeLock(_ clusterConfig, _ []tbls.TSS, _ []p2p.Peer) error {
+	// TODO(corver): Create lock file.
```
corverroos, line 381: @xenowits this is for you #528

app/app_test.go
```diff
@@ -271,31 +272,37 @@ func startBootnode(ctx context.Context, t *testing.T) (string, <-chan error) {
 }
 
 // tcpAddrFromENR returns the "<ip4>:<tcp>" address stored in the ENR.
-func tcpAddrFromENR(t *testing.T, record enr.Record) string {
+func tcpAddrFromENR(t *testing.T, record string) string {
 	t.Helper()
 
+	r, err := p2p.DecodeENR(record)
```
leolara, line 278: I would suggest a more descriptive name than 'r'

p2p/enr_test.go
```diff
@@ -20,20 +20,21 @@ import (
...
 	"github.com/obolnetwork/charon/p2p"
 )
 
 func TestDecodeENR(t *testing.T) {
-	manifest, _, _ := app.NewClusterForT(t, 1, 3, 4, 0)
+	lock, _, _ := cluster.NewForT(t, 1, 3, 4, 0)
 
-	for _, p := range manifest.Peers {
-		enrStr, err := p2p.EncodeENR(p.ENR)
+	for _, o := range lock.Operators {
```
leolara, line 30: Perhaps operator instead of "o"

cluster/distvalidator.go
```diff
@@ -58,3 +64,13 @@ func (v DistValidator) HashTreeRootWith(hh *ssz.Hasher) error {
...
 	return nil
 }
+
+// PublicKey returns the validator group public key.
+func (v DistValidator) PublicKey() (*bls_sig.PublicKey, error) {
+	return tblsconv.KeyFromCore(core.PubKey(v.PubKey))
+}
+
+// PublicShare returns a peer's threshold BLS public share.
+func (v DistValidator) PublicShare(peerIdx int) (*bls_sig.PublicKey, error) {
```
leolara, line 74: If we have this method perhaps we don't need PubShares to be public

corverroos: it is a DTO/value so fields are public

p2p/enr_test.go
corverroos: I think short names for are fine is the scope is super small

dB2510 (approved):

app/app.go
```diff
@@ -227,29 +237,33 @@ func wireP2P(ctx context.Context, life *lifecycle.Manager, conf Config, manifest
...
+		}
 
-		corePubkey, err := tblsconv.KeyToCore(dv.PublicKey())
+		corePubkey, err := tblsconv.KeyToCore(pubkey)
 		if err != nil {
 			return err
 		}
 
-		pubkey, err := tblsconv.KeyToETH2(dv.PublicKey())
+		pk, err := tblsconv.KeyToETH2(pubkey)
```
xenowits, line 261: should colocate this `pk`. code may be confusing because there are two pubkeys: `pubkey` and `pk`.

```go
pk, err := tblsconv.KeyToETH2(pubkey)
if err != nil {
   return err
}

pubkeys = append(pubkeys, pk)
```

leolara: think about what makes them different and name them with longer identifiers that represent that

Issue #568, opened 2022-05-20
Implement compose to debug local docker-compose clusters


corverroos:
## Problem to be solved

Since charon-docker-compose has been repurposed to be more customer/user facing, advanced features has been removed from it. It therefore cannot do things like "run locally built binary", "DKG using docker compose", or "mock validators", or "simnet".

As charon core developers, we need a way to test all the features of charon locally using docker-compose.
  
## Proposed solution

- Write a go program called `compose`
- It is the successor of  the "old" charon-docker-compose for developers.
- The aim is to be able to test charon locally using docker-compose.
- It is different from the "new" charon-docker-compose in that it supports "advanced" features like build locally binary, split keys, docker-compose DKG, mock validators, etc.
- It is aimed at Charon core developers, not external customers or users. 
- The "old" charon-docker-compose used a makefile to configure the cluster (make create-cluster, make split-keys, make disable-simnet, etc)
- This is fine, but since I want to write this tool in Golang, we can improve on the UX a lot.
-  I propose to serve a simple web UI form to configure the cluster 
- The tool would serve this ( via `compose define`)
- On submission, the tool will create a  `compose.yml` file which contains the cluster definition and all the other stuff in the webui form. 
- The tool will also be configurable by direct manipulation of the `compose.yml`
- The tool will live in the charon repo at package `testutil/compose`

Example flow:
 - compose define : populate form and submit, creates compose.yml which contains the cluster definition and other config. 
 - compose lock && docker-compose up: creates docker-compose.yml for key generation and then creates keys and cluster lock
 - compose run && docker-compose up : creates docker-compose.yml for running the cluster and runs it.

corverroos:
Example web ui 
<img width="379" alt="image" src="https://user-images.githubusercontent.com/29249923/169480764-0b5f3e17-dd47-47fe-9aaf-ba496e968b5a.png">

PR #571, opened 2022-05-20
testutil/compose: add run command

corverroos:
Implements the `compose run` command that generates a docker-compose.yml for running a cluster.

category: feature
ticket: #568

PR #570, opened 2022-05-20
testutil/compose: add lock command

corverroos:
Adds the first version of the lock command that creates a docker-compose file to run `charon create cluster` to generate keys and cluster lock file.

category: feature
ticket: #568

PR #575, opened 2022-05-20
testutil/compose: add dkg lock

corverroos:
Adds support for DKG key generation to compose lock command.

category: feature
ticket: #568


corverroos:
This PR is stacked on:
 - #571 which is stacked on
 - #570

dB2510 (approved):

Issue #583, opened 2022-05-23
cmd: ensure valid network during create dkg


corverroos:
## Problem to be solved

The `charon create dkg` command doesn't validate network/forkversion, not does it have a default. This results in empty values by default.

## Proposed solution

- Add a flag to define the network in the `charon create dkg` command.
- Add a default value (mainnet?)
- Verify the network when provided.

PR #598, opened 2022-05-23
cmd: ensures valid network during create dkg

dB2510:
Ensures valid network during create dkg using network flag which is then used to get fork version to create cluster definition.

category: bug
ticket: #583

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -43,7 +43,6 @@ type Config struct {
 	DataDir string
 	P2P     p2p.Config
 	Log     log.Config
-	Network string
 
 	TestDef     *cluster.Definition
 	TestSigning bool
@@ -63,6 +62,11 @@ func Run(ctx context.Context, conf Config) (err error) {
 	}
 
 	def, err := loadDefinition(conf)
+	if err != nil {
+		return err
+	}
+
+	network, err := forkVersionToNetwork(def.ForkVersion)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


dkg/dkg.go
```diff
@@ -67,6 +66,11 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	network, err := forkVersionToNetwork(def.ForkVersion)
```
corverroos, line 69: I think we should do it the other way round, provide a network and map it to fork version. Requiring people to input fork versions is not a great UX.

xenowits: agreed

dB2510: Done!
Modified in createdkg command flags

xenowits (approved):
LGTM🚀

dkg/dkg.go
```diff
@@ -565,3 +569,20 @@ func waitConnect(ctx context.Context, tcpNode host.Host, p peer.ID) (time.Durati
 
 	return 0, false
 }
+
+func forkVersionToNetwork(forkVersion string) (string, error) {
```
xenowits, line 573: @corverroos do you think functions like `forkVersionToNetwork()` and `networkToForkVersion()` should be separated out somewhere?

dB2510: This is very small piece of code, can be duplicated imo

corverroos: Indeed, also remember the rule of 3, only once it is used in 3 places or more, is it worth extracting.

corverroos (approved):

PR #600, opened 2022-05-24
testutil/compose: introduce new and auto commands

corverroos:
Refactors compose:
 - Decouple `new` from `define`: `new` only creates config.json file, `define` is not identical to `lock` and `run`; it only creates docker-compose files.
 - This adds support for using existing config files, and generating clusters from them
 - Add convenience function `compose auto` that runs `compose define && compose lock && compose run`, since this will be the most common use-case.
 
 Result:
 ```
 compose new --keygen==dkg && compose auto
 ``` 

category: refactor 
ticket: #568


dB2510 (approved):
LGTM

Issue #602, opened 2022-05-24
Integrate QBFT into charon run


corverroos:
## Problem to be solved

We have a QBFT implementation, but it isn't wired yet.

## Proposed solution

Integrate it.

PR #603, opened 2022-05-24
core/consensus: integrate qbft

corverroos:
Integrates consensus QBFT component into charon run.

Also adds unit test for consensus component.

category: feature 
ticket: #602 
feature_flag: qbft_consensus

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -50,6 +50,7 @@ import (
 	"github.com/obolnetwork/charon/core"
 	"github.com/obolnetwork/charon/core/aggsigdb"
 	"github.com/obolnetwork/charon/core/bcast"
+	"github.com/obolnetwork/charon/core/consensus"
 	"github.com/obolnetwork/charon/core/dutydb"
 	"github.com/obolnetwork/charon/core/fetcher"
 	"github.com/obolnetwork/charon/core/leadercast"
@@ -151,7 +152,16 @@ func Run(ctx context.Context, conf Config) (err error) {
 	}
 	lockHashHex := hex.EncodeToString(lockHash[:])[:7]
 
-	tcpNode, localEnode, err := wireP2P(ctx, life, conf, lock)
+	p2pKey := conf.TestConfig.P2PKey
+	if p2pKey == nil {
+		var err error
+		p2pKey, err = p2p.LoadPrivKey(conf.DataDir)
+		if err != nil {
+			return err
+		}
+	}
+
+	tcpNode, localEnode, err := wireP2P(ctx, life, conf, lock, p2pKey)
 	if err != nil {
 		return err
 	}
@@ -182,7 +192,7 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 	wireMonitoringAPI(life, conf.MonitoringAddr, localEnode)
 
-	if err := wireCoreWorkflow(ctx, life, conf, lock, nodeIdx, tcpNode); err != nil {
+	if err := wireCoreWorkflow(ctx, life, conf, lock, nodeIdx, tcpNode, p2pKey); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -413,6 +411,40 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
...
+	peers, err := lock.Peers()
+	if err != nil {
+		return nil, nil, err
+	}
+	peerIDs, err := lock.PeerIDs()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if featureset.Enabled(featureset.QBFTConsensus) {
```
corverroos, line 427: this is how feature flags are used

core/consensus/component.go
```diff
@@ -43,9 +43,7 @@ const (
 )
 
 // NewComponent returns a new consensus QBFT component.
-func NewComponent(tcpNode host.Host, peers []p2p.Peer,
-	peerIdx int64, p2pKey *ecdsa.PrivateKey,
```
corverroos, line 47: removing peerIdx, rather inferring it from tcphost to avoid mistakes.

core/consensus/msg.go
```diff
@@ -112,10 +119,6 @@ func (m msg) ToConsensusMsg() *pbv1.ConsensusMsg {
 
 // hashProto returns a deterministic ssz hash root of the proto message.
 func hashProto(msg proto.Message) ([32]byte, error) {
-	if msg == nil {
-		return [32]byte{}, nil
-	}
```
corverroos, line 117: Checking if an interface is nil is bad, since nil pointer implementations of the interface isn't nil...

leolara (changes requested):

core/consensus/component.go
```diff
@@ -43,9 +43,7 @@ const (
...
-) (*Component, error) {
+func NewComponent(tcpNode host.Host, peers []p2p.Peer, p2pKey *ecdsa.PrivateKey) (*Component, error) {
```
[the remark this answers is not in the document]
leolara: That way it will read `consensu.QBFT` in other packages.

app/app.go
```diff
@@ -191,16 +201,8 @@ func Run(ctx context.Context, conf Config) (err error) {
 }
 
 // wireP2P constructs the p2p tcp (libp2p) and udp (discv5) nodes and registers it with the life cycle manager.
-func wireP2P(ctx context.Context, life *lifecycle.Manager, conf Config, lock cluster.Lock,
+func wireP2P(ctx context.Context, life *lifecycle.Manager, conf Config, lock cluster.Lock, p2pKey *ecdsa.PrivateKey,
 ) (host.Host, *enode.LocalNode, error) {
-	p2pKey := conf.TestConfig.P2PKey
```
dB2510, line 196: why this was moved from here to Run function?

app/app.go
dB2510: cool :)

core/consensus/component.go
```diff
@@ -42,8 +42,8 @@ const (
 	protocolID = "/charon/consensus/qbft/1.0.0"
 )
 
-// NewComponent returns a new consensus QBFT component.
-func NewComponent(tcpNode host.Host, peers []p2p.Peer, p2pKey *ecdsa.PrivateKey) (*Component, error) {
```
dB2510: NewComponent looks more relevant 🙃

dB2510 (approved):
LGTM 🚀

app/app.go
corverroos: previously this was only used in p2p wiring, now it is used in p2p and qbft (need to sign messages). So moved it one level up.

core/consensus/component.go
corverroos: 😉

leolara: if we name components "component", everything would be named component, that is why we should avoid those types of generics in naming

Issue #596, opened 2022-05-23
refactor: replace verifiers with pubshares


xenowits:
## Problem to be solved

Replace `Verifiers` (the coefficients of the public polynomial) with `Pubshares` (the public shares corresponding to each participant).

## Proposed solution

`Verifiers` is now deprecated and is currently used in packages such as `tss`, `keycast` and `createcluster`. Refactor code to achieve the same.

PR #604, opened 2022-05-24
tbls: replace verifier with pubshares

dB2510:
Refactors tbls to use pubshares from start and remove verifier from the workflow.

category: refactor
ticket: #596

### DIFF
--- a/cluster/testdata/TestEncode_lock_json.golden
+++ b/cluster/testdata/TestEncode_lock_json.golden
@@ -31,20 +31,12 @@
  },
  "distributed_validators": [
   {
-   "distributed_public_key": "0x7b182e046410f44bc4b0f3f03a0d06820a30f257",
-   "threshold_verifiers": [
-    "NCyLgFXEZtiGRB0lmQbWms2JS5aK6fDrnZZc5qRpPE4=",
-    "vogVAbfZhGtm6wK1flzae2y6aJHWFr1obDe4NGE6yLo="
-   ]
+   "distributed_public_key": "0x7b182e046410f44bc4b0f3f03a0d06820a30f257"
   },
   {
-   "distributed_public_key": "0xa22c008ffe688352734ae4e3f1217acd5f832708",
-   "threshold_verifiers": [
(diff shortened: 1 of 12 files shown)


corverroos (approved):
🚀

cluster/testdata/TestEncode_lock_json.golden
```diff
@@ -31,20 +31,12 @@
  },
  "distributed_validators": [
   {
-   "distributed_public_key": "0x7b182e046410f44bc4b0f3f03a0d06820a30f257",
-   "threshold_verifiers": [
```
xenowits, line 35: shouldn't `threshold_verifiers` be replaced with `pubshares`?

dkg/frost.go
```diff
@@ -219,7 +219,7 @@ func makeShares(
 	targetID uint32,
 ) ([]share, error) {
 	// Get set of public shares for each validator.
-	pubShares := make(map[uint32]map[uint32]*bls_sig.PublicKey) // map[ValIdx]map[SourceID]*bls_sig.PublicKey
+	pubShares := make(map[uint32]map[int]*bls_sig.PublicKey) // map[ValIdx]map[SourceID]*bls_sig.PublicKey
```
xenowits, line 222: nit: why `int` instead of `uint32`?

dkg/keycast.go
```diff
@@ -185,9 +175,9 @@ func createShares(numValidators, numNodes, threshold int, random io.Reader) ([][
 
 		for ni := 0; ni < numNodes; ni++ {
 			resp[ni] = append(resp[ni], share{
-				PubKey:      pubkey,
-				Verifier:    verifier,
-				SecretShare: shares[ni],
+				PubKey:       tss.PublicKey(),
+				PublicShares: tss.PublicShares(),
+				SecretShare:  shares[ni],
```
xenowits, line 180: nit: can we rename the loop variable `ni`?

dkg/frost.go
dB2510: while going through the code i didn't found any specific reason to use uint32 for dkg, that's why i went with fundamental types such as int

can change if there was any specific reason behind the usage of uint32 for dkg purposes

PR #607, opened 2022-05-24
testutil/compose: add feature set flags

corverroos:
Adds the ability to enable different feature sets to compose.

Also fix issue with teku vc config.

category: feature
ticket: #568


xenowits (approved):
Looks good!

Issue #609, opened 2022-05-25
Simnet integration tests


corverroos:
## Problem to be solved

We do have a bunch of simnet integration tests in the `app` package, but they mostly use in-memory transports, not real networking.

Now that we have `compose` adding a bunch of **simnet** integration tests using docker(-compose) networking would add a lot of value as basic integration smoke tests. It would allow testing:
 - alpha/beta/stable versions of the charon run
 - `charon create dkg` + `charon dkg` + `charon run` as a flow.
 - `charon create cluster` + `charon run` as a flow.
 - cluster with teku and lighthouse and validator mock
  
 Note the aim is to create a go package that runs the tests: `go test testutil/compose/tests`, so similar to `app` packages's "integration unit tests".

 Assertion and evaluation of can be achieved via grafana alerts on prometheus metrics. 

 ## Proposed solution

The following features need to be added to `compose`:
- Grafana alerts that track cluster operation via prometheus metrics:
   - Number of errors should be zero
   - Number of warnings should be low
   - Number of attestations and block proposals should be higher then some threshold per minute
   - Triggered duties minus completed duties should be lower than some threshold per minute
   - Number of validatorapi non-200s should be zero/low.
 - These alerts should be "sent" to compose via webhooks.
 - `compose run -alerts` should launch simple web server that listens for webhook alerts.
 - `compose run -timeout=2m` should stop `docker-compose up` after the timeout.
 - `compose run -alerts` should fail if it has received any alerts by timeout.
 - `compose run` and other commands should fail if `docker-compose up` has non-zero exit codes.
 - Write a table test that runs `compose auto` for all the test cases mentioned above.

### Out of Scope

Real beacon nodes isn'' in scope. Only simnet.

leolara:
@corverroos this overlaps  with the work that I am doing with @eth2devops in automatic tests for actual real VC clients and that I have to talk with @OisinKyne about.

OisinKyne:
This sounds like a use case for prometheus alert-manager rather than adding features to `compose` personally.

corverroos:
The overlap with E2E is debatable, this is only simnet, so doesn't include real beacon nodes or ETH1. I would say this is a smoke test that runs before E2E.

corverroos:
@OisinKyne it does more than add alerts:
 - it adds some alerts to `compose` static grafana config
 - it catches them in `compose auto` 
 - it writes a bunch of smoke test using different cluster configurations

leolara:
[...]

PR #618, opened 2022-05-26
testutil/compose: support alerts and timeout

corverroos:
Add support for `--alert-timeout` to `compose auto` that starts a webhook server that collects alerts and stops docker-compose after the timeout and reports any received alerts.

This add supports for grafana alerts based smoke test evaluators.

category: feature 
ticket: #609

### DIFF
--- a/testutil/compose/compose/alert.go
+++ b/testutil/compose/compose/alert.go
@@ -1,0 +1,73 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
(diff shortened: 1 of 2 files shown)


testutil/compose/compose/alert.go
```diff
@@ -0,0 +1,73 @@
...
+
+// startAlertCollector starts a server that accepts alert webhooks until the context is closed and returns
+// a channel on which the received webhooks will be sent.
+func startAlertCollector(ctx context.Context, port int) (chan []byte, error) {
+	l, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", port))
+	if err != nil {
+		return nil, errors.Wrap(err, "new listener")
+	}
+
+	bodies := make(chan []byte)
```
dB2510, line 40: nit: probably a better variable name 😅

dB2510 (approved):

testutil/compose/compose/alert.go
corverroos: 💀

corverroos: Once I have the json format of the alerts, I'mm probably make it into a proper types struct.

dB2510: cool

PR #625, opened 2022-05-27
app/retryer: remove error on expected timeout

corverroos:
QBFT blocks until the context is cancelled and therefore always returns a `context.DeadlineExceeded` error. This is expected, so should not result in errors. So only log "Timeout retrying" on second to third attempts.

category: bug
ticket: none


corverroos:
All slots currently result in these errors, even though consensus was succesful:
image

xenowits (approved):

dB2510 (approved):

PR #626, opened 2022-05-27
app: enable proposer duties for simnet

corverroos:
Enables deterministic block proposer duties for simnet. This allows compose (and all other cases of simnet) to trigger block proposal.

category: feature
ticket: none


dB2510 (approved):
LGTM

PR #627, opened 2022-05-30
app/retryer: refactor context error log

corverroos:
Refactors how context deadline errors are handled in retryer:
 - The previous change #625 removed error logs on context timeout during first attempt.
 - But this isn't correct since other blocking use-cases SHOULD error if they block and then timeout (like fetcher DutyPropser).
 - Revert that change, logging all timeout errors.
 - Instead refactor qbft consensus to not return an error on timeout, since that is expected behaviour.

category: refactor
ticket: none


xenowits (approved):
Looks good!

PR #628, opened 2022-05-30
core/qbft: fix round change issue

corverroos:
After adding a sanity check to the `isDuplicatedRule` function, I discovered that there was one case where the `changRound` function wasn't used.

category: bug
ticket: none


core/qbft/qbft.go
```diff
@@ -295,7 +303,7 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
 			}
 
 		case <-timerChan: // Algorithm 3:1
-			round++
+			changeRound(round + 1)
```
corverroos, line 306: this is the bug fix

leolara (approved):

PR #649, opened 2022-06-01
precommit: enable buf breaking change check

corverroos:
Enables `buf` check for breaking changes.

category: misc
ticket: none


.github/workflows/pre-commit.yml
```diff
@@ -11,6 +11,8 @@ jobs:
       SKIP: golangci-lint,run-go-tests,no-commit-to-branch
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
corverroos, line 15: required so `buf` can checkout main for comparison.

Issue #643, opened 2022-05-31
p2p: add explanation of enr types in --p2p-bootnodes flag


xenowits:
## Problem to be solved

 `--p2p-bootnodes` strings :  Comma-separated list of discv5 bootnode URLs or ENRs. `Example: enode://<hex node id>@10.3.58.6:30303?discport=30301`.

There are multiple variants/schemas of ENRs: the `enr://<enr_hex>` format and the URL formats (ex: `http://bootnode.gcp.obol.tech:16000/enr`).

## Proposed solution

Add an example of both in the flag description.

xenowits:
Removed the explanations entirely since both ENRs and URLs have standard formats

PR #658, opened 2022-06-02
flags: add enr type explanations in p2p-bootnodes flag

xenowits:
1. Add explanation for different enr variants in `--p2p-bootnodes` flag.
2. Add `http://bootnode.gcp.obol.tech:16000/enr` as the default bootnode in charon run and charon dkg.

category: refactor
ticket: #643

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -74,7 +74,7 @@ func bindDataDirFlag(flags *pflag.FlagSet, dataDir *string) {
 }
 
 func bindP2PFlags(flags *pflag.FlagSet, config *p2p.Config) {
-	flags.StringSliceVar(&config.UDPBootnodes, "p2p-bootnodes", []string{"http://bootnode.gcp.obol.tech:16000/enr"}, "Comma-separated list of discv5 bootnode URLs/ENRs. Example: enr://<enr_hex>,http://<enr_url>.")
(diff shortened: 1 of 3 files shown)


cmd/run.go
```diff
@@ -74,7 +74,7 @@ func bindDataDirFlag(flags *pflag.FlagSet, dataDir *string) {
 }
 
 func bindP2PFlags(flags *pflag.FlagSet, config *p2p.Config) {
-	flags.StringSliceVar(&config.UDPBootnodes, "p2p-bootnodes", nil, "Comma-separated list of discv5 bootnode URLs or ENRs. Example: enode://<hex node id>@10.3.58.6:30303?discport=30301.")
+	flags.StringSliceVar(&config.UDPBootnodes, "p2p-bootnodes", []string{"http://bootnode.gcp.obol.tech:16000/enr"}, "Comma-separated list of discv5 bootnode URLs or ENRs. Example: enode://<hex node id>@10.3.58.6:30303?discport=30301.")
```
corverroos, line 77: I think we can remove the exmaple all together. ENRS have one a single format which is pretty well known (same for HTTP)

corverroos (approved):

cmd/run.go
xenowits: yeah, makes sense

PR #668, opened 2022-06-05
eth2util: move and refactor epoch hash root

corverroos:
Move `EpochMerkle` to `eth2util` since it is used by different packages. Also simplify it to a function `EpochHashRoot`.

category: refactor
ticket: none

### DIFF
--- a/core/consensus/msg.go
+++ b/core/consensus/msg.go
@@ -122,12 +122,16 @@ func hashProto(msg proto.Message) ([32]byte, error) {
 	hh := ssz.DefaultHasherPool.Get()
 	defer ssz.DefaultHasherPool.Put(hh)
 
+	index := hh.Index()
+
 	// Do deterministic marshalling.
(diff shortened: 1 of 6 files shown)


core/consensus/msg.go
```diff
@@ -122,13 +122,17 @@ func hashProto(msg proto.Message) ([32]byte, error) {
...
+	index := hh.Index()
+
 	// Do deterministic marshalling.
 	b, err := proto.MarshalOptions{Deterministic: true}.Marshal(msg)
 	if err != nil {
 		return [32]byte{}, errors.Wrap(err, "marshal proto")
 	}
 	hh.PutBytes(b)
 
+	hh.Merkleize(index)
```
corverroos, line 134: this doesn't have an effect, but it seems best practice to merkelize ssz hashes.

xenowits (approved):
LGTM🚀

PR #670, opened 2022-06-05
core/parsigdb: fix mutex issue

corverroos:
Fixes issue where parsigdb mutex was locked while calling downstream subscribers. 

This aligns with best practice of decreasing locking scope to absolute minimum. 

category: refactor
ticket: none

### DIFF
--- a/core/parsigdb/memory.go
+++ b/core/parsigdb/memory.go
@@ -78,24 +78,8 @@ func (db *MemDB) StoreInternal(ctx context.Context, duty core.Duty, signedSet co
 
 // StoreExternal stores an externally received partially signed duty data set.
 func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet core.ParSignedDataSet) error {
-	db.mu.Lock()
-	defer db.mu.Unlock()
-
 	for pubkey, sig := range signedSet {
-		k := key{Duty: duty, PubKey: pubkey}
-		sigs := db.entries[k]
-
-		var exists bool
-		for _, s := range sigs {
-			if s.ShareIdx == sig.ShareIdx {
-				exists = true
-				break
-			}
-		}
-		if !exists {
-			sigs = append(sigs, sig)
-			db.entries[k] = sigs
-		}
+		sigs := db.store(key{Duty: duty, PubKey: pubkey}, sig)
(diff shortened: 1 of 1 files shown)


core/parsigdb/memory.go
```diff
@@ -78,24 +78,8 @@ func (db *MemDB) StoreInternal(ctx context.Context, duty core.Duty, signedSet co
 
 // StoreExternal stores an externally received partially signed duty data set.
 func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet core.ParSignedDataSet) error {
-	db.mu.Lock()
-	defer db.mu.Unlock()
```
corverroos, line 82: this lock was held while calling `or _, sub := range db.threshSubs {` below. This is bad practice since it unnecessarily increases the locking scope.

core/parsigdb/memory.go
```diff
@@ -116,6 +100,31 @@ func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet co
 	return nil
 }
 
+// store adds the value to the list of signatures at the provided key
+// and returns a copy of the resulting list.
+func (db *MemDB) store(k key, value core.ParSignedData) []core.ParSignedData {
+	db.mu.Lock()
+	defer db.mu.Unlock()
```
corverroos, line 107: only lock while reading/writing shared data.

core/parsigdb/memory.go
dB2510: how did you found out this?

dB2510 (approved):

core/parsigdb/memory.go
```diff
@@ -116,6 +106,27 @@ func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet co
 	return nil
 }
 
+// store return true if the value was added to the list of signatures at the provided key
```
xenowits, line 109: nit:
```suggestion
// store returns true if the value was added to the list of signatures at the provided key
```

xenowits (approved):

core/parsigdb/memory.go
corverroos: just reading through the code

PR #672, opened 2022-06-06
dkg: move writing of files at the end

dB2510:
Moves writing of files like keystores, cluster-lock and deposit-data at the end of p2p exchange, so that we won't get files written to the disk if any peer drops in between

category: bug
ticket: none


xenowits (approved):
Looks good!

corverroos (approved):

Issue #635, opened 2022-05-31
p2p: robust intra-cluster broadcasting


corverroos:
## Problem to be solved

DutyAttester and DutyProposer doesn't work in a charon cluster with one or more nodes being unavailable/down.

This is due to `parsigex` (and probably other libp2p components) synchronously and sequentially sending a message to all other peers, one after the other. If sending to one node errors, then the error is caught and sending to the next node is continued. This is fine. But networking is IO, and IO can block. So sending to a down node, can and does result in the call to send blocking for multiple seconds. When the duty times out during this period, the context is cancelled and we log "partial send success" without sending to all nodes. 

The other issue is that we calculate "success count" incorrectly, so it looks like we sent all available nodes. But we actually only sent to a few other nodes, then blocked, then exitted.

A related problem is warning/error log storm generated when a single node is down. In large clusters, when multiple nodes are down, these warnings will spam the logs.

## Proposed solution

Introduce a "libp2p-sender", a thing that solves both these problems (name TBD):
- **Async sending**: components like parsigex uses libp2p-sender to send async messages. `sender.SendAsync(ctx, targetPeer, protocol, msg)`
- This will prevent sends blocking other sends when a node is down and IO blocks.
- **Warn log filtering**: It doesn't log warnings when sending to a known down node fails.
- It keeps track of higher level networking metrics, and logs summaries and state changes only.
- Wire this thing into all components doing p2p sending.

corverroos:
Out of scope (follow up PR): Introduce a `Broadcast` method to the p2psender so each component doesn't need to duplicate broadcasting logic.

PR #675, opened 2022-06-07
p2p: implement sender log filter

corverroos:
Implement `p2p.Sender` that filters p2p sending logs per peer based on state changes.

Still need to wire and refactor other components to also use `p2p.Sender.`

category: feature
ticket: #635

### DIFF
--- a/p2p/sender.go
+++ b/p2p/sender.go
@@ -17,6 +17,7 @@ package p2p
 
 import (
 	"context"
+	"sync"
 
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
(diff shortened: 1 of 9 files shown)


xenowits (approved):

p2p/sender.go
```diff
@@ -26,18 +27,78 @@ import (
...
+	state.buffer = append(state.buffer, err)
+	if len(state.buffer) > senderBuffer { // Trim buffer
+		state.buffer = state.buffer[len(state.buffer)-senderBuffer:]
 	}
+
+	failure := err != nil
+	success := !failure
+
+	if success && state.failing {
+		// See if we have senderHysteresis successes i.o.t. change state to success.
```
xenowits, line 66: `i.o.t` means?

p2p/sender.go
```diff
@@ -26,18 +27,78 @@ import (
...
-	if err != nil {
-		log.Warn(ctx, "Failed sending p2p message", err)
+const (
+	senderHysteresis = 3
+	senderBuffer     = senderHysteresis + 1
+)
+
+type peerState struct {
+	failing bool
+	buffer  []error
```
xenowits, line 40: nit: can rename to `bufErrs`

p2p/sender.go
corverroos: in order to

p2p/sender.go
corverroos: I prefer to call the thing by what it is/does, rather than adding type suffixes, especially if it is there is only one of them

p2p/sender.go
```diff
@@ -26,18 +27,78 @@ import (
...
+		// First attempt failed or state changed to failing
+		log.Warn(ctx, "P2P sending failing", err, z.Str("peer", PeerName(peerID)))
+		state.failing = true
+	}
+
+	s.states.Store(peerID, state) // Note there is a race if two results for the same peer is added at the same time, but this isn't critical.
+}
+
+// SendAsync returns nil sends a libp2p message asynchronously. It logs results on state change (success to/from failure).
+func (s *Sender) SendAsync(ctx context.Context, tcpNode host.Host, protoID protocol.ID, peerID peer.ID, msg proto.Message) error {
```
xenowits, line 91: discussion: why does this method `SendAsync()` even return an error?

corverroos: so that it has the same API as `Send`, so components can use an abstract SendFunc and either use sync or async

corverroos: I'll move SendFunc here

xenowits: okay alright 👍

Issue #584, opened 2022-05-23
clashing read only cluster definition and lock files.


corverroos:
## Problem to be solved

When running `charon create dkg` or `charon dkg` or `charon create cluster`, readonly `cluster-definition.json` and `cluster-lock.json` files are created (perms `0o400`). 

It is readonly to indicate that customers should not modify the files, since they contain hashes that become invalid if the file is editted.

When running the command again (when the output files are present), the command fails with obscure errors. The UX around this isn't great.

 ## Proposed solution

  - Improve error handling and UX.
  - The command should check for clashes (and all other disk write issues) at the start and fail with a proper message.
  - Failing at the start (instead of at the end) will ensure that DKG cannot be partially completed, either all succeed or all fail.
  - This allows for adding deterministic human friendly names to output.
  - Note that unit testing/integration testing might want to specify the output filenames, since otherwise they need to scan the folder to figure out what was output.
  
  
  Previous Options Considered
  -----
  Option 1:
 - Add a timestamp suffix to the filename. 
 - This will indeed make files unique and avoid clashes, but it can be confusing to know which file is the correct one. This is especially an issue for `cluster_lock` which will have different names on the different peer's machines, even though it is the same file.
   
Option 2:
 - Add cluster name or some deterministic derived human friendly suffix to the file.
 - Since the names are deterministic, this will still result in clashes.

corverroos:
Make perms `0o444`

PR #676, opened 2022-06-07
dkg: check disk writes

dB2510:
Check disk writes by creating random sample files and then removing sample ones if successful.

category: bug
ticket: #584

### DIFF
--- a/dkg/disk.go
+++ b/dkg/disk.go
@@ -153,7 +153,7 @@ func checkWrites(dataDir string, def cluster.Definition) error {
 		sigs[pk] = sig
 	}
 
-	if err := writeDepositData(sigs, "0x0000000000000000000000000000000000000000", "prater", dataDir); err != nil {
+	if err := writeDepositData(sigs, testutil.RandomETHAddress(), "prater", dataDir); err != nil {
 		return err
 	}
 


dkg/disk.go
```diff
@@ -112,10 +113,32 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
+	err = os.WriteFile(depositPath, bytes, 0o444) // read-only
 	if err != nil {
 		return errors.Wrap(err, "write deposit data")
 	}
 
 	return nil
 }
+
+// fileExists returns error if keystores, cluster-lock and deposit-data already exists in given dataDir.
+func fileExists(dataDir string, vals int) error {
```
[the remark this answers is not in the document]
corverroos: as per ticket:
```
The command should check for clashes (and all other disk write issues) at the start and fail with a proper message.
```

dkg/disk.go
```diff
@@ -112,10 +116,74 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
+		return errors.Wrap(err, "remove sample deposit-data.json")
+	}
+
+	for i := 0; i < def.NumValidators; i++ {
+		if err := os.Remove(path.Join(dataDir, fmt.Sprintf("keystore-%d.json", i))); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("remove sample keystore-%d.json", i))
+		}
+
+		if err := os.Remove(path.Join(dataDir, fmt.Sprintf("keystore-%d.txt", i))); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("remove sample keystore-%d.txt", i))
```
corverroos, line 180: avoid sprintf in errors, rather use `z.Str("file", filename)`

dkg/disk.go
```diff
@@ -112,10 +116,74 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
+
+		sig, err := testutil.RandomBLSSignature()
+		if err != nil {
+			return err
+		}
+
+		sigs[pk] = sig
+	}
+
+	if err := writeDepositData(sigs, "0x0000000000000000000000000000000000000000", "prater", dataDir); err != nil {
```
corverroos, line 156: no const for this withdrawal addr avialble?

dkg/disk.go
```diff
@@ -112,10 +116,74 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
+		return err
```
corverroos, line 157: Are these error message human friendly? Will people understand what the problem is?

dkg/disk.go
```diff
@@ -112,10 +116,74 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
 	}
 
 	return nil
 }
+
+// checkWrites writes sample files to check disk writes and removes sample files after verification.
+func checkWrites(dataDir string, def cluster.Definition) error {
+	var shares []share
+	sigs := make(map[core.PubKey]*bls_sig.Signature)
+	for i := 0; i < def.NumValidators; i++ {
```
corverroos, line 131: another option which is simpler is:
```
files := getSampleFiles(def.NumValidators)
for f in files {
  err := os.WriteFile("sample", f, 0x444)
  if err ! nil { "cannot write file", z.Str("file",f) }
  
  err := os.Remove(f)
  if err ! nil { "cannot delete file", z.Str("file",f) }
} 
```

corverroos (approved):

dkg/disk.go
dB2510: can be replaced with `testutil.RandomETHAddress()`

Issue #320, opened 2022-03-30
Auto-generate dynamic changelog in github actions


corverroos:
## Problem to be solved

We have a github action `release.yml` that is triggered when a commit is tagged. It generates a static changelog. But we also have a tool `go run testutil/genchangelog/main.go` that auto generates a changelog between two tags. But this is still run manually.

## Proposed solution

Run `genchangelog` as part of `release.yml` and push its output as the release changlog.

PR #677, opened 2022-06-07
github/workflows: auto generate changelog

corverroos:
Auto-generates changelog in github action. Also simplify by removing usage of docker image and just use local code.

category: misc
ticket: #320

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,31 +8,11 @@ jobs:
     runs-on: ubuntu-latest
     name: Publish Release
     steps:
-
-    # Wait for docker build to complete since we need pegged cli-reference.
-    - uses: fountainhead/action-wait-for-check@v1.0.0
-      with:
-        token: ${{ secrets.GITHUB_TOKEN }}
-        checkName: Build Docker Image
-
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -8,31 +8,11 @@ jobs:
...
-      run: |
-        tag=$(echo "${{ github.ref }}" | cut -d / -f 3)
-        echo "::set-output name=tag::$tag"
-
-    - run: docker run ghcr.io/obolnetwork/charon:${{steps.get-tag.outputs.tag}} run --help > cli-reference.txt
-
-    - run: cat cli-reference.txt
-
+    - run: go run . --help > cli-reference.txt
+    - run: go run testutil/genchangelog/main.go
```
xenowits, line 13: so, ig the readme in `branching.md` needs to be updated?

xenowits (approved):

.github/workflows/release.yml
xenowits: Step 6, 7

corverroos: cool

PR #693, opened 2022-06-10
precommit: check log and error capitalisation

corverroos:
Adds pre-commit hooks that checks:
 - log messages are capitalised
 - error messages are not capitalised

category: misc
ticket: none

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,7 +8,7 @@ linters-settings:
     default-signifies-exhaustive: true
   forbidigo:
     forbid:
-      - 'fmt\.Print.*'            # Prevent debug logging
+      - 'fmt\.Print.*(# Avoid debug logging)?'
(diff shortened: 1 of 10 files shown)


.golangci.yml
```diff
@@ -8,7 +8,7 @@ linters-settings:
     default-signifies-exhaustive: true
   forbidigo:
     forbid:
-      - 'fmt\.Print.*'            # Prevent debug logging
+      - 'fmt\.Print.*(# Avoid debug logging)?'
```
corverroos, line 11: forbidigo supports UX messages inside regexp

corverroos: forbidigo only supports checking single identifiers. It cannot check function calls with arguments in a single regexp.

dB2510 (approved):

PR #694, opened 2022-06-10
cmd: change default datadir

dB2510:
Changes default datadir to .charon and default name of `p2pkey` to `charon-enr-private-key`. Also read and write keystore to `/validator_keys` in .charon. This changes are also been made to `create cluster command` as well.
The files would look as follows if default values are used:
Screenshot 2022-06-10 at 8 32 52 PM
Screenshot 2022-06-13 at 11 34 22 AM

Ref: https://github.com/ObolNetwork/charon-distributed-validator-node/issues/1

category: refactor
ticket: #702


cmd/createenr.go
```diff
@@ -59,7 +59,7 @@ func runCreateEnrCmd(w io.Writer, config p2p.Config, dataDir string) error {
 	}
 	defer db.Close()
 
-	_, _ = fmt.Fprintf(w, "Created ENR private key: %s/p2pkey\n", dataDir)
+	_, _ = fmt.Fprintf(w, "Created ENR private key: %s/charon-enr-private-key\n", dataDir)
```
xenowits, line 62: nit: can shorten to `enr-private-key`

OisinKyne: We went with `charon-enr-private-key` so I think it's accurate as is.

OisinKyne (approved):
LGTM

Only feedback is maybe we change `/keys` to `/validator_keys`, and some suggestions for more verbose errors and a question around why we might be trying to wipe keystores in code.

dkg/disk.go
```diff
@@ -66,9 +65,12 @@ func writeKeystores(datadir string, shares []share) error {
...
 
-	err := keystore.StoreKeys(secrets, datadir)
-	if err != nil {
-		return err
+	if err := os.Mkdir(path.Join(datadir, "/keys"), os.ModePerm); err != nil {
+		return errors.Wrap(err, "mkdir /keys")
+	}
+
+	if err := keystore.StoreKeys(secrets, path.Join(datadir, "/keys")); err != nil {
+		return errors.Wrap(err, "store keystores")
```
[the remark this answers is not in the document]
dB2510: full error is given by `err `variable. `store keystores` is like a wrapper text to the actual error text.

dkg/disk.go
```diff
@@ -171,14 +173,8 @@ func checkWrites(dataDir string, def cluster.Definition) error {
 		return errors.Wrap(err, "remove sample deposit-data.json")
```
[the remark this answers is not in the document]
dB2510: Here also `remove sample deposit-data.json` is just a wrapper text around actual error message given err variable. There can be any number of errors whereas `existing deposit-data.json file already present` is one of them. 
This function is written to remove sample files that we create before starting dkg to check disk writes.

testutil/compose/compose/smoke_internal_test.go
```diff
@@ -89,19 +89,20 @@ func TestSmoke(t *testing.T) {
...
 
 				data.Nodes[2].ImageTag = "v0.5.0" // TODO(corver): Update this with new releases.
-				data.Nodes[2].Entrypoint = containerBinary
+				// TODO(dhruv): Using default entrypoint for now need to update with data.Nodes[1].Entrypoint = containerBinary
+				// Use default entrypoint
 
 				data.Nodes[3].ImageTag = "v0.5.0"
-				data.Nodes[3].Entrypoint = containerBinary
+				// TODO(dhruv): Using default entrypoint for now need to update with data.Nodes[1].Entrypoint = containerBinary
+				// Use default entrypoint
```
corverroos, line 105: this is misleading since default entrypoint is using local pre-built binary so not that version. Suggest rather commenting out the whole test.

testutil/compose/compose/smoke_internal_test.go
```diff
@@ -89,19 +89,20 @@ func TestSmoke(t *testing.T) {
 				conf.KeyGen = compose.KeyGenDKG
 			},
 			TmplFunc: func(data *compose.TmplData) {
-				const containerBinary = "/usr/local/bin/charon"
-
 				data.Nodes[0].ImageTag = "latest"
-				data.Nodes[0].Entrypoint = "" // Use locally pre-built binary
+				// Use default entrypoint
```
corverroos, line 93: // Use default entrypoint which is pre-built binary if provided

Issue #688, opened 2022-06-09
Proxy event stream from beacon node


OisinKyne:
## Problem to be solved

Teku complains that it can't communicate with the BN event stream

```
teku_1        | 2022-06-09 14:49:32.587 ERROR - Validator   *** Error while connecting to beacon node event stream
teku_1        | java.net.SocketTimeoutException: timeout
teku_1        | at okio.SocketAsyncTimeout.newTimeoutException(JvmOkio.kt:143) ~[okio-jvm-2.8.0.jar:?]
teku_1        | at okio.AsyncTimeout.access$newTimeoutException(AsyncTimeout.kt:162) ~[okio-jvm-2.8.0.jar:?]
teku_1        | at okio.AsyncTimeout$source$1.read(AsyncTimeout.kt:335) ~[okio-jvm-2.8.0.jar:?]
teku_1        | at okio.RealBufferedSource.indexOf(RealBufferedSource.kt:427) ~[okio-jvm-2.8.0.jar:?]
```

## Proposed solution

Proxy the event stream from BN to validator-api charon component. Mutate objects as required to fool the downstream validator as to its private key. 

### Out of Scope

Testing whether we can use charon issued events on this event stream to force/prompt the VC into behaviour that could help us (e.g. tricking it into asking for a block early with a fork updated event)

corverroos:
At this point on teku does events, and it only asks for "slot"/"head" events, so we do not need to mutate data.

PR #696, opened 2022-06-10
core/validatorapi: fix proxy response flushing

corverroos:
Fix event stream proxy issue. Since we wrapped the response writer in our own type, it didn't implement http.Flusher, so streaming responses were not proxied/flushed to the client.

category: bug
ticket: #688

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -654,7 +654,7 @@ type writeFlusher interface {
 	http.Flusher
 }
 
-// proxyResponseWriter wraps a http response writer and instruments errors.
+// proxyResponseWriter wraps the writeFlusher interface and instruments errors.
 type proxyResponseWriter struct {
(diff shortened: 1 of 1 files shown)


core/validatorapi/router.go
```diff
@@ -647,9 +647,16 @@ func hexQuery(query url.Values, name string) ([]byte, error) {
...
 }
 
+// writeFlusher is copied from /net/http/httputil/reverseproxy.go.
+// It is required to flush streaming responses.
+type writeFlusher interface {
+	http.ResponseWriter
+	http.Flusher
+}
+
 // proxyResponseWriter wraps a http response writer and instruments errors.
```
xenowits, line 657:
```suggestion
// proxyResponseWriter wraps the writeFlusher interface and instruments errors.
```

xenowits (approved):

Issue #698, opened 2022-06-12
core: refactor signed data types to interfaces


corverroos:
## Problem to be solved

See design doc: https://docs.google.com/document/d/1uGoYuYWROAfWiRwVtcd4DO7DIOVLsXkddrQarsHidhg/edit#

Core workflow data types `ParSignedData` and `AggSignedData` are partially unstructured containing signatures and raw json (bytes) data. This works well to decouple data from logic, and promotes the charon middleware architecture as a “transparent” data-signing layer in the eth2 stack. 

This approach has however the following drawbacks:
- The logic working with the data (like extracting, replacing, verifying signatures) is spread out throughout the different components. 
- Most components have type switches that then decode and encode different types.
- Charon isn’t actually transparent wrt data, since we do encode and decode.

## Proposed solution

Refactoring these types into interfaces with proper typed implementations for the different duties would provide the following benefits:
- Data wrangling logic moved out of components and colocated in one place; the interface implementation.
- Adding new duties, would mostly be adding a new implementation.
- Components focus on what they need to do, not how it is done.

PR #699, opened 2022-06-12
core: add new SignedData interface and impls

corverroos:
Step 1 of N: Adds the new interface and implementations and helper functions. Nothing wired or integrated yet.

category: refactor
ticket: #698

### DIFF
--- a/core/signeddata.go
+++ b/core/signeddata.go
@@ -75,6 +75,7 @@ type VersionedSignedBeaconBlock struct {
 
 func (b VersionedSignedBeaconBlock) Signature() Signature {
 	switch b.Version {
+	// No block nil checks since `NewVersionedSignedBeaconBlock` assumed.
 	case spec.DataVersionPhase0:
 		return SigFromETH2(b.Phase0.Signature)
 	case spec.DataVersionAltair:
@@ -99,6 +100,7 @@ func (b VersionedSignedBeaconBlock) SetSignature(sig Signature) (SignedData, err
 	}
 
 	switch b.Version {
+	// No block nil checks since `NewVersionedSignedBeaconBlock` assumed.
 	case spec.DataVersionPhase0:
 		resp.Phase0.Signature = sig.ToETH2()
 	case spec.DataVersionAltair:
@@ -106,7 +108,7 @@ func (b VersionedSignedBeaconBlock) SetSignature(sig Signature) (SignedData, err
 	case spec.DataVersionBellatrix:
 		resp.Bellatrix.Signature = sig.ToETH2()
 	default:
-		panic("unknown version") // Note this is avoided by using `NewVersionedSignedBeaconBlock`.
+		return nil, errors.Wrap(err, "unknown type")
 	}
 
 	return resp, nil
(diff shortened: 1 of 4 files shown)


core/signeddata.go
```diff
@@ -0,0 +1,251 @@
...
+	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
+
+	"github.com/obolnetwork/charon/app/errors"
+)
+
+var (
+	_ SignedData = VersionedSignedBeaconBlock{}
+	_ SignedData = Attestation{}
+	_ SignedData = Signature{}
+	_ SignedData = SignedVoluntaryExit{}
```
corverroos, line 33: The 4 implementations of the new `SignedData` interface

core/types.go
```diff
@@ -209,6 +246,24 @@ type AttestationData struct {
...
+}
+
+// ParSignedData2 is a partially signed duty data only signed by a single threshold BLS share.
+// TODO(corver): Rename and place ParSignedData.
+type ParSignedData2 struct {
+	SignedData
+	// ShareIdx returns the threshold BLS share index.
+	ShareIdx int
+}
+
```
corverroos, line 266: these are the two new types (one interface and one struct)

core/signeddata.go
```diff
@@ -0,0 +1,277 @@
...
+func NewPartialAttestation(att *eth2p0.Attestation, shareIdx int) ParSignedData2 {
+	return ParSignedData2{
+		SignedData: NewAttestation(att),
+		ShareIdx:   shareIdx,
+	}
+}
+
+// Attestation is a signed attestation and implements SignedData.
+type Attestation struct {
+	eth2p0.Attestation
```
corverroos, line 197: cannot subtype since `eth2p0.Attestation`'s field `Signature` clashes with interface method `Signature`

core/types.go
```diff
@@ -169,9 +170,45 @@
...
+// NewPartialSignature is a convenience function that returns a new partially signature.
+func NewPartialSignature(sig Signature, shareIdx int) ParSignedData2 {
+	return ParSignedData2{
+		SignedData: sig,
+		ShareIdx:   shareIdx,
+	}
+}
+
+// Signature is a BLS12-381 Signature. It implements SignedData.
 type Signature []byte
```
dB2510, line 182: If this is an implementation of SignedData then why it is not present in signeddata.go?

core/signeddata.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/attestantio/go-eth2-client/spec/bellatrix"
+	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
+
+	"github.com/obolnetwork/charon/app/errors"
+)
+
+var (
+	_ SignedData = VersionedSignedBeaconBlock{}
+	_ SignedData = Attestation{}
+	_ SignedData = Signature{}
```
dB2510, line 32: I assume Signature{} is to used by Randao and exchanger in dkg. Can we be more specific about those types and also accomodate those cases as well? Since in exchanger we require to share some data(like cluster-lock and deposit-data) not just signature.

core/signeddata.go
dB2510: can we make this attestation private since we already have getter and setter methods(Signature and SetSignature)?

dB2510: Although I don't have any strong opinion on this.

dB2510 (approved):
Looks Great!

core/signeddata.go
corverroos: it cannot be made private since some components will cast and use it, like bcast, and aggsigdb

PR #703, opened 2022-06-13
cluster: fix bug in hash tree root calculation

xenowits:
The current calculation of hash tree root is incorrect as it doesn't call `*ssz.hasher.MerkleizeWithMixin()` method when serializing arrays.

We discovered it when comparing charon (fastssz) hash outputs with typescript (chainsafe/ssz) hashes.

category: bug
ticket: none


cluster/cluster_test.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/obolnetwork/charon/testutil"
 )
 
-//go:generate go test . -v -clean -update
```
xenowits, line 33: `-clean` deletes everything inside the `testdata` dir including the versioned definition and lock json files

cluster/definition.go
```diff
@@ -195,11 +195,16 @@ func (d Definition) HashTreeRootWith(hh *ssz.Hasher) error {
...
+	// Field (9) 'Operators'
+	{
+		subIndx := hh.Index()
+		num := uint64(len(d.Operators))
+		for _, operator := range d.Operators {
+			if err := operator.HashTreeRootWith(hh); err != nil {
+				return err
+			}
 		}
+		hh.MerkleizeWithMixin(subIndx, num, num)
```
xenowits, line 207: See https://github.com/prysmaticlabs/prysm/blob/4cbb69602f35c11611be446b1b0a06bd3c3c7e8e/proto/eth/v1/generated.ssz.go#L1193

corverroos (approved):

PR #704, opened 2022-06-13
core: poc signed data visitor

corverroos:
This is a PoC of adding a "signed data duty visitor" to the `SignedData` interface to further decouple logic from data without the use of type switches.

This is starting to feel super OO.... not sure this is better than a plain switch...?

category: refactor
ticket: #699

### DIFF
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -125,6 +125,66 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 	}
 }
 
+// Broadcast2 broadcasts the aggregated signed duty data object to the beacon-node.
+func (b Broadcaster) Broadcast2(ctx context.Context, duty core.Duty,
+	pubkey core.PubKey, aggData core.SignedData,
+) (err error) {
+	ctx = log.WithTopic(ctx, "bcast")
+	ctx = log.WithCtx(ctx, z.Any("pubkey", pubkey))
+	defer func() {
+		if err == nil {
+			instrumentDuty(duty, pubkey, b.delayFunc(duty.Slot))
+		}
+	}()
+
+	v := signedDataVisitor{eth2Cl: b.eth2Cl, delayFunc: b.delayFunc}
+
+	return aggData.VisitDuty(ctx, duty, v)
+}
+
+type signedDataVisitor struct {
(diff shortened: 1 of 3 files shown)


core/bcast/bcast.go
```diff
@@ -125,6 +125,66 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 	}
 }
 
+// Broadcast2 broadcasts the aggregated signed duty data object to the beacon-node.
+func (b Broadcaster) Broadcast2(ctx context.Context, duty core.Duty,
```
corverroos, line 129: example of what broadcast would look like using a visitor pattern

core/bcast/bcast.go
```diff
@@ -125,6 +125,66 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
...
+			instrumentDuty(duty, pubkey, b.delayFunc(duty.Slot))
+		}
+	}()
+
+	v := signedDataVisitor{eth2Cl: b.eth2Cl, delayFunc: b.delayFunc}
+
+	return aggData.VisitDuty(ctx, duty, v)
+}
+
+type signedDataVisitor struct {
```
corverroos, line 145: the nice thing is the each type's logic is defined in its own function. The pad thing is that we cannot define other input parameters our output parameters to the function since Go generics doesn't support "function generics", it only support "type generics"...

core/types.go
```diff
@@ -199,6 +200,17 @@ type SignedData interface {
...
+}
+
+// SignedDataDutyVisitor abstract a signed data implementation logic switch.
+// Upon calling SignedData.VisitDuty the implementation will cal its associated method on this visitor.
+type SignedDataDutyVisitor interface {
+	VisitSignature(context.Context, Duty, Signature) error
+	VisitAttestation(context.Context, Duty, Attestation) error
+	VisitBlock(context.Context, Duty, VersionedSignedBeaconBlock) error
+	VisitExit(context.Context, Duty, SignedVoluntaryExit) error
 }
```
corverroos, line 214: The visitor pattern.....

corverroos:
Not going down this route as a plain old `switch` statement is much simpler and supports all use-cases.

PR #705, opened 2022-06-13
*: integrate signed data interface types

corverroos:
Integrates new `SignedData` interface and `ParSignedData` type.

category: refactor
ticket: #699


core/sigagg/sigagg.go
```diff
@@ -43,47 +41,32 @@ func New(threshold int) *Aggregator {
...
-		firstRoot   eth2p0.Root
-	)
-	for i, parSig := range parSigs {
-		root, err := getSignedRoot(duty.Type, parSig)
-		if err != nil {
-			return err
-		}
-		if i == 0 {
-			firstParSig = parSig
-			firstRoot = root
```
corverroos, line 75: removed root comparison, since it doesn't add much value. We will verify signatures in ParSigDB, and aggregation will fail if roots do not match.

core/aggsigdb/memory.go
```diff
@@ -123,19 +124,34 @@ func (db *MemDB) Run(ctx context.Context) {
...
 	key := memDBKey{command.duty, command.pubKey}
 
-	curData, ok := db.data[key]
-	if ok && !curData.Equal(command.data) {
-		command.response <- errors.New("trying to update entry in aggsigdb/MemDB")
-		close(command.response)
-
-		return
-	} else if !ok {
+	exiting, ok := db.data[key]
```
dB2510, line 127:
```suggestion
	existing, ok := db.data[key]
```

core/fetcher/fetcher.go
```diff
@@ -142,16 +142,17 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, argSet core
 			Slot: slot,
 			Type: core.DutyRandao,
 		}
-		randao, err := f.aggSigDBFunc(ctx, dutyRandao, pubkey)
+		signedData, err := f.aggSigDBFunc(ctx, dutyRandao, pubkey)
```
dB2510, line 145: maybe rename it to `randaoData` rather naming it same as return type of the function

core/fetcher/fetcher.go
```diff
@@ -142,16 +142,17 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, argSet core
...
 		if err != nil {
 			return nil, err
 		}
-		randaoEth2 := core.DecodeRandaoAggSignedData(randao)
+
+		randao := signedData.Signature().ToETH2()
 
 		// TODO(dhruv): what to do with graffiti?
 		// passing empty graffiti since it is not required in API
 		var graffiti [32]byte
```
dB2510, line 154: Let's have graffiti something which includes a string like:
`Charon Distributed Validator`
This shows up on beaconcha.in alongside Public key when our DV proposes a block.

core/sigagg/sigagg.go
```diff
@@ -43,47 +41,32 @@ func New(threshold int) *Aggregator {
...
 // Subscribe registers a callback for aggregated signed duty data.
-func (a *Aggregator) Subscribe(fn func(context.Context, core.Duty, core.PubKey, core.AggSignedData) error) {
+func (a *Aggregator) Subscribe(fn func(context.Context, core.Duty, core.PubKey, core.SignedData) error) {
 	a.subs = append(a.subs, fn)
 }
 
 // Aggregate aggregates the partially signed duty data for the DV.
-func (a *Aggregator) Aggregate(ctx context.Context, duty core.Duty, pubkey core.PubKey, parSigs []core.ParSignedData) error {
-	if len(parSigs) < a.threshold {
+func (a *Aggregator) Aggregate(ctx context.Context, duty core.Duty, pubkey core.PubKey, shares []core.ParSignedData) error {
```
dB2510, line 53: `shares` doesn't fit here i think, since we also call partial key shares as `shares`. Maybe rename it to `pSigData` or revert this change

dB2510 (approved):
Looks Great! 🚀

corverroos:
Abandoning this approach add it breaks immutability. The values passed around are now mutable, since go-eth2-client types are composed using pointers. The core workflow architecture is built on the concept of "self-describing immutable values" flowing through the different components. Mutable values will result in a lot of very hard to find bugs. Sure it can be achieved with the interface pattern a `Clone()` method, but that is not enforceable and therefore brittle. Sure the old raw `Data []byte` structs are also technically mutable, but devs are enforced to used `Encode/Decode` methods when communication between components, so the eth2client types cannot leak mutability.

corverroos:
Reopening after discussion. `Clone` is a common pattern and makes explicit that messages should be treated as immutable. The raw `Data []byte` achieves the same result, it just isn't as obvious why and how.

core/fetcher/fetcher.go
corverroos: sure, another PR though?

PR #716, opened 2022-06-15
core/consensus: add tracing to qbft

corverroos:
Adds a tracing span to qbft cosensus. End the span earlier when consensus is decided, not only when returning from Propose since that only happens at duty deadline.

category: misc
ticket: none

### DIFF
--- a/core/consensus/component.go
+++ b/core/consensus/component.go
@@ -25,6 +25,7 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
+	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/obolnetwork/charon/app/errors"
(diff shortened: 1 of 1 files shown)


core/consensus/component.go
```diff
@@ -289,3 +296,18 @@ func (c *Component) getPeerIdx() (int64, error) {
 func isContextErr(err error) bool {
 	return errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)
 }
+
+type spanKey struct{}
```
dB2510, line 300: why empty struct?

dB2510 (approved):

core/consensus/component.go
corverroos: See `context.WithValue` godoc:
```
// The provided key must be comparable and should not be of type
// string or any other built-in type to avoid collisions between
// packages using context. Users of WithValue should define their own
// types for keys. To avoid allocating when assigning to an
// interface{}, context keys often have concrete type
// struct{}. Alternatively, exported context key variables' static
// type should be a pointer or interface.
```

dB2510: cool

PR #718, opened 2022-06-15
core/consensus: fix tracing span

corverroos:
Fixes consensus tracing span by cancelling the parent span instead.

category: bug
ticket: none


corverroos:
See resulting trace example:

image

dB2510 (approved):

Issue #684, opened 2022-06-08
dkg: implement sync protocol


corverroos:
## Problem to be solved

DKG uses the built-in libp2p ping protocol to detect connections to peers. We however need the following features from the networking layer:
 - Reconnect on "stream reset". This is due to relay connection timeouts which we should support.
 - Soft shutdown. When need to distinguish between node connection failure and nodes shutting down at the end of successful DKG.
 - Assert identical cluster definition hash between all peers with proper errors on failure.
 - context timeout on peer connection loss.

## Proposed solution

Implement a `dkg.Sync` object that implements a custom libp2p protocol that achieves the above. Enhance on the ping pattern.

corverroos:
Please break this up into multiple PRs, e.g.:
 - Add proto message types
 - Implement naive Server/Client (no shutdown, no AwaitConnected methods) (add tests)
 - Add Shutdown to client and AwaitAllShutdown server (add tests)
 - Add AwaitConnected to server (add tests)
 - Integrate into DKG.

PR #735, opened 2022-06-21
dkg/sync: implement client.AwaitConnected

dB2510:
Implements AwaitConnected method of sync.Client

category: feature
ticket: #684

### DIFF
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -18,18 +18,22 @@ package sync
 import (
 	"bufio"
 	"context"
+	"time"
 
 	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/app/z"
 	pb "github.com/obolnetwork/charon/dkg/dkgpb/v1"
 	"github.com/obolnetwork/charon/p2p"
 )
 
-type Result struct {
+type result struct {
+	rtt       time.Duration
 	timestamp string
 	error     string
 }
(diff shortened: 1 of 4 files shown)


dkg/sync/client.go
```diff
@@ -16,84 +16,159 @@
...
+
 // NewClient starts a goroutine that establishes a long lived connection to a p2p server and returns a new Client instance.
-func NewClient(ctx context.Context, tcpNode host.Host, server p2p.Peer, hashSig []byte, onFailure func(), ch chan *pb.MsgSyncResponse) Client {
-	go func() {
-		s, err := tcpNode.NewStream(ctx, server.ID, syncProtoID)
-		if err != nil {
-			log.Error(ctx, "Open new stream with server", err)
-		}
-		defer s.Close()
+// TODO(dhruv): call onFailure on permanent failure.
```
xenowits, line 118: can remove this comment to somewhere more appropriate

dkg/sync/client.go
```diff
@@ -16,84 +16,159 @@
...
-			log.Error(ctx, "Marshal msg", err)
-		}
+	ctx, cancel := context.WithCancel(ctx)
 
-		if _, err = s.Write(b); err != nil {
-			log.Error(ctx, "Write msg to stream", err)
-		}
+	out := make(chan result)
+	go func() {
+		defer close(out)
```
xenowits, line 140: discussion: should this statement be outside of the goroutine?🤔

dkg/sync/server.go
```diff
@@ -63,75 +62,76 @@ func NewServer(ctx context.Context, tcpNode host.Host, peers []p2p.Peer, defHash
...
+				return
+			}
+
+			pID := s.Conn().RemotePeer()
+			log.Debug(ctx, "Message received from client", z.Any("server", p2p.PeerName(pID)))
+
+			pubkey, err := pID.ExtractPublicKey()
+			if err != nil {
+				log.Error(ctx, "Get client public key", err)
+				err = s.Reset()
```
xenowits, line 96: discussion: these two lines are repeated in each `if err != nil {}` block
```go
err = s.Reset()
log.Error(ctx, "Stream reset", err)
```
Can we do something to prevent the duplication?

xenowits (approved):
Looks good. Need a few changes

dkg/sync/client.go
```diff
@@ -16,84 +16,159 @@
...
+	server    p2p.Peer
+	results   chan result
 }
 
 // AwaitConnected blocks until the connection with the server has been established or returns an error.
-func (*Client) AwaitConnected() error {
+func (c *Client) AwaitConnected() error {
+	results := c.results
+	for resp := range results {
+		if resp.error == "Invalid Signature" {
```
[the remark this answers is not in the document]
dB2510: sure

dkg/sync/client.go
dB2510: let it be here, will be removed in subsequent PRs

dkg/sync/client.go
dB2510: nope, this go routine is responsible of writing into this channel so thats why it needs to close the channel once context is called or ctx.Done is closed

dkg/sync/server.go
dB2510: yeah, will remove these and add a `defer s.Close()` at the beginning of this function

dkg/sync/client.go
```diff
@@ -16,84 +16,159 @@
...
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/app/z"
 	pb "github.com/obolnetwork/charon/dkg/dkgpb/v1"
 	"github.com/obolnetwork/charon/p2p"
 )
 
+type result struct {
```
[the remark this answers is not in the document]
dB2510: any suggestions?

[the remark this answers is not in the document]
dB2510: this package is inspired from: https://github.com/libp2p/go-libp2p/blob/707100a521dace4a7a87fd17c04042ea483380b5/p2p/protocol/ping/ping.go#L93
hence the type name too

dkg/sync/client.go
```diff
@@ -16,84 +16,159 @@
 package sync
```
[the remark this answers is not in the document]
xenowits: // Package sync provides APIs that ensure robust network connectivity between all peers in the DKG.
does this sound good?

PR #746, opened 2022-06-24
refactor: change cluster-lock default dir

xenowits:
Change `cluster-lock` default dir to `.charon/cluster/cluster-lock.json`

category: refactor
ticket: none


OisinKyne:
What's the reason for this? Not sure I prefer it

xenowits:
> What's the reason for this? Not sure I prefer it

it was a misunderstanding

PR #748, opened 2022-06-25
dkg/sync: implement await shutdown

dB2510:
Implements await shutdown methods on sync.Client and sync.Server

category: feature
ticket: #684

### DIFF
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -35,6 +35,7 @@ import (
 type result struct {
 	rtt       time.Duration
 	timestamp string
+	shutdown  bool
 	error     error
 }
 
@@ -44,6 +45,7 @@ type Client struct {
 	tcpNode   host.Host
 	server    p2p.Peer
 	results   chan result
+	stream    network.Stream
 }
 
 // AwaitConnected blocks until the connection with the server has been established or returns an error.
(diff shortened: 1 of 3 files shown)


dkg/sync/client.go
```diff
@@ -65,71 +67,84 @@ func (c *Client) AwaitConnected() error {
...
+	return result{
+		rtt:       time.Since(before),
+		timestamp: resp.SyncTimestamp.String(),
+	}
+}
+
+func (c *Client) send(msg *pb.MsgSync) (*pb.MsgSyncResponse, error) {
+	wb, err := proto.Marshal(msg)
+	if err != nil {
+		return nil, errors.Wrap(err, "unmarshal msg")
```
xenowits, line 109:
```suggestion
		return nil, errors.Wrap(err, "marshal msg")
```

dkg/sync/server.go
```diff
@@ -69,7 +70,18 @@ func (s *Server) AwaitAllConnected() error {
...
+	for len(msgs) < len(s.peers) {
+		select {
+		case <-s.ctx.Done():
+			return s.ctx.Err()
+		case msg := <-s.receiveShutdown:
+			msgs = append(msgs, msg)
+		}
+	}
+
+	log.Info(s.ctx, "All Clients Shutdown Successfully 🎉", z.Any("clients", len(msgs)))
```
xenowits, line 84:
```suggestion
	log.Info(s.ctx, "All clients shutdown successfully 🎉", z.Any("clients", len(msgs)))
```

dkg/sync/server.go
```diff
@@ -124,6 +137,28 @@ func NewServer(ctx context.Context, tcpNode host.Host, peers []p2p.Peer, defHash
...
+				resBytes, err := proto.Marshal(resp)
+				if err != nil {
+					log.Error(ctx, "Marshal server response", err)
+					return
+				}
+
+				_, err = s.Write(resBytes)
+				if err != nil {
+					log.Error(ctx, "Send response to client", err, z.Any("client", p2p.PeerName(pID)))
+					return
```
xenowits, line 155: discussion: why not return the `error` as well?

xenowits (approved):

dkg/sync/server.go
dB2510: stream handler function doesn't return error, that's why only log is there

Issue #752, opened 2022-06-29
`create dkg` doensn't verify ENRs


OisinKyne:
## Problem to be solved

We managed to do a cluster creation with an invalid ENR supplied by Chris

## Proposed solution

Don't let this happen, verify the legitimacy of ENRs before lettting `create dkg` process run

PR #757, opened 2022-06-30
cmd/createdkg: verify enr in create dkg

dB2510:
Verifies operator ENRs to be in standard form at the start of create dkg. Also make frost dkg as the default one.

category: refactor
ticket: #752

### DIFF
--- a/cmd/createdkg.go
+++ b/cmd/createdkg.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	crand "crypto/rand"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path"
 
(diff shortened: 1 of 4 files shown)


cmd/createdkg.go
```diff
@@ -90,7 +92,11 @@ func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
 	}()
 
 	var operators []cluster.Operator
-	for _, opENR := range conf.OperatorENRs {
+	for i, opENR := range conf.OperatorENRs {
+		_, err := p2p.DecodeENR(opENR)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("invalid ENR of operator %d", i))
```
corverroos, line 98: you should just be able to use `z.Int("index",i))` instead of `fmt.Sprintf`

corverroos (approved):
One or two small tests would be awesome

cmd/createdkg.go
dB2510: done

Issue #709, opened 2022-06-14
Add alive and ready monitoring endpoints


corverroos:
## Problem to be solved

Scheduling software needs to be able to determine the health of a charon node. 

## Proposed solution
Follow k8s reference: https://kubernetes.io/docs/reference/using-api/health-checks/

Add the following endpoints to charon monitoring API:
 - `/livez`: returns 200 always
 - `/readyz`: return 200 if beacon node sync status=synced and quorum peers can be pinged (do all calls in parallel and use 1s timeout) .

corverroos:
https://kubernetes.io/docs/reference/using-api/health-checks/ Kubernetes has deprecated `healthz` and recommends `/livez` and `/readyz`

PR #760, opened 2022-07-02
app: add readyz and alivez endpoints to monitoring api

xenowits:
Add `/readyz` and `/livez` monitoring endpoints

category: feature
ticket: #709


app/monitoringapi.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/obolnetwork/charon/app/z"
+	"github.com/obolnetwork/charon/cluster"
+	"github.com/obolnetwork/charon/p2p"
+)
+
+// ready returns a http.Handler which returns 200 when both the beacon node is synced and all quorum peers can be pinged  in parallel within a timeout. Returns 500 otherwise.
+func ready(ctx context.Context, conf Config, life *lifecycle.Manager, tcpNode host.Host, lock cluster.Lock) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// check if beacon node is fully synced
+		syncing := beaconNodeSynced(ctx, conf, life)
```
corverroos, line 39: The comment of this function says "// beaconNodeReady returns true if the beacon node is fully synced." you instantiate "syncing", suggest:
```
if ok := beaconNodeSyncing; !ok {
  w.WriteHeader(http.StatusInternalServerError)
  _, _ = w.Write([]byte("Beacon node not synced"))
}
```

app/monitoringapi.go
```diff
@@ -0,0 +1,121 @@
...
+		p := p
+
+		go func() {
+			for result := range ping.Ping(ctx, tcpNode, p.ID) {
+				if result.Error != nil {
+					pingErrs <- result
+					log.Error(ctx, "Peer ping failed", err, z.Str("peer", p2p.PeerName(p.ID)))
+				} else {
+					pings = append(pings, result)
+					if len(pings) == len(peers)-1 {
```
corverroos, line 101: You cannot access a slice concurrently from different goroutines, pass the results in the channel and let the parent goroutine handle the counting and logic

app/monitoringapi.go
```diff
@@ -0,0 +1,121 @@
...
+			continue // don't ping self
+		}
+
+		p := p
+
+		go func() {
+			for result := range ping.Ping(ctx, tcpNode, p.ID) {
+				if result.Error != nil {
+					pingErrs <- result
+					log.Error(ctx, "Peer ping failed", err, z.Str("peer", p2p.PeerName(p.ID)))
```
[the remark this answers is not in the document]
corverroos: Actually, lets simplify, just do not log anything on ping errors. We already have a ping service that does logging, it will log. This monitoringaoi can just be quiet. No need to use p2p.Sender in this case.

app/monitoringapi.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/lifecycle"
+	"github.com/obolnetwork/charon/cluster"
+	"github.com/obolnetwork/charon/p2p"
+)
+
+// newReadyHandler returns a http.HandlerFunc which returns 200 when both the beacon node is synced and all quorum peers can be pinged  in parallel within a timeout. Returns 500 otherwise.
+func newReadyHandler(ctx context.Context, conf Config, life *lifecycle.Manager, tcpNode host.Host, lock cluster.Lock) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		eth2Svc, _, err := newETH2Client(ctx, conf, life, nil)
```
corverroos, line 37: pass in eth2client rather

app/monitoringapi.go
```diff
@@ -0,0 +1,122 @@
...
+		if err != nil {
+			writeResponse(w, http.StatusInternalServerError, "Failed to get beacon sync state")
+		}
+
+		if syncing {
+			writeResponse(w, http.StatusInternalServerError, "Beacon node not synced")
+			return
+		}
+
+		peers, err := lock.Peers()
```
corverroos, line 52: pass in peers

app/monitoringapi.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	for {
+		select {
+		case res := <-results:
+			if res.Error != nil {
+				return res.Error
+			}
+
+			pings = append(pings, res)
```
corverroos, line 108: suggest just keeping a count variable

app/monitoringapi.go
xenowits: does that mean initializing `eth2Cl` and `peers` in the `wireMonitoringAPI` function and then passing them to the `newReadyHandler` function?

xenowits: done

app/app.go
```diff
@@ -191,9 +191,24 @@ func Run(ctx context.Context, conf Config) (err error) {
...
 	initStartupMetrics()
 
-	wireMonitoringAPI(life, conf.MonitoringAddr, localEnode)
+	pubkeys, err := eth2PubKeys(lock.Validators)
+	if err != nil {
+		return err
+	}
 
-	if err := wireCoreWorkflow(ctx, life, conf, lock, nodeIdx, tcpNode, p2pKey); err != nil {
+	eth2Cl, beaconAddr, err := newETH2Client(ctx, conf, life, pubkeys)
```
corverroos, line 199: rather pass in lock here and calcualte pubkeys inside eth2client

app/app.go
```diff
@@ -191,9 +191,24 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	peerIDs, err := lock.PeerIDs()
+	if err != nil {
+		return err
+	}
+
+	wireMonitoringAPI(ctx, life, conf.MonitoringAddr, localEnode, tcpNode, eth2Cl.(eth2client.NodeSyncingProvider), peerIDs)
```
corverroos, line 209: rather pass in eth2cl directly, and then cast inside and catch and return the error

app/app.go
```diff
@@ -482,17 +508,24 @@ func createMockValidators(pubkeys []eth2p0.BLSPubKey) beaconmock.ValidatorSet {
 
 // wireMonitoringAPI constructs the monitoring API and registers it with the life cycle manager.
 // It serves prometheus metrics, pprof profiling and the runtime enr.
-func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.LocalNode) {
+func wireMonitoringAPI(ctx context.Context, life *lifecycle.Manager, addr string, localNode *enode.LocalNode, tcpNode host.Host, eth2Cl eth2client.NodeSyncingProvider, peerIDs []peer.ID) {
```
corverroos, line 511: move this to `monitoringapi.go` as well

app/monitoringapi.go
```diff
@@ -0,0 +1,105 @@
...
+
+// newReadyHandler returns a http.HandlerFunc which returns 200 when both the beacon node is synced and all quorum peers can be pinged  in parallel within a timeout. Returns 500 otherwise.
+func newReadyHandler(ctx context.Context, tcpNode host.Host, eth2Cl eth2client.NodeSyncingProvider, peerIDs []peer.ID) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		syncing, err := beaconNodeSyncing(ctx, eth2Cl)
+		if err != nil {
+			writeResponse(w, http.StatusInternalServerError, "Failed to get beacon sync state")
+		}
+
+		if syncing {
```
corverroos, line 39: nit: `else if syncing`

app/monitoringapi.go
```diff
@@ -0,0 +1,105 @@
...
+			return
+		}
+
+		err = peersReady(ctx, peerIDs, tcpNode)
+		if err != nil {
+			writeResponse(w, http.StatusInternalServerError, "Couldn't ping all peers")
+			return
+		}
+
+		writeResponse(w, http.StatusOK, "ok")
```
corverroos, line 50: for livez you return "alive", for readyz you return "ok", suggest returning "ok" in both.

app/monitoringapi.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return state.IsSyncing, nil
+}
+
+// peersReady returns nil if all quorum peers can be pinged in parallel within a timeout. Returns error otherwise.
+func peersReady(ctx context.Context, peerIDs []peer.ID, tcpNode host.Host) error {
+	pingCnt := 0
```
corverroos, line 66: can just do `var resultCount int`

app/monitoringapi.go
```diff
@@ -0,0 +1,105 @@
...
+	state, err := eth2Cl.NodeSyncing(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	return state.IsSyncing, nil
+}
+
+// peersReady returns nil if all quorum peers can be pinged in parallel within a timeout. Returns error otherwise.
+func peersReady(ctx context.Context, peerIDs []peer.ID, tcpNode host.Host) error {
```
corverroos, line 65: you need to cancel the context when returning to stop the long lived ping connections:
```
ctx, cancel := context.WithCancel(ctx)
defer cancel()
```

app/monitoringapi.go
```diff
@@ -0,0 +1,105 @@
...
+	for _, pID := range peerIDs {
+		if tcpNode.ID() == pID {
+			continue // don't ping self
+		}
+
+		go func(pID peer.ID) {
+			for result := range ping.Ping(ctx, tcpNode, pID) {
+				results <- result
+
+				break // no retries, just break on first ping result
```
corverroos, line 79: Start comments with a capital.

app/monitoringapi.go
```diff
@@ -0,0 +1,105 @@
...
+			if res.Error != nil {
+				return res.Error
+			}
+
+			pingCnt++
+
+			if pingCnt == len(peerIDs)-1 { // all pings successful
+				return nil
+			}
+		case <-time.After(1 * time.Second):
```
corverroos, line 96: this will reset the timeout after each message...

app/monitoringapi.go
xenowits: makes sense!

app/monitoringapi.go
xenowits: isn't that what we want? we want all pings to be done within this timeout

corverroos: we want total 1 second to get all responses. we do not want to reset the 1second timeout every time we get a response. That can result in `<number of peers>*1s` before final timeout.

xenowits: gotcha! As you suggested, i have replaced `time.After` with `context.WithTimeout()`

app/monitoringapi.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return state.IsSyncing, nil
+}
+
+// peersReady returns nil if all quorum peers can be pinged in parallel within a timeout. Returns error otherwise.
+func peersReady(ctx context.Context, peerIDs []peer.ID, tcpNode host.Host) error {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
corverroos, line 114: can remove `1*`

corverroos (approved):

Issue #764, opened 2022-07-04
scheduler: sync with beacon node clock to mitigate slot timing issues


corverroos:
## Problem to be solved

Even though slot timing (start time) is deterministic, we sometimes get errors from the beacon node due to querying "future" slots.  We do retry, but this can cause "getting stuck in retries" if there is a significant skew in clocks between the charon node and beacon node.

We need a way to "sync" with clock of the beacon node. By continuously adjusting an offset.

## Proposed solution

The only timing related endpoint provided by the beacon API is the "events" streaming API. By streaming "head" we should get notified of new slots, and this should be able to allow us to calculate the offset with the beacon node. Note that Infura API is backed by a cluster, and the node serving the events stream might be different from the node being queried....

PR #766, opened 2022-07-05
core/schedule: implement beacon node clock sync

corverroos:
Implements a beacon node clock sync feature in alpha. It uses median offset of 10 last head events slots. 

category: feature 
ticket: #764 

### DIFF
--- a/core/scheduler/clocksync.go
+++ b/core/scheduler/clocksync.go
@@ -1,0 +1,97 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package scheduler
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"sync"
+	"time"
+
+	eth2client "github.com/attestantio/go-eth2-client"
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
+	"github.com/jonboulle/clockwork"
+
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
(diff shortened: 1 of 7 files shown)


core/scheduler/clocksync.go
```diff
@@ -0,0 +1,97 @@
...
+
+		startTime := genesis.Add(time.Duration(head.Slot) * slotDuration)
+		newOffset := clock.Since(startTime)
+
+		offsets = append(offsets, newOffset)
+		if len(offsets) < offsetCount {
+			return
+		}
+
+		offsets = offsets[len(offsets)-offsetCount:]
```
dB2510, line 67: what does this mean?
need a comment here

core/scheduler/clocksync.go
```diff
@@ -0,0 +1,97 @@
...
+
+		clone := append([]time.Duration(nil), offsets...)
+		sort.Slice(clone, func(i, j int) bool {
+			return clone[i] < clone[j]
+		})
+
+		median := clone[len(clone)/2]
+		syncMedianGauge.Set(medianOffset.Seconds())
+		if median < -maxOffset || median > maxOffset {
```
dB2510, line 76: how can median be negative? If we are querying too early?

core/scheduler/clocksync.go
```diff
@@ -0,0 +1,97 @@
...
+		mu.Lock()
+		medianOffset = median
+		mu.Unlock()
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return func() time.Duration {
+		mu.Lock()
```
dB2510, line 92: why locking and unlocking here?

dB2510 (approved):

core/scheduler/clocksync.go
```diff
@@ -0,0 +1,97 @@
...
+
+	eth2client "github.com/attestantio/go-eth2-client"
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
+	"github.com/jonboulle/clockwork"
+
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+// newClockSyncer returns a function that returns the current median beacon node clock sync offset.
```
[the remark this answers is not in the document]
corverroos: Sure will add: `The clock sync offset is the difference in time we need to add to our clock to sync with the beacon node clock`

core/scheduler/clocksync.go
dB2510: but we are just returning right?

dB2510: and you are locking and unlocking on line 83 and 85

corverroos: > // newClockSyncer returns a function that returns the current median beacon node clock sync offset.

We read the current median offset every time we schedule a slot

PR #769, opened 2022-07-06
feat: add livez monitoring api to prometheus

xenowits:
Add `livez` and `readyz` monitoring endpoints to `prometheus`

category: feature
ticket: none

### DIFF
--- a/app/metrics.go
+++ b/app/metrics.go
@@ -61,8 +61,8 @@ var (
 func initStartupMetrics() {
 	versionGauge.WithLabelValues(version.Version).Set(1)
 	startGauge.SetToCurrentTime()
-	livezGauge.Set(1)
-	readyzGauge.Set(1)
+	livezGauge.Set(0)
+	readyzGauge.Set(0)
 
 	hash, _ := GitCommit()
 	gitGauge.WithLabelValues(hash).Set(1)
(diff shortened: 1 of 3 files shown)


app/metrics.go
```diff
@@ -42,11 +42,19 @@ var (
...
+		Subsystem: "monitoring",
+		Name:      "livez",
+		Help:      "Set to 1 if `/livez` endpoint returns 200, 0 otherwise",
+	})
 )
 
 func initStartupMetrics() {
 	versionGauge.WithLabelValues(version.Version).Set(1)
 	startGauge.SetToCurrentTime()
+	livezGauge.SetToCurrentTime()
```
corverroos, line 57: mmmm, see you own comment: "Set to 1 if `/livez` endpoint returns 200, 0 otherwise"

app/metrics.go
```diff
@@ -42,11 +42,19 @@ var (
 		Name:      "start_time_secs",
 		Help:      "Gauge set to the app start time of the binary in unix seconds",
 	})
+
+	livezGauge = promauto.NewGauge(prometheus.GaugeOpts{
```
corverroos, line 46: you should also add `readyz`

app/metrics.go
corverroos: you are setting it to boot timestamp...

corverroos (approved):

Issue #762, opened 2022-07-04
DKG sync: Move connected logs to server


corverroos:
## Problem to be solved

We currently log "incoming" and "outgoing" connected to peer logs as part of DKG sync protocol. This isn't very useful to users. The previous "connected to X of Y" peers was better. 

## Proposed solution

- Remove the current "Connected to incoming/outgoing peer" logs.
- Replace with approximate `Connected to X of Y peers" in the `Server`.
- Ignore the fact that client connections are not logged, and that we assume that if we have a server connection, that there will also be a client connection.  

### Out of Scope

Implementing some higher layer log aggregator that combines client and server logs.

PR #771, opened 2022-07-07
dkg/sync: move connected logs to server

dB2510:
Moves connected logs to server and removes inbound and outbound logs from client and server.

category: bug
ticket: #762

### DIFF
--- a/dkg/sync/server.go
+++ b/dkg/sync/server.go
@@ -20,6 +20,7 @@ package sync
 import (
 	"context"
 	"encoding/binary"
+	"fmt"
 	"io"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


xenowits (approved):

dkg/sync/server.go
```diff
@@ -109,11 +110,12 @@ func (s *Server) isConnected(pID peer.ID) bool {
 }
 
 // setConnected sets the shared connected state for the peer.
-func (s *Server) setConnected(pID peer.ID) {
+func (s *Server) setConnected(ctx context.Context, pID peer.ID) {
```
corverroos, line 113: another way is to return the length and log with context on the outside

corverroos (approved):

dkg/sync/server.go
dB2510: Done!

Issue #761, opened 2022-07-04
DKG Sync Protocol: Fail server side if client sends wrong definition hash


corverroos:
## Problem to be solved

The server side of the DKG sync protocol only response with an error if the client sends a mismatching cluster definition hash. It doesn't actually fail fast and error and exit the DKG. Only the client fails and exits the DKG when it receives a error response from a server.

Note that this is actually a race condition:
 - Lets say we have 2 nodes, X and Y.
 - X and Y both have sync clients and servers.
 - Both X and Y will fail and exit if their clients connect to the other server and receive a error response.
 - But if X's client connects (and then exits) before Y's, then Y's client will never be able to connect and will hang.
 - So this ticket is a workaround for X failing so fast, that Y can never connect and fail as well.

## Proposed solution

Add a shared `errResponse bool` field to the `sync.Server`. If any `pb.SyncResponse.Error` is set, this field should be set to `true`. Upon call to `AwaitAllConnected`, return an error if `errResponse==true`.

PR #772, opened 2022-07-07
dkg/sync: fail server side if client sends wrong hash_sig

dB2510:
Fails server if client sends wrong definition hash signature by flipping errResponse shared variable.

category: bug
ticket: #761

### DIFF
--- a/dkg/sync/server.go
+++ b/dkg/sync/server.go
@@ -85,7 +85,7 @@ func (s *Server) AwaitAllConnected(ctx context.Context) error {
 	}
 }
 
-// isError checks if there was any error in between the server flow and returns error.
+// isError checks if there was any error in between the server flow.
 func (s *Server) isError() bool {
 	s.mu.Lock()
 	defer s.mu.Unlock()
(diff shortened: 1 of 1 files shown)


dkg/sync/server.go
```diff
@@ -73,6 +74,13 @@ func (s *Server) AwaitAllConnected(ctx context.Context) error {
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-timer.C:
+			s.mu.Lock()
+			if s.errResponse {
+				s.mu.Unlock()
```
xenowits, line 79: why unlock here and then unlock on line 82 again?

xenowits: can simply defer unlock

dB2510: line 79 unlocks the mutex if the condition is true because after that we have to return the error while line 82 unlocks if the condition is false so that other threads can lock mutex

xenowits (approved):

corverroos (approved):

PR #774, opened 2022-07-07
core/scheduler: improve clock sync logging

corverroos:
Rather than including offsets in every slot tick log. Split it to its own periodic (+-1m) debug logs that makes it clear what it means and also includes if the feature is enabled.

category: misc
ticket: #764 

### DIFF
--- a/core/scheduler/clocksync.go
+++ b/core/scheduler/clocksync.go
@@ -31,7 +31,7 @@ import (
 )
 
 const (
-	maxOffset   = time.Second * 2
+	maxOffset   = time.Second * 6 // Half a slot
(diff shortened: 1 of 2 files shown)


core/scheduler/clocksync.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 const (
-	maxOffset   = time.Second * 2
+	maxOffset   = time.Second * 6 // Half a slot
```
corverroos, line 34: Bump this to half a slot, since then it obviously bad!

dB2510 (approved):
LGTM

PR #778, opened 2022-07-08
core/scheduler: tweak clock sync

corverroos:
Few small tweaks to beacon node clock sync:
 - Remove waiting for minimum 10 samples.
 - Fix logging issue (mod vs div)
 - Update metric name
 - Add sync offset panel to Grafana dashboard

category: misc
ticket: #764 

### DIFF
--- a/testutil/compose/define.go
+++ b/testutil/compose/define.go
@@ -257,7 +257,7 @@ func buildLocal(ctx context.Context, dir string) error {
 	log.Info(ctx, "Building local charon binary", z.Str("repo", repo), z.Str("target", target))
 
 	cmd := exec.CommandContext(ctx, "go", "build", "-o", target)
-	cmd.Env = append(os.Environ(), "GOOS=linux", "GOARCH=amd64")
+	cmd.Env = append(os.Environ(), "GOOS=linux", "GOARCH=amd64", "CGO_ENABLED=0")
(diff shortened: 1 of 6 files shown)


corverroos:
image

testutil/compose/define.go
```diff
@@ -257,7 +257,7 @@ func buildLocal(ctx context.Context, dir string) error {
 	log.Info(ctx, "Building local charon binary", z.Str("repo", repo), z.Str("target", target))
 
 	cmd := exec.CommandContext(ctx, "go", "build", "-o", target)
-	cmd.Env = append(os.Environ(), "GOOS=linux", "GOARCH=amd64")
+	cmd.Env = append(os.Environ(), "GOOS=linux", "GOARCH=amd64", "CGO_ENABLED=0")
```
xenowits, line 260: discussion: why do we need to set `CGO_ENABLED` flag here?

xenowits (approved):
Looks good!

dB2510 (approved):
LGTM

testutil/compose/define.go
corverroos: when building go programs, it is best practice to explicitly disable CGO if not required, otherwise some C libraries and stuff is included in the build which can impact compatibility.

Issue #781, opened 2022-07-10
Publish docker image to dockerhub


dB2510:
## Problem to be solved

We are publishing our image to ghcr.io but that might cause authentication problems for some of our users. To solve that we need to publish our docker image to dockerhub to make it accessible to everyone.

## Proposed solution

Publish docker image to dockerhub alongwith ghcr.io. The `build-push-deploy` workflow needs to be updated. Also `charon-distributed-validator-node`, `charon-distributed-validator-cluster` and `obol-docs` repositories needs to be updated with appropriate link.

corverroos:
Success: https://hub.docker.com/r/obolnetwork/charon/tags

PR #782, opened 2022-07-10
github: push charon to official docker hub registry

OisinKyne:
Pushes the charon image to docker hub after ghcr. Needs to be tested by being merged. Potentially we don't want to send dev builds here too and might want to split this to only run on tags not commits to main. 

category: misc
ticket: #781


dB2510 (approved):

corverroos (approved):
Does this build once for each target repository? It would be great if the binaries are built once and then pushed to the different targets...

corverroos (approved):

Issue #768, opened 2022-07-05
Explicit failed duty tracking


corverroos:
## Problem to be solved

We have metrics and logs for when duty is successfully broadcasted/executed. But we do not have metrics and logs if duties are not successful. This is because duties are not managed by a single entity. Duties are just data flowing through the core workflow. Duties fail by getting stuck somewhere in the workflow. 

Examples of how duties can fail:
 - Scheduler: duty never resolved
 - Fetcher: beacon node not available
 - Consensus: Quorum not reached
 - ValidatorAPI: VC not querying duties or submitting signature
 - ParSigEx: Peers down
 - SigAgg: Not enough partial signatures
 - Bcast: Beacon node unavailable

We need to provide logs and metrics that tracks whether each duty is successful or at what point in the workflow it got stuck.

## Proposed solution

Implement a new `tracker` component as a sub package of `core`. It will be registered as a subscriber all other other core component outputs. It can therefore track which steps are completed for each duty. On duty deadline, it can determine where the duty got stuck and the log and instrument that.

## Out of Scope

Do not track "unresolved duties" since that needs digging into internals of Scheduler. Keep tracking to component output for now.

corverroos:
After chatting to Dhruv about "instrumenting cluster participation", we thought that `tracker` can also do that.

So tracker would also use the events (output from validatorapi and parsigex to determine which VCs submitted partial signatures for a duty and instrument that as "who is correctly participating the cluster".

 So tracker should store all events for a duty, not just the latest. It should also include the pubkey.
 
```
type event struct {
  Duty core.Duty
  Pubkey core.Pubkey
  Component component
}

...

var events map[core.Duty][]event
```


Tracker will then analyse and report (log and metrics) two different "rules":
 - failed duties
 - cluster participation

corverroos:
Components and ordering is: Scheduler, Fetcher, Consensus, ValidatorAPI, ParSigEx, ParSigDB, SigAgg.

PR #802, opened 2022-07-15
core/tracker: add tracker component

dB2510:
Adds tracker component with deadliner interface. Tests would be implemented in subsequent PRs alongwith Deadliner implementation.

category: feature
ticket: #768


core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+		scheduler:         "scheduler",
+		fetcher:           "fetcher",
+		consensus:         "consensus",
+		validatorAPI:      "validatorAPI",
+		parSigDBInternal:  "parSigDBInternal",
+		parSigEx:          "parSigEx",
+		parSigDBThreshold: "parSigDBThreshold",
+		sigAgg:            "sigAgg",
+	}[c]
+}
```
corverroos, line 54: can replace with `//go generate stringer -type=component`

core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+			t.analyzeDuty(ctx, duty)
+			t.trimDuty(duty)
+		}
+	}
+}
+
+// storeEvent stores the event as value with the duty as key. Also, adds duty to the deadliner.
+func (t *Tracker) storeEvent(e event) {
+	t.events[e.duty] = append(t.events[e.duty], e)
+	t.deadliner.Add(e.duty)
```
corverroos, line 107: can al oinline in `Run`, since this doesn't do much

core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+
+	events := t.events[duty]
+
+	// Sort in reverse order (see order above).
+	sort.Slice(events, func(i, j int) bool {
+		return events[i].component > events[j].component
+	})
+
+	// Case of failed duties
+	if events[0].component != sigAgg {
```
corverroos, line 123: avoid nesting: 
```
if events[0].component == sigAgg {
  // Duty completed succesfully
  return
}
```

core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+		log.Error(ctx, "Duty stuck", errors.New("duty stuck"), z.Str("component", failedComponent), z.Str("duty", duty.String()))
+
+		return
+	}
+	// TODO(dhruv): Case of cluster participation (duty success)
+}
+
+// trimDuty trims the events for a given duty.
+func (t *Tracker) trimDuty(duty core.Duty) {
+	delete(t.events, duty)
```
corverroos, line 135: can inline since this doesn't do much

core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+		default:
+			t.input <- event{
+				duty:      duty,
+				component: scheduler,
+				pubkey:    pubkey,
+			}
+		}
+	}
+
+	log.Debug(ctx, "Sent events to tracker", z.Str("component", scheduler.string()))
```
corverroos, line 153: this will spam logs, suggest removing all these

core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+// storeEvent stores the event as value with the duty as key. Also, adds duty to the deadliner.
+func (t *Tracker) storeEvent(e event) {
+	t.events[e.duty] = append(t.events[e.duty], e)
+	t.deadliner.Add(e.duty)
+}
+
+// analyzeDuty analyzes the events for a given duty after the duty's deadline is exceeded.
+func (t *Tracker) analyzeDuty(ctx context.Context, duty core.Duty) {
+	// Deadline crossed for this duty.
+	log.Debug(ctx, "Deadline exceeded", z.Str("duty", duty.String()))
```
corverroos, line 113: Suggest only logging failures

core/tracker/tracker.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// storeEvent stores the event as value with the duty as key. Also, adds duty to the deadliner.
+func (t *Tracker) storeEvent(e event) {
+	t.events[e.duty] = append(t.events[e.duty], e)
+	t.deadliner.Add(e.duty)
+}
+
+// analyzeDuty analyzes the events for a given duty after the duty's deadline is exceeded.
+func (t *Tracker) analyzeDuty(ctx context.Context, duty core.Duty) {
```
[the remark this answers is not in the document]
corverroos: Then report via a testable function call:
```
failed, failedComponent, failedMsg := analyseFailedDauty(duty, t.events[duty])
t.reportFailedDuty(duty, failed, failedComponent, failedMsg)
```

core/validatorapi/validatorapi.go
```diff
@@ -217,6 +218,11 @@ func (c *Component) RegisterAwaitBeaconBlock(fn func(ctx context.Context, slot i
 	c.awaitBlockFunc = fn
 }
 
+// RegisterTracker registers the tracker component.
+func (c *Component) RegisterTracker(fn func(context.Context, core.Duty, core.ParSignedDataSet)) {
```
corverroos, line 222: no need for this, can use existing `RegisterParSigDB` (should probably rename to `Subscribe`)

corverroos (approved):

core/interfaces.go
```diff
@@ -207,7 +207,7 @@ func Wire(sched Scheduler,
 		VAPIRegisterAwaitAttestation:    vapi.RegisterAwaitAttestation,
 		VAPIRegisterGetDutyDefinition:   vapi.RegisterGetDutyDefinition,
 		VAPIRegisterPubKeyByAttestation: vapi.RegisterPubKeyByAttestation,
-		VAPIRegisterParSigDB:            vapi.RegisterParSigDB,
+		VAPIRegisterParSigDB:            vapi.Subscribe,
```
corverroos, line 210: should also rename this `VAPIRegisterParSigDB`

core/tracker/tracker.go
```diff
@@ -0,0 +1,252 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case e := <-t.input:
+			// TODO(dhruv): do not store events for expired duties
+			t.events[e.duty] = append(t.events[e.duty], e)
+			t.deadliner.Add(e.duty)
+		case duty := <-t.deadliner.C():
+			failed, failedComponent, failedMsg := analyseFailedDuty(duty, t.events[duty])
+			if failed {
```
corverroos, line 87: you should probably call report in all cases, so you can test positive and negative cases. The report function can just ignore non-failed.

core/tracker/tracker.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+			// TODO(dhruv): Case of cluster participation (duty success)
+			// t.analyseClusterParticipation()
+			delete(t.events, duty)
+		}
+	}
+}
+
+// analyseFailedDuty analyzes the events for a given duty after the duty's deadline is exceeded.
+func analyseFailedDuty(duty core.Duty, events []event) (bool, component, string) {
```
corverroos, line 98: suggest improve godoc to indicate what it returns, which isn't clear from the function or comment

core/tracker/tracker.go
```diff
@@ -0,0 +1,252 @@
...
+	// Sort in reverse order (see order above).
+	sort.Slice(events, func(i, j int) bool {
```
corverroos, line 100: note this modifies the original events map which isn't great, suggest coping the slice before modifying it.

core/tracker/tracker.go
```diff
@@ -0,0 +1,252 @@
...
+		return events[i].component > events[j].component
+	})
+
+	if events[0].component == sigAgg {
```
corverroos, line 104: this will panic if the slice is empty

core/tracker/tracker.go
```diff
@@ -0,0 +1,252 @@
...
+		// Duty completed successfully
+		return false, sigAgg, ""
+	}
+
+	return true, events[0].component + 1, fmt.Sprintf("%s is failed in %s component", duty.String(), (events[0].component + 1).String())
```
corverroos, line 109: I think we can improve the error message. Be more specific. If it got stuck during validatorAPI, we can say "the VC didn't submit a signed duty", if it got stuck at parsigdb then "peers didn't share any/sufficient partial signatures", etc.

core/tracker/tracker.go
```diff
@@ -0,0 +1,252 @@
...
+
+func (*Tracker) reportFailedDuties(core.Duty, component, string) {
+	// TODO(dhruv): instrument failed duty
+}
+
+// SchedulerEvent inputs event from core.Scheduler component.
+func (t *Tracker) SchedulerEvent(_ context.Context, duty core.Duty, defSet core.DutyDefinitionSet) error {
+	for pubkey := range defSet {
+		select {
+		case <-t.quit:
```
corverroos, line 120: should also add case for ctx timeout

core/tracker/tracker.go
dB2510: yeah right, slices are pass by reference

core/tracker/tracker.go
xenowits: Agreed.

We can have specific analysis in a separate PR.

PR #805, opened 2022-07-16
core/deadline: add deadline implementation

dB2510:
Adds implementation for deadliner interface

category: feature
ticket: #768

### DIFF
--- a/core/deadline.go
+++ b/core/deadline.go
@@ -17,6 +17,8 @@ package core
 
 import (
 	"context"
+	"math"
+	"sync"
 	"time"
 
 	eth2client "github.com/attestantio/go-eth2-client"
@@ -45,7 +47,119 @@ type Deadliner interface {
 	C() <-chan Duty
 }
 
+// Deadline implements the Deadliner interface.
+type Deadline struct {
+	mu           sync.Mutex
+	genesisTime  time.Time
+	slotDuration time.Duration
+
+	dutyChan     chan Duty
+	deadlineChan chan Duty
+
+	// duties represents a set since duty deduplication is required.
+	duties map[Duty]bool
+}
+
+// NewDeadliner returns a new instance of Deadline.
+func NewDeadliner(ctx context.Context, eth2Cl slotTimeProvider) (*Deadline, error) {
+	genesis, err := eth2Cl.GenesisTime(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	duration, err := eth2Cl.SlotDuration(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	d := &Deadline{
+		genesisTime:  genesis,
+		slotDuration: duration,
+		dutyChan:     make(chan Duty),
+		deadlineChan: make(chan Duty),
+		duties:       make(map[Duty]bool),
+	}
+
+	go func() {
+		var (
+			deadlineTimer <-chan time.Time
+			minDuty       Duty
+		)
+
+		for {
+			select {
+			case <-ctx.Done():
+				// Killed by parent context
+				return
+			case duty := <-d.dutyChan:
+				if minDuty.Type == DutyUnknown {
+					// Initialise minDuty and deadlineTimer
+					minDuty = duty
+					deadlineTimer = time.After(time.Until(d.deadlineTime(minDuty)))
+				}
+				d.duties[duty] = true
+			case <-deadlineTimer:
+				// Send deadlined duty to receiver
+				d.deadlineChan <- minDuty
+
+				// Delete duty whose deadline has passed
+				delete(d.duties, minDuty)
+
+				// New min duty for next deadline
+				md, ok := d.getMinDuty()
+				minDuty = md
+				if !ok {
+					continue
+				}
+				deadlineTimer = time.After(time.Until(d.deadlineTime(minDuty)))
+			}
+		}
+	}()
+
+	return d, nil
+}
+
+// Add adds a duty to be notified of the deadline.
+func (d *Deadline) Add(duty Duty) {
+	d.dutyChan <- duty
+}
+
+// C returns the deadline channel.
+func (d *Deadline) C() <-chan Duty {
+	return d.deadlineChan
+}
+
+// getMinDuty gets the first duty to process.
+func (d *Deadline) getMinDuty() (Duty, bool) {
+	minDuty := Duty{Slot: math.MaxInt64, Type: dutySentinel}
+	for duty := range d.duties {
+		if duty.Slot < minDuty.Slot || ((duty.Slot == minDuty.Slot) && (duty.Type < minDuty.Type)) {
+			minDuty = duty
+		}
+	}
+
+	if minDuty.Slot == math.MaxInt64 {
+		return minDuty, false
+	}
+
+	return minDuty, true
+}
+
+// deadlineTime returns time at which duty is supposed to hit deadline.
+func (d *Deadline) deadlineTime(duty Duty) time.Time {
+	if duty.Type == DutyExit {
+		// Do not timeout exit duties.
+		return time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
+	}
+
+	start := d.genesisTime.Add(d.slotDuration * time.Duration(duty.Slot))
+	end := start.Add(d.slotDuration * time.Duration(lateFactor))
+
+	return end
+}
+
 // NewDutyDeadlineFunc returns the function that provides duty deadlines.
+// TODO(dhruv): replace "NewDutyDeadlineFunc" with above deadliner implementation wherever used.
 func NewDutyDeadlineFunc(ctx context.Context, eth2Svc eth2client.Service) (func(Duty) time.Time, error) {
 	eth2Cl, ok := eth2Svc.(slotTimeProvider)
 	if !ok {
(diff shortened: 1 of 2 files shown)


core/deadline.go
```diff
@@ -45,7 +47,119 @@ type Deadliner interface {
...
+	go func() {
+		var (
+			deadlineTimer <-chan time.Time
+			minDuty       Duty
+		)
+
+		for {
+			select {
+			case <-ctx.Done():
+				// Killed by parent context
```
xenowits, line 92: can remove this comment

core/deadline.go
```diff
@@ -45,7 +47,119 @@ type Deadliner interface {
...
+	}
+
+	if minDuty.Slot == math.MaxInt64 {
+		return minDuty, false
+	}
+
+	return minDuty, true
+}
+
+// deadlineTime returns time at which duty is supposed to hit deadline.
```
xenowits, line 148:
```suggestion
// deadlineTime returns the deadline time of the duty.
```

core/deadline.go
```diff
@@ -45,7 +47,119 @@ type Deadliner interface {
...
+			case <-deadlineTimer:
+				// Send deadlined duty to receiver
+				d.deadlineChan <- minDuty
+
+				// Delete duty whose deadline has passed
+				delete(d.duties, minDuty)
+
+				// New min duty for next deadline
+				md, ok := d.getMinDuty()
+				minDuty = md
```
xenowits, line 110: move this after the `if !ok` check

core/deadline.go
```diff
@@ -45,7 +47,119 @@ type Deadliner interface {
...
+			minDuty       Duty
+		)
+
+		for {
+			select {
+			case <-ctx.Done():
+				// Killed by parent context
+				return
+			case duty := <-d.dutyChan:
+				if minDuty.Type == DutyUnknown {
```
xenowits, line 95: why do you assume that the first duty to get into `dutyChan` would be a `DutyUnknown`?

core/deadline.go
dB2510: sure

core/deadline.go
dB2510: here i am checking minDuty for its zero value to initialise minDuty with first duty obtained

core/deadline.go
dB2510: sure

xenowits (approved):

core/deadline.go
```diff
@@ -34,18 +36,135 @@ type slotTimeProvider interface {
...
+			}
+		}
+	}()
+
+	return d, nil
+}
+
+// Add adds a duty to be notified of the deadline.
+func (d *Deadline) Add(duty Duty) {
+	d.dutyChan <- duty
```
corverroos, line 128: This will block forever if the ctx was cancelled, suggest using quit pattern.

core/deadline.go
```diff
@@ -34,18 +36,135 @@ type slotTimeProvider interface {
...
+			case <-ctx.Done():
+				return
+			case duty := <-d.dutyChan:
+				if minDuty.Type == DutyUnknown {
+					// Initialise minDuty and deadlineTimer
+					minDuty = duty
+					deadlineTimer = time.After(time.Until(d.deadlineTime(minDuty)))
+				}
+				d.duties[duty] = true
+			case <-deadlineTimer:
```
corverroos, line 104: Note that this will block forever when DutyExit is selected as minDuty....

core/deadline.go
```diff
@@ -34,18 +36,135 @@ type slotTimeProvider interface {
...
+
+	if minDuty.Slot == math.MaxInt64 {
+		return minDuty, false
+	}
+
+	return minDuty, true
+}
+
+// deadlineTime returns the deadline time of the duty.
+func (d *Deadline) deadlineTime(duty Duty) time.Time {
```
corverroos, line 154: do not duplicate this logic, suggest passing in `deadlineFunc` as argument to constructor.

core/deadline.go
```diff
@@ -34,18 +36,135 @@ type slotTimeProvider interface {
...
+	}
+
+	start := d.genesisTime.Add(d.slotDuration * time.Duration(duty.Slot))
+	end := start.Add(d.slotDuration * time.Duration(lateFactor))
+
+	return end
+}
+
 // NewDutyDeadlineFunc returns the function that provides duty deadlines.
+// TODO(dhruv): replace "NewDutyDeadlineFunc" with above deadliner implementation wherever used.
```
corverroos, line 167: not sure I agree, I would rather say we keep deadlineFunc as it is super simple, and Deadliner wraps deadlineFunc.

corverroos (changes requested):
Blocking since DutyExit will break this

core/deadline.go
dB2510: sure, i was thinking the same initially but was confused

core/deadline.go
dB2510: sure

core/deadline.go
corverroos: So you basiaclly need to update the current deadline timer if a duty is received with an earlier deadline. So if you are currently on super long DutyExit deadline, and you receive DutyAttestation for the next slot (which has earlier deadline), then you need to replace that deadline with the new duty.

suggest:
```
var (
  currentTimer     *time.Timer
  currentDuty       Duty
  currentDeadline time.Time
)

    case duty := <-d.dutyChan:
       deadline := deadlineFunc(duty)
       if currDeadline.After(deadline) {
           // Replace with newer deadline
           currentTimer.Stop()
           currentDuty = duty
	   currentTimer = time.NewTimer(time.Until(d.deadlineFunc(duty)))
	   currentDeadline = deadline
       }
```

core/deadline.go
```diff
@@ -34,17 +34,130 @@ type slotTimeProvider interface {
...
+		dutyDeadline := deadlineFunc(duty)
+		if currDeadline.After(dutyDeadline) {
+			currDuty = duty
+			currDeadline = dutyDeadline
+		}
+	}
+
+	if currDuty.Type == DutyUnknown {
+		return Duty{}, false
+	}
```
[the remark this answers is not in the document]
corverroos: you can call this on startup (with empty duty map) to initialise the variables.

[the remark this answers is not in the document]
corverroos:
```
setCurrState := func() {
			currDuty, currDeadline = getCurrDuty(duties, deadlineFunc)
			if !currDeadline.IsZero {
			  currTimer.Stop()
			}
			currTimer = time.NewTimer(time.Until(currDeadline))
		}
```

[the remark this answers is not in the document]
dB2510: Done!

core/deadline.go
```diff
@@ -34,17 +34,130 @@ type slotTimeProvider interface {
...
+			select {
+			case <-ctx.Done():
+				return
+			case duty := <-d.dutyChan:
+				if currDeadline.IsZero() {
+					// Initialise if there are no duties.
+					setCurrState(duty)
+				} else if currDeadline.After(d.deadlineFunc(duty)) {
+					// Set current deadline to the latest one.
+					currTimer.Stop()
```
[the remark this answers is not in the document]
dB2510: yup

core/deadline.go
```diff
@@ -34,17 +34,106 @@ type slotTimeProvider interface {
...
 	// It should only be called by a single goroutine.
 	C() <-chan Duty
 }
 
+// Deadline implements the Deadliner interface.
+type Deadline struct {
+	dutyChan     chan Duty
+	deadlineChan chan Duty
+	quit         chan struct{}
+	deadlineFunc func(Duty) time.Time
```
corverroos, line 55: don't think you need deadlineFunc here actually.

core/deadline.go
```diff
@@ -34,17 +34,106 @@ type slotTimeProvider interface {
...
+		currDuty, currDeadline := getCurrDuty(duties, d.deadlineFunc)
+		currTimer := time.NewTimer(time.Until(currDeadline))
+
+		defer func() {
+			close(d.quit)
+			currTimer.Stop()
+		}()
+
+		setCurrState := func() {
+			if !currDeadline.IsZero() {
```
corverroos, line 80: think we can remove this if statement now, since currDeadline is never zero, or rather, currTimer is always non-nil.

core/deadline.go
```diff
@@ -34,17 +34,106 @@ type slotTimeProvider interface {
...
+				currTimer.Stop()
+			}
+			currDuty, currDeadline = getCurrDuty(duties, d.deadlineFunc)
+			currTimer = time.NewTimer(time.Until(currDeadline))
+		}
+
```
corverroos, line 86: add a TODO here maybe: `optimise getCurrDuty and updating current state if earlier deadline detected`

core/deadline.go
```diff
@@ -34,17 +34,106 @@ type slotTimeProvider interface {
...
+	dutyChan     chan Duty
+	deadlineChan chan Duty
+	quit         chan struct{}
+	deadlineFunc func(Duty) time.Time
+}
+
+// NewDeadliner returns a new instance of Deadline.
+// It runs a goroutine which is responsible for reading and storing duties,
+// and sending the deadlined duty to receiver's deadlineChan.
+func NewDeadliner(ctx context.Context, deadlineFunc func(Duty) time.Time) (*Deadline, error) {
```
corverroos, line 61: can remove error response param, since never non-nil.

PR #806, opened 2022-07-18
core/fetcher: add hardcoded graffiti

dB2510:
Adds "Obol Distributed Validator" as graffiti in every beacon block by charon which would be replaced with graffiti obtained from cluster-lock.json

category: misc
ticket: none

### DIFF
--- a/core/fetcher/fetcher.go
+++ b/core/fetcher/fetcher.go
@@ -149,9 +149,10 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, defSet core
 
 		randao := randaoData.Signature().ToETH2()
 
-		// TODO(dhruv): what to do with graffiti?
-		// passing empty graffiti since it is not required in API
+		// TODO(dhruv): replace hardcoded graffiti with the one from cluster-lock.json
 		var graffiti [32]byte
+		s := []byte("Obol Distributed Validator")
+		copy(graffiti[:], s)
 		block, err := f.eth2Cl.BeaconBlockProposal(ctx, eth2p0.Slot(uint64(slot)), randao, graffiti[:])
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


core/fetcher/fetcher.go
```diff
@@ -149,9 +149,10 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, defSet core
 
 		randao := randaoData.Signature().ToETH2()
 
-		// TODO(dhruv): what to do with graffiti?
-		// passing empty graffiti since it is not required in API
+		// TODO(dhruv): replace hardcoded graffiti with the one from cluster-lock.json
 		var graffiti [32]byte
+		s := []byte("Obol Distributed Validator")
+		copy(graffiti[:], s)
```
xenowits, line 155:
```suggestion
		copy(graffiti[:], []byte("Obol Distributed Validator"))
```

xenowits (approved):

corverroos (approved):

core/fetcher/fetcher.go
```diff
@@ -149,9 +149,10 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, defSet core
 
 		randao := randaoData.Signature().ToETH2()
 
-		// TODO(dhruv): what to do with graffiti?
-		// passing empty graffiti since it is not required in API
+		// TODO(dhruv): replace hardcoded graffiti with the one from cluster-lock.json
 		var graffiti [32]byte
+		s := []byte("Obol Distributed Validator")
```
OisinKyne, line 154: Proposal. Checkout the default grafitti for  all of the clients running here https://codebeautify.org/string-hex-converter

Would it be long term  more  value  add to make the string  `charon/version-sha` so we can see what versions of old clusters are still out there and live as they produce blocks?

OisinKyne (approved):

PR #808, opened 2022-07-19
core/tracker: add tests for tracker component

xenowits:
- Add unit tests for tracker component.
- Add `reportDutyFunc` to `NewTracker`.

It will be followed by a PR for wiring tracker along with the integration tests.

category: test
ticket: #768

### DIFF
--- a/core/tracker/tracker.go
+++ b/core/tracker/tracker.go
@@ -58,15 +58,20 @@ type Tracker struct {
 	events    map[core.Duty][]event
 	deadliner core.Deadliner
 	quit      chan struct{}
+
+	// reportDutyFunc instruments the duty. It ignores non-failed duties.
+	reportDutyFunc func(core.Duty, bool, string, string)
 }
 
 // NewTracker returns a new Tracker.
-func NewTracker(deadliner core.Deadliner) *Tracker {
+// TODO(xenowits): implement reportDutyFunc to emit metrics.
+func NewTracker(deadliner core.Deadliner, reportDutyFunc func(failedDuty core.Duty, isFailed bool, component string, msg string)) *Tracker {
 	t := &Tracker{
-		input:     make(chan event),
-		events:    make(map[core.Duty][]event),
-		quit:      make(chan struct{}),
-		deadliner: deadliner,
+		input:          make(chan event),
+		events:         make(map[core.Duty][]event),
+		quit:           make(chan struct{}),
+		deadliner:      deadliner,
+		reportDutyFunc: reportDutyFunc,
 	}
 
 	return t
@@ -82,13 +87,13 @@ func (t *Tracker) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return ctx.Err()
 		case e := <-t.input:
-			// TODO(dhruv): Do not store events for expired duties.
+			// TODO(xenowits): Do not store events for expired duties.
 			t.events[e.duty] = append(t.events[e.duty], e)
 			t.deadliner.Add(e.duty)
 		case duty := <-t.deadliner.C():
-			failed, failedComponent, failedMsg := analyseFailedDuty(duty, t.events[duty])
+			failed, failedComponent, failedMsg := analyseDutyFailed(duty, t.events[duty])
 
-			t.reportFailedDuty(duty, failed, failedComponent, failedMsg)
+			t.reportDutyFunc(duty, failed, failedComponent.String(), failedMsg)
 
 			// TODO(dhruv): Case of cluster participation (duty success)
 			// t.analyseClusterParticipation()
@@ -97,9 +102,9 @@ func (t *Tracker) Run(ctx context.Context) error {
 	}
 }
 
-// analyseFailedDuty analyzes the events for a deadlined duty. It returns true if the duty didn't complete the sigagg component.
+// analyseDutyFailed analyzes the events for a deadlined duty. It returns true if the duty didn't complete the sigagg component.
 // If it failed, it also returns the component it failed at and a human friendly error message.
-func analyseFailedDuty(duty core.Duty, es []event) (bool, component, string) {
+func analyseDutyFailed(duty core.Duty, es []event) (bool, component, string) {
 	events := make([]event, len(es))
 	copy(events, es)
 
@@ -117,12 +122,9 @@ func analyseFailedDuty(duty core.Duty, es []event) (bool, component, string) {
 		return false, sigAgg, ""
 	}
 
+	// TODO(xenowits): Improve message to have more specific details.
+	//  Ex: If the duty got stuck during validatorAPI, we can say "the VC didn't successfully submit a signed duty").
 	return true, events[0].component + 1, fmt.Sprintf("%s failed in %s component", duty.String(), (events[0].component + 1).String())
-}
-
-// reportFailedDuty logs and instruments the duty. It ignores non-failed duties.
-func (*Tracker) reportFailedDuty(core.Duty, bool, component, string) {
-	// TODO(dhruv): instrument failed duty
 }
 
 // SchedulerEvent inputs event from core.Scheduler component.
@@ -186,13 +188,13 @@ func (t *Tracker) ConsensusEvent(ctx context.Context, duty core.Duty, data core.
 }
 
 // ValidatorAPIEvent inputs events from core.ValidatorAPI component.
-func (t *Tracker) ValidatorAPIEvent(ctx context.Context, duty core.Duty, data core.ParSignedDataSet) {
+func (t *Tracker) ValidatorAPIEvent(ctx context.Context, duty core.Duty, data core.ParSignedDataSet) error {
 	for pubkey := range data {
 		select {
 		case <-ctx.Done():
-			return
+			return ctx.Err()
 		case <-t.quit:
-			return
+			return nil
 		default:
 			t.input <- event{
 				duty:      duty,
@@ -201,6 +203,8 @@ func (t *Tracker) ValidatorAPIEvent(ctx context.Context, duty core.Duty, data co
 			}
 		}
 	}
+
+	return nil
 }
 
 // ParSigExEvent inputs event from core.ParSigEx component.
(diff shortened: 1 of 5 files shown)


core/tracker/tracker.go
```diff
@@ -82,13 +95,16 @@ func (t *Tracker) Run(ctx context.Context) error {
...
 			return ctx.Err()
 		case e := <-t.input:
-			// TODO(dhruv): Do not store events for expired duties.
-			t.events[e.duty] = append(t.events[e.duty], e)
-			t.deadliner.Add(e.duty)
+			// Do not store events for expired duties.
+			if time.Now().Before(t.deadlineFunc(e.duty)) {
+				t.events[e.duty] = append(t.events[e.duty], e)
+				t.deadliner.Add(e.duty)
+				log.Println(e.component)
```
dB2510, line 102: remove this log, it will spam the logs

core/tracker/tracker.go
```diff
@@ -58,18 +59,29 @@ type Tracker struct {
...
+
+	deadlineFunc   func(duty core.Duty) time.Time
+	reportDutyFunc func(core.Duty, bool, string, string)
 }
 
 // NewTracker returns a new Tracker.
-func NewTracker(deadliner core.Deadliner) *Tracker {
+// TODO(xenowits): implement reportDutyFunc to emit metrics.
+func NewTracker(ctx context.Context, deadlineFunc func(core.Duty) time.Time, reportDutyFunc func(core.Duty, bool, string, string)) (*Tracker, error) {
+	deadliner, err := core.NewDeadliner(ctx, deadlineFunc)
```
corverroos, line 70: pass in deadliner, we should not need both deadliner and deadlineFunc, so remove deadlineFunc

core/tracker/tracker.go
```diff
@@ -58,15 +58,20 @@ type Tracker struct {
...
 	quit      chan struct{}
+
+	// reportDutyFunc instruments the duty. It ignores non-failed duties.
+	reportDutyFunc func(core.Duty, bool, string, string)
 }
 
 // NewTracker returns a new Tracker.
-func NewTracker(deadliner core.Deadliner) *Tracker {
+// TODO(xenowits): implement reportDutyFunc to emit metrics.
+func NewTracker(deadliner core.Deadliner, reportDutyFunc func(failedDuty core.Duty, isFailed bool, component string, msg string)) *Tracker {
```
corverroos, line 68: `NewTracker` should not take `reportDutyFunc` as param, it should set the production function. `NewTrackerForT` should take it as a param, since we only override in testing.

core/tracker/tracker.go
```diff
@@ -97,9 +102,9 @@ func (t *Tracker) Run(ctx context.Context) error {
 	}
 }
 
-// analyseFailedDuty analyzes the events for a deadlined duty. It returns true if the duty didn't complete the sigagg component.
+// analyseDutyFailed analyzes the events for a deadlined duty. It returns true if the duty didn't complete the sigagg component.
```
corverroos, line 105: suggest not renaming since this analysis is specifically for "failed duties", the other analysis will be for "duty participation"

core/tracker/tracker_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func TestNewTracker(t *testing.T) {
+	duty, defSet := trackerHelper(t)
+
+	deadlineFunc := func(startTime time.Time) func(core.Duty) time.Time {
+		return func(duty core.Duty) time.Time {
+			duration := time.Millisecond
+			lateFactor := 1
+
+			if duty.Type == core.DutyExit {
```
corverroos, line 40: you do not need this logic in tests if you do not actually actaully test DutyExit.

core/tracker/tracker_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/obolnetwork/charon/core"
+	"github.com/obolnetwork/charon/core/tracker"
+	"github.com/obolnetwork/charon/testutil"
+)
+
+func TestNewTracker(t *testing.T) {
+	duty, defSet := trackerHelper(t)
+
+	deadlineFunc := func(startTime time.Time) func(core.Duty) time.Time {
```
corverroos, line 35: why pass in starttime? seems like you always use `time.Now()`, so can just:
```
deadlineFuncProvider := func() func(core.Dity) time.Time {
  startTime := time.Now()
  return  func(core.Dity) time.Time {
    ...
  } 
}
```

core/tracker/tracker.go
xenowits: clarification: `analyseFailedDuty` seems like "analysing a failed duty" which means it already assumes that the input duty is a failed one, which maynot be the case.

Shouldn't the function signature be 
`func analyseFailedDuty(core.Duty, []event) (component, string)`

instead of 

`func analyseFailedDuty(core.Duty, []event) (bool, component, string)`

core/tracker/tracker_test.go
corverroos: also since your duration is so small, you can simplify this a lot by just returning time.Time{}, which should result in immediate timeout...?

xenowits: Yeah, but i'm actually passing a `DutyExit` in the test

core/tracker/tracker_test.go
```diff
@@ -0,0 +1,154 @@
...
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
+	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/stretchr/testify/require"
+
+	"github.com/obolnetwork/charon/core"
+	"github.com/obolnetwork/charon/core/tracker"
+	"github.com/obolnetwork/charon/testutil"
+)
+
+func TestNewTracker(t *testing.T) {
```
corverroos, line 32: you are not testing NewTracker, you are testing the whole Tracker

core/tracker/tracker_test.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+
+			start := startTime.Add(duration * time.Duration(duty.Slot))
+			end := start.Add(duration * time.Duration(lateFactor))
+
+			return end
+		}
+	}
+
+	pubkey := testutil.RandomCorePubKey(t)
```
corverroos, line 52: suggest adding this stuff to helper, suggest renaming helper to `setupData`

core/tracker/tracker_test.go
```diff
@@ -0,0 +1,154 @@
...
+		reportDutyFunc := func(failedDuty core.Duty, isFailed bool, component string, msg string) {
+			require.Equal(t, duty, failedDuty)
+			require.False(t, isFailed)
+			require.Equal(t, "sigAgg", component)
+			cancel()
+		}
+
+		tr := tracker.NewTracker(deadliner, reportDutyFunc)
+
+		go func() {
```
corverroos, line 100: this is going to flap, since there is not gaurentee that timeout only happens AFTER you added all the events. Relying on timeing is bad. Make it explicit, but providing a customer deadliner implemenation where you call C explicitly.

corverroos (changes requested):

core/tracker/tracker_test.go
xenowits: What about reducing it to a single line?
```go
	deadlineFuncProvider := func() func(duty core.Duty) time.Time {
		return func(duty core.Duty) time.Time {
			return time.Now().Add(time.Millisecond * time.Duration(duty.Slot))
		}
	}
```

xenowits: Resolving this conversation as we now use a mock deadliner implementation.

core/tracker/tracker_test.go
xenowits: Understood!

core/tracker/tracker.go
corverroos: So if we have two different analysis:
 - Detect if a duty failed or succeeded and if it failed indicate the reason
 - Detect which peers participated in a duty
 
 What would you call those two different analysis and what would the signatures be?

xenowits: Probably i'd go for:

```go
// analyseDutyFailed detects if a duty failed. It returns true if the duty failed and returns the component
// where it failed and a message why.
func analyseDutyFailed(core.Duty, []event) (bool, component, string) {}

// analysePeerParticipation returns the pubkeys of the peers who participated in the duty.
func analysePeerParticipation(core.Duty, []event) []core.Pubkey {}
```

corverroos: sounds good

nit:
```
// analyseParticipation returns the share indexes of peers that participated in this duty.
func analyseParticipation(core.Duty, []event) []int {}
```

core/tracker/tracker.go
```diff
@@ -58,20 +58,32 @@ type Tracker struct {
...
+		quit:           make(chan struct{}),
+		deadliner:      deadliner,
+		reportDutyFunc: reportDutyFunc,
 	}
 
 	return t
 }
 
+// NewForT returns a new Tracker for use in tests.
+func NewForT(deadliner core.Deadliner, reportDutyFunc func(failedDuty core.Duty, isFailed bool, component string, msg string)) *Tracker {
```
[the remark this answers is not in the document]
corverroos: not sure making component a string is worth having external tests, maybe the tests should be internal, then we also do not need the NewForT function, the test can directly override the reportDutyFunc.

core/tracker/tracker.go
```diff
@@ -58,20 +58,32 @@ type Tracker struct {
...
+	t := New(deadliner)
+	t.reportDutyFunc = reportDutyFunc
```
corverroos, line 82: nit: please rename `reportDutyFunc` to something like `failedDutyReporter`

corverroos (approved):
LGTM

core/tracker/tracker.go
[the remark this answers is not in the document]
xenowits: agreed, this could well be an internal test

Issue #809, opened 2022-07-19
Duty Builder Proposer - Builder Spec Implementation


ciaranmcveigh5:
## Problem to be solved

Charon clusters should be compatible with the builder-specs to enable builder/proposer separation. One of the requirements for validators is the implementation of SignedBlindedBeaconBlock https://github.com/ethereum/builder-specs/blob/main/specs/builder.md#signedblindedbeaconblock.
 
## Proposed solution

A new Duty called DutyBuilderProposer, will closely follow the implementation seen in DutyProposer.

PRs
- [x]  Addition of the Duty in core/types.go
- [x] Unsigned data implementation in core/unisgneddata.go
- [x] Signed data implementation in core/signeddata.go
- [x] Validator api implementation in core/validatorapi/validatorapi.go
- [x] Scheduler implementation in core/scheduler/scheduler.go
- [x] Fetcher implementation in core/fetcher/fetcher.go
- [x] Dutydb implementation in core/dutydb/memory.go
- [x] app & cmd implementation of builder flag (not wired up)
- [x] Broadcast implementation in core/bcast/bcast.go
- [x] Protobuff implementation in core/proto.go
- [x] Interfaces implementation in core/interfaces.go
- [x] eth2types validatorapi in core/validatorapi/eth2types.go
- [x] router validatorapi in core/validatorapi/router.go
- [x] blindedbeaconblockproposalprovider in testutil/validatormock/validator mock.go 
- [x] Documentation

### Out of Scope

n/a

corverroos:
Closing to include in v0.10.0 release notes

PR #810, opened 2022-07-19
core: addition of the DutyBuilderProposer in types.go

ciaranmcveigh5:
- Addition of the DutyBuilderProposer in types.go 
- Updated related tests in types_test.go

Unsure of team preference on feature flag.

category: feature
ticket: #809

### DIFF
--- a/core/types.go
+++ b/core/types.go
@@ -31,14 +31,15 @@ type DutyType int
 const (
 	// DutyType enums MUST not change, it will break backwards compatibility.
 
-	DutyUnknown  DutyType = 0
-	DutyProposer DutyType = 1
-	DutyAttester DutyType = 2
-	DutyRandao   DutyType = 3
-	DutyExit     DutyType = 4
+	DutyUnknown         DutyType = 0
+	DutyProposer        DutyType = 1
+	DutyAttester        DutyType = 2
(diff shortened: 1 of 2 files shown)


core/types.go
```diff
@@ -47,11 +48,12 @@ func (d DutyType) Valid() bool {
...
-		DutyUnknown:  "unknown",
-		DutyAttester: "attester",
-		DutyProposer: "proposer",
-		DutyRandao:   "randao",
-		DutyExit:     "exit",
+		DutyUnknown:         "unknown",
+		DutyAttester:        "attester",
+		DutyProposer:        "proposer",
+		DutyRandao:          "randao",
+		DutyBuilderProposer: "builderProposer",
```
ciaranmcveigh5, line 55: Not sure what case type is wanted for "builderProposer"

corverroos:
So feature flags are primarily for "rolling out features" via alpha->beta->stable. In this case, we will probably not release builder-api to stable any time soon, but rather keep it as opt-in feature via its own flag `--enable-builder-api`. 

We could still combine it with a feature flag especially to group PRs in the release notes under "Alpha Features", but we haven't actually done that grouping yet.

So lets skip the feature flag for now, we can add a command line flag when we actually integrate all the code to enable it.

PR #817, opened 2022-07-19
core: unsigned data implementation for duty builder proposer

ciaranmcveigh5:
- unsigned data implementation of blinded blocker proposer
- implementation of clone test
- Addition to random.go to generate random blinded block body and blinded block (required for clone test)

This touches core & testutil not sure on convention for PR header when this is the case - have gone with core for now

category: feature
ticket: #809

### DIFF
--- a/core/unsigneddata.go
+++ b/core/unsigneddata.go
@@ -277,6 +277,13 @@ func UnmarshalUnsignedData(typ DutyType, data []byte) (UnsignedData, error) {
 		}
 
 		return resp, nil
+	case DutyBuilderProposer:
+		var resp VersionedBlindedBeaconBlock
+		if err := json.Unmarshal(data, &resp); err != nil {
+			return nil, errors.Wrap(err, "unmarshal block")
+		}
+
+		return resp, nil
 	default:
 		return nil, errors.New("unsupported unsigned data duty type")
 	}
--- a/core/unsigneddata.go
+++ b/core/unsigneddata.go
@@ -18,6 +18,7 @@ package core
 import (
 	"encoding/json"
 
+	eth2api "github.com/attestantio/go-eth2-client/api"
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec"
 	"github.com/attestantio/go-eth2-client/spec/altair"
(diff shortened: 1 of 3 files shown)


core/unsigneddata.go
```diff
@@ -180,6 +182,83 @@ func (b *VersionedBeaconBlock) UnmarshalJSON(input []byte) error {
...
 }
 
+type VersionedBlindedBeaconBlock struct {
+	eth2api.VersionedBlindedBeaconBlock
+}
+
+// NewVersionedBlindedBeaconBlock validates and returns a new wrapped VersionedBlindedBeaconBlock.
+func NewVersionedBlindedBeaconBlock(block *eth2api.VersionedBlindedBeaconBlock) (VersionedBlindedBeaconBlock, error) {
+	switch block.Version {
+	case spec.DataVersionBellatrix:
```
ciaranmcveigh5, line 192: Phase0 and Altair don't exist on the VersionedBlindedBeaconBlock struct so have been omitted from the switch statement

https://github.com/attestantio/go-eth2-client/blob/master/api/versionedblindedbeaconblock.go

core/unsigneddata.go
```diff
@@ -18,6 +18,7 @@ package core
 import (
 	"encoding/json"
 
+	eth2api "github.com/attestantio/go-eth2-client/api"
```
ciaranmcveigh5, line 21: let me know if you'd like this naming (eth2api) changed

core/unsigneddata.go
corverroos: cool

testutil/random.go
```diff
@@ -225,6 +270,25 @@ func RandomExecutionPayLoad() *bellatrix.ExecutionPayload {
 	}
 }
 
+func RandomExecutionPayloadHeader() *bellatrix.ExecutionPayloadHeader {
+	return &bellatrix.ExecutionPayloadHeader{
+		ParentHash:       RandomArray32(),
+		FeeRecipient:     bellatrix.ExecutionAddress{},
+		StateRoot:        RandomArray32(),
+		ReceiptsRoot:     RandomArray32(),
+		LogsBloom:        [256]byte{},
```
corverroos, line 279: you can omit "zero" values, since they are automatically assigned if not specified. So can remove all the `...{}` as well as `0`

corverroos (approved):
LGTM

testutil/random.go
ciaranmcveigh5: cool this was copied from RandomExecutionPayLoad only diff is payload has transactions whereas header has tranactionsRoot

do you want me to omit the zero values on RandomExecutionPayLoad as well?

Issue #710, opened 2022-06-14
Add timestamp to cluster definition file


corverroos:
## Problem to be solved

It is difficult to look at a cluster definition file (or lock file containing a definition) and know if this is the one you want. It is hard to distinguish multiple versions of lock files with similar config (operator and validator counts).

## Proposed solution

Add a `timestamp` field to `cluster_definition` so it is clear when the definition was created. It should be a human friendly string field without strict precision requirements, so RFC3339 is good. Ensure it is added to all hash calculations.

Then also add a cluster version bump and add version logic so parsing older versions is supported. This probably entails keeping older version structs and logic around and using version to switch on that.

## Out of scope

Adding a timestamp the lock file (outside the definition) is hard since that would need to be included in consensus.

OisinKyne:
So would this be included or excluded from the config hash? Would we want to replace UUID with a nanosecond timestamp? Do we want a human readable timestamp format?

dB2510:
If we are adding timestamp field then it can surely replace existing UUID field. Since this would be constant throughout, it would be included as part of config_hash. It can be in any format I think, although I am inclined towards unix timestamp format.

corverroos:
I would say human readable, since the aim is to make a definition/lock file identifiable by looking at it. I want to be able to look at the file and say "ah, yes, this was the one we created during that previous attempt,  this other file is older/newer and isn't the one I am searching for".

We can remove `UUID` if the timestamp is nanosecond granularity. So suggest `RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"`
 
It should be included in the `config_hash` AND `definition_hash` I think.

PR #818, opened 2022-07-19
cluster: add timestamp to definition

corverroos:
Add human readable `Timestamp` to cluster definition (and therefore cluster lock) and bump version to v1.0.1. This is backwards compatible with v1.0.0 (but not vice versa), see `TestBackwardsCompatability`.

category: feature
ticket: #710


cluster/definition.go
```diff
@@ -19,19 +19,26 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
+	"time"
 
 	ssz "github.com/ferranbt/fastssz"
 	"github.com/libp2p/go-libp2p-core/peer"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/z"
```
OisinKyne, line 28: What is this?

OisinKyne (approved):
I thought we agreed to save adding a timestamp to a config file until we had the (version upgrade + config upgrade + cluster upgrade) code working such that we could test that process out. Have many questions about the rollout of this that I am asking here with the aim of improving our release management across org (now that we have Thomas's help). 

- [ ] Should  this be a `v1.1.0` if it's adding a feature not a patch release?
- [ ] Does this work with `create dkg` and `create cluster`? Do we have tests of them in CI such that if we did break them we would find out?
- [ ] Does a v1.0.0 cluster config format work with a v1.0.1 format supporting charon image? 
- [ ] Does an older image of charon handle this unexpected  format successfully? 
- [ ] Does an older image of charon handle this unexpected format with a helpful error message?
- [ ] Can a mix of old and new charon images do a dkg together with the old format?
- [ ] Can a mix of old and new charon images do a dkg together with the new format?
- [ ] Does the api support this structure  on all three endpoints? 
- [ ] Does the launchpad front end react code support this structure for sending  to the API? 
- [ ] Are the obol-docs and charon-docs updated with this change?

cluster/definition.go
dB2510: It is a package that provides support for structured logging and structured errors. For ex:
```
log.Info(ctx, "Charon starting",
		z.Str("version", version.Version),
		z.Str("git_commit_hash", hash),
		z.Str("git_commit_time", timestamp),
	)
```

Issue #814, opened 2022-07-19
Ignore deadlined duties


xenowits:
## Problem to be solved

Currently, we don't check if a duty has deadlined before:
- adding it to the deadliner's duty channel (see core.Deadline.Add(duty))
- adding events to tracker's input channel (see core.Tracker.input)

Note that the `Deadliner` will be used in other components as well (to trim data from state) and they may also have the requirement to ignore expired duties.

So, we need a solution to determine if a duty has deadlined before "using" it.

## Proposed solution

```go

        // Deadline implements the Deadliner interface.
        type Deadline struct {
                // .... Add another field
                // Receivers listen to addedChan to find out if their call to Add(duty) succeeded
	        addedChan    chan bool
        }

	case duty := <-d.dutyChan:
		deadline := deadlineFunc(duty)
		if deadline.Before(time.Now()) {
			d.addedChan <- false
			return
		}

		duties[duty] = true
		d.addedChan <- true

		if deadline.Before(currDeadline) {
			setCurrState()
		}
		duties[duty] = true
		setCurrState()
```

### Out of Scope

None.

PR #819, opened 2022-07-20
core/deadline: ignore deadlined duties

xenowits:
- Modify `Deadliner` to ignore deadlined duties.
- Modify `Add(duty)` to return true if a duty is added successfully.
- Fix deadline tests.

category: refactor
ticket: #814

### DIFF
--- a/core/deadline.go
+++ b/core/deadline.go
@@ -65,10 +65,9 @@ type Deadline struct {
 }
 
 // NewForT returns a Deadline for use in tests.
-func NewForT(t *testing.T, deadlineFunc func(Duty) time.Time, clock clockwork.Clock) (*Deadline, context.CancelFunc) {
+func NewForT(ctx context.Context, t *testing.T, deadlineFunc func(Duty) time.Time, clock clockwork.Clock) *Deadline {
 	t.Helper()
 
-	ctx, cancel := context.WithCancel(context.Background())
 	d := &Deadline{
 		inputChan:    make(chan deadlineInput),
 		deadlineChan: make(chan Duty),
@@ -78,7 +77,7 @@ func NewForT(t *testing.T, deadlineFunc func(Duty) time.Time, clock clockwork.Cl
 
 	go d.run(ctx, deadlineFunc)
 
-	return d, cancel
+	return d
 }
 
 // NewDeadliner returns a new instance of Deadline.
--- a/core/deadline.go
+++ b/core/deadline.go
@@ -17,6 +17,7 @@ package core
 
 import (
 	"context"
+	"testing"
 	"time"
 
 	eth2client "github.com/attestantio/go-eth2-client"
@@ -64,7 +65,10 @@ type Deadline struct {
 }
 
 // NewForT returns a Deadline for use in tests.
-func NewForT(ctx context.Context, deadlineFunc func(Duty) time.Time, clock clockwork.Clock) *Deadline {
+func NewForT(t *testing.T, deadlineFunc func(Duty) time.Time, clock clockwork.Clock) (*Deadline, context.CancelFunc) {
+	t.Helper()
+
+	ctx, cancel := context.WithCancel(context.Background())
 	d := &Deadline{
 		inputChan:    make(chan deadlineInput),
 		deadlineChan: make(chan Duty),
@@ -74,7 +78,7 @@ func NewForT(ctx context.Context, deadlineFunc func(Duty) time.Time, clock clock
 
 	go d.run(ctx, deadlineFunc)
 
-	return d
+	return d, cancel
 }
 
 // NewDeadliner returns a new instance of Deadline.
@@ -133,7 +137,12 @@ func (d *Deadline) run(ctx context.Context, deadlineFunc func(Duty) time.Time) {
 			}
 		case <-currTimer.Chan():
 			// Send deadlined duty to receiver.
-			d.deadlineChan <- currDuty
+			select {
+			case <-ctx.Done():
+				return
+			case d.deadlineChan <- currDuty:
+			}
+
 			delete(duties, currDuty)
 			setCurrState()
 		}
--- a/core/deadline.go
+++ b/core/deadline.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	eth2client "github.com/attestantio/go-eth2-client"
+	"github.com/jonboulle/clockwork"
 
 	"github.com/obolnetwork/charon/app/errors"
 )
@@ -38,20 +39,42 @@ type slotTimeProvider interface {
 // may only be used by a single goroutine. So, multiple instances are required
 // for different components and use cases.
 type Deadliner interface {
-	// Add adds a duty to be notified of the Deadline via C.
-	// Note that duties will be deduplicated and only a single duty will be provided via C.
-	Add(duty Duty)
+	// Add returns true if the duty was added for future deadline scheduling. It is idempotent
+	// and returns true if the duty was previously added and still awaits deadline scheduling. It
+	// returns false if the duty has already expired and cannot therefore be added for scheduling.
+	Add(duty Duty) bool
 
 	// C returns the same read channel every time and contains deadlined duties.
 	// It should only be called by a single goroutine.
 	C() <-chan Duty
 }
 
+// deadlinerInput represents the input to inputChan.
+type deadlineInput struct {
+	duty    Duty
+	success chan<- bool
+}
+
 // Deadline implements the Deadliner interface.
 type Deadline struct {
-	dutyChan     chan Duty
+	inputChan    chan deadlineInput
 	deadlineChan chan Duty
+	clock        clockwork.Clock
 	quit         chan struct{}
+}
+
+// NewForT returns a Deadline for use in tests.
+func NewForT(ctx context.Context, deadlineFunc func(Duty) time.Time, clock clockwork.Clock) *Deadline {
+	d := &Deadline{
+		inputChan:    make(chan deadlineInput),
+		deadlineChan: make(chan Duty),
+		clock:        clock,
+		quit:         make(chan struct{}),
+	}
+
+	go d.run(ctx, deadlineFunc)
+
+	return d
 }
 
 // NewDeadliner returns a new instance of Deadline.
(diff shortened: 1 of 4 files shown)


core/deadline.go
```diff
@@ -45,10 +45,16 @@ type Deadliner interface {
 	// C returns the same read channel every time and contains deadlined duties.
 	// It should only be called by a single goroutine.
 	C() <-chan Duty
+
+	// AddedChan returns the same read channel every time and contains a bool which is set to
+	// true when call to Add(duty) succeeds. It should only be called by a single goroutine.
+	AddedChan() <-chan bool
```
dB2510, line 51: we decided to return bool from Add function. Not adding a new method. remove this

core/deadline.go
```diff
@@ -45,10 +45,16 @@ type Deadliner interface {
...
 }
 
 // Deadline implements the Deadliner interface.
 type Deadline struct {
+	// Consumers listen on addedChan to find out if their call to Add(duty) succeeded.
+	addedChan    chan bool
```
dB2510, line 57: Add function will return the response of this channel. Modified interface:
```
// Deadliner provides duty Deadline functionality. The C method isn’t thread safe and
// may only be used by a single goroutine. So, multiple instances are required
// for different components and use cases.
type Deadliner interface {
	// Add adds a duty to be notified of the Deadline via C.
	// Note that duties will be deduplicated and only a single duty will be provided via C.
        // It also expects a boolean response which signals if added duty is a expired one or not.
	Add(duty Duty) bool

	// C returns the same read channel every time and contains deadlined duties.
	// It should only be called by a single goroutine.
	C() <-chan Duty
}
```

core/deadline.go
xenowits: done!

core/deadline.go
```diff
@@ -87,8 +90,20 @@ func NewDeadliner(ctx context.Context, deadlineFunc func(Duty) time.Time) *Deadl
 			case <-ctx.Done():
 				return
 			case duty := <-d.dutyChan:
+				deadline := deadlineFunc(duty)
+
+				// Ignore expired duties.
+				if deadline.Before(time.Now()) {
+					d.addedChan <- false
+					return
```
dB2510, line 98: replace return with continue with return. return will exit from this goroutine while we still want to add duties

core/deadline.go
```diff
@@ -102,13 +117,18 @@ func NewDeadliner(ctx context.Context, deadlineFunc func(Duty) time.Time) *Deadl
...
-		return
+		return false
 	default:
 		d.dutyChan <- duty
 	}
+
+	select {
+	case ok := <-d.addedChan:
+		return ok
+	}
```
dB2510, line 131: suggest:
```
        select {
	case <-d.quit:
		return false
	default:
		d.dutyChan <- duty
		return <-d.addedChan
	}
```

core/deadline_test.go
```diff
@@ -29,23 +29,18 @@ func TestDeadliner(t *testing.T) {
...
-			// Do not timeout exit duties.
-			return time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
+	deadlineFuncProvider := func() func(duty core.Duty) time.Time {
+		return func(duty core.Duty) time.Time {
+			if duty.Type == core.DutyExit {
+				// Do not timeout exit duties.
+				return time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
+			}
+
+			return time.Now().Add(time.Millisecond * time.Duration(duty.Slot))
```
dB2510, line 39:
```suggestion
	deadlineFuncProvider := func(duty core.Duty) time.Time {
			if duty.Type == core.DutyExit {
				// Do not timeout exit duties.
				return time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
			}

			return time.Now().Add(time.Millisecond * time.Duration(duty.Slot))
```

dB2510: can add custom deadlineFunc which will not depend of any time assumptions and deadline duties in desired order

dB2510: Also add tests to ignore expired duties

xenowits: > can add custom deadlineFunc which will not depend of any time assumptions and deadline duties in desired order

yes, but i don't think it's required in this test as duties are received in the same order as the order in which they are added

core/deadline.go
```diff
@@ -38,9 +38,9 @@ type slotTimeProvider interface {
 // may only be used by a single goroutine. So, multiple instances are required
 // for different components and use cases.
 type Deadliner interface {
-	// Add adds a duty to be notified of the Deadline via C.
+	// Add adds a duty to be notified of the Deadline via C. It returns true if the duty is added successfully.
```
corverroos, line 41: Suggest: `Add returns true if the duty was added for future deadline scheduling. It is idempotent and returns true if the duty was previously added and still awaits deadline scheduling. It returns false if the duty has already expired and cannot therefore be added for scheduling.`

core/deadline.go
```diff
@@ -49,6 +49,8 @@ type Deadliner interface {
 
 // Deadline implements the Deadliner interface.
 type Deadline struct {
+	// Consumers listen on addedChan to find out if their call to Add(duty) succeeded.
+	addedChan    chan bool
```
corverroos, line 53: This isn't going to work. You cannot guarantee which reader will receive the message if multiple are listening. Follow the. pattern in `aggsigdb.MemDB.commands.writeCommand`. but rename to:

```
type deadlineInput struct {
  Duty core.Duty
  Success chan <- bool 
}

// Note we can rename and unexport deadliner implementation since we only ever export the interface.
type deadliner struct {
  input chan deadlineInput
  ...
}
```

core/deadline_test.go
xenowits: > Also add tests to assert ignored expired duties

yes, we do need to also add some expired duties to test that flow, i'm adding that

core/deadline.go
dB2510: Based on this i would suggest:
```
func (d *Deadline) Add(duty Duty) {
	select {
	case <-d.quit:
		return
	default:
               input := deadlineInput{duty: duty, success: make(chan bool)}
               d.input <- input
               return <-input.success
	}
}
```

corverroos: Please look closely at aggsigdb.MemDB.commands.writeCommand, the above will not work

dB2510:
```
func (d *Deadline) Add(duty Duty) bool {
        response := make(chan bool)
	select {
	case <-d.quit:
		return false
	default:
               d.input <- deadlineInput{duty: duty, success: response}
	}

        select {
	case <-d.quit:
		return false
	default:
                return <-response
	}
}
```

corverroos: Question, why not used buffered channel `response := make(chan bool, 1)` like aggsigdb?

dB2510: Buffered channel would be useful to unblock deadliner's central goroutine to not wait for the reader of this channel. In case if add goroutine is exited, central go routine should also be exited not to wait for the reader to read response to add.

We should use buffered channel 👍

corverroos: I think either is actually fine in this specific use-case. "Unbuffered" handover is fine since Add goroutine will always already be listening. AggSigDB uses a buffered channel since its Store goroutine may exit early when its context is closed so might not be listening at all. In this case, `Add` doesn't have a context so it cannot exit early so buffered channel isn't required.

core/deadline_test.go
```diff
@@ -29,48 +31,87 @@ func TestDeadliner(t *testing.T) {
...
+				return startTime.Add(-1 * time.Hour)
+			}
+
+			return startTime.Add(time.Duration(duty.Slot) * time.Second)
 		}
+	}
+
+	deadliner := core.NewForT(ctx, deadlineFuncProvider(), clock)
+
+	// Add our duties to the deadliner.
```
xenowits, line 54: will be adding a `sync.WaitGroup` to wait for these goroutines to finish adding their duties.

xenowits: Done!

core/deadline.go
```diff
@@ -38,77 +39,122 @@ type slotTimeProvider interface {
...
 type Deadline struct {
-	dutyChan     chan Duty
+	inputChan    chan deadlineInput
 	deadlineChan chan Duty
+	clock        clockwork.Clock
 	quit         chan struct{}
 }
 
+// NewForT returns a Deadline for use in tests.
+func NewForT(ctx context.Context, deadlineFunc func(Duty) time.Time, clock clockwork.Clock) *Deadline {
```
corverroos: please add `*testing.T` after ctx

core/deadline_test.go
```diff
@@ -16,56 +16,106 @@
...
+	})
+
+	require.Equal(t, nonExpiredDuties, actualDuties)
+}
+
+// sendDuties runs a goroutine which adds the duties to the deadliner channel.
+func addDuties(t *testing.T, wg *sync.WaitGroup, duties []core.Duty, expected bool, deadliner *core.Deadline) {
+	t.Helper()
+
+	go func(duties []core.Duty, expected bool) {
```
xenowits, line 81: different duties are added in random order

core/deadline_test.go
```diff
@@ -29,43 +32,93 @@ func TestDeadliner(t *testing.T) {
...
 	}
 
-	deadliner := core.NewDeadliner(ctx, deadlineFuncProvider())
+	deadliner := core.NewForT(ctx, deadlineFuncProvider(), clock)
 
-	expectedDuties := []core.Duty{
-		core.NewVoluntaryExit(2),
-		core.NewAttesterDuty(2),
+	wg := &sync.WaitGroup{}
+	wg.Add(3)
```
corverroos, line 53: move wg.Add(1) into `addDuties`, since then it adds and remove itself

core/deadline_test.go
```diff
@@ -29,43 +32,93 @@ func TestDeadliner(t *testing.T) {
...
+
+	// Add our duties to the deadliner.
+	addDuties(t, wg, expiredDuties, false, deadliner)
+	addDuties(t, wg, nonExpiredDuties, true, deadliner)
+	addDuties(t, wg, voluntaryExits, true, deadliner)
+
+	wg.Wait()
+
+	clock.Advance(1 * time.Second)
```
corverroos, line 62: why is this required?

core/tracker/tracker.go
```diff
@@ -86,9 +86,10 @@ func (t *Tracker) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return ctx.Err()
 		case e := <-t.input:
-			// TODO(xenowits): Do not store events for expired duties.
-			t.events[e.duty] = append(t.events[e.duty], e)
-			t.deadliner.Add(e.duty)
+			if t.deadliner.Add(e.duty) {
```
corverroos, line 89: avoid nesting `if !t.deadliner.Add(e.duty) { // Ignore expired duties; continue }`

corverroos (approved):

core/deadline_test.go
xenowits: To initialize the fake timer

PR #820, opened 2022-07-20
core: add changes for signed data for builder proposer

ciaranmcveigh5:
- signed data implementation of blinded blocker proposer
- implementation of sigagg test

An issue with pre commit check specifically golangci-lint "dupl", appeared to be giving a false positive

`core/sigagg/sigagg_test.go:346: 346-400 lines are duplicate of `core/sigagg/sigagg_test.go:261-315` (dupl)`

This issue https://github.com/golangci/golangci-lint/issues/1372 discussed updating the dupl threshold in the config which I've done. Not sure if I'm missing something and this is a legitimate dupl or if this has happened to you guys before and you have a better solution. 

category: feature
ticket: #809

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,9 @@ linters-settings:
   cyclop:
     max-complexity: 15
     skip-tests: true
+  dupl:
+    # tokens count to trigger issue, 150 by default
+    threshold: 350
   exhaustive:
     default-signifies-exhaustive: true
   forbidigo:
--- a/core/sigagg/sigagg_test.go
+++ b/core/sigagg/sigagg_test.go
@@ -20,6 +20,8 @@ import (
 	"crypto/rand"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


.golangci.yml
```diff
@@ -4,6 +4,9 @@ linters-settings:
   cyclop:
     max-complexity: 15
     skip-tests: true
+  dupl:
+    # tokens count to trigger issue, 150 by default
+    threshold: 350
```
ciaranmcveigh5, line 9: scaled upwards in jumps of 50 so threshold at 300 was failing for me

corverroos: yeah, we have duplication in tests. This is fine for now, until we can write some common test code to dedup.

corverroos (approved):

.golangci.yml
corverroos: From [Go Proverbs](https://go-proverbs.github.io/): _A little copying is better than a little dependency_.

PR #822, opened 2022-07-20
core: validator api implementation for builder proposer

ciaranmcveigh5:
- validator api implementation of blinded blocker proposer
- implementation of relevant tests

category: feature
ticket: #809


core/validatorapi/validatorapi.go
```diff
@@ -388,6 +398,104 @@ func (c Component) SubmitBeaconBlock(ctx context.Context, block *spec.VersionedS
...
+		return err
+	}
+
+	err = c.verifyBlindedBlockSignature(ctx, block, pubkey, slot)
+	if err != nil {
+		return err
+	}
+
+	// Save Partially Signed Block to ParSigDB
+	duty := core.NewProposerDuty(int64(slot))
```
corverroos, line 478: `core.NewBuilderProposerDuty`

core/validatorapi/validatorapi.go
```diff
@@ -388,6 +398,104 @@ func (c Component) SubmitBeaconBlock(ctx context.Context, block *spec.VersionedS
...
+	ctx = log.WithCtx(ctx, z.Any("duty", duty))
+
+	log.Debug(ctx, "Beacon block submitted by validator client")
```
corverroos, line 481: "Blinded beacon block ..."

core/validatorapi/validatorapi.go
```diff
@@ -483,6 +591,29 @@ func (c Component) verifyBlockSignature(ctx context.Context, block *spec.Version
...
+		return err
+	}
+
+	var sig eth2p0.BLSSignature
+	if block.Version == spec.DataVersionBellatrix {
+		if block.Bellatrix.Signature == sig {
+			return errors.New("no bellatrix signature")
+		}
+		sig = block.Bellatrix.Signature
+	}
```
corverroos, line 606: should probably add a default error case (`verifyBlockSignature` also needs one).

eth2util/signing/signing.go
```diff
@@ -29,10 +29,11 @@ import (
 type DomainName string
 
 const (
-	DomainBeaconProposer DomainName = "DOMAIN_BEACON_PROPOSER"
-	DomainBeaconAttester DomainName = "DOMAIN_BEACON_ATTESTER"
-	DomainRandao         DomainName = "DOMAIN_RANDAO"
-	DomainExit           DomainName = "DOMAIN_VOLUNTARY_EXIT"
+	DomainBeaconProposer        DomainName = "DOMAIN_BEACON_PROPOSER"
+	DomainBeaconBuilderProposer DomainName = "DOMAIN_BEACON_BUILDER_PROPOSER"
```
corverroos, line 33: mmm not sure about this. These domains are defined in the eth2 spec (it isn't our own thing). I guess blinded blocks are signed with normal  `DomainBeaconProposer` actually. Can you confirm? If so, you can then remove this.

corverroos (approved):

eth2util/signing/signing.go
ciaranmcveigh5: yep you're correct https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md double checked on the eth research discord and they confirmed "DOMAIN_BEACON_PROPOSER" is to be used on both

PR #827, opened 2022-07-21
core/scheduler: add support for duty builder proposer

ciaranmcveigh5:
- scheduler implementation for builder proposer

category: feature
ticket: #809

### DIFF
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -245,9 +245,17 @@ func (s *Scheduler) resolveDuties(ctx context.Context, slot slot) error {
 		return err
 	}
 
-	err = s.resolveProDuties(ctx, slot, vals)
-	if err != nil {
-		return err
+	// TODO: replace if false with check for builder api flag
+	if false {
+		err = s.resolveBuilderProDuties(ctx, slot, vals)
+		if err != nil {
+			return err
+		}
+	} else {
+		err = s.resolveProDuties(ctx, slot, vals)
+		if err != nil {
+			return err
+		}
 	}
 
 	s.setResolvedEpoch(uint64(slot.Epoch()))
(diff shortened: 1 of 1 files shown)


ciaranmcveigh5:
The proposer duty type originates in Scheduler (in Charon)

As a first pass on the builder api for me it makes sense to exclusively schedule normal blocks or builder blocks

We mentioned having a flag to enable the builder-api this flag would disable normal blocks from being scheduled

thoughts?

core/scheduler/scheduler.go
```diff
@@ -245,9 +245,17 @@ func (s *Scheduler) resolveDuties(ctx context.Context, slot slot) error {
 		return err
 	}
 
-	err = s.resolveProDuties(ctx, slot, vals)
-	if err != nil {
-		return err
+	// TODO: replace if false with check for builder api flag
```
ciaranmcveigh5, line 248: This would be some check on the builder api flag not sure how we want to handle it for now

core/scheduler/scheduler.go
```diff
@@ -317,7 +325,12 @@ func (s *Scheduler) resolveProDuties(ctx context.Context, slot slot, vals valida
 			continue
 		}
 
+		// TODO: replace if false with check for builder api flag
```
ciaranmcveigh5, line 328: We have 2 options for checking for the builder api flag

in the resolveDuties function and implement the resolveBuilderProDuties function in this PR or have one generic resolveProDuties and have the check for the builder api flag in there

In the future i think you'll probably want to schedule both and only execute 1

Some notes from flashbots

- MEV Boost has a BUILDER_PROPOSAL_DELAY_TOLERANCE which is the deadline within the slot for a builder to have proposed a block (External Block Build)
- After this deadline the “External Block Build” will be aborted in favour of a “Local Block Build”
- From MEV Boost documentation, both “External Block Build” and “Local Block Build” should begin as soon as they are able to. The Charon client must ensure one or the other NOT both are submitted.

core/scheduler/scheduler.go
corverroos: yeah, I think you can just add a `builderAPI bool` variable/field to `New` and `Scheduler`. We can add the flag itself in another PR. But it would basically be adding `BuilderAPI bool` field to `app.Config` and wiring it up in the `cmd` and `app` packages.

core/scheduler/scheduler.go
corverroos: I think we schedule just 1 for now to simplify things. Once we know and understand more, we can look at scheduling both and executing one. 

Can switch inside `resolveProDuties`, just flip the duty type basically based on the flag.

corverroos (approved):

Issue #552, opened 2022-05-17
Multiple Beacon Nodes


OisinKyne:
## Problem to be solved

Eth2-client from Jim can natively support multiple beacon nodes. It can also decide to either give the fastest response back or the most profitable response. We don't use these features at all yet, and we should to provide more fault tolerance to each charon client. 

## Proposed solution

Add a flag `--beacon-node-endpoints` that takes an array of strings. Each of these should be used as beacon node clients. 

### Out of Scope

Subsequently, we should configure variants of charon clusters with the most valuable strategy and the fast strategy, and run some tests with sophisticated operators that have access to multiple CL clients for the same testnet. We can see if the `fast` or `best` strategy makes a noticeable difference to each clusters returns. My gut feel would be on fast being more important that best in a DV context.

corverroos:
Just some small comments on the out-of-scope "best" idea:
- Selecting "best" requires using some scoring/rating logic on the unsigned block. This might not be trivial. Or rather, I do not know how to do this (yet).
- In the context of a distributed validator cluster, we cannot control which node's "best" will be selected. So we will not be selecting the "best" of the cluster, only the "best" of one random node.

PR #830, opened 2022-07-21
cmd: support multiple beacon nodes

corverroos:
Adds support for multiple beacon nodes via eth2 multi client. Adds the `--beacon-node-endpoints` flags and deprecates `--beacon-node-endpoint` flag. Updates validatorapi proxy to use active eth2client address.

category: feature
ticket: #552


dB2510 (approved):
LGTM

PR #836, opened 2022-07-21
github: fix changelog generation

corverroos:
Fix automatic changelog generation by doing a deep checkout so full git history is available to `genchangelog`. 

category: misc
ticket: none

### DIFF
--- a/.github/workflows/build-push-deploy.yml
+++ b/.github/workflows/build-push-deploy.yml
@@ -11,7 +11,8 @@ jobs:
     name: Build Docker Image
     steps:
     - uses: actions/checkout@v3
-
+      with:
+        fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/build-push-deploy.yml
```diff
@@ -11,7 +11,8 @@ jobs:
     name: Build Docker Image
     steps:
     - uses: actions/checkout@v3
-
+      with:
+        fetch-depth: 0
```
xenowits, line 15: So, it means "fetch all history for all branches and tags"

xenowits (approved):

.github/workflows/build-push-deploy.yml
corverroos: yup

PR #837, opened 2022-07-21
core: proto implementation for builder proposer

ciaranmcveigh5:
- proto implementation for builder proposer
- relevant tests

category: feature
ticket: #809

### DIFF
--- a/core/proto.go
+++ b/core/proto.go
@@ -54,6 +54,12 @@ func ParSignedDataFromProto(typ DutyType, data *pbv1.ParSignedData) (ParSignedDa
 			return ParSignedData{}, errors.Wrap(err, "unmarshal block")
 		}
 		signedData = b
+	case DutyBuilderProposer:
+		var b VersionedSignedBlindedBeaconBlock
+		if err := json.Unmarshal(data.Data, &b); err != nil {
+			return ParSignedData{}, errors.Wrap(err, "unmarshal blinded block")
+		}
+		signedData = b
 	case DutyExit:
 		var e SignedVoluntaryExit
 		if err := json.Unmarshal(data.Data, &e); err != nil {
--- a/core/proto_test.go
+++ b/core/proto_test.go
@@ -20,6 +20,8 @@ import (
 	"math/rand"
 	"testing"
 
+	eth2api "github.com/attestantio/go-eth2-client/api"
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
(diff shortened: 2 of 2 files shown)


core/proto.go
```diff
@@ -54,6 +54,12 @@ func ParSignedDataFromProto(typ DutyType, data *pbv1.ParSignedData) (ParSignedDa
 			return ParSignedData{}, errors.Wrap(err, "unmarshal block")
 		}
 		signedData = b
```
ciaranmcveigh5, line 56: i looked for this ParSignedDataFromProto function in the repo and this is it's only declaration so not sure if this needs to be implemented here

core/proto_test.go
```diff
@@ -110,6 +124,22 @@ func TestSetBlockSig(t *testing.T) {
 	require.NotEqual(t, clone.Signature(), block.Signature())
 }
 
+func TestSetBlindedBlockSig(t *testing.T) {
```
ciaranmcveigh5, line 127: may be wrong but this isn't testing anything in the proto.go file as far as i can see

should it be included it's copied from the TestSetBlockSig

corverroos: yeah, these tests should move to `signeddata_test` since it tests whether `SignedData` implementation is immutable and that `SetSignature` returns a new instance.

core/proto_test.go
```diff
@@ -64,6 +66,18 @@ func TestParSignedDataSetProto(t *testing.T) {
...
 		},
+		{
+			Type: core.DutyBuilderProposer,
+			Data: core.VersionedSignedBlindedBeaconBlock{
+				VersionedSignedBlindedBeaconBlock: eth2api.VersionedSignedBlindedBeaconBlock{
+					Version: spec.DataVersionBellatrix,
+					Bellatrix: &eth2v1.SignedBlindedBeaconBlock{
+						Message:   testutil.RandomBellatrixBlindedBeaconBlock(t),
+						Signature: testutil.RandomEth2Signature(),
+					},
```
corverroos, line 77: can extract this to `testutil.RandomCoreVersionSignedBlindedBeaconBlock` is you want to decrease creating these from scratch every time.

core/proto.go
corverroos: `ParSignedDataFromProto ` is only called from `ParSignedData<Set>FromProto` which is called from parsigex to unmarshal protos received over the wire. But yeah, adding it here is correct.

corverroos (approved):

PR #839, opened 2022-07-22
core/app: configure fallback beacon node address

dB2510:
Adds a fallback beacon node address to mitigate no beacon node active error from go-eth2-client in case of a single beacon node endpoint.

category: bug
ticket: none

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -478,13 +478,7 @@ func proxyHandler(eth2Cl eth2client.Service, fallbackBeaconAddr string) (http.Ha
 func getBeaconNodeAddress(eth2Cl eth2client.Service, fallbackBeaconAddr string) (*url.URL, error) {
 	addr := eth2Cl.Address()
 	if addr == "none" {
-		// eth2multi returns "none" if no active clients.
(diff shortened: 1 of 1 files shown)


core/validatorapi/router.go
```diff
@@ -475,11 +475,16 @@ func proxyHandler(eth2Cl eth2client.Service) (http.HandlerFunc, error) {
...
 	addr := eth2Cl.Address()
 	if addr == "none" {
 		// eth2multi returns "none" if no active clients.
-		return nil, errors.New("no active beacon node clients")
+		fallbackURL, err := url.Parse(fallbackBeaconAddr)
+		if err != nil {
+			return nil, errors.Wrap(err, "invalid beacon node address", z.Str("address", addr))
+		}
+
+		return fallbackURL, nil
```
corverroos, line 487: can replace with:
```
if addr == "none" {
  addr := fallbackBeaconAddr
}
```

corverroos (approved):

core/validatorapi/router.go
dB2510: sure

PR #840, opened 2022-07-22
app/eth2wrap: only use eth2multi if required

corverroos:
Only use eth2multi is multiple beacon nodes configured. This mitigates an open issue in eth2multi when only a single node is configured, see https://github.com/attestantio/go-eth2-client/issues/20.

category: bug
ticket: #552

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -28,7 +28,6 @@ import (
 	eth2client "github.com/attestantio/go-eth2-client"
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2http "github.com/attestantio/go-eth2-client/http"
-	eth2multi "github.com/attestantio/go-eth2-client/multi"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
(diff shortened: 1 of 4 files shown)


app/app.go
```diff
@@ -66,6 +65,8 @@ import (
 	"github.com/obolnetwork/charon/testutil/validatormock"
 )
 
+const eth2ClientTimeout = time.Second * 2
```
corverroos, line 68: make timeout explicit (instead of using default eth2client value)

xenowits (approved):
LGTM!

app/eth2wrap/genwrap/genwrap.go
```diff
@@ -49,13 +50,28 @@ import (
...
 
+// Interface assertions
+var (
+    _ eth2client.Service = (*Service)(nil)
+
+	{{range .Providers}} _ eth2client.{{.}} = (*Service)(nil)
+    {{end -}}
+)
+
+type eth2Provider interface {
```
corverroos, line 61: had to introduce this interface since we cannot rely on implementation anymore since we use either eth2http or eth2multi.

dB2510 (approved):

Issue #798, opened 2022-07-13
Implement v1 priority protocol


corverroos:
## Problem to be solved

Charon clusters need a way to "to make cluster wide decisions". This can be used to for many different use-cases, from picking a fork, to enabling cluster wide versions/upgrades to adding and removing operators.  

The priority protocol has been designed to support all these use-cases by basically aggregating and resolving "cluster wide priorities". This is achieved by each node proposing its priorities (per topic). And then for the protocol to calculate cluster wide priorities  proposed by quorum peers. Topics and priorities are just strings, so can be used for many use cases.

See [Design Doc](https://docs.google.com/document/d/1dJI9yoX_iU93ngy5fvfm86RH21y-lMfULbzb9vAQclk/edit#) for more details.

## Proposed solution

Implement the prioritiser component as per design doc.

### Out of Scope

Integration with core workflow and use-cases of the priority protocol is not included yet.

PR #843, opened 2022-07-22
core/priority: implement cluster priority calculation

corverroos:
Implements the prioritiser's deterministic cluster wide priority function. 

category: feature
ticket: #798


core/priority/calculate.go
```diff
@@ -0,0 +1,158 @@
...
+// priorities given the priorities of each peer. Priorities are included in the
+// result if minRequired peers provided them and are ordered by number of peers
+// then by overall priority.
+//
+// Peer messages are validated such that
+// they do not contain duplicate peers, they contain identical slots,
+// individual peers may not contain duplicate topics, individual topics
+// may not contain duplicate priorities or more than 1000 priorities.
+//nolint:gocognit // Mostly just map reduce.
+func calculateResults(msgs []*pbv1.PriorityMsg, minRequired int) (*pbv1.PriorityResult, error) {
```
dB2510, line 41: It would be great if you mention the formula that you are using to calculate result in comments

core/priority/calculate.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+		}
+
+		// Order by score decreasing
+		sort.Slice(allPriorities, func(i, j int) bool {
+			return scores[allPriorities[i]] > scores[allPriorities[j]]
+		})
+
+		// Extract scores with min required count.
+		minScore := (minRequired - 1) * countWeight
```
dB2510, line 123: why "-1"?

dB2510 (approved):
LGTM

core/priority/calculate.go
dB2510, line 41: `minRequired` can be renamed to quorum

core/priority/calculate.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+
+	return &pbv1.PriorityResult{
+		Msgs:   input,
+		Topics: topicResults,
+	}, nil
+}
+
+// newDeduper returns a new generic named deduplicater.
+func newDeduper[T comparable](name string) func(t T) error {
```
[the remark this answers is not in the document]
corverroos: because duplicates are invalid, it means someone is trying to cheat.

core/priority/calculate.go
```diff
@@ -0,0 +1,158 @@
...
+		input      []*pbv1.PriorityMsg
+		dedupPeers = newDeduper[string]("peer") // Peers may not be duplicated
+	)
+	for _, msg := range msgs {
+		if slot == 0 {
+			slot = msg.Slot
+		} else if msg.Slot != slot {
+			return nil, errors.New("mismatching slots")
+		}
+		if err := dedupPeers(msg.PeerId); err != nil {
```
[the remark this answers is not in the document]
corverroos: cool

core/priority/calculate.go
corverroos: because we subtract priority, all peers will have a score less than `count * countWeight`

core/priority/calculate.go
corverroos: technically this algo doesn't know about quorum or consensus or qbft, it just needs a minimum threshold. So I do not want to create coupling to things outside the algo.

corverroos: Isn't this is?  That is all you need to know I feel
```
Priorities are included in the
// result if minRequired peers provided them and are ordered by number of peers
// then by overall priority.
```

xenowits (approved):
Looks good!

PR #847, opened 2022-07-25
core/validatorapi: validator api builder proposer 

ciaranmcveigh5:
- Additional changes for validator api
- Router changes for blinded routes

category: feature
ticket: #809

### DIFF
--- a/core/validatorapi/router_internal_test.go
+++ b/core/validatorapi/router_internal_test.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"testing"
 
+	eth2api "github.com/attestantio/go-eth2-client/api"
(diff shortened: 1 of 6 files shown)


core/validatorapi/router_internal_test.go
```diff
@@ -458,6 +459,29 @@ func TestRouter(t *testing.T) {
 		testRouter(t, handler, callback)
 	})
 
+	t.Run("submit_blinded_block_bellatrix", func(t *testing.T) {
```
corverroos, line 462: need to also add a testcase for router's `proposeBlindedBlock` 🙏

corverroos (approved):

PR #850, opened 2022-07-25
docs: architecture md update for builder proposer

ciaranmcveigh5:
- Architecture document update

category: feature
ticket: #809


ciaranmcveigh5:
Few areas in the architecture doc where builderproposer could be added - not sure how much detail you want added / if you'd rather have the architecture doc focus on the main flow


for example context about the cmd flag and that only the DutyProposer will be acted on when the flag is not active and only DutyBuilderProposer will be acted on when it is active

corverroos (approved):

dB2510 (approved):
Screenshot 2022-07-26 at 1 12 26 PM
Screenshot 2022-07-26 at 1 12 58 PM

Few indentation corrections.

Issue #848, opened 2022-07-25
Refactor cluster definition and lock fields to hex


corverroos:
## Problem to be solved

We currently use `golang/json` package's default formatting of `bytes` in our cluster definition and cluster lock json objects which is base64 standard.

This is causing problems for DV-launchpad as it uses the definition config hash for routing to its backend and base64 standard isn't URL safe.

## Proposed solution

The following formats are available to choose from:
 - hex: We leave hex for "eth2" things like group pubkey and ENRs. We want to use a different format for charon/obol specific things so users are less easily confused. Like thining the DV pubkey is one of the pubshares since those would also look like addresses.
 - base64 URL: base64 URL addresses the problem of using base64 in URLs be replacing "/" with "_" and "+" with "-". This could work but the resulting encoding still contain "special characters" which can be hard to work and look at for humans. 
 - base58: base58 is both URL safe and human friendly. It only contains alpha numeric characters exluding look-alikes like "I" and "l". Base58 is also a common format in other crypto currencies, but not ETH (yet).
 
 Final proposal:
  - We therefore suggest refactoring all byte fields from base64 standard to )x prefixed hex since that is the Ethereum standard which simplifies the config and aligns it with our industry.
  - These fields include:
      - `cluster-definition.config_hash`
      - `cluster-definition.definition_hash`
      - `cluster-definition.operators.config_signature`
      - `cluster-definition.operators.enr_signature`
      - `cluster-lock.lock_hash`
      - `cluster-lock.signature_aggregate`
      - `cluster-lock.distributed_validators.public_shares`
  - Increase the version from 1.1.0 to 1.2.0. 
  - Since this is only cosmetic (the data is unchanged), the new logic should be backwards compatible with v1.0.0 and v1.1.0 files. Note the new file format v1.2.0 will not be compatible with old logic.
  - This can be achieved by created 2 sets of "json formatters", e.g. `lockJSON_v1_1` (for old) and `lockJSON_v1_2` (for new). All child formatters (definition, operators, distributed validators)  would require the same.
  - funcs `func unmarshalLock_v1_1(*Lock, []byte) error` should then implemented to contain the separate logic.
  - Add backwards compatibility unit and smoke tests

### Out of Scope

No other config upgrades included.

corverroos:
@Oisín Kyne I think we should actually upgrade `cluster-definition.operators.config_signature` and 
[...]

PR #851, opened 2022-07-25
cluster: refactor base64 fields to base58

corverroos:
Introduces **draft** cluster config version `1.2.0`. EIP712 signatures are refactored to industry standard hex. Other base64 fields are refactored to base58 to be URL safe and easy for humans to work with.

`v1.2.0` will contain more upgrades, so not "released" yet.

category: feature
ticket: #848


cluster/definition.go
```diff
@@ -87,7 +92,7 @@ type Definition struct {
 	Version string
 
 	// Timestamp is the human readable timestamp of this definition.
-	// Note this was added in v1.0.1, so may be empty for older versions.
+	// Note this was added in v1.1.0, so may be empty for older versions.
```
xenowits, line 95:
```suggestion
	// Note that this was added in v1.1.0, so may be empty for older versions.
```

cluster/definition.go
```diff
@@ -246,27 +285,51 @@ func (d Definition) MarshalJSON() ([]byte, error) {
...
-		DKGAlgorithm:        d.DKGAlgorithm,
-		ForkVersion:         d.ForkVersion,
-		Operators:           d.Operators,
-		ConfigHash:          configHash[:],
-		DefinitionHash:      defHash[:],
-	})
-	if err != nil {
-		return nil, errors.Wrap(err, "marshal lock")
-	}
+	if isJSONv1x1(d.Version) {
```
dB2510, line 288: switch case would be more cleaner + readable imo

cluster/distvalidator.go
```diff
@@ -75,3 +80,90 @@ func (v DistValidator) PublicKey() (*bls_sig.PublicKey, error) {
...
+type distValidatorJSONv1x1 struct {
+	PubKey              string   `json:"distributed_public_key"`
+	PubShares           [][]byte `json:"public_shares,omitempty"`
+	FeeRecipientAddress string   `json:"fee_recipient_address,omitempty"`
+}
+
+// distValidatorJSONv1x2 is the json formatter of DistValidator for versions v1.2.0 and later.
+type distValidatorJSONv1x2 struct {
+	PubKey              ethHex   `json:"distributed_public_key"`
+	PubShares           []base58 `json:"public_shares,omitempty"`
```
dB2510, line 94: why are we changing format of PubShares?
Since PubShares are internal public keys of individual co-validator we won't be using it anyways anywhere. Seems like a premature change.

we can keep it the same

dB2510 (approved):
some small comments

dB2510:
need to run `go mod tidy` for pre-commits to pass

cluster/definition.go
```diff
@@ -30,13 +30,18 @@ import (
...
 
 const (
-	definitionVersion = "v1.1.0"
-	dkgAlgo           = "default"
+	currentVersion = v1_1
+	dkgAlgo        = "default"
+
+	v1_2 = "v1.2.0" // WIP
+	v1_1 = "v1.1.0"
+	v1_0 = "v1.0.0"
```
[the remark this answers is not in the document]
corverroos: yeah, good idea

cluster/definition.go
corverroos: fair

cluster/distvalidator.go
corverroos: we are upgrading all base64 types to either base58 or hex. It is in the ticket.

cluster/definition.go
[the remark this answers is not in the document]
corverroos: actually not sure this is going to work, since we need to do string comparison, which doesn't work nice with enums.

PR #857, opened 2022-07-26
app: wire tracker

corverroos:
Wire tracker in core workflow. Implement failed duty reporter.

Also fix concurrency deadlock between Tracker and Deadliner.

category: feature
ticket: #768

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
       - run: docker pull consensys/teku:latest
-      - run: go test -timeout=5m -v github.com/obolnetwork/charon/app -integration -slow
+      - run: go test -timeout=5m github.com/obolnetwork/charon/app -integration -slow
 
   compose_tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
       - run: docker pull consensys/teku:latest
-      - run: go test -timeout=5m github.com/obolnetwork/charon/app -integration -slow
+      - run: go test -v -timeout=5m github.com/obolnetwork/charon/app -integration -slow
 
   compose_tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
       - run: docker pull consensys/teku:latest
-      - run: go test -timeout=5m github.com/obolnetwork/charon/app -integration -slow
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
       - run: docker pull consensys/teku:latest
-      - run: go test -timeout=5m github.com/obolnetwork/charon/app -integration -slow
+      - run: go test -timeout=5m -v github.com/obolnetwork/charon/app -integration -slow
```
xenowits, line 47: discussion: any reason why we need verbose logs?

app/app.go
```diff
@@ -321,6 +322,11 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 		pubshares = append(pubshares, eth2Share)
 	}
 
+	peers, err := lock.Peers()
```
xenowits, line 325: i think `peers` is needed as a param for `wireTracker` so why not colocate them?

core/tracker/tracker.go
```diff
@@ -67,11 +67,11 @@ type Tracker struct {
 	deadliner core.Deadliner
 	quit      chan struct{}
 
-	// failedDutyReporter instruments the duty. It ignores non-failed duties.
-	failedDutyReporter func(core.Duty, bool, string, string)
+	// failedDutyReporter instruments duty failures.
```
xenowits, line 70: discussion: what's the preferred approach when declaring function literals? omitting the names of the parameters or explicitly naming the parameters?

core/tracker/tracker.go
```diff
@@ -146,9 +150,18 @@ func analyseDutyFailed(duty core.Duty, es []event) (bool, component, string) {
 	return true, events[0].component + 1, fmt.Sprintf("%s failed in %s component", duty.String(), (events[0].component + 1).String())
 }
 
-// failedDutyReporter instruments the duty. It ignores non-failed duties.
-// TODO(xenowits): Implement logic for reporting duties.
-func failedDutyReporter(core.Duty, bool, string, string) {}
+// failedDutyReporter instruments failed duties.
```
xenowits, line 153: logs and instruments failed duties?

xenowits (approved):
Looks good!

.github/workflows/test.yml
corverroos: should remove sorry, it was failing so was just for deubbing.

core/tracker/tracker.go
corverroos: only explicitly name parameters if it isn't obvious from the types what they mean. In this case `bool, string, string` doesn't indicate what they mean.

core/tracker/tracker.go
corverroos: I would argue instrumentation is a umbrella term term that includes metrics and logs and tracing.

app/app.go
corverroos: in general yes, but this is such low signal code, I put it at the start with all the "prep" variables.

PR #867, opened 2022-07-27
app: push readyz in background

dB2510:
This change pushed readyz metrics every 1 second. Earlier we were pushing this metric only when someone calls readyz endpoint. If no-one calls that endpoint it will show inactive on readyz grafana panel.

category: bug
ticket: #880

### DIFF
--- a/app/monitoringapi.go
+++ b/app/monitoringapi.go
@@ -35,7 +35,7 @@ import (
 
 // wireMonitoringAPI constructs the monitoring API and registers it with the life cycle manager.
 // It serves prometheus metrics, pprof profiling and the runtime enr.
-func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.LocalNode, tcpNode host.Host, eth2Svc eth2client.Service, peerIDs []peer.ID) error {
+func wireMonitoringAPI(ctx context.Context, life *lifecycle.Manager, addr string, localNode *enode.LocalNode, tcpNode host.Host, eth2Svc eth2client.Service, peerIDs []peer.ID) error {
 	mux := http.NewServeMux()
 
 	// Serve prometheus metrics
@@ -72,6 +72,24 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
 
 	life.RegisterStart(lifecycle.AsyncBackground, lifecycle.StartMonitoringAPI, httpServeHook(server.ListenAndServe))
 	life.RegisterStop(lifecycle.StopMonitoringAPI, lifecycle.HookFunc(server.Shutdown))
+
+	go func() {
+		ticker := time.NewTicker(time.Second)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				syncing, err := beaconNodeSyncing(ctx, eth2Cl)
+				if err != nil || syncing || peersReady(ctx, peerIDs, tcpNode) != nil {
+					readyzGauge.Set(0)
+				} else {
+					readyzGauge.Set(1)
+				}
+			}
+		}
+	}()
 
 	return nil
 }
--- a/app/monitoringapi.go
+++ b/app/monitoringapi.go
@@ -35,6 +35,12 @@ import (
 	"github.com/obolnetwork/charon/app/lifecycle"
 )
 
+var (
+	errReadyPingFailing      = errors.New("couldn't ping all peers")
+	errReadySyncing          = errors.New("beacon node not synced")
+	errReadyBeaconNodeFailed = errors.New("failed to get beacon sync state")
+)
+
 // wireMonitoringAPI constructs the monitoring API and registers it with the life cycle manager.
 // It serves prometheus metrics, pprof profiling and the runtime enr.
 func wireMonitoringAPI(ctx context.Context, life *lifecycle.Manager, addr string, localNode *enode.LocalNode, tcpNode host.Host, eth2Svc eth2client.Service, peerIDs []peer.ID) error {
@@ -101,23 +107,32 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
 			case <-ctx.Done():
 				return
 			case <-ticker.Chan():
-				mu.Lock()
 				syncing, err := beaconNodeSyncing(ctx, eth2Cl)
 				if err != nil {
-					readyErr = errors.New("failed to get beacon sync state")
+					mu.Lock()
+					readyErr = errReadyBeaconNodeFailed
+					mu.Unlock()
+
 					readyzGauge.Set(0)
 				} else if syncing {
-					readyErr = errors.New("beacon node not synced")
+					mu.Lock()
+					readyErr = errReadySyncing
+					mu.Unlock()
+
 					readyzGauge.Set(0)
 				} else if peersReady(ctx, peerIDs, tcpNode) != nil {
-					readyErr = errors.New("couldn't ping all peers")
+					mu.Lock()
+					readyErr = errReadyPingFailing
+					mu.Unlock()
+
 					readyzGauge.Set(0)
 				} else {
+					mu.Lock()
 					readyErr = nil
+					mu.Unlock()
+
 					readyzGauge.Set(1)
 				}
-
-				mu.Unlock()
 			}
 		}
 	}()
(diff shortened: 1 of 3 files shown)


app/monitoringapi.go
```diff
@@ -73,6 +73,24 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
 	life.RegisterStart(lifecycle.AsyncBackground, lifecycle.StartMonitoringAPI, httpServeHook(server.ListenAndServe))
 	life.RegisterStop(lifecycle.StopMonitoringAPI, lifecycle.HookFunc(server.Shutdown))
 
+	go func() {
+		ticker := time.NewTicker(time.Second)
```
dB2510, line 77: we can discuss on the duration part, what will be the right duration to push metrics

app/monitoringapi.go
```diff
@@ -73,6 +73,24 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
...
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				syncing, err := beaconNodeSyncing(ctx, eth2Cl)
```
dB2510, line 84: avoided logging errors here, since that would spam the logs

app/monitoringapi.go
```diff
@@ -73,6 +73,24 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
 	life.RegisterStart(lifecycle.AsyncBackground, lifecycle.StartMonitoringAPI, httpServeHook(server.ListenAndServe))
 	life.RegisterStop(lifecycle.StopMonitoringAPI, lifecycle.HookFunc(server.Shutdown))
 
+	go func() {
```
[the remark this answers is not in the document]
corverroos: Also add a test for `isRead` please 🙏

corverroos (approved):

app/monitoringapi.go
corverroos: metrics are not pushed, they are pull. So you do not need to update more than pull period, which is 15s. Suggest making this 10s.

app/monitoringapi.go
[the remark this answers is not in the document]
corverroos: Maybe we should ONLY have a async "isReady" resolver, and then `readyz` just reads the latest value..?

app/monitoringapi.go
```diff
@@ -133,19 +158,26 @@ func peersReady(ctx context.Context, peerIDs []peer.ID, tcpNode host.Host) error
...
+		errCount int
 	)
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		case res := <-results:
 			if res.Error != nil {
-				continue
+				errCount++
```
dB2510, line 171: added error counter to return error if are unable to ping quorum number of peers

app/monitoringapi.go
```diff
@@ -76,32 +87,46 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
...
+	)
+	go func() {
+		ticker := clock.NewTicker(10 * time.Second)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.Chan():
+				mu.Lock()
```
corverroos, line 104: remember to decrease mutex locked scope to absolute minimum, and remember to NEVER do IO (disk/network) calls while a lock is held.

app/monitoringapi.go
```diff
@@ -133,19 +158,26 @@ func peersReady(ctx context.Context, peerIDs []peer.ID, tcpNode host.Host) error
 
 	var (
 		// Require quorum successes (excluding self). Formula from IBFT 2.0 paper https://arxiv.org/pdf/1909.10194.pdf
-		require = int(math.Ceil(float64(len(peerIDs)*2)/3)) - 1
-		actual  int
+		require  = int(math.Ceil(float64(len(peerIDs)*2)/3)) - 1
+		actual   int
+		errCount int
```
corverroos, line 163: suggest: `okCount, errCount`

app/monitoringapi_internal_test.go
```diff
@@ -0,0 +1,195 @@
...
+			hosts[i].Peerstore().AddAddrs(hostsInfo[k].ID, hostsInfo[k].Addrs, peerstore.PermanentAddrTTL)
+		}
+	}
+
+	clock := clockwork.NewFakeClock()
+	readyErrFunc := startReadyChecker(ctx, hosts[0], bmock, peers, clock)
+
+	// We wrap the Advance() calls with blockers to make sure that the ticker
+	// can go to sleep and produce ticks without time passing in parallel.
+	clock.BlockUntil(1)
```
corverroos, line 74: BlockUntil only works for calls to `Sleep` and `After`, not Timer.Chan().. so you'll need to another way.

app/monitoringapi_internal_test.go
```diff
@@ -0,0 +1,195 @@
...
+	for {
+		err := readyErrFunc()
+		if err != nil {
+			require.EqualError(t, err, "beacon node not synced")
+			break
+		}
+	}
+}
+
+func TestStartCheckerPingFail(t *testing.T) {
```
corverroos, line 139: lots of duplication in these tests, can't we make it a table test rather?

app/monitoringapi.go
```diff
@@ -76,32 +87,46 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
...
+				mu.Lock()
+				syncing, err := beaconNodeSyncing(ctx, eth2Cl)
+				if err != nil {
+					readyErr = errors.New("failed to get beacon sync state")
+					readyzGauge.Set(0)
+				} else if syncing {
+					readyErr = errors.New("beacon node not synced")
+					readyzGauge.Set(0)
+				} else if peersReady(ctx, peerIDs, tcpNode) != nil {
+					readyErr = errors.New("couldn't ping all peers")
```
corverroos, line 113: suggest using error sentinels (`var errReadyNoPing = errors.New("blah blah")`) and use them for comparison in tests

app/monitoringapi_internal_test.go
dB2510: yeah sure

app/monitoringapi.go
```diff
@@ -56,7 +64,16 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
 		return errors.New("invalid eth2 service")
 	}
 
-	mux.Handle("/readyz", newReadyHandler(tcpNode, eth2Cl, peerIDs))
+	readyErrFunc := startReadyChecker(ctx, tcpNode, eth2Cl, peerIDs, clockwork.NewRealClock())
+	mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+		readyErr := readyErrFunc()
+		if readyErr != nil {
```
corverroos, line 70: can inline

app/monitoringapi.go
```diff
@@ -76,32 +93,55 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
...
-			writeResponse(w, http.StatusInternalServerError, "Failed to get beacon sync state")
-			return
-		} else if syncing {
-			writeResponse(w, http.StatusInternalServerError, "Beacon node not synced")
-			return
+// startReadyChecker returns function which returns an error resulting from ready checks periodically.
+func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client.NodeSyncingProvider, peerIDs []peer.ID, clock clockwork.Clock) func() error {
+	var (
+		mu       sync.Mutex
+		readyErr error
```
corverroos, line 100: suggest initialising the error with non-nil, "errReadyUninit = "ready check uninitialised"

app/monitoringapi.go
```diff
@@ -76,32 +93,55 @@ func wireMonitoringAPI(life *lifecycle.Manager, addr string, localNode *enode.Lo
...
+				if err != nil {
+					mu.Lock()
+					readyErr = errReadyBeaconNodeFailed
+					mu.Unlock()
+
+					readyzGauge.Set(0)
+				} else if syncing {
+					mu.Lock()
+					readyErr = errReadySyncing
+					mu.Unlock()
```
corverroos, line 120: rather do locking in one place.

suggest:
```
		... } else if syncing {
					err = errReadySyncing
					readyzGauge.Set(0)
		} ...
		
		mu.Lock()
		readyErr = err
		mu.Unlock()
```

corverroos (approved):

app/monitoringapi.go
dB2510: yeah but i avoided this because it needs send response as error string which will result in calling readErrFunc() twice and again asking for mutex.

PR #872, opened 2022-07-28
p2p: fix empty allow list

corverroos:
When `--p2p-allowlist` is not defined, it should default to "allow all". But an empty string results in a empty netlist which results in "allow none". This is only now a problem since geth's discv5 previously ignored `NetRestrict`. 

category: bug
ticket: none


corverroos:
See https://github.com/ethereum/go-ethereum/pull/25304.
This should result in these tests passing: https://github.com/ObolNetwork/charon/pull/868

dB2510 (approved):

PR #875, opened 2022-07-28
core/priority: add scores to output

corverroos:
Adds explicit scores to resulting priorities. This solves the tie-breaker (if scores are identical) problem by allowing the user of the priority protocol to decide how to handle ties based on their use-case. 

- This simplifies the protocol, no fancy pseudo random shuffling.
- Users can however still decide to implement fancy pseudo random shuffling themselves if scores are identical
- Or users can use the priority result ordering defined by peers with lowest peer IDs.

category: feature
ticket: #798


.pre-commit/run_buf.sh
```diff
@@ -7,4 +7,4 @@ fi
 
 buf generate
 buf lint
-buf breaking --against '.git#branch=origin/main'
```
corverroos, line 10: disabled this since these are breaking changes, but the protos are still in draft, so this breaking changes are ok.

dB2510: don't forget to uncomment this later

core/corepb/v1/priority.proto
```diff
@@ -23,8 +23,22 @@ message PriorityMsg {
   bytes signature = 9;
 }
 
-// PriorityTopic defines a single peers priorities for a single topic in the Prioritiser protocol.
-message PriorityTopic {
+// PriorityTopic defines a single peers proposed priorities for a single topic in the Prioritiser protocol.
```
dB2510, line 26:
```suggestion
// PriorityTopicProposal defines a single peers proposed priorities for a single topic in the Prioritiser protocol.
```

core/corepb/v1/priority.proto
```diff
@@ -23,8 +23,22 @@ message PriorityMsg {
...
+message PriorityTopicProposal {
   string topic = 1;
   repeated string priorities = 2;
 }
+
+// PriorityTopic defines the cluster wide resulting priorities for a
```
dB2510, line 32:
```suggestion
// PriorityTopicResult defines the cluster wide resulting priorities for a
```

dB2510 (approved):

xenowits (approved):
LGTM!

PR #876, opened 2022-07-28
dkg: write sample file in checkWrites

dB2510:
Replaces writing keystores, cluster-lock and deposit-data with just a sample.txt in checkWrites before starting DKG. This is done to avoid errors if number of operators are less than threshold in cluster-definition.
ref: https://discord.com/channels/849256203614945310/936387516251127838/1001384008715882576

category: bug
ticket: none

### DIFF
--- a/dkg/disk.go
+++ b/dkg/disk.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"os"
 	"path"
+	"strings"
 
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/coinbase/kryptology/pkg/signatures/bls/bls_sig"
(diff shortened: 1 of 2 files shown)


dkg/disk.go
```diff
@@ -127,58 +124,13 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
-
-	// Cleanup sample files
-	if err := os.Remove(path.Join(dataDir, "deposit-data.json")); err != nil {
-		return errors.Wrap(err, "remove sample deposit-data.json")
-	}
-
-	if err := os.RemoveAll(path.Join(dataDir, "/validator_keys")); err != nil {
-		return errors.Wrap(err, "remove keys")
+func checkWrites(dataDir string) error {
+	if err := os.WriteFile(path.Join(dataDir, "sample.txt"), []byte{}, 0o444); err != nil {
```
corverroos, line 128: suggest using the real filenames, can also use nil instead of `[]byte{}` maybe?

dkg/disk.go
```diff
@@ -127,58 +124,37 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
+func checkWrites(dataDir string) error {
+	checkBody := []byte("delete me: dummy file used to check write permissions")
 
-		sig, err := testutil.RandomBLSSignature()
-		if err != nil {
-			return err
-		}
-
-		sigs[pk] = sig
+	// Write all sample files and directories.
```
corverroos, line 130: use a for loop, suggest:
```
files := []striung{"cluster-lock.json", "deposit-data.json", "validator_keys/keystore-0.json"}
for file in file {
  if file contains "/" {
    err := os.MkdirAll(dataDir+filepath.Dir(file))
  }
  err := os.WriteFile(dataDir+filepath)
  err := os.RemoveAll(dataDir+filepath)   
  if file contains "/" {
    err := os.MkdirAll(dataDir+filepath.Dir(file))
  }
  }
}
```

dkg/disk.go
```diff
@@ -127,58 +124,37 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
-		return errors.Wrap(err, "remove sample deposit-data.json")
+	// Remove all sample files and directories.
+	if err := os.RemoveAll(path.Join(dataDir, "validator_keys")); err != nil {
+		return errors.Wrap(err, "check writes")
 	}
 
-	if err := os.RemoveAll(path.Join(dataDir, "/validator_keys")); err != nil {
-		return errors.Wrap(err, "remove keys")
+	if err := os.Remove(path.Join(dataDir, "cluster-lock.json")); err != nil {
+		return errors.Wrap(os.Remove(path.Join(dataDir, "sample.txt")), "remove sample file")
```
corverroos, line 153: best not do fancy stuff when handling errors. Keep it simple, trying to do cleanup adds too much complexity. Also, you are dropping the first error

dkg/disk.go
dB2510: done

dkg/disk.go
dB2510: yeah

dkg/disk.go
```diff
@@ -127,58 +125,26 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
-func checkWrites(dataDir string, def cluster.Definition) error {
-	var shares []share
-	sigs := make(map[core.PubKey]*bls_sig.Signature)
-	for i := 0; i < def.NumValidators; i++ {
-		tss, sks, err := tbls.GenerateTSS(def.Threshold, len(def.Operators), rand.Reader)
-		if err != nil {
-			return err
+func checkWrites(dataDir string) error {
+	checkBody := []byte("delete me: dummy file used to check write permissions")
+	files := []string{"cluster-lock.json", "deposit-data.json", "validator_keys/keystore-0.json"}
```
corverroos, line 130: you might need to support windows as well which has a different path separator. So suggest using `filepath.Join` and `filepath.Dir() != ""` to check for separator below.

dkg/disk.go
```diff
@@ -127,58 +125,26 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
+	for _, file := range files {
+		if strings.Contains(file, "/") {
+			if err := os.MkdirAll(path.Join(dataDir, path.Dir(file)), 0o777); err != nil {
+				return errors.Wrap(err, "mkdir check writes")
```
corverroos, line 134: you should probably add the filename as z.Str to the errors so it is clear what file is failing.

dkg/disk.go
```diff
@@ -127,58 +125,26 @@ func writeDepositData(aggSigs map[core.PubKey]*bls_sig.Signature, withdrawalAddr
...
 		}
-
-		sig, err := testutil.RandomBLSSignature()
-		if err != nil {
-			return err
+		if err := os.Remove(path.Join(dataDir, file)); err != nil {
+			return errors.Wrap(err, "remove file")
+		}
+		if strings.Contains(file, "/") {
+			if err := os.RemoveAll(path.Join(dataDir, path.Dir(file))); err != nil {
```
corverroos, line 144: note that this directory might have existed before with other files, which you ar now deleting... probably not a good idea

corverroos (approved):

dkg/disk.go
corverroos: I would just skip deleting, since we are going to create the folder in anycase after this. Just add a `TODO: should we should delete the validator keys folder if we created it above?`

dkg/disk.go
dB2510: done

dkg/disk.go
dB2510: done

dkg/disk.go
dB2510: done

PR #878, opened 2022-07-29
cmd/createcluster: add fee-recipient and name flags

dB2510:
Adds fee recipient and name flags and use withdrawal address, fee recipient address and name in cluster lock creation. This will be useful for those who want to create a cluster without going through dkg process preferably centralised stakers.

category: refactor
ticket: #879

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -43,7 +43,6 @@ import (
 )
 
 const (
-	clusterName           = "local"
 	defaultWithdrawalAddr = "0x0000000000000000000000000000000000000000"
 	defaultNetwork        = "prater"
 )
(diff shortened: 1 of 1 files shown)


cmd/createcluster.go
```diff
@@ -81,9 +83,11 @@ func newCreateClusterCmd(runFunc func(io.Writer, clusterConfig) error) *cobra.Co
 }
 
 func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
+	flags.StringVar(&config.Name, "name", clusterName, "Optional cosmetic cluster name")
```
corverroos, line 86: Suggest not defaulting to "", suggest using current default that is `clusterName=local`.

cmd/createcluster.go
```diff
@@ -352,7 +356,7 @@ func newLock(conf clusterConfig, dvs []tbls.TSS, peers []p2p.Peer) (cluster.Lock
 		})
 	}
 
-	def := cluster.NewDefinition(clusterName, len(dvs), conf.Threshold, "", "", "", ops, rand.Reader)
+	def := cluster.NewDefinition(conf.Name, len(dvs), conf.Threshold, conf.FeeRecipient, conf.WithdrawalAddr, networkToForkVersion[conf.Network], ops, rand.Reader)
```
corverroos, line 359: split very long lines

cmd/createcluster.go
corverroos: Also, technically not optional if there is a default, suggest removing "Optional" from description.

corverroos (approved):

cmd/createcluster.go
corverroos: Or if empty cluster name is fine, then at least remove `clusterName` const

Issue #849, opened 2022-07-25
Duty Builder Registration - Builder Spec Implementation


ciaranmcveigh5:
## Problem to be solved

Charon clusters should be compatible with the builder-specs to enable builder/proposer separation. One of the requirements for validators is the implementation of SignedValidatorRegistrationV1 https://github.com/ethereum/builder-specs/blob/main/specs/builder.md#signedvalidatorregistrationv1

## Proposed solution

A new Duty called DutyBuilderRegistration will closely follow the implementation seen in DutyExit / DutyRandao.

PRs

- [x] Addition of the Duty in core/types.go
- [x] signeddata implementation
- [x] bcast implementation
- [x] validatorapi implementation
- [x] proto.go
- [x] sigagg.go
- [x] Documentation
### Out of Scope

n/a

corverroos:
Closing to include in v0.10.0 release notes

PR #881, opened 2022-07-29
core/bcast: implementation for builder registration

ciaranmcveigh5:
- bcast implementation for builder registration
- relevant test case

category: feature
ticket: #849

### DIFF
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -130,7 +130,7 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 	case core.DutyBuilderRegistration:
 		registration, ok := aggData.(core.VersionedSignedValidatorRegistration)
 		if !ok {
-			return errors.New("invalid exit")
+			return errors.New("invalid validator registration")
 		}
 
 		err = b.eth2Cl.SubmitValidatorRegistrations(ctx, []*eth2api.VersionedSignedValidatorRegistration{&registration.VersionedSignedValidatorRegistration})
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	eth2client "github.com/attestantio/go-eth2-client"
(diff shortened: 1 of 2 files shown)


core/bcast/bcast_test.go
```diff
@@ -127,6 +127,31 @@ func TestBroadcastBlindedBeaconBlock(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	mock, err := beaconmock.New()
+	require.NoError(t, err)
+
+	registration := testutil.RandomCoreVersionedSignedValidatorRegistration(t).VersionedSignedValidatorRegistration
+
+	aggData := core.VersionedSignedValidatorRegistration{VersionedSignedValidatorRegistration: registration}
+
+	mock.SubmitValidatorRegistrationsFunc = func(ctx context.Context, registrations []*eth2api.VersionedSignedValidatorRegistration) error {
+		require.Equal(t, aggData.VersionedSignedValidatorRegistration, *registrations[0])
```
ciaranmcveigh5, line 140: this bcast is submitting 1 validator registration so we compare registrations[0]

should we be constructing an array with mutliple registrations and checking that array is equal to the registrations (array) in the mocksSubmitValidatorRegistrationsFunc

core/bcast/bcast.go
```diff
@@ -125,6 +127,21 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty,
 
 		return err
 
+	case core.DutyBuilderRegistration:
+		registration, ok := aggData.(core.VersionedSignedValidatorRegistration)
+		if !ok {
+			return errors.New("invalid exit")
```
corverroos, line 133: "invalid registration"

core/bcast/bcast_test.go
corverroos: can do that or just check the length is 1

corverroos (approved):

corverroos (approved):

PR #892, opened 2022-08-02
p2p: decrease circuit relay reservation frequency

corverroos:
This reduces libp2p circuit relay reservations to the required minimum. Previously, we re-reserved on "client connection timeout" of 2s which is not required since server-side reservations remain open when client-side connections are recycled. This resulted in multiple open reservations which quickly results in "reservation-rejected" errors. 

category: bug
ticket: none

### DIFF
--- a/p2p/relay.go
+++ b/p2p/relay.go
@@ -84,16 +84,20 @@ func NewRelayReserver(tcpNode host.Host, relay Peer) lifecycle.HookFunc {
 				continue
 			}
 
+			// Note a single long-lived reservation (created by server-side) is mapped to
+			// many short-lived limited client-side connections.
+			// When the reservation expires, the server needs to re-reserve.
+			// When the connection expires (stream reset error), then client needs to reconnect.
+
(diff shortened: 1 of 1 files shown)


p2p/relay.go
```diff
@@ -84,16 +84,20 @@ func NewRelayReserver(tcpNode host.Host, relay Peer) lifecycle.HookFunc {
...
-				z.Any("limit_data_mb", resv.LimitData/(1<<20)),
+				z.Any("reservation_expire", resv.Expiration),        // Server side reservation expiry (long)
+				z.Any("connection_duration", resv.LimitDuration),    // Client side connection limit (short)
+				z.Any("connection_data_mb", resv.LimitData/(1<<20)), // Client side connection limit (short)
 			)
 
 			select {
 			case <-ctx.Done():
 				return nil
-			case <-time.After(resv.LimitDuration - time.Second*10):
```
corverroos, line 96: removed this

dB2510 (approved):

Issue #890, opened 2022-08-02
Enable libp2p hole punching


corverroos:
## Problem to be solved

We currently require nodes to be publicly accessible via a static IP address (determined at startup), otherwise nodes must connect via a bootnode relay (using libp2p circuit relay protocol) which introduces another network hop to all those connections and introduces the complication of relay connections being recycled every few minutes requiring error handling and reconnects.

Libp2p has a set of features called at [Hole Punching](https://github.com/libp2p/specs/blob/master/connections/hole-punching.md). Enabling hole punching would result in "some" relay connections being upgraded to "direct connections"  when possible. It leverages "circuit-relay", "autonat", "identify" and "dcutr" protocols under the hood.

Direct connections have the following benefits over relay connections:
 - They are long lived (not recycled every few minutes)
 - They are only only a single hop, which decreases network latency.
 - They remove a point of failure (the relay) while they are active.
 - This feature can also be used by publicly accessible nodes with dynamic IPs.

## Proposed solution

Research and investigate how to enable hole punching and ensure networking remains robust.

### Out of Scope

Refactoring ENRs to native LibP2P identity if not required (doesn't seem like it)

thomasheremans:
Need to setup a meeting with libp2p team @OisinKyne

PR #896, opened 2022-08-02
app: enable libp2p hole punch in alpha

corverroos:
Enables libp2p holepunching as alpha feature. This is still very experimental, but we need to get first iterations out and tested in the wild to collect more data.

This will not actually do anything until this libp2p issue is merged https://github.com/libp2p/go-libp2p/issues/1664

Need to figure out why tests are failing, they pass locally...

category: feature 
ticket: #890 
feature_flag: p2p_hole_punch


dkg/sync/client.go
```diff
@@ -142,7 +142,7 @@ func (c *Client) isConnected() bool {
 
 // sendMsgs sends period sync protocol messages on the stream until error or shutdown.
 func (c *Client) sendMsgs(ctx context.Context, stream network.Stream) (bool, error) {
-	timer := time.NewTicker(time.Second)
+	timer := time.NewTicker(time.Second * 10)
```
corverroos, line 145: had to increase period messages to more than 1s, otherwise bad temp addresses are never removed from the peer store.

p2p/ping.go
```diff
@@ -57,7 +57,7 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID,
 	for ctx.Err() == nil {
 		pingPeerOnce(ctx, svc, p, logFunc, callback)
 
-		const backoff = time.Second
+		const backoff = time.Second * 10
```
corverroos, line 60: this is causing tests to fail since we ping much slower and then the tests time out...

corverroos:
Putting this on ice until later

PR #915, opened 2022-08-04
app: add registration simnet test

corverroos:
Adds builder registration to simnet tests.

category: test
ticket: #849

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	eth2client "github.com/attestantio/go-eth2-client"
+	eth2api "github.com/attestantio/go-eth2-client/api"
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2http "github.com/attestantio/go-eth2-client/http"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
@@ -108,8 +109,8 @@ type TestConfig struct {
 	BroadcastCallback func(context.Context, core.Duty, core.PubKey, core.SignedData) error
 	// DisablePromWrap disables wrapping prometheus metrics with cluster identifiers.
 	DisablePromWrap bool
-	// BuilderAPI enables the builder API opt-in feature
-	BuilderAPI bool
+	// BuilderRegistration provides a channel for tests to trigger builder registration by the validator mock,
+	BuilderRegistration <-chan *eth2api.VersionedValidatorRegistration
 }
 
 // Run is the entrypoint for running a charon DVC instance.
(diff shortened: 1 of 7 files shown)


testutil/validatormock/validatormock.go
```diff
@@ -338,6 +338,43 @@ func ProposeBlindedBlock(ctx context.Context, eth2Cl Eth2Provider, signFunc Sign
 	return eth2Cl.SubmitBlindedBeaconBlock(ctx, signedBlock)
 }
 
+// this is register (singular) but the underlying submit function is registrations should it be changed to handle multiple registrations
```
ciaranmcveigh5, line 341: this is register (singular) but the underlying submit function is registrations should it be changed to handle multiple registrations

testutil/validatormock/validatormock.go
```diff
@@ -338,6 +338,43 @@ func ProposeBlindedBlock(ctx context.Context, eth2Cl Eth2Provider, signFunc Sign
...
+func Register(ctx context.Context, eth2Cl Eth2Provider, signFunc SignFunc, registration *eth2api.VersionedValidatorRegistration, pubkey eth2p0.BLSPubKey) error {
+	// TODO(corver): Sign and submit registration.
+
+	sigRoot, err := registration.V1.HashTreeRoot()
+	if err != nil {
+		return err
+	}
+
+	// epoch required for the GetDataRoot function not sure what we should set that as
+	sigData, err := signing.GetDataRoot(ctx, eth2Cl, signing.DomainApplicationBuilder, eth2p0.Epoch(100), sigRoot)
```
ciaranmcveigh5, line 351: epoch required for the GetDataRoot function not sure what we should set that as

testutil/validatormock/validatormock.go
```diff
@@ -338,6 +338,43 @@ func ProposeBlindedBlock(ctx context.Context, eth2Cl Eth2Provider, signFunc Sign
...
+	if err != nil {
+		return err
+	}
+
+	// pubkey required for signature
+	sig, err := signFunc(ctx, pubkey, sigData[:])
```
ciaranmcveigh5, line 357: if function is passed in multiple registrations and multiple pubkeys i'm guessing we would just loop through and use pubkey index i for registration index i?

app/app.go
```diff
@@ -642,6 +643,35 @@ func wireValidatorMock(conf Config, pubshares []eth2p0.BLSPubKey, sched core.Sch
...
+				log.Error(ctx, "Cannot connect to validatorapi", err)
+				return
+			}
+			eth2Cl, ok := eth2Svc.(validatormock.Eth2Provider)
+			if !ok {
+				log.Error(ctx, "Invalid eth2 service", nil)
+			}
+
+			// what pubkey should this be - should we be handing in registrations & pubkeys
+			err = validatormock.Register(ctx, eth2Cl, signer, registration, pubshares[0])
```
ciaranmcveigh5, line 666: This is receiving registrations from the chan - i've just used pubshares[0] for now 

how would you want to associate a pubkey with a registration that is received through the channel

would have used indexes but range on chan doesn't return index - not sure if theres a work around

ciaranmcveigh5 (approved):

PR #918, opened 2022-08-05
docs: fix readme

xenowits:
Remove EIP712 from comments on `enr_signature` or `config_signature`.

category: docs
ticket: none


corverroos (changes requested):
not sure about this. See https://github.com/ObolNetwork/charon/blob/main/cluster/definition.go#L135-L156

Issue #886, opened 2022-08-02
Flapy test in monitoringapi_internal_test


dB2510:
## Problem to be solved

A test named peer_ping_failing is flapping in monitoringapi_internal_test: https://github.com/ObolNetwork/charon/runs/7615741168?check_suite_focus=true which is responsible to assert error obtained from ready function if peer is not able to connect to quorum number of peers. The problem lies here:
https://github.com/ObolNetwork/charon/blob/e7e9998e2b1f0650e598f189f1012e19078196b1/app/monitoringapi_internal_test.go#L113

which expects test to complete in 100ms. This test is non-deterministic since it depends on mutex to result in desired error.

## Proposed solution

Make this test deterministic by determining what is the ideal duration to check in require.Eventually function.

PR #934, opened 2022-08-08
app: fix flapy test in monitoringapi_internal_test

dB2510:
Fixes flapy test in monitoringapi_internal_test.

category: test
ticket: #886

### DIFF
--- a/app/monitoringapi_internal_test.go
+++ b/app/monitoringapi_internal_test.go
@@ -113,7 +113,7 @@ func TestStartChecker(t *testing.T) {
 			if tt.err != nil {
 				require.Eventually(t, func() bool {
 					err = readyErrFunc()
-					if !errors.Is(err, errReadyUnInit) {
+					if !errors.Is(err, errReadyUninitialised) {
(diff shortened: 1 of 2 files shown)


app/monitoringapi_internal_test.go
```diff
@@ -113,7 +113,7 @@ func TestStartChecker(t *testing.T) {
 			if tt.err != nil {
 				require.Eventually(t, func() bool {
 					err = readyErrFunc()
-					if err != nil {
+					if !errors.Is(err, errReadyUnInit) {
```
dB2510, line 116: it was failing because we are checking for err != nil while we have `errReadyUnInit` as default initially.

corverroos: `...UnInit` doesn't make sense

corverroos (approved):

app/monitoringapi_internal_test.go
dB2510: was suggested by you 🙃

PR #938, opened 2022-08-08
p2p: autonat status

corverroos:
Adds autonat reachability metrics. Also add panel to simnet dash,

category: feature
ticket: none

### DIFF
--- a/p2p/p2p.go
+++ b/p2p/p2p.go
@@ -26,6 +26,7 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/libp2p/go-libp2p"
 	libp2pcrypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/event"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/network"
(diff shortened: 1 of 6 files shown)


corverroos:
Example dashboard: 
image

p2p/p2p.go
```diff
@@ -182,6 +184,34 @@ func multiAddrViaRelay(relayPeer Peer, peerID peer.ID) (ma.Multiaddr, error) {
 	return transportAddr.Encapsulate(relayAddr), nil
 }
 
+func NewEventCollecter(tcpNode host.Host) lifecycle.HookFuncCtx {
+	return func(ctx context.Context) {
+		sub, err := tcpNode.EventBus().Subscribe(new(event.EvtLocalReachabilityChanged))
+		if err != nil {
+			log.Error(ctx, "Subscribe libp2p events", err)
```
OisinKyne, line 191: Nit,  probably not worth doing if it's an error deep in prom logs and not user facing  code.  
```suggestion
			log.Error(ctx, "There was an error subscribing to libp2p LocalReachabilityChanged events", err)
```

OisinKyne (approved):
LGTM

p2p/p2p.go
corverroos: yeah, this is deep inside, should never happen in practise

PR #944, opened 2022-08-09
testutil/validatormock: mitigate future slot issues

corverroos:
Similar to `scheduler` mitigate against `eth2http: failed to obtain validators` errors which happen when beacon node sync is behind by querying active validators by "head" instead of explicit "slot".

category: test
ticket: none


corverroos:
See: https://github.com/ObolNetwork/charon/blob/main/core/scheduler/scheduler.go#L518-L527

xenowits (approved):

xenowits:
also, why is this PR categorised as a test?

corverroos:
Validator mock is only used in testing

Issue #217, opened 2022-03-15
Move partial signature verification to parsigdb


corverroos:
## Problem to solve

We currently verify partial signatures submitted by the VC in the validatorapi. This doesn't prevent invalid signatures being added by our peers.

## Proposed solution

- Decouple verification by providing a `VerifyFunc` to parsigDB.
- `type VerifyFunc func(ctx context.Context, duty core.Duty, pubkey core.PubKey, parSig core.ParSignedData) error`
- Pass in verifyFunc into NewParSigDB, call this function before storing.
- Extract verification logic from `validatorapi` and move to `eth2util/signing`, providing a `Veirfy(...) error` funciton.
- For DKG, we implement a custom `Varify` function that can do lockhash and depositdata verifucation based on custom duty types.

corverroos:
Note we should probably do verification in parsigex and validagtorAPI, as close to the source as possible (fail fast). Only doing this in parsigdb results in invalid "events" being generated and complicated `Tracker`.

corverroos:
Note verification in DKG and Charon Run is different. Will need to pass in the verification function as param to `parsigex.New()`

PR #953, opened 2022-08-10
core/validatorapi: add verification functions to validatorapi

dB2510:
Adds verification functions for each duty in eth2util/signing and use them in validatorapi.

category: refactor
ticket: #217

### DIFF
--- a/eth2util/signing/signing.go
+++ b/eth2util/signing/signing.go
@@ -126,7 +126,7 @@ func GetDataRoot(ctx context.Context, eth2Cl Eth2Provider, name DomainName, epoc
 //
 //}
 
-func VerifyAttestation(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_sig.PublicKey, att eth2p0.Attestation) error {
+func VerifyAttestation(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_sig.PublicKey, att *eth2p0.Attestation) error {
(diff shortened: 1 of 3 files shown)


eth2util/signing/signing.go
```diff
@@ -128,121 +126,111 @@ func GetDataRoot(ctx context.Context, eth2Cl Eth2Provider, name DomainName, epoc
...
-// If it is partially signed, provide the pubshare.
-// If it is aggregate signed, provide the group pubkey.
-func VerifySignedData(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_sig.PublicKey,
-	eth2SignedType interface{},
-) error {
-	// eth2util shouldn't import core package, so can't use core.SignedData.
-	// To avoid pointer vs non-pointer issues, always get value if pointer is provided.
-	if reflect.TypeOf(eth2SignedType).Kind() == reflect.Pointer {
-		eth2SignedType = reflect.ValueOf(eth2SignedType).Elem().Interface()
+func VerifyAttestation(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_sig.PublicKey, att eth2p0.Attestation) error {
```
corverroos, line 129: I think all these eth2 types should be pointers, since that is how goeth2 users them, always as pointers.

corverroos (approved):

eth2util/signing/signing.go
dB2510: sure

PR #955, opened 2022-08-10
cmd: increase bootnode max connection limits

corverroos:
Increases bootnode max connections from 256 inbound to 1024 in and out.

category: misc
ticket: none

### DIFF
--- a/cmd/bootnode.go
+++ b/cmd/bootnode.go
@@ -45,6 +45,7 @@ type BootnodeConfig struct {
 	AutoP2PKey    bool
 	P2PRelay      bool
 	MaxResPerPeer int
+	MaxConns      int
 	RelayLogLevel string
 }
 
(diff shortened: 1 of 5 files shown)


cmd/bootnode.go
```diff
@@ -133,7 +135,19 @@ func RunBootnode(ctx context.Context, config BootnodeConfig) error {
 			}
 		}
 
-		tcpNode, err := p2p.NewTCPNode(config.P2PConfig, key, p2p.NewOpenGater(), udpNode, nil, nil)
+		// Increase resource limits
+		const maxConns = 1024
```
dB2510, line 139: can we add a flag for this in bootnode command?

dB2510 (approved):

cmd/bootnode.go
OisinKyne: +1 for flag on the bootnode command. Its more 12factor than code commits for config changes. Maybe something like `--max-peers`?

xenowits (approved):

cmd/bootnode.go
dB2510: I have added it as `--p2p-max-connections` with description: "Libp2p maximum number of peers that can connect to this bootnode."

PR #958, opened 2022-08-11
p2p: instrument relay connections

corverroos:
Instrument relay connectedness, also re-reserve if connection lost.

category: feature
ticket: none


thomasheremans (approved):
Trusting you on this one :D

PR #963, opened 2022-08-11
app: simplify monitoring ready checker

corverroos:
Simplify readyz monitoring endpoint ready checker. 

Instead of actually pinging, just check if we have existing connections. If we do not have enough connections for 3 rounds, then assume connection issues. Since temporary blips should be ignored.

That also addresses the issue of dial error reasons being swallowed by this logic which resulted in logs only showing `Dial Backoff` which is not informative. This therefore removes a source of swallowed real reasons.

category: refactor 
ticket: none

### DIFF
--- a/app/monitoringapi_internal_test.go
+++ b/app/monitoringapi_internal_test.go
@@ -24,7 +24,6 @@ import (
 	"github.com/jonboulle/clockwork"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
-	"github.com/libp2p/go-libp2p-core/peerstore"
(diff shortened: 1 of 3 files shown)


app/monitoringapi_internal_test.go
```diff
@@ -119,7 +120,7 @@ func TestStartChecker(t *testing.T) {
 					}
 
 					return false
-				}, 100*time.Millisecond, time.Millisecond)
+				}, time.Second, 100*time.Millisecond)
```
dB2510, line 123: with one second duration, this becomes a slow test i suppose

dB2510 (approved):
LGTM

PR #964, opened 2022-08-11
core/scheduler: remove beacon clock sync experiment

corverroos:
The Beacon clock sync feature used "head events" to infer beacon node clock offset. But head events are not sent on slot boundaries, rather on "beacon chain head block events" when new head blocks are added to the chain, sure they are tied to slots, but they are not sent on slot boundaries.

So remove this feature as there is not easy way to infer beacon node clock offset.
 
category: refactor 
ticket: none


dB2510 (approved):

dkg/dkg.go
```diff
@@ -52,7 +52,6 @@ type Config struct {
 }
 
 // Run executes a dkg ceremony and writes secret share keystore and cluster lock files as output to disk.
-
```
corverroos, line 55: sneaking this fix in

PR #966, opened 2022-08-11
dkg: fix golangci-lint

dB2510:
Fixes golangci-lint on main.

category: fixbuild
ticket: none


corverroos:
Not going to work, since the change introduced a race condition, fix in #965

PR #967, opened 2022-08-11
dkg: add verification of partial signature

dB2510:
Adds partial signature verification before aggregation.

category: refactor
ticket: #217

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -321,7 +321,12 @@ func signAndAggLockHash(ctx context.Context, shares []share, def cluster.Definit
 		Validators: dvs,
 	}
 
-	sigLockHash, err := signLockHash(lock, nodeIdx.ShareIdx, shares)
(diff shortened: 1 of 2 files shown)


dkg/dkg.go
```diff
@@ -379,7 +384,17 @@ func signAndAggDepositData(ctx context.Context, ex *exchanger, shares []share, w
 		return nil, err
 	}
 
-	aggSigDepositData, err := aggDepositDataSigs(peerSigs)
+	pubkeyToShares := make(map[core.PubKey]map[int]*bls_sig.PublicKey)
```
corverroos, line 387: would pass in shares into `aggDepositDataSigs` and create this map there since this function is getting too big

corverroos (approved):
Maybe one or two invalid signature tests as well?

dkg/dkg.go
dB2510: done

PR #969, opened 2022-08-13
app: correct charon status calculation

dB2510:
Corrects calculation of charon status where it needs to exclude self when comparing the number of connected peers with threshold. Also updates tests.

category: bug
ticket: none

### DIFF
--- a/app/monitoringapi_internal_test.go
+++ b/app/monitoringapi_internal_test.go
@@ -59,6 +59,12 @@ func TestStartChecker(t *testing.T) {
 			absentPeers: 3,
 			err:         errReadyTooFewPeers,
 		},
+		{
+			name:        "success",
+			isSyncing:   false,
+			numPeers:    4,
+			absentPeers: 1,
+		},
 	}
 
 	for _, tt := range tests {
@@ -91,8 +97,8 @@ func TestStartChecker(t *testing.T) {
 			}
 
 			// connect each host with its peers
-			for i := 0; i < tt.numPeers; i++ {
-				for k := tt.absentPeers; k < tt.numPeers; k++ {
+			for i := 0; i < tt.numPeers-tt.absentPeers; i++ {
+				for k := 0; k < tt.numPeers-tt.absentPeers; k++ {
 					if i == k {
 						continue
 					}
--- a/app/monitoringapi_internal_test.go
+++ b/app/monitoringapi_internal_test.go
@@ -96,7 +96,7 @@ func TestStartChecker(t *testing.T) {
 				hosts = append(hosts, h)
 			}
 
-			// connect each host with its peers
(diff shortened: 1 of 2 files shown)


dB2510:
Screenshot 2022-08-13 at 1 04 54 PM
Here it is showing inactive while I am connected to 2 peers out of 3.

app/monitoringapi_internal_test.go
```diff
@@ -91,8 +97,8 @@ func TestStartChecker(t *testing.T) {
 			}
 
 			// connect each host with its peers
-			for i := 0; i < tt.numPeers; i++ {
-				for k := tt.absentPeers; k < tt.numPeers; k++ {
+			for i := 0; i < tt.numPeers-tt.absentPeers; i++ {
```
corverroos, line 100: note I swapped this since we now open connections, not just add addresses, and adding a connection results in both sides of the connection being "active". so you are effectively connecting peer 0 to all peers even if absentPeers > 0

corverroos (changes requested):

app/monitoringapi_internal_test.go
dB2510: so if number of peers are 4 and absent peers is 1 then according this:
0 is connected to 1 and 2
1 is connected to 0 and 2 
2 is connected to 0 and 1
3 is connected to none

peer 0 is not connected to peer 3

dB2510: I have added tt.numPeers-tt.absentPeers condition to both the for loops "since we now open connections".

dB2510: According to your logic, if we have 4 peers and absent peers is 1 then:
0 is connected to 1, 2 and 3
1 is connected to 2 and 3
2 is connected to 1 and 3
3 is connected to 1 and 2

we call `startReadyChecker` for host[0] for which count in `quorumPeersConnected` function comes out to be 3. However, 1 peer should be absent and count should be 2.

corverroos: Maybe to simplify, since we only test node0, is to only connect node0 to N peers. Do not also connect the other needs to each other.

corverroos (approved):

app/monitoringapi_internal_test.go
dB2510: sure

dB2510: done

app/monitoringapi_internal_test.go
```diff
@@ -91,15 +97,9 @@ func TestStartChecker(t *testing.T) {
 			}
 
 			// connect each host with its peers
```
corverroos, line 99: connect node0 to others

dB2510: done

PR #976, opened 2022-08-15
core/parsigex: instrument signature verification

corverroos:
Adds tracing spans to signature verification in parsigex.

- Also increase parsigex handle timeout from 1s to 5s since signature verification takes 200ms per sig...
- Also allow compose to bind to ports to allow pprof on charon.

category: misc
ticket: none


dB2510 (approved):
Increasing timeout to 5s would mean that we can only accommodate 25 verifications?

Issue #974, opened 2022-08-15
Memory leak 


corverroos:
## Problem to be solved

There seems to be a memory leak in charon. The metric `go_memstats_alloc_bytes` shows exponential growth resulting in docker container hanging.
 
<img width="1443" alt="image" src="https://user-images.githubusercontent.com/29249923/184597603-7af2e987-7dd6-4228-89c3-530fc2cdb410.png">

## Proposed solution

Find the leak and plug it.

corverroos:
This seems to be fixed for now
<img width="1448" alt="image" src="https://user-images.githubusercontent.com/29249923/185049813-1d9653c8-8780-48b4-b423-40bf4940ecb5.png">

PR #978, opened 2022-08-15
core/dutydb: trim expired duties

corverroos:
Trim expired duties in dutydb.

category: feature 
ticket: #974

### DIFF
--- a/core/dutydb/memory.go
+++ b/core/dutydb/memory.go
@@ -29,13 +29,15 @@ import (
 )
 
 // NewMemDB returns a new in-memory dutyDB instance.
-func NewMemDB() *MemDB {
+func NewMemDB(deadliner core.Deadliner) *MemDB {
 	return &MemDB{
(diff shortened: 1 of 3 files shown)


core/dutydb/memory.go
```diff
@@ -101,6 +109,24 @@ func (db *MemDB) Store(_ context.Context, duty core.Duty, unsignedSet core.Unsig
...
 
+	// Delete all expired duties.
+	for {
+		var deleted bool
+		select {
+		case duty := <-db.deadliner.C():
+			err := db.deleteDutyUnsafe(duty)
+			if err != nil {
+				return err
+			}
```
xenowits, line 120:
```suggestion
			err := db.deleteDutyUnsafe(duty); if err != nil {
				return err
			}
```

corverroos: not sure we have converged on a style for this pattern yet, we use both styles

dB2510 (approved):

PR #979, opened 2022-08-15
core/scheduler: improve builder-api config error

corverroos:
Improve scheduler errors when querying mismatching proposer duties ito builder-api config.

category: misc
ticket: none

### DIFF
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -139,6 +139,12 @@ func (s *Scheduler) Run() error {
 
 // GetDutyDefinition returns the definition for a duty if resolved already, otherwise an error.
 func (s *Scheduler) GetDutyDefinition(ctx context.Context, duty core.Duty) (core.DutyDefinitionSet, error) {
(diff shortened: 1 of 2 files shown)


core/scheduler/scheduler.go
```diff
@@ -139,6 +139,12 @@ func (s *Scheduler) Run() error {
 
 // GetDutyDefinition returns the definition for a duty if resolved already, otherwise an error.
 func (s *Scheduler) GetDutyDefinition(ctx context.Context, duty core.Duty) (core.DutyDefinitionSet, error) {
+	if duty.Type == core.DutyBuilderProposer && !s.builderAPI {
+		return nil, errors.New("builder-api not enabled, but duty builder proposer requested")
+	} else if duty.Type == core.DutyProposer && s.builderAPI {
```
dB2510, line 144: does this mean if builder api is enabled then we won't use DutyProposer at all?

dB2510 (approved):

core/scheduler/scheduler.go
corverroos: yup

PR #982, opened 2022-08-16
app/expbackoff: implement exponential backoff package

corverroos:
Implements an exponential backoff package based on `google.golang.org/grpc@v1.48.0/backoff`. This is to be used in many places where we have `// TODO(corver): improve backoff`

category: feature
ticket: none

### DIFF
--- a/app/expbackoff/expbackoff.go
+++ b/app/expbackoff/expbackoff.go
@@ -1,0 +1,205 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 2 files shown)


app/expbackoff/expbackoff.go
```diff
@@ -0,0 +1,205 @@
...
+	backoff := float64(config.BaseDelay)
+	max := float64(config.MaxDelay)
+
+	for backoff < max && retries > 0 {
+		backoff *= config.Multiplier
+		retries--
+	}
+	if backoff > max {
+		backoff = max
+	}
```
dB2510, line 170:
```suggestion
	backoff = math.Max(backoff * math.Pow(config.Multiplier, retries), max)
```

dB2510 (approved):
LGTM

app/expbackoff/expbackoff.go
corverroos: simpler is better, math is hard to read

corverroos:
This is required for #985 and #986

PR #988, opened 2022-08-17
core/validatorapi: teku proposer config updates

ciaranmcveigh5:
- Use fee recipient from cluster lock or withdrawal address if fee recipient not set for builder api registrations
- Add gas limit to proposer config as hardcoded const - while the VC's will default to a value this value may be different between VC clients so between to hardcode it to a known value

category: feature
ticket: #849

### DIFF
--- a/core/validatorapi/teku.go
+++ b/core/validatorapi/teku.go
@@ -36,16 +36,6 @@ type TekuBuilder struct {
 	Overrides map[string]string `json:"registration_overrides,omitempty"`
 }
 
-var feeRecipient = "0x000000000000000000000000000000000000dead"
-
-// var cluster.Lock lock
-
-// if address, ok := lock["fee_recipient"]; ok {
-//     feeRecipient = lock["fee_recipient"]
-// } else {
-// 	feeRecipient = lock["withdrawal_address"]
-// }
-
 const gasLimit = 30000000
 
 type TekuProposerConfigProvider interface {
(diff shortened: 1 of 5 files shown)


core/validatorapi/teku.go
```diff
@@ -32,10 +32,21 @@ type TekuProposerConfig struct {
...
 type TekuBuilder struct {
 	Enabled   bool              `json:"enabled"`
+	GasLimit  uint              `json:"gas_limit"`
 	Overrides map[string]string `json:"registration_overrides,omitempty"`
 }
 
-const dead = "0x000000000000000000000000000000000000dead"
+var feeRecipient = ""
+
+// var cluster.Lock lock
```
ciaranmcveigh5, line 41: Need to uncomment this not sure whats the variable/ file i need to import to gain access to whats in the cluster lock 

saw "github.com/obolnetwork/charon/cluster"

can see it's passed in via 

where conf can be used with 

lock, err := loadLock(conf)

ciaranmcveigh5:
integration tests are failing because fee recipient is 0 bytes and so can't be converted to an "Address" this will be resolved once the lock config code is uncommented

corverroos:
I pushed a fix, but to the wrong branch since I didn't push to your repo, see https://github.com/ObolNetwork/charon/commit/5b0e4de445eef7fceb7ac05d008caddf85540ffa.

corverroos (approved):

PR #990, opened 2022-08-17
core/validatorapi: add lighthouse route for validator definitions

ciaranmcveigh5:
- add lighthouse route for validator defintions, returned in JSON user will have to convert to yaml

category: feature
ticket: #849

### DIFF
--- a/core/validatorapi/lighthouse.go
+++ b/core/validatorapi/lighthouse.go
@@ -15,7 +15,7 @@
 
 package validatorapi
 
-type LighthouseValidatorDefinitionJSON struct {
+type LighthouseValidatorDefinition struct {
 	Enabled                    bool   `json:"enabled"`
 	VotingPublicKey            string `json:"voting_public_key"`
 	Type                       string `json:"type"`
@@ -27,39 +27,27 @@ type LighthouseValidatorDefinitionJSON struct {
 	BuilderPubkeyOverride      string `json:"builder_pubkey_override"`
 }
 
-// type LighthouseValidatorDefinitionYAML struct {
-// 	Enabled                    bool   `yaml:"enabled"`
-// 	VotingPublicKey            string `yaml:"voting_public_key"`
-// 	Type                       string `yaml:"type"`
-// 	VotingKeystorePath         string `yaml:"voting_keystore_path"`
-// 	VotingKeystorePasswordPath string `yaml:"voting_keystore_password_path"`
-// 	SuggestedFeeRecipient      string `yaml:"suggested_fee_recipient"`
-// 	GasLimit                   uint   `yaml:"gas_limit"`
-// 	BuilderProposals           bool   `yaml:"builder_proposals"`
-// 	BuilderPubkeyOverride      string `yaml:"builder_pubkey_override"`
-// }
-
 const gasLimit = 30000000
 
 type LighthouseValidatorDefinitionsProvider interface {
-	LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error)
+	LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinition, error)
 }
 
-func (c Component) LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error) {
-	resp := []LighthouseValidatorDefinitionJSON{}
+func (c Component) LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinition, error) {
+	var resp []LighthouseValidatorDefinition
 
 	for pubkey, pubshare := range c.sharesByKey {
-		resp = append(resp, LighthouseValidatorDefinitionJSON{
+		resp = append(resp, LighthouseValidatorDefinition{
 			Enabled:         true,
 			VotingPublicKey: string(pubshare),
 			// Asking whether these need to be defined in lighthouse discord
 			// Type: dead,
 			// VotingKeystorePath: dead,
 			// VotingKeystorePasswordPath: dead,
-			SuggestedFeeRecipient: string(pubkey),
+			SuggestedFeeRecipient: c.feeRecipient,
 			GasLimit:              gasLimit,
 			BuilderProposals:      true,
-			BuilderPubkeyOverride: dead,
+			BuilderPubkeyOverride: string(pubkey),
 		})
 	}
 
--- a/core/validatorapi/lighthouse.go
+++ b/core/validatorapi/lighthouse.go
@@ -1,0 +1,69 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package validatorapi
+
+import (
+	"context"
+	"fmt"
+)
+
+type LighthouseValidatorDefinitionsResponse struct {
+	Proposers map[string]TekuProposerConfig `json:"proposer_config"`
+	Default   TekuProposerConfig            `json:"default_config"`
+}
+
+type LighthouseValidatorDefinitions struct {
+	FeeRecipient string      `json:"fee_recipient"`
+	Builder      TekuBuilder `json:"builder"`
+}
+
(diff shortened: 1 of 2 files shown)


core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+	Type                       string `json:"type"`
+	VotingKeystorePath         string `json:"voting_keystore_path"`
+	VotingKeystorePasswordPath string `json:"voting_keystore_password_path"`
+	SuggestedFeeRecipient      string `json:"suggested_fee_recipient"`
+	GasLimit                   uint   `json:"gas_limit"`
+	BuilderProposals           bool   `json:"builder_proposals"`
+	BuilderPubkeyOverride      string `json:"builder_pubkey_override"`
+}
+
+// type LighthouseValidatorDefinitionYAML struct {
```
ciaranmcveigh5, line 30: can potentially remove this for now and have the user make the coversion

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (c Component) LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error) {
+	resp := []LighthouseValidatorDefinitionJSON{}
+
+	for pubkey, pubshare := range c.sharesByKey {
+		resp = append(resp, LighthouseValidatorDefinitionJSON{
+			Enabled:         true,
+			VotingPublicKey: string(pubshare),
+			// Asking whether these need to be defined in lighthouse discord
+			// Type: dead,
```
ciaranmcveigh5, line 56: asking on whether some of these can be omitted

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+// 	BuilderPubkeyOverride      string `yaml:"builder_pubkey_override"`
+// }
+
+const gasLimit = 30000000
+
+type LighthouseValidatorDefinitionsProvider interface {
+	LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error)
+}
+
+func (c Component) LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error) {
```
ciaranmcveigh5, line 48: no need for context since the timestamp isn't in this file - might be something we want them to include

core/validatorapi/lighthouse.go
corverroos, line 30: yeah, we have a `yaml.JSONToYAML` which can trigger in `validatorapi.writeResponse` if the `Accept: application/yaml` header is present in the request, but lets push that out to another PR once we really need it.

core/validatorapi/router.go
```diff
@@ -548,6 +554,12 @@ func tekuProposerConfig(p TekuProposerConfigProvider) handlerFunc {
 	}
 }
 
+func lighthouseValidatorDefinitions(p LighthouseValidatorDefinitionsProvider) handlerFunc {
+	return func(ctx context.Context, _ map[string]string, _ url.Values, _ []byte) (interface{}, error) {
+		return p.LighthouseValidatorDefinitions()
```
corverroos, line 559: always good to pass context, even though it isn't used yet.

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+			Enabled:         true,
+			VotingPublicKey: string(pubshare),
+			// Asking whether these need to be defined in lighthouse discord
+			// Type: dead,
+			// VotingKeystorePath: dead,
+			// VotingKeystorePasswordPath: dead,
+			SuggestedFeeRecipient: string(pubkey),
+			GasLimit:              gasLimit,
+			BuilderProposals:      true,
+			BuilderPubkeyOverride: dead,
```
corverroos, line 62: this should probably be `string(pubkey)`?

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for pubkey, pubshare := range c.sharesByKey {
+		resp = append(resp, LighthouseValidatorDefinitionJSON{
+			Enabled:         true,
+			VotingPublicKey: string(pubshare),
+			// Asking whether these need to be defined in lighthouse discord
+			// Type: dead,
+			// VotingKeystorePath: dead,
+			// VotingKeystorePasswordPath: dead,
+			SuggestedFeeRecipient: string(pubkey),
```
corverroos, line 59: this should probably be `c.feeRecipient` see https://github.com/ObolNetwork/charon/commit/5b0e4de445eef7fceb7ac05d008caddf85540ffa

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package validatorapi
+
+type LighthouseValidatorDefinitionJSON struct {
```
corverroos, line 18: I'd drop the JSON suffix, since we only do that if there is a non-json equivalent.

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,67 @@
...
+// }
+
+const gasLimit = 30000000
+
+type LighthouseValidatorDefinitionsProvider interface {
+	LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error)
+}
+
+func (c Component) LighthouseValidatorDefinitions() ([]LighthouseValidatorDefinitionJSON, error) {
+	resp := []LighthouseValidatorDefinitionJSON{}
```
corverroos, line 49: nit: `var resp []LighthouseValidatorDefinitionJSON`

corverroos (commented):
Can also add a test like in https://github.com/ObolNetwork/charon/commit/5b0e4de445eef7fceb7ac05d008caddf85540ffa

ciaranmcveigh5:
To be honest im not sure the value in this endpoint given lighthouse doesn't allow you to define validator_defintions.yml as a  url and the fact that the file path field is something the operator is going to have to appended / transform to yaml with some script anyway

effectively the information we're scraping is

```
{
   "distributed_public_key": "0xa878c8ec402799536f0b94967e578fdbcd84828f564d604f0db491979438357b797491399be1f22de8a44673f14c087e",
   "public_shares": [
    "pkadKH8m7LNgSbebQI4lc4oOFZmA8y+2WRdEFrng6Pf47MVdAaVFKMFsE4uxIB6v",
    "h4E/J3cWJXDyO2epebBEDeGn8LUOhzxWr6GRmmoQEGh4QhCHMeg3wbi6iXOwnunm",
    "s/CZz1rpJATz0auDPWIkaM0aFLJxnl2qo2kNlg42ZG59Y93bjqLB33tCsSlq1iAi"
   ]
  }
  
  etc.
  ```
  
which is what we have in cluster-lock.json only difference being we want the pubkeyshares in their key format rather than b64 encoded
 
wonder if this is just added bloat to the codebase without adding much value

File: core/validatorapi/lighthouse.go [hunk 9]
```diff
@@ -0,0 +1,51 @@
...
+func (c Component) LighthouseValidatorDefinitions(_ context.Context) ([]LighthouseValidatorDefinition, error) {
+	var resp []LighthouseValidatorDefinition
+
+	for pubkey, pubshare := range c.sharesByKey {
+		resp = append(resp, LighthouseValidatorDefinition{
+			Enabled:               true,
+			VotingPublicKey:       string(pubshare),
+			SuggestedFeeRecipient: c.feeRecipient,
+			GasLimit:              gasLimit,
+			BuilderProposals:      true,
```
corverroos, line 45: we should probably wire the `--builder-api` flag from `app.Config` similar to how we ire fee_recipient from lock file.

core/validatorapi/lighthouse.go
```diff
@@ -0,0 +1,51 @@
...
+
+type LighthouseValidatorDefinitionsProvider interface {
+	LighthouseValidatorDefinitions(ctx context.Context) ([]LighthouseValidatorDefinition, error)
+}
+
+func (c Component) LighthouseValidatorDefinitions(_ context.Context) ([]LighthouseValidatorDefinition, error) {
+	var resp []LighthouseValidatorDefinition
+
+	for pubkey, pubshare := range c.sharesByKey {
+		resp = append(resp, LighthouseValidatorDefinition{
```
corverroos, line 40: Maybe add a comment about why we do not set keystore values, and what about "type"?

corverroos (approved):
Please add a test similar to teku config test

corverroos:
Yeah, we are going to refactor the lock file pubshares from base64 to hex in the upcoming release, so then transforming the lock file itself would be trivial I guess, so happy to not add this endpoint if that works better.

corverroos:
As discussed, closing this as this can be done via script

PR #997, opened 2022-08-18
precommit: add timeout to pre-commit tests

dB2510:
Adds timeout flag to pre-commit tests so that they timeout early rather than blocking forever.

category: misc
ticket: none

### DIFF
--- a/.pre-commit/run_tests.sh
+++ b/.pre-commit/run_tests.sh
@@ -4,4 +4,4 @@
 MOD=$(go list -m)
 PKGS=$(echo "$@"| xargs -n1 dirname | sort -u | sed -e "s#^#${MOD}/#")
 
-go test $PKGS
+go test -timeout=5m $PKGS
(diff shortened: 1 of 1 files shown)


.pre-commit/run_tests.sh
```diff
@@ -4,4 +4,4 @@
 MOD=$(go list -m)
 PKGS=$(echo "$@"| xargs -n1 dirname | sort -u | sed -e "s#^#${MOD}/#")
 
-go test $PKGS
+go test -timeout=5m $PKGS
```
corverroos, line 7: What is the problem you are trying to solve, since there is already a default timeout of 10m, so I doubt changing it to 5min is going to help...?

```
	-timeout d
	    If a test binary runs longer than duration d, panic.
	    If d is 0, the timeout is disabled.
	    The default is 10 minutes (10m).
```

corverroos (changes requested):
Doubt this will help with "tests blocking forever"

corverroos (approved):

PR #1015, opened 2022-08-22
readme: list all of the rest of the clients we need to support

OisinKyne:
Update lighthouse to mark BlindedBeaconBlock complete. Also list the progress on the rest of the clients

category: docs
ticket: none


README.md
```diff
@@ -61,13 +61,13 @@ It is still early days for the Obol Network and things are under active developm
 We are moving fast so check back in regularly to track the progress.
 
 Charon is a distributed validator, so its main responsibility is performing validation duties.
```
xenowits, line 63: we can also say what are the blockers for integrating prysm, nimbus or vouch.

OisinKyne: There's a table for that above these changes :)

Issue #919, opened 2022-08-05
Refactor randao core signed data implementation


corverroos:
## Problem to be solved

While refactoring signature verification into its own standalone function, we discovered that randao signatures cannot be verified on their own, since they do not contain the raw data that we signed, which is the epoch. We previously relied on the associated duty/slot. This creates additional coupling and means we cannot extract and decouple signature verification properly. 

Before #217 can be completed, we need to address this issue.

## Proposed solution

- Introduce our own custom ETH2 type called `eth2util.SignedEpoch`:
```
type SignedEpoch struct {
  Epoch eth2p0.Epoch
  Signature eth2p0.BLSSignature
}
```

- Refactor `eth2util.EpochHashRoot` as a method of this type.

- Introduce a new `core.SignedEpoch` wrapper of `eth2util.SignedEpoch` implementing `core.SignedData`, exactly the same as the other siigned data implemenations. 

- Rename `DutyRando` to `DutySignature`, use that duty in DKG.

- Add a new duty type, `DutyRandao`, use this duty in charon run. (this is a breaking change)

- Use `core.SignedEpoch` for `DutyRandao` instead of `core.Signature`.

- Note we are not staying backwards compatible, DutyRandao (and therefore DutyPropose) will not work in cluster with a mix of previous and later versions since the wire value of DutyRandao is changing. 

 - Decouple Randao signature verification from core.Duty.Slot

thomasheremans:
Paused until v1.0

corverroos:
Yeah, this will require breaking changes to the API. 

To keep backwards wire compatibility, we need to refactor randao to a new duty type enum, and send both enums from the new servers and convert the old duty type to the new duty in parsigex (for charon run only not dkg) which introduces more coupling, which is what we are trying to fix.

If this can go in with breaking change, then it would be best.

dB2510:
Closed by #1019

PR #1019, opened 2022-08-23
core: refactor randao to SignedRandao

dB2510:
Refactors DutyRandao to use SignedRandao and move away from Signature type. Also introduces DutySignature which will be used by DKG.

category: refactor
ticket: #919


core/types.go
```diff
@@ -39,6 +39,7 @@ const (
 	DutyExit                DutyType = 4
 	DutyBuilderProposer     DutyType = 5
 	DutyBuilderRegistration DutyType = 6
+	DutySignature           DutyType = 7
```
corverroos, line 42: instead of breaking backwards compatibility in DKG and charon run, only break charon run, so like issue mentions: DutyRandao should be 7 and DutySignature should remain at 3.

core/signeddata.go
```diff
@@ -637,6 +637,81 @@ func (r *VersionedSignedValidatorRegistration) UnmarshalJSON(input []byte) error
...
+		SignedData: SignedRandao{
+			Epoch:        epoch,
+			BLSSignature: sig,
+		},
+		ShareIdx: shareIdx,
+	}
+}
+
+// SignedRandao is a signed Randao Reveal which implements SignedData.
+type SignedRandao struct {
```
corverroos, line 652: like issue mentions, please create a SignedEpoch type in eth2util and wrap that type here and make `EpochHashRoot` a method of that type

core/signeddata.go
```diff
@@ -637,6 +637,81 @@ func (r *VersionedSignedValidatorRegistration) UnmarshalJSON(input []byte) error
...
+	var resp SignedRandao
+	err := cloneJSONMarshaler(s, &resp)
+	if err != nil {
+		return SignedRandao{}, errors.Wrap(err, "clone randao")
+	}
+
+	return resp, nil
+}
+
+type signedRandaoJSON struct {
```
corverroos, line 710: extra json type not required if using ethutil.SignedData type I think

eth2util/signing/signing.go
```diff
@@ -166,13 +166,7 @@ func VerifyBlindedBlock(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_si
...
 }
 
-func VerifyRandao(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_sig.PublicKey, randao eth2p0.BLSSignature, slot eth2p0.Slot) error {
-	// Calculate slot epoch
-	epoch, err := epochFromSlot(ctx, eth2Cl, slot)
-	if err != nil {
-		return err
-	}
-
+func VerifyRandao(ctx context.Context, eth2Cl Eth2Provider, pubkey *bls_sig.PublicKey, randao eth2p0.BLSSignature, epoch eth2p0.Epoch) error {
```
corverroos, line 169: this should take `eth2util.SignedRandao`

core/signeddata.go
```diff
@@ -637,6 +638,77 @@ func (r *VersionedSignedValidatorRegistration) UnmarshalJSON(input []byte) error
 	return nil
 }
 
+// NewPartialSignedRandao is a convenience function that returns a new partially signed Randao Reveal.
+func NewPartialSignedRandao(randao eth2util.SignedEpoch, shareIdx int) ParSignedData {
```
corverroos, line 642: probably fine to take epoch, sig, index directly like previously since it seems like we never have a `eth2util.SignedEpoch` available.

core/signeddata.go
corverroos: can remove this type and move json un/marshalling to eth2util

core/validatorapi/validatorapi.go
```diff
@@ -409,16 +419,26 @@ func (c Component) BlindedBeaconBlockProposal(ctx context.Context, slot eth2p0.S
...
-		return signing.VerifyRandao(ctx, c.eth2Cl, pubshare, randao, slot)
+		return signing.VerifyRandao(ctx, c.eth2Cl, pubshare, sigEpoch)
 	})
 	if err != nil {
 		return nil, err
 	}
 
 	duty := core.NewRandaoDuty(int64(slot))
-	parSig := core.NewPartialSignature(core.SigFromETH2(randao), c.shareIdx)
+	parSig := core.NewPartialSignedRandao(sigEpoch, c.shareIdx)
```
corverroos, line 441: can move this up, then do `signing.VerifyRandao(ctx, ..., parSig.SignedEpoch0`

core/validatorapi/validatorapi.go
```diff
@@ -613,8 +633,16 @@ func (c Component) SubmitVoluntaryExit(ctx context.Context, exit *eth2p0.SignedV
 }
 
 func (c Component) submitRandaoDuty(ctx context.Context, pubKey core.PubKey, slot eth2p0.Slot, randao eth2p0.BLSSignature) error {
```
corverroos, line 635: this isn't used, can remove?

eth2util/types.go
```diff
@@ -22,14 +22,20 @@ import (
...
 
+// SignedEpoch represents signature of corresponding epoch.
+type SignedEpoch struct {
+	Epoch     eth2p0.Epoch
+	Signature eth2p0.BLSSignature
+}
+
 // EpochHashRoot returns the ssz hash root of the epoch.
-func EpochHashRoot(epoch eth2p0.Epoch) ([32]byte, error) {
+func (s SignedEpoch) EpochHashRoot() ([32]byte, error) {
```
corverroos, line 32: can refactor this to align with eth2client pattern of getting the hash root of any type
image

corverroos (approved):

core/signeddata.go
```diff
@@ -637,6 +638,68 @@ func (r *VersionedSignedValidatorRegistration) UnmarshalJSON(input []byte) error
...
+	return resp, nil
+}
+
+func (s *SignedRandao) UnmarshalJSON(input []byte) error {
+	if err := json.Unmarshal(input, &s.SignedEpoch); err != nil {
+		return errors.Wrap(err, "unmarshal signed randao")
+	}
+
+	return nil
+}
```
corverroos, line 691: what about same as attestation?

```
func (a Attestation) MarshalJSON() ([]byte, error) {
	return a.Attestation.MarshalJSON()
}

func (a *Attestation) UnmarshalJSON(b []byte) error {
	return a.Attestation.UnmarshalJSON(b)
}
```

eth2util/types.go
```diff
@@ -0,0 +1,81 @@
...
+	s.Epoch = resp.Epoch
+	s.Signature = resp.Signature
+
+	return nil
+}
+
+type signedEpochJSON struct {
+	Epoch     eth2p0.Epoch        `json:"epoch"`
+	Signature eth2p0.BLSSignature `json:"signature"`
+}
```
corverroos, line 81: this type isn't required, just add json stuff to real type above

corverroos (approved):

core/signeddata.go
dB2510: sure

eth2util/types.go
dB2510: i added this type to avoid getting into infinite loop if i add json stuff to the real type since it will have its own Unmarshal JSON method which results in recursive calling

Issue #1025, opened 2022-08-23
Locally created clusters should verify


corverroos:
## Problem to be solved

When creating cluster lock files with `charon create cluster` they should be valid and should not require `--no-verify` since this is an official way to create a cluster.

## Proposed solution

Cluster lock files without operator addresses and signatures and only ENRs should be valid.

### Out of Scope

Can always add some kind of signing to clusters created via `charon create cluster` later.

PR #1026, opened 2022-08-23
cluster: skip verification if locally created

corverroos:
Verification of unsigned `charon create cluster` lock files succeed avoiding the need for `--no-verify`.

category: bug
ticket: #1025

### DIFF
--- a/cluster/definition.go
+++ b/cluster/definition.go
@@ -121,7 +121,14 @@ func (d Definition) Verify() error {
 		return errors.Wrap(err, "config hash")
 	}
 
+	var noSigs int
 	for _, o := range d.Operators {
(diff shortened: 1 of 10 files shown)


cluster/definition.go
```diff
@@ -155,6 +162,10 @@ func (d Definition) Verify() error {
 		}
 	}
 
+	if noSigs > 0 && noSigs != len(d.Operators) {
```
xenowits, line 165: so, either no one signs OR all of them sign, only then we consider it verified?

corverroos: yup

xenowits (approved):
Looks good!

PR #1031, opened 2022-08-24
cmd/createenr: make ENR creation deteministic

dB2510:
Makes ENR creation deterministic by creating ENRs directly from enr.Record.

category: refactor
ticket: none

### DIFF
--- a/cmd/createenr.go
+++ b/cmd/createenr.go
@@ -16,10 +16,14 @@
 package cmd
 
 import (
+	"crypto/ecdsa"
 	"fmt"
 	"io"
+	"net"
 	"strings"
 
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/spf13/cobra"
 
 	"github.com/obolnetwork/charon/app/errors"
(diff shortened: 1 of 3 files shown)


cmd/createenr_internal_test.go
```diff
@@ -32,3 +36,22 @@ func TestRunCreateEnr(t *testing.T) {
...
+		UDPAddr:  "127.0.0.1:3630",
+	}
+
+	key, err := ecdsa.GenerateKey(crypto.S256(), rand.New(rand.NewSource(time.Now().Unix())))
+	require.NoError(t, err)
+
+	originalENR, err := createENR(key, config)
+	require.NoError(t, err)
+
+	for i := 0; i < 100; i++ {
```
corverroos, line 52: there is nothing random to this function, so no need to do 100 times, can just do twice

cmd/createenr.go
```diff
@@ -61,22 +65,47 @@ func runCreateEnrCmd(w io.Writer, config p2p.Config, dataDir string) error {
...
 
+func createENR(key *ecdsa.PrivateKey, config p2p.Config) (string, error) {
+	tcpAddrs, err := config.ParseTCPAddrs()
+	if err != nil {
+		return "", err
+	}
+
+	var r enr.Record
+	for _, addr := range tcpAddrs {
+		r.Set(enr.IP(addr.IP))
```
corverroos, line 91: we should not have two different ways of calculating ENRs from config. for example, you are not including `ExternalIP` or `ExternalHost`. We need to think a bit more about what we encode into ENRs and how they are used.

corverroos (changes requested):
Not sure about this, this just ignores a bunch of things that do in `NewLocalEnode`. Pretty sure this breaks some existing use-cases

cmd/createenr.go
```diff
@@ -61,22 +63,36 @@ func runCreateEnrCmd(w io.Writer, config p2p.Config, dataDir string) error {
...
 
+func createENR(key *ecdsa.PrivateKey, config p2p.Config) (string, error) {
+	node, _, err := p2p.NewLocalEnode(config, key)
+	if err != nil {
+		return "", err
+	}
+
+	record := node.Node().Record()
+	record.SetSeq(0)
+	if err = enode.SignV4(record, key); err != nil {
```
dB2510, line 89: need to sign enr again since localenode returns signed enr record

cmd/createenr.go
```diff
@@ -61,22 +63,36 @@ func runCreateEnrCmd(w io.Writer, config p2p.Config, dataDir string) error {
...
 	_, _ = fmt.Fprintf(w, "Created ENR private key: %s\n", keyPath)
-	_, _ = fmt.Fprintln(w, localEnode.Node().String())
+	_, _ = fmt.Fprintln(w, enrStr)
 
 	writeEnrWarning(w, keyPath)
 
 	return nil
 }
 
+func createENR(key *ecdsa.PrivateKey, config p2p.Config) (string, error) {
```
corverroos, line 81: add a comment since it isn't obvious why we do stuff twice

cmd/enr.go
```diff
@@ -64,20 +64,18 @@ func runNewENR(w io.Writer, config p2p.Config, dataDir string, verbose bool) err
...
-	newEnr := localEnode.Node().String()
-	_, _ = fmt.Fprintln(w, newEnr)
+	_, _ = fmt.Fprintln(w, enrStr)
 
 	if !verbose {
 		return nil
 	}
 
-	r, err := p2p.DecodeENR(newEnr)
+	r, err := p2p.DecodeENR(enrStr)
```
corverroos, line 78: another round of encoding and decoding, feel like `createENR` should return `enr.Record`

corverroos (approved):

cmd/createenr.go
dB2510: sure

cmd/enr.go
dB2510: sure

PR #1052, opened 2022-08-29
p2p: decrease mutable udp node lock scope

corverroos:
Decreases `p2p.MutableUDPNode` lock scope in `maybeRefresh`.

category: refactor 
ticket: none


dB2510 (approved):

corverroos:
Not as easy as that it seems, results in `Recreate discv5 udp node: listen udp: listen udp 127.0.0.1:54976: bind: address already in use`. Leaving as is for now.

Issue #1045, opened 2022-08-25
`charon dkg` can take a remote URI for a DKG cluster-definition


OisinKyne:
## Problem to be solved

Copying a definition file to .charon always goes poorly in user testing. Designing the launchpad with this in mind might not be such a good idea. 

## Proposed solution

We already want to move away from a `--data-dir` flag to a `--definition-file` flag as per https://github.com/ObolNetwork/charon/issues/994

Allow specifying a URI to a remote definition file. (Oisin will get the develop.kyne.eu/ api endpoint running on an `api.obol.tech` domain so it looks pretty). 

### Out of Scope

Nothing to do with `charon run` or cluster-lock.json

corverroos:
Note we already have a `--definition-file` flag. `--data-dir` is only used to load the `charon-enr-private-key`

PR #1070, opened 2022-09-01
dkg/disk: fetch definition file from remote uri

xenowits:
Add support to fetch `cluster-definition` file from a remote URI.

category: feature
ticket: #1045


dkg/disk.go
```diff
@@ -53,6 +61,27 @@ func loadDefinition(conf Config) (cluster.Definition, error) {
 	return res, nil
 }
 
+// fetchDefinition fetches cluster definition file from a remote URI.
+func fetchDefinition(url string) (cluster.Definition, error) {
+	resp, err := http.Get(url) //nolint:gosec,noctx
```
xenowits, line 66: discuss: should i pass a `http.Client` as a function parameter?

dkg/disk.go
```diff
@@ -53,6 +61,27 @@ func loadDefinition(conf Config) (cluster.Definition, error) {
...
+	}
+	defer resp.Body.Close()
+
+	buf, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return cluster.Definition{}, errors.Wrap(err, "read response body")
+	}
+
+	def := new(cluster.Definition)
+	if err := def.UnmarshalJSON(buf); err != nil {
```
corverroos, line 78: rather do standard: `json.Unmarshal`

dkg/disk.go
```diff
@@ -53,6 +61,27 @@ func loadDefinition(conf Config) (cluster.Definition, error) {
...
+		return cluster.Definition{}, errors.Wrap(err, "fetch file")
+	}
+	defer resp.Body.Close()
+
+	buf, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return cluster.Definition{}, errors.Wrap(err, "read response body")
+	}
+
+	def := new(cluster.Definition)
```
corverroos, line 77: nit: `var def cluster.Definition` then `json.Unmarshal(data, &def)`

dkg/disk.go
```diff
@@ -33,12 +36,17 @@ import (
 	"github.com/obolnetwork/charon/tbls/tblsconv"
 )
 
-// loadDefinition returns the cluster definition from disk (or the test definition if configured).
+// loadDefinition returns the cluster definition from disk or network. It returns the test definition if configured.
 func loadDefinition(conf Config) (cluster.Definition, error) {
```
corverroos, line 40: add an internal test for this function please

dkg/disk.go
```diff
@@ -33,12 +36,17 @@ import (
 	"github.com/obolnetwork/charon/tbls/tblsconv"
 )
 
-// loadDefinition returns the cluster definition from disk (or the test definition if configured).
+// loadDefinition returns the cluster definition from disk or network. It returns the test definition if configured.
```
corverroos, line 39: also please update the flag description

corverroos, line 39: `from disk or a HTTP URL`

dkg/disk.go
```diff
@@ -33,12 +36,17 @@ import (
...
 func loadDefinition(conf Config) (cluster.Definition, error) {
 	if conf.TestDef != nil {
 		return *conf.TestDef, nil
 	}
 
+	_, err := url.ParseRequestURI(conf.DefFile)
```
corverroos, line 45: suggest checking scheme for http or https

corverroos (approved):

dkg/disk.go
corverroos: what happens for `file://local/definition.jon`?

dkg/disk.go
corverroos: just pass in context and use a 5s timeout

dkg/disk.go
xenowits: it should be a valid URI with either`HTTP` or `HTTPS` scheme. It rejects other schemes like `file` or `ftp`

dkg/disk.go
xenowits: done. for both `loadDefinition` and `fetchDefinition`

dkg/disk.go
corverroos:
```
func fetchDefinition(ctx context.Context, url string) (cluster.Definition, error) {
    ctx, cancel := context.WithCancel(ctx, time.Second*5)
    defer cancel()
    
    req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
    
    resp, err = new(http.Client).Do(req)
```

dkg/disk_internal_test.go
```diff
@@ -0,0 +1,161 @@
...
+			b, _ := validDef.MarshalJSON()
+			_, _ = w.Write(b)
+		case "/invalidDef":
+			b, _ := invalidDef.MarshalJSON()
+			_, _ = w.Write(b)
+		}
+	}))
+	defer server.Close()
+
+	type args struct {
```
corverroos, line 51: no need to this type, can just make it a string

dkg/disk_internal_test.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := fetchDefinition(context.Background(), tt.args.url)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("fetchDefinition() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
corverroos, line 80: use `require.Equal`

dkg/disk_internal_test.go
```diff
@@ -0,0 +1,161 @@
...
+	invalidFile := "invalid-cluster-definition.json"
+	err = os.WriteFile(invalidFile, []byte{1, 2, 3}, 0o666)
+	require.NoError(t, err)
+
+	defer func() {
+		require.NoError(t, os.Remove(validFile))
+		require.NoError(t, os.Remove(invalidFile))
+	}()
+
+	type args struct {
```
corverroos, line 108: same, just make it a string (DefFile values)

dkg/disk_internal_test.go
```diff
@@ -0,0 +1,161 @@
...
+			name:    "Fetch invalid definition",
+			args:    args{url: fmt.Sprintf("%s/%s", server.URL, "invalidDef")},
+			want:    invalidDef,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := fetchDefinition(context.Background(), tt.args.url)
+			if (err != nil) != tt.wantErr {
```
corverroos, line 76: nit: use `require` package
```
if test.wantErr {
    require.Error(t, err)
    return
}
```

Issue #994, opened 2022-08-18
Drop data-dir in favour of explicit paths


corverroos:
## Problem to be solved

We currently have a flag `--data-dir` which is only used to implicitly load/save the `charon-enr-private-key`. While we have explicit flags for `--lock-file` and `--definition-file`. 

Making the private key file explicit has the following benefits:
  - Allows using different filenames: e.g. bootnode-enr-private-key or any user defined name
  - Makes config more obvious, `--data-dir` implies there are other things in that folder as well...
  - Aligns with `--lock-file` and `--definition-file` which is explicitly defined

## Proposed solution

- Add a `--private-key` flag with default value `.charon/charon-enr-private-key`
- Deprecate `--data-dir` and add warning if it is populated.
- Be backwards compatible
- Add a `--simnet-validator-keys` flag that contains directory to lead private key shares, default value `.charon/validator_keys/`

xenowits:
Note that the ticket is concerned only about `charon run` and not other subcommands.

PR #1074, opened 2022-09-03
cmd: drop data-dir flag

xenowits:
* Deprecate `--data-dir` flag
* Add `--private-key` and `--simnet-validator-keys` flags

category: refactor
ticket: #994

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -114,6 +114,8 @@ func bindPrivKeyFlag(cmd *cobra.Command, dataDir, privKeyFile *string) { //nolin
 			if _, err := os.Open(*privKeyFile); errors.Is(err, os.ErrNotExist) {
 				return errors.New("private key file doesn't exist")
 			}
+
+			return nil
 		}
 
 		// --data-dir present but --private-key absent
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -48,7 +48,7 @@ func newRunCmd(runFunc func(context.Context, app.Config) error) *cobra.Command {
 
 	bindRunFlags(cmd, &conf)
 	bindNoVerifyFlag(cmd.Flags(), &conf.NoVerify)
-	bindDataDirFlag(cmd.Flags(), &conf.DataDir)
+	bindDataDirFlag(cmd, &conf.DataDir)
 	bindP2PFlags(cmd.Flags(), &conf.P2P)
 	bindLogFlags(cmd.Flags(), &conf.Log)
 	bindFeatureFlags(cmd.Flags(), &conf.Feature)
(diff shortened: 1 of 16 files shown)


OisinKyne:
Can we alias `private-key` to `enr` too?

cmd/run.go
```diff
@@ -71,7 +71,9 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
 	cmd.Flags().StringVar(&config.JaegerService, "jaeger-service", "charon", "Service name used for jaeger tracing.")
 	cmd.Flags().BoolVar(&config.SimnetBMock, "simnet-beacon-mock", false, "Enables an internal mock beacon node for running a simnet.")
 	cmd.Flags().BoolVar(&config.SimnetVMock, "simnet-validator-mock", false, "Enables an internal mock validator client when running a simnet. Requires simnet-beacon-mock.")
+	cmd.Flags().StringVar(&config.SimnetValidatorKeys, "simnet-validator-keys", ".charon/validator_keys", "The path to the directory containing simnet validator key shares.")
 	cmd.Flags().BoolVar(&config.BuilderAPI, "builder-api", false, "Enables the builder api. Will only produce builder blocks. Builder API must also be enabled on the validator client. Beacon node must be connected to a builder-relay to access the builder network.")
+	cmd.Flags().StringVar(&config.PrivKeyFile, "private-key", ".charon/charon-enr-private-key", "The path to the enr private key.")
```
corverroos, line 76: suggest decoupling flag migration/backwards compatibility from the logic.

```
func bindPrivateKeyFlag(cmd *cobra.Command, privateKeyFile *string) {
  // define `--private-key-file` and `--data-dir` flags here, do overrides and resolving to actual privateKeyFile here. Log or error if anything weird
}
```

Think about edge cases:
 - `datadir` configured and file present, `private-key` configured and not file present
 - `datadir` configred and file not present, `private-key` configured and file present
  - etc etc etc

corverroos (changes requested):

corverroos:
> Can we alias `private-key` to `enr` too?

@OisinKyne not sure what you mean? an ENR isn't a private key, it is a public key at most?

cmd/run.go
xenowits: `bindPrivKeyFlag` should only `RESOLVE` to actual private-key-file. It should not handle scenarios in which the keyfile doesn't exist at the `private-key-file` path. This is because all the individual subcommands, ex, `charon create enr`, `charon run`, `charon enr`, `charon bootnode` etc. function differently on the keyfile input. For ex, `charon create enr` would `CREATE` a new file if one doesn't already exist. While, `charon run` will throw an error on startup as it won't create a new one.

cmd/run.go
```diff
@@ -95,15 +96,38 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
...
+	preRunE := cmd.PreRunE // Allow multiple wraps of PreRunE.
+	cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
+		ctx := log.WithTopic(cmd.Context(), "cmd")
+		if *dataDir != "" {
+			log.Warn(ctx, "Deprecated flag 'data-dir' used, please use new flag 'private-key'.", nil)
+
+			if *privKeyFile == charonEnrPrivKey { // use --data-dir as --private-key is absent
+				*privKeyFile = p2p.KeyPath(*dataDir)
+			} else { // override --data-dir with --private-key as both are provided
+				log.Warn(ctx, "'private-key' flag would take precedence as both 'data-dir' and 'private-key' flags are provided.", nil)
```
xenowits, line 114: pls suggest appropriate text

Issue #1064, opened 2022-08-31
Add `DutyPrepareAggregator` to DutyDB


dB2510:
## Problem to be solved

We need to introduce `DutyPrepareAggregator` which will be responsible for determining whether a validator is an aggregator and requests the beacon node to temporarily subscribe to the committee subnet that will be queried in the next step  to perform actual aggregation. We need to add support to store `BeaconCommitteeSubscriptionResponse` in DutyDB.

## Proposed solution

- Introduce a new duty type named `DutyPrepareAggregator` to core/types.go after DutyRandao.
- Add a `core.SignedData` implementation that wraps the proposed new type `v2.BeaconCommitteeSubscription` (Refer [design doc](https://docs.google.com/document/d/1_J8Y9M4FsONmEENSPwzzZW9n_zrnmAPGoSO2sm6DIO8)). This will be submitted by VCs as a partial signature.
- Add support for `DutyPrepareAggregator` to DutyDB, adding a new query endpoint `AwaitCommitteeSubnetResponse(ctx context.Context, validatorIndex eth2p0.ValidatorIndex, slot eth2p0.Slot)`. This is required by validatorapi to return `BeaconCommitteeSubscriptionResponse` to VC.
- Add unit tests for changes in DutyDB.

PR #1076, opened 2022-09-05
core: add signed data type for BeaconCommitteeSubscription

dB2510:
Adds DutyPrepareAggregator and signed data type for BeaconCommitteeSubscription.

category: feature
ticket: #1064

### DIFF
--- a/core/signeddata.go
+++ b/core/signeddata.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/eth2util"
+	"github.com/obolnetwork/charon/eth2util/eth2exp"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


core/signeddata.go
```diff
@@ -691,6 +692,56 @@ func (s SignedRandao) clone() (SignedRandao, error) {
 	return resp, nil
 }
 
+// NewPartialSignedBeaconCommitteeSubscription is a convenience function which returns new partially signed BeaconCommitteeSubscription.
+func NewPartialSignedBeaconCommitteeSubscription(sub eth2exp.BeaconCommitteeSubscription, shareIdx int) ParSignedData {
```
corverroos, line 696: wondering if we can drop the `Beacon` part without loss of context?

core/signeddata_test.go
```diff
@@ -67,3 +68,20 @@ func TestSetVersionedValidatorRegistrationSig(t *testing.T) {
 	require.NoError(t, err)
 	require.NotEqual(t, clone.Signature(), registration.Signature())
 }
+
+func TestSetBeaconCommitteeSubscription(t *testing.T) {
```
corverroos, line 72: can this be refactored into a table test?

corverroos (approved):

core/signeddata.go
dB2510: i suggest we should keep this to align with underlying type

core/signeddata_test.go
dB2510: done

PR #1081, opened 2022-09-06
core/dutydb: fix lint

dB2510:
Fixes golangci-lint on main.

category: fixbuild
ticket: none


core/dutydb/memory.go
```diff
@@ -70,6 +70,8 @@ func (db *MemDB) Shutdown() {
 }
 
 // Store implements core.DutyDB, see its godoc.
+//
+//nolint:gocognit
```
dB2510, line 74: cognitive complexity of Store is just right, i don't this can be improved any further

Issue #1084, opened 2022-09-06
Bypass definition hash verification with `--no-verify` flag


dB2510:
## Problem to be solved

We are having issues with generating definition hash on DV launchpad side. Due to this, currently DV launchpad supports cluster-definition.json without definition_hash field. We need to bypass definition_hash verification for `charon dkg` command using `--no-verify` flag.

## Proposed solution

Move config hash verification and definition hash verification from UnmarshalJSON method to Verify method of Definition type. Add unit test to verify above usecase. Also move lock hash verification from UnmarshalJSON method to Verify method of Lock type.

PR #1090, opened 2022-09-07
dkg: add --no-verify for hash verifications

dB2510:
Configure --no-verify for config_hash and definition_hash verifications.

category: refactor
ticket: #1084

### DIFF
--- a/cluster/cluster_test.go
+++ b/cluster/cluster_test.go
@@ -83,7 +83,6 @@ func TestEncode(t *testing.T) {
 		require.NoError(t, err)
 
 		require.Equal(t, b1, b2)
-		require.Equal(t, definition, definition2)
 
 		lock := cluster.Lock{
 			Definition:         definition,
@@ -127,7 +126,6 @@ func TestEncode(t *testing.T) {
 		require.NoError(t, err)
 
 		require.Equal(t, b1, b2)
-		require.Equal(t, lock, lock2)
 	}
 }
 
--- a/cluster/definition.go
+++ b/cluster/definition.go
@@ -16,6 +16,7 @@
 package cluster
 
 import (
+	"bytes"
 	"encoding/json"
 	"io"
 	"time"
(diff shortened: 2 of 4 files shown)


dkg/disk.go
```diff
@@ -59,6 +60,18 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
 		return cluster.Definition{}, errors.Wrap(err, "unmarshal definition")
 	}
 
+	// Verify config hash and definition hash from json string and resultant cluster.Definition.
+	if !conf.NoVerify {
+		expected, err := json.MarshalIndent(res, "", " ")
```
corverroos, line 65: requiring a fixed json indentation as part of verification isn't correct. The json indentation shouldn't matter.

corverroos (changes requested):
Comparing raw json with fixed indentation is not robust

dkg/disk.go
dB2510: yeah right, corrected 👍

cluster/definition.go
```diff
@@ -169,6 +169,58 @@ func (d Definition) Verify() error {
 	return nil
 }
 
+func (d Definition) VerifyHashesJSON(data []byte) error {
```
corverroos, line 172: this API doesn't feel good. Needing to pass the raw bytes again doesn't feel intuitive.

I would go for something like:
```
var d Definition
err = json.Unmarshal(rawBytes, &d)

err = d.VerifyHashes()
```

For this to work, you need to unmarshal the hashes into private unexported fields so you can calculate the actual hashes and compare.

dB2510: i was also thinking the same to include this as private fields but was hesitant to do any changes in Definition type

dB2510: for this to work we would need to do json marshal every time when new Definition is called

dB2510: suggest: keep `VerifyHashesJSON(data []byte) error` but as stateless function which would introduce two function parameters as `VerifyHashesJSON(rawBytes, configHash, defHash []byte) error` ?

cluster/cluster_test.go
```diff
@@ -83,7 +83,6 @@ func TestEncode(t *testing.T) {
 		require.NoError(t, err)
 
 		require.Equal(t, b1, b2)
-		require.Equal(t, definition, definition2)
```
dB2510, line 86: if raw bytes are equal then it doesn't make sense for this require

dkg/disk_internal_test.go
```diff
@@ -91,49 +92,88 @@ func TestLoadDefinition(t *testing.T) {
...
...,"config_signature":"0xc8a306ec102a83dce35be9a552d0b30a09e74b1c1b2316596245c0e8f2ffa5d063051373993ecb694d82e2d65a398ecd5ef06599661903470385998d19a702f001","enr_signature":"0x61c44ee84dfc3991fe9de0d25f2bc3cad7feb0d3341208c894f455a233bcb1d2647b4fd29c376ceb2e5d2990fb2af7b6eae10bfe5b55963404ea2571222b350c01"},{"address":"0xfDdd1CF7733Fd8a638020e963792f9Fcc0182Bf4","enr":"enr:-Iu4QNgEtRy6wbpdPCXrj52_rF4Ur7OQf6mOg1xfRpmzPgRYW-QSA-oUslOTmIPL8etUIg95quQoRJg9FIILIa6990uAgmlkgnY0gmlwhH8AAAGJc2VjcDI1NmsxoQLM5474DZHwbqwSbQFLrAO8PNh2AdZOXTYGy1ZItyDJaoN0Y3CC52uDdWRwguds","config_signature":"0xddb059b67e0603c0073536225e3a5ae5a7eeae178ec4c31dda521ed9b0209dc90b7e3295ee5ad4f35fbe0b8d62adf4aa4cfbddd3106f7563d234e0b134d9e93701","enr_signature":"0x3c556e60c4b44a4ea015f860759fa01e2e8664d105455b71e55845f451d7ec5e048ee56999c9f5333024bfcc087471fcb7d8e8a2685cb59d63266b4df092104501"}],"uuid":"04513690-AA41-CE01-6281-7901E9FB6D87","version":"v1.2.0","timestamp":"2022-09-07T18:46:30+05:30","num_validators":1,"threshold":2,"fee_recipient_address":"0xd805a5fcea20d3d27d3eee59d5dd5749e3271617","withdrawal_address":"0x75e896f172869cf3ade31c97f681cc1a4015ceed","dkg_algorithm":"default","fork_version":"0x00000000"}`)
```
corverroos, line 101: not a fan of this wall of text. Making changes to this is impossible. Rather:
 - unmarshal valid definition above from disk into map[string]any
 - delete the config and definition hashes
 - marshal to disk

corverroos (approved):

cluster/cluster_test.go
corverroos: rather add an ``Equal` method to Definition

dkg/disk_internal_test.go
```diff
@@ -91,49 +92,88 @@ func TestLoadDefinition(t *testing.T) {
...
 				return
 			}
 			require.NoError(t, err)
-			require.Equal(t, tt.want, got)
+			require.True(t, compareDefinitions(t, got, tt.want))
 		})
 	}
 }
+
+func compareDefinitions(t *testing.T, a, b cluster.Definition) bool {
```
corverroos, line 169: rather add an `Equal` method to Definition

dB2510:
taken care by https://github.com/ObolNetwork/charon/pull/1099

PR #1099, opened 2022-09-08
dkg: add --no-verify for hash verifications

dB2510:
Configure --no-verify for config_hash and definition_hash verifications by adding VerifyHashes method to Definition.

category: refactor
ticket: #1084

### DIFF
--- a/cluster/definition.go
+++ b/cluster/definition.go
@@ -305,27 +305,32 @@ func (d *Definition) UnmarshalJSON(data []byte) error {
 		return errors.New("unsupported version")
 	}
 
+	*d = def
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


cluster/definition.go
```diff
@@ -305,28 +305,33 @@ func (d *Definition) UnmarshalJSON(data []byte) error {
 		return errors.New("unsupported version")
 	}
 
+	*d = def
+
+	return nil
+}
+
+// VerifyHashes returns an error if hashes populated from json object doesn't matches actual hashes.
+func (d Definition) VerifyHashes() error {
```
corverroos, line 314: `VerifyDefinitionHashes` since we will add `VerifyLockHashes` and both method appear on the Lock struct

corverroos (approved):

PR #1102, opened 2022-09-08
tbls: add signatures verification benchmark test

corverroos:
Adds a bls signature verification benchmark test.

category: test
ticket: none


xenowits (approved):

dB2510 (approved):
3.095s is slow for 100 verifications.

corverroos:
> 3.095s is slow for 100 verifications.

That is not the number to look at, the relevant number is `8223595 ns/op` or `8ms per verification`.

Issue #1093, opened 2022-09-07
Add support for `v1.BeaconCommitteeSubscriptionsSubmitter` to bcast


xenowits:
## Problem to be solved

Add support for `v1.BeaconCommitteeSubscriptionsSubmitter` to `bcast` component.

## Proposed solution

* If broadcast `DutyPrepareAggregator` using `SubmitBeaconCommitteeSubscriptionsV2` fails with a 404 error (if the beacon node doesn’t support v2 yet):
  * is_aggregator = `eth2exp.CalculateCommitteeSubscriptionResponse()`
  * Create a new `v1.BeaconCommitteeSubscription{ ..., IsAggregator: is_aggregator}` object
  *  Broadcast using `SubmitBeaconCommitteeSubscriptions` (v1).

### Out of Scope

None

PR #1112, opened 2022-09-11
core/bcast: add support for v1.BeaconCommitteeSubscriptionsSubmitter to bcast

dB2510:
Adds support for v1.BeaconCommitteeSubscriptionsSubmitter to bcast and beaconmock.

category: feature
ticket: #1093


core/bcast/bcast.go
```diff
@@ -144,14 +145,35 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, pubkey core.
...
 		if !ok {
-			return errors.New("invalid beacon committee subscription")
+			return errors.New("invalid beacon committee sub")
 		}
 
-		_, err = b.eth2Cl.SubmitBeaconCommitteeSubscriptionsV2(ctx, []*eth2exp.BeaconCommitteeSubscription{&subscription.BeaconCommitteeSubscription})
-		if err == nil {
-			log.Info(ctx, "Beacon committee subscription successfully submitted to beacon node", nil)
+		_, err = b.eth2Cl.SubmitBeaconCommitteeSubscriptionsV2(ctx, []*eth2exp.BeaconCommitteeSubscription{&sub.BeaconCommitteeSubscription})
+		if err != nil && strings.Contains(err.Error(), "404") {
```
corverroos, line 154: I think we can just try on any error, little simpler

corverroos: can also do:

```
if err == nil {
  return nil
}
log.Debug(ctx, "v2 failed ..."
// do v1
```

corverroos (approved):

core/bcast/bcast.go
dB2510: Done!

PR #1140, opened 2022-09-20
core: add table tests for unsigned data

dB2510:
Adds table tests for unsigned data implementations.

category: test
ticket: none

### DIFF
--- a/core/unsigneddata_test.go
+++ b/core/unsigneddata_test.go
@@ -51,14 +51,7 @@ func TestUnsignedDataClone(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			clone, err := test.data.Clone()
 			require.NoError(t, err)
-
-			b1, err := test.data.MarshalJSON()
-			require.NoError(t, err)
-
-			b2, err := clone.MarshalJSON()
-			require.NoError(t, err)
-
-			require.Equal(t, b1, b2)
+			require.Equal(t, test.data, clone)
 		})
 	}
 }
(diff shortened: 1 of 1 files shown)


core/unsigneddata_test.go
```diff
@@ -24,30 +24,41 @@ import (
...
-	clone, err := block.Clone()
-	require.NoError(t, err)
-	block2 := clone.(core.VersionedBlindedBeaconBlock)
-	slot2, err := block2.Slot()
-	require.NoError(t, err)
+			b2, err := clone.MarshalJSON()
+			require.NoError(t, err)
 
-	require.Equal(t, slot1, slot2)
+			require.Equal(t, b1, b2)
```
xenowits, line 61: Let's also add another line of test for `UnmarshalJSON()`. Like, for VersionedBeaconBlock:
```go
	x := new(core.VersionedBeaconBlock)
	require.NoError(t, x.UnmarshalJSON(b1))
	require.Equal(t, *x, test.data)
```

xenowits (approved):

core/unsigneddata_test.go
dB2510: i don't think we need this since Clone() method already takes care of this

xenowits: Btw, Clone() also uses the `MarshalJSON()` method, so actually, we don't need to test the marshal/unmarshal methods explicitly.

corverroos (approved):

core/unsigneddata_test.go
dB2510: sure

PR #1155, opened 2022-09-22
dkg: increase GET remote definition timeout

dB2510:
Increases GET remote definition timeout to 10s from 5s. Also add logs.

category: misc
ticket: none

### DIFF
--- a/dkg/disk.go
+++ b/dkg/disk.go
@@ -54,6 +54,8 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
 		if err != nil {
 			return cluster.Definition{}, errors.Wrap(err, "read definition")
 		}
+
+		log.Info(ctx, "Definition file loaded from remote URI", z.Str("URI", conf.DefFile))
 	} else {
 		buf, err := os.ReadFile(conf.DefFile)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


dkg/disk.go
```diff
@@ -63,6 +65,8 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
 		if err = json.Unmarshal(buf, &def); err != nil {
 			return cluster.Definition{}, errors.Wrap(err, "unmarshal definition")
 		}
+
+		log.Info(ctx, "Definition file loaded from disk", z.Str("location", conf.DefFile))
```
xenowits, line 69:
```suggestion
		log.Info(ctx, "Definition file loaded from disk", z.Str("path", conf.DefFile))
```

xenowits (approved):

dkg/disk.go
```diff
@@ -54,6 +54,8 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
 		if err != nil {
 			return cluster.Definition{}, errors.Wrap(err, "read definition")
 		}
+
+		log.Info(ctx, "Definition file loaded from remote URI", z.Str("URI", conf.DefFile))
```
corverroos, line 58: move logging of definition hash here (remove from later log)

corverroos: suggest: `Cluster definition downloaded from URL`

URL is more accurate since we only allow HTTP and more common

dkg/disk.go
corverroos, line 69: same: `Cluster definition loaded from disk` and move definition hash here

corverroos (approved):

dkg/disk.go
dB2510: done

dkg/disk.go
dB2510: done

PR #1167, opened 2022-09-25
app: remove cluster_name from labels

xenowits:
Remove `cluster_name` from default labels as we already have a unique identifier for cluster (`cluster_hash`). Moreover, `cluster_name` values are not supposed to be unique as they are user picked.

category: misc
ticket: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -190,7 +190,6 @@ func Run(ctx context.Context, conf Config) (err error) {
 	// Instrument prometheus metrics with cluster and node identifiers.
 	promauto.WrapAndRegister(prometheus.Labels{
 		"cluster_hash": lockHashHex,
-		"cluster_name": lock.Name,
 		"cluster_peer": p2p.PeerName(tcpNode.ID()),
 	})
 


eth2devops (approved):
LGTM

corverroos (changes requested):
This is a nice to have, I'd keep it.

alysaleha:
@corverroos @xenowits I agree that it is nice to have for while when we run a few hundreds of clusters, but when we run thousands of clusters it will be useless, as we will not care about names, and specially this label is not a unique id and can be repeated. Also removing it (and any extra labels) will optimize the cardinality.
What do you think guys?

xenowits:
> @corverroos @xenowits I agree that it is nice to have for while when we run a few hundreds of clusters, but when we run thousands of clusters it will be useless, as we will not care about names, and specially this label is not a unique id and can be repeated. Also removing it (and any extra labels) will optimize the cardinality. What do you think guys?

Aligned on less labels. I don't have any strong opinion to include/remove the label. I also wonder what value including this label brings

Issue #937, opened 2022-08-08
Verify charon-enr-private-key matches lock file before p2p setup


corverroos:
## Problem to be solved

Currently we only verify that the charon-enr-private-key matches the lock file after p2p setup (discv5+bootnode+libp2p). This can result in bootnode thrashing if users have a some random lock file but with a non-matching charon-enr-private-key.

## Proposed solution

Verify the charon-enr-private-key is included in the lock file before doing p2p setup.

PR #1181, opened 2022-09-27
dkg: verify private key before p2p setup

corverroos:
This was only added to `charon run`, also doing this in dkg now.

category: feature
ticket: #937

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -18,7 +18,6 @@
 package app
 
 import (
-	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"encoding/hex"
@@ -165,7 +164,12 @@ func Run(ctx context.Context, conf Config) (err error) {
 		}
 	}
 
-	if err := verifyP2PKey(lock, p2pKey); err != nil {
+	peers, err := lock.Peers()
+	if err != nil {
+		return err
+	}
+
+	if err := p2p.VerifyP2PKey(peers, p2pKey); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


app/app.go
```diff
@@ -675,31 +679,3 @@ func (h httpServeHook) Call(context.Context) error {
 
 	return nil
 }
-
-// verifyP2PKey returns an error if the p2pkey doesn't match any lock operator ENR.
-func verifyP2PKey(lock cluster.Lock, key *ecdsa.PrivateKey) error {
```
corverroos, line 680: moved this to p2p package so DKG can access it

p2p/peer.go
```diff
@@ -139,3 +140,26 @@ func (p *MutablePeer) Subscribe(sub func(Peer)) {
...
+		if err != nil {
+			return errors.Wrap(err, "key to bytes")
+		}
+
+		if bytes.Equal(wantBytes, gotBytes) {
+			return nil
+		}
+	}
+
+	return errors.New("private key not matching any operator ENR") // This message seems to know something about the context in which it is used :(
```
xenowits, line 164: i didn't understand the comment ;)

xenowits (approved):

p2p/peer.go
corverroos: How about: `unknown private key provided, it doesn't match any public key encoded inside the operator ENRs`

corverroos: This functions isn't provided with any operators or ENRs so the fact that it mentions it in the error message is assuming specific implicit external use-case

PR #1184, opened 2022-09-27
core/fetcher: support aggregate attestation not found

corverroos:
Fixes panic in fetcher when `AggregateAttestation` isn't found and go-eth2-client returns `nil,nil`. This can happen when using infura and the `SubmitBeaconCommitteeSubscription` was processed by a different node. Just retry in that case.

Also use proper timeout for eth2exp calls (2s is too short).

category: bug
ticket: none

### DIFF
--- a/app/eth2wrap/success.go
+++ b/app/eth2wrap/success.go
@@ -25,7 +25,7 @@ func isSyncStateOk(s *apiv1.SyncState) bool {
 	return !s.IsSyncing
 }
 
-// isAggregateAttestationOk returns tue if the aggregate attestation is not nil (which can happen if the subscription wasn't done).
+// isAggregateAttestationOk returns true if the aggregate attestation is not nil (which can happen if the subscription wasn't successful).
 func isAggregateAttestationOk(att *phase0.Attestation) bool {
 	return att != nil
 }
--- a/app/eth2wrap/success.go
+++ b/app/eth2wrap/success.go
@@ -15,9 +15,17 @@
 
 package eth2wrap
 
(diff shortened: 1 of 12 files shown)


app/eth2wrap/success.go
```diff
@@ -15,9 +15,17 @@
...
+	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+)
 
 // isSyncStateOk returns tue if the sync state is not syncing.
 func isSyncStateOk(s *apiv1.SyncState) bool {
 	return !s.IsSyncing
 }
+
+// isAggregateAttestationOk returns tue if the aggregate attestation is not nil (which can happen if the subscription wasn't done).
```
xenowits, line 28:
```suggestion
// isAggregateAttestationOk returns true if the aggregate attestation is not nil (which can happen if the subscription wasn't successful).
```

testutil/validatormock/attest.go
```diff
@@ -310,17 +314,18 @@ func attest(ctx context.Context, eth2Cl eth2wrap.Client, signFunc SignFunc, slot
 	return datas, nil
 }
 
-// aggregate does attestation aggregation for the provided validators, selections and attestation datas.
+// aggregate does attestation aggregation for the provided validators, selections and attestation datas and returns true.
+// It returns false if the no aggregation is required.
```
xenowits, line 318:
```suggestion
// It returns false if aggregation is not required.
```

xenowits (approved):

Issue #1183, opened 2022-09-27
Add verification for `SyncCommitteeMessage`


xenowits:
## Problem to be solved

Verify `SyncCommitteeMessage`.

## Proposed solution

Add a new function `VerifySyncMessage()` to `ethutil/signing/signing.go`. Add corresponding test in `signing_test.go`.

### Out of Scope

None.

PR #1216, opened 2022-10-03
eth2util/signing: add sync message verify

dB2510:
Adds signature verification for sync committee message.

category: feature
ticket: #1183

### DIFF
--- a/eth2util/signing/signing_test.go
+++ b/eth2util/signing/signing_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
+	"github.com/attestantio/go-eth2-client/spec/altair"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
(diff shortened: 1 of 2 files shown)


eth2util/signing/signing_test.go
```diff
@@ -263,12 +264,35 @@ func TestVerifyAggregateAndProof(t *testing.T) {
...
 
 	sig, pubkey := sign(t, sigData[:])
-	agg.Signature = tblsconv.SigToETH2(sig)
+	agg.Signature = sig
 
 	require.NoError(t, signing.VerifyAggregateAndProof(context.Background(), bmock, pubkey, agg))
 }
 
-func sign(t *testing.T, data []byte) (*bls_sig.Signature, *bls_sig.PublicKey) {
+func TestVerifySyncCommitteeMessage(t *testing.T) {
```
corverroos, line 272: can we make this a table test rather?

corverroos (approved):

eth2util/signing/signing_test.go
dB2510: since each test require different signing data and calls different function, i don't think it would add much value converting these tests to table test.

Issue #1203, opened 2022-09-29
Upgrade EIP712 signature structure


corverroos:
## Problem to be solved

The cluster definition EIP712 signature structure needs to be updated for improved UX in the DV Launchpad.

## Proposed solution

See this [design doc](https://docs.google.com/document/d/1lYP1IU56JLtadrdaZTA9J1edGJjEaG8hLesyQlNxjCc/edit#).

TL;DR:
 - Create two `PrimaryTypes` signatures, `ENR` and `ConfigHash`.
 - Only include `name=Obol`, `version=1`,`chainID=<charinID>` in the `EIP712Domain` structure.
 - Only include `config_hash=string(configHashHex)` in the ConfigHash message.
 - Only include `end=string(enrString)` in the ENR message.
 
 This is a breaking change ito signature verification. But since there are no actual signed lock files or definition files in the wild, we are deprecating support for older signatures versions.
 
 Since v1.3 is still in draft, add this as part of v1.3.
 Error when unmarshalling any older version (1.0-1.2) that contains signatures.
 Skip signature verification for older versions (1.0-1.2).

We need to document how these signatures are created since it is part of the cluster definition spec.
See if we can find examples of other spec of EIP712 signatures. Add that to the spec in the `configuration.md` file.
### Out of Scope

We need to release v1.3 at some point before v0.11 release.

dB2510:
Closed with final PR #1234

PR #1217, opened 2022-10-03
cluster: skip sig verification for versions earlier than v1.3.0

xenowits:
* For older definition versions (`< v1.3.0`):
* Skip verification of operator enr and config signatures 
* Error with "older version signatures not supported" when unmarshalling old definitions
* Add tests

category: feature
ticket: https://github.com/ObolNetwork/charon/issues/1203


cluster/definition.go
```diff
@@ -139,6 +140,12 @@ func (d Definition) NodeIdx(pID peer.ID) (NodeIdx, error) {
 
 // VerifySignatures returns true if all config signatures are fully populated and valid. A verified definition is ready for use in DKG.
 func (d Definition) VerifySignatures() error {
+	// Skip signature verification for definition versions earlier than v1.3. This is because v1.3 introduced a breaking change to how EIP712
+	// signatures are calculated. https://github.com/ObolNetwork/charon/issues/1203
+	if semver.Compare(d.Version, v1_3) == -1 {
+		return nil
+	}
```
corverroos, line 147: best not introduce additional libraries without consulting with the team, please remove and rather use simple comparison: `if d.Version==v1_0 || d.Version==v1_1 || d.Version==v1_2`.

corverroos: Also, as mentioned earlier please error if signatures are present in this case, i.e., require empty signatures

corverroos: can extract the above condition into a function `supportEIP712Sigs`

cluster/cluster_test.go
```diff
@@ -49,20 +50,29 @@ func TestEncode(t *testing.T) {
 					{
 						Address:         testutil.RandomETHAddress(),
 						ENR:             fmt.Sprintf("enr://%x", testutil.RandomBytes32()),
-						ConfigSignature: testutil.RandomSecp256k1Signature(),
-						ENRSignature:    testutil.RandomSecp256k1Signature(),
+						ConfigSignature: []byte{},
```
corverroos, line 53: Suggest adding them as that is default, then removing if version doesn't support it.

dB2510:
Cleared and completed in #1226

Issue #1222, opened 2022-10-04
Consensus upgrades not backwards compatible


corverroos:
# 🐞 Bug Report

## Description

The latest (main branch) upgrades to consensus wire protobufs are not backwards compatible with v0.10.1 or earlier.

## 🔬 Minimal Reproduction

compose `run version matrix with dkg` has "invalid signature" errors.

PR #1223, opened 2022-10-04
core/consensus: fix wire compatibility issue

corverroos:
A second signature wasn't the correct approach for backwards compatibility. Since we marshal the proto to bytes when calculating the hash, unknown fields are actually retained. That means that hashing the whole proto always results in the same hash (and signature) even if the wire contained unknown fields.

category: bug
ticket: #1222


core/corepb/v1/consensus.proto
```diff
@@ -15,10 +15,23 @@ message QBFTMsg {
   core.corepb.v1.UnsignedDataSet value_legacy          = 5 [deprecated = true]; // Prefer value.
   int64                          prepared_round        = 6;
   core.corepb.v1.UnsignedDataSet prepared_value_legacy = 7 [deprecated = true]; // Prefer prepared_value.
-  bytes                          signature_legacy      = 8 [deprecated = true]; // Signature containing legacy values.
+  bytes                          signature             = 8;
   google.protobuf.Any            value                 = 9;
   google.protobuf.Any            prepared_value        = 10;
-  bytes                          signature             = 11; // Signature containing new values.
```
corverroos, line 21: second signature removed

dB2510 (approved):

xenowits (approved):

PR #1226, opened 2022-10-05
cluster: skip sig verification for versions earlier than v1.3.0

dB2510:
Skip EIP712 signature verification for versions earlier than v1.3.0.

category: feature
ticket: #1203

### DIFF
--- a/cluster/cluster_test.go
+++ b/cluster/cluster_test.go
@@ -34,7 +34,7 @@ import (
 //go:generate go test . -v -update -clean
 
 func TestEncode(t *testing.T) {
-	for _, version := range cluster.SupportedVersionsForT(t) {
+	for _, version := range cluster.SupportedVersionsForT() {
 		t.Run(version, func(t *testing.T) {
 			vStr := strings.ReplaceAll(version, ".", "_")
 			rand.Seed(1)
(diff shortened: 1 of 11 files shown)


cluster/version.go
```diff
@@ -53,7 +51,7 @@ func isJSONv1x3(version string) bool {
 }
 
 // SupportedVersionsForT returns the supported definition versions for testing purposes only.
-func SupportedVersionsForT(*testing.T) []string {
```
corverroos, line 56: please do not remove this, it enforces use in testing only

cluster/cluster_test.go
```diff
@@ -64,6 +64,14 @@ func TestEncode(t *testing.T) {
 			)
 			require.NoError(t, err)
 			definition.Version = version
+
+			if version == "v1.0.0" || version == "v1.1.0" || version == "v1.2.0" {
```
corverroos, line 68: add comment why

cluster/definition.go
```diff
@@ -140,6 +140,11 @@ func (d Definition) NodeIdx(pID peer.ID) (NodeIdx, error) {
 
 // VerifySignatures returns true if all config signatures are fully populated and valid. A verified definition is ready for use in DKG.
 func (d Definition) VerifySignatures() error {
+	// Skip signature verification for definition versions earlier than v1.3 since there are no EIP712 signatures before v1.3.0.
+	if !supportEIP712Sigs(d.Version) {
+		return nil
```
corverroos, line 145: add `TODO: maybe error if signatures present`

cluster/definition.go
```diff
@@ -437,6 +451,10 @@ func unmarshalDefinitionV1x2or3(data []byte) (def Definition, err error) {
 	return def, nil
 }
 
+func supportEIP712Sigs(version string) bool {
```
corverroos, line 454: add godoc

corverroos (approved):

cluster/definition.go
dB2510: we error in UnmarshalJSON though

corverroos: sure, but we should also maybe error here. If I create a old version with signatures, it shouldn't be valid. On that note, we should also error when marshalling probably.

dB2510: done

PR #1238, opened 2022-10-07
cluster: fix function comment

xenowits:
Fix function comment and add tests.

category: refactor
ticket: none


cluster/definition.go
```diff
@@ -438,10 +438,10 @@ func unmarshalDefinitionV1x2or3(data []byte) (def Definition, err error) {
 	return def, nil
 }
 
-// supportEIP712Sigs returns true if version is earlier than v1.3.0.
```
xenowits, line 441: should have been "returns false if version is earlier than v1.3.0"

Issue #1188, opened 2022-09-28
Add support for `SyncCommitteeMessage` to validatorAPI


xenowits:
## Problem to be solved

Add support for `POST /eth/v1/beacon/pool/sync_committees` via the `eth2client.SyncCommitteeMessagesSubmitter` interface to `validatorapi`, verifying the signature and storing a new `DutySyncCommitteeMessage` to parsigdb.

## Proposed solution

* Add a method `SubmitSyncCommitteeMessages()` to `core/validatorapi/validatorapi.go`.
* Add `eth2client.SyncCommitteeMessagesSubmitter` to the `Handler` interface.
* Add endpoint `POST /eth/v1/beacon/pool/sync_committees` to `NewRouter`.
* Add a new method `submitSyncCommitteeMessages()` that calls `SyncCommitteeMessagesSubmitter.SubmitSyncCommitteeMessages()` method.
* Add a new method `SubmitSyncCommitteeMessages()` on `Component` .
* Add appropriate tests in `router_internal_test.go` and `validatorapi_test.go`.

### Out of Scope

None.

PR #1244, opened 2022-10-10
core/validatorapi: add support for `DutySyncMessage`

xenowits:
Adds support for `DutySyncMessage` to validatorapi component. Also, adds tests.

category: feature
ticket: #1188

### DIFF
--- a/core/validatorapi/router_internal_test.go
+++ b/core/validatorapi/router_internal_test.go
@@ -207,7 +207,7 @@ func TestRawRouter(t *testing.T) {
 	})
 }
 
-//nolint:maintidx // This function a test of tests, so analyses as "complex".
+//nolint:maintidx // This function is a test of tests, so analysed as "complex".
 func TestRouter(t *testing.T) {
 	t.Run("attesterduty", func(t *testing.T) {
 		handler := testHandler{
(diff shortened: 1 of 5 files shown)


core/validatorapi/router_internal_test.go
```diff
@@ -548,6 +548,23 @@ func TestRouter(t *testing.T) {
 
 		testRouter(t, handler, callback)
 	})
+
+	t.Run("submit_sync_committee_messages", func(t *testing.T) {
+		msgs := []*altair.SyncCommitteeMessage{testutil.RandomSyncCommitteeMessage()}
```
corverroos, line 553: submit two to test the slice

core/validatorapi/validatorapi_test.go
```diff
@@ -1321,3 +1321,39 @@ func TestComponent_SubmitAggregateAttestations(t *testing.T) {
...
+
+	vapi.Subscribe(func(_ context.Context, duty core.Duty, set core.ParSignedDataSet) error {
+		require.Equal(t, core.NewSyncMessageDuty(int64(msg.Slot)), duty)
+
+		pk, err := core.PubKeyFromBytes(pubkey[:])
+		require.NoError(t, err)
+
+		data, ok := set[pk]
+		require.True(t, ok)
+		require.Equal(t, core.NewPartialSignedSyncMessage(msg, 0), data)
```
corverroos, line 1353: how do you know this code is actually called? (please do not use ctx cancel pattern, anymore, rather use something like

```
done := make(chan struct)
...
close(done)
...
<-done   
```

corverroos (approved):

core/validatorapi/validatorapi_test.go
xenowits: we can add a counter which is incremented each time the subscription function is called. we can use it to assert that this code block is actually called.

xenowits: Sure, but this test doesn't use the ctx cancel pattern

corverroos: as long as it is threadsafe then sure

Issue #1100, opened 2022-09-08
Handle feeRecipient gracefully


OisinKyne:
# 🐞 Bug Report

### Description

It's not clear we use the `feeRecipient` field barely at all bar MEV-boost registration. We should assert that we use it properly from the cluster-lock and nowhere else. 

### Has this worked before in a previous version?

No

## 🌍  Your Environment

**What version of Charon are you running? (Which release)**

v0.10

**Anything else relevant (validator index / public key)?**

## Solution

Log feeRecipient in charon run.

OisinKyne:
Outstanding Qs:

- How do we override default validator set fee recipients
- Can we default feeRecipient to withdrawalAddress? 
- Can we enforce a single feeRecipient

corverroos:
Similar to builder registrations, the beacon api has a `POST /eth/v1/validator/prepare_beacon_proposer` endpoint to provide your desired fee recipient for block proposal. See https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/prepareBeaconProposer. 

We should:
 - Call this endpoint with the lock file fee recipient
 - Ensure the correct fee recipient is received when fetching a block

PR #1246, opened 2022-10-10
app: handle fee recipient

dB2510:
Handles fee recipient address by adding a slot subscriber to scheduler which calls prepare_beacon_proposer endpoint at the start of each epoch.

category: feature
ticket: #1100

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -687,8 +687,11 @@ func wireTracing(life *lifecycle.Manager, conf Config) error {
 
 // setFeeRecipient returns a slot subscriber for scheduler which calls prepare_beacon_proposer endpoint at start of each epoch.
 func setFeeRecipient(eth2Cl eth2wrap.Client, pubkeys []eth2p0.BLSPubKey, feeRecipient string) func(ctx context.Context, slot core.Slot) error {
+	onStartup := false
+
 	return func(ctx context.Context, slot core.Slot) error {
-		if !slot.FirstInEpoch() {
+		// Either call if it is first slot in epoch or on charon startup.
+		if onStartup && !slot.FirstInEpoch() {
 			return nil
 		}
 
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -674,6 +685,37 @@ func wireTracing(life *lifecycle.Manager, conf Config) error {
 	return nil
 }
 
+// setFeeRecipient returns a slot subscriber for scheduler which calls prepare_beacon_proposer endpoint at start of each epoch.
+func setFeeRecipient(eth2Cl eth2wrap.Client, pubkeys []eth2p0.BLSPubKey, feeRecipient string) func(ctx context.Context, slot core.Slot) error {
+	return func(ctx context.Context, slot core.Slot) error {
+		if !slot.FirstInEpoch() {
```
corverroos, line 691: also do this on startup (ie the first time it is called)

core/fetcher/fetcher.go
```diff
@@ -224,6 +227,15 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, defSet core
 			return nil, err
 		}
 
+		// Ensure fee recipient is correctly populated in block.
+		if block.Version == spec.DataVersionBellatrix {
+			actual := fmt.Sprintf("%#x", block.Bellatrix.Body.ExecutionPayload.FeeRecipient)
+			if actual != f.feeRecipientAddress {
+				log.Warn(ctx, "Fee recipient address different than expected", nil,
```
corverroos, line 234: `Proposing block with unexpected fee recipient address`

corverroos (approved):

app/app.go
```diff
@@ -674,6 +685,43 @@ func wireTracing(life *lifecycle.Manager, conf Config) error {
 	return nil
 }
 
+// setFeeRecipient returns a slot subscriber for scheduler which calls prepare_beacon_proposer endpoint at start of each epoch.
+// TODO(dhruv): move this somewhere else once more use-cases like this becomes clear.
+func setFeeRecipient(eth2Cl eth2wrap.Client, pubkeys []eth2p0.BLSPubKey, feeRecipient string) func(ctx context.Context, slot core.Slot) error {
+	onStartup := false
```
corverroos, line 691: suggest:
```
onStartup := true
...
if !onStartup && !slot.FirstInEpoch() {
  return nil
}
onStartup = false

```

corverroos (approved):

Issue #995, opened 2022-08-18
Log all flags on startup


corverroos:
## Problem to be solved

Charon has a lot of flags, many of which has default values. It can be hard to debug incorrect config or weird errors for values not defined by the users. 

E.g. `.charon/charon-enr-private-key not found` or `error resolving bootnode ENR bootnode.gcp.obol.tech`even thought they never specific anything of that sort.
  
## Proposed solution

Log all flag (config) values at startup at info level.
Either do this a single log (hard to read) or each value on its own line (very verbose).
Do this for `bootnode`, `dkg` and `run`.

PR #1249, opened 2022-10-11
cmd: log all flags on startup

dB2510:
Add logging of all flags on startup to bootnode, dkg and run commands.

category: feature
ticket: #995

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -127,6 +127,23 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	log.Info(ctx, "Charon run config",
+		z.Str("lock-file", conf.LockFile),
+		z.Any("beacon-node-endpoints", conf.BeaconNodeAddrs),
+		z.Str("validator-api-address", conf.ValidatorAPIAddr),
+		z.Str("monitoring-address", conf.MonitoringAddr),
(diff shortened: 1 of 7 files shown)


app/app.go
```diff
@@ -127,6 +127,23 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	log.Info(ctx, "Charon run config",
```
dB2510, line 130: Went with single log because multi line looked very verbose. Also this can go out of screen, in that case we can use ctrl + F or cmd + F to find the value of desired flag.

dB2510: Screenshot 2022-10-11 at 1 04 10 PM

dB2510: done

cmd/cmd.go
```diff
@@ -141,3 +144,12 @@ func titledHelp(cmd *cobra.Command) {
 		titledHelp(child)
 	}
 }
+
+func printFlags(ctx context.Context, flags *pflag.FlagSet, msg string) {
```
corverroos, line 148: add godoc please

corverroos (approved):

dB2510:
This is how it looks like:
Screenshot 2022-10-11 at 8 27 00 PM
Screenshot 2022-10-11 at 8 27 12 PM

Issue #998, opened 2022-08-19
Remove p2p address flag defaults


corverroos:
## Problem to be solved

We currently default the following `charon run/dkg` address flags:
```
"p2p-tcp-address", []string{"127.0.0.1:3610"}
"p2p-udp-address", "127.0.0.1:3630"
```

I feel that this doesn't add any value and that it actually introduces issues/noise in some cases and that defaulting to empty values will actually improve networking and libp2p since unreachable IPs are not present by default.

- Static `localhost:{port}` bindings NEVER work, "it isn't a thing":
  - Charon is never run is isolation, it always runs as part of a cluster.
  - That means that if other nodes are on different machines (or in different containers), then `localhost` bindings doesn't work.
  - When run locally without containers, this will result in clashing ports

- It gives the idea to users that localhost "is a thing" while it is not.
- It results in localhost being added to ENRs created by `charon create enr` which doesn't add value and introduces noise and issues in charon needing to filter this out.
- It isn't clear that users MUST configure these if they want to accept incoming connections.

- Empty defaults are better since:
   - We can skip starting a discv5 udpnode since it doesn't do anything if it cannot listen to incoming connections.
   - libp2p will not be configured with invalid "local addresses" that need to be filtered out. We can then actually start using "local addresses" as valid things.
   - We will not advertise localhost libp2p addresses that are unreachable if discv5 is actually enabled and configured.
   - We can log "--p2p-udp-address not defined, so not starting Discv5 UDP node, so only relying on libp2p relays for discovery" 

## Proposed solution

- Remove the default values for `p2p-tcp-address` and `p2p-udp-address`.
- Print all ENR values in `charon enr -v` since noisy addresses will not be present anymore.
- Do not start a discv5 udpnode if `p2p-udp-address` not defined.
- Do not add `libp2p.ListenAddrs(addrs...),` if `p2p-tcp-address` empty.
- Error on startup if `--p2p-udp-address==""` and `--p2p-bootnode-relay=false` and `--p2p-bootnodes-from-lockfile==false` since then this node is undiscoverable.

PR #1250, opened 2022-10-11
cmd: output all info in `charon enr --verbose`

xenowits:
Output all information pertaining to a decoded ENR for `charon enr --verbose`.

category: feature
ticket: #998

### DIFF
--- a/cmd/enr.go
+++ b/cmd/enr.go
@@ -20,9 +20,11 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net"
 	"strings"
 
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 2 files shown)


xenowits:
Screenshot 2022-10-11 at 12 51 59 PM
Screenshot 2022-10-11 at 12 59 24 PM

cmd/enr.go
```diff
@@ -109,3 +113,42 @@ func pubkeyHex(pubkey ecdsa.PublicKey) string {
 func bindEnrFlags(flags *pflag.FlagSet, verbose *bool) {
 	flags.BoolVar(verbose, "verbose", false, "Prints the expanded form of ENR.")
 }
+
+// enrNetworkingKeys returns a string containing the non-empty networking keys (ips and ports) present in the ENR record.
+func enrNetworkingKeys(r enr.Record) string {
```
corverroos, line 118: does this work?
```
entries :=  []enr.Entry{
		new(enr.IPv4),
		new(enr.IPv6),
		new(enr.TCP),
		new(enr.TCP6),
		new(enr.UDP),
		new(enr.UDP6),
	}
		
	var sb strings.Builder
	for _, entry := range entries {
		if err := r.Load(entry); err == nil {
			_, _ = sb.WriteString(fmt.Sprintf("%s: %v\n", entry.ENRKey(), entry))
		}
	}
	
	return sb.String()
```

corverroos, line 118: please add a test for the output 🙏

corverroos (approved):

cmd/enr.go
xenowits: unfortunately it doesn't work. your approach seems quite concise though

Issue #1190, opened 2022-09-28
Spec Prysm support


OisinKyne:
## Problem to be solved

We want to support all validator client implementations. 

## Proposed solution

Setup time with the prysm team to dig deep into their VC implementation. Understand what parts of their implementation mandate the use of grpc and access to the full ~50MB state object (via _debug methods only in HTTP). Write a design doc for what a `valdiatorgrpc` component would look like in charon. 

### Out of Scope

Implementing the proposal is not part of this ticket. This is just to get a first cut at a design doc after research with the prysm team and review by the Obol and prysm teams.

corverroos:
Closing this as Prysm is going to add support for BeaconAPI

PR #1251, opened 2022-10-11
prysm/grpc: pseudo code adding prysm beacon node validator API

corverroos:
Psuedo code implementing the Prysm beacon node validator API gRPC server.

TL;DR:
 - 🎉 Most endpoints map directly to eth2client endpoints with only type conversions. We can therefore simply proxy these to our validatorAPI.
 - 🕰 A few "streaming" queries require scheduler slot events, 🎉 but otherwise queries validatorAPI 
 - 🎉 `ValidatorStatus` endpoint can also proxy to validatorAPI, 🙈 but some deposit related data will be empty (that is either not used or only used for logging)
 - 🎉 `Proposer` related endpoints can proxy to valiatgorAPI, 🔧 but requires access to config flag`--builder-api` since blinded and normal blocks use the same endpoint.
 - 🆙 `AttestationAggregation` endpoints require similar upgrades to our proposed `BeaconAPI` changes. But also require timing changes.
 - 😔 `SyncCommitteeMessages` BeaconBlockRoot is also implicit similar to BeaconAPI,  🎉 but otherwise can proxied to valiatgorAPI.
 -  🆙 `SyncContribution` endpoints require similar upgrades to our proposed `BeaconAPI` changes. But also require timing changes.

> Note this will never be merged as it imports prysm.

category: misc
ticket: #1190

### DIFF
--- a/core/validatorapi/grpc/validator.go
+++ b/core/validatorapi/grpc/validator.go
@@ -63,6 +63,7 @@ type BeaconNodeValidatorServer struct {
 	aggSigDB core.AggSigDB
 }
 
+// GetDuties 🎉 can proxy to valiatgorAPI (multiple calls)
 func (s *BeaconNodeValidatorServer) GetDuties(ctx context.Context, req *prysmpb.DutiesRequest) (*prysmpb.DutiesResponse, error) {
 	pubkeys := toGroupPubkeys(req.PublicKeys)
 
@@ -75,6 +76,7 @@ func (s *BeaconNodeValidatorServer) GetDuties(ctx context.Context, req *prysmpb.
 	}, nil
 }
 
+// StreamDuties 🕰 requires scheduler epoch events, 🎉 but can then just query validatorAPI.
 func (s *BeaconNodeValidatorServer) StreamDuties(req *prysmpb.DutiesRequest, server prysmpb.BeaconNodeValidator_StreamDutiesServer) error {
 	ctx := server.Context()
 	first, _ := s.GetDuties(ctx, req)
@@ -89,6 +91,7 @@ func (s *BeaconNodeValidatorServer) StreamDuties(req *prysmpb.DutiesRequest, ser
 	return nil
 }
 
+// DomainData 🎉 can proxy to valiatgorAPI
 func (s *BeaconNodeValidatorServer) DomainData(ctx context.Context, req *prysmpb.DomainRequest) (*prysmpb.DomainResponse, error) {
 	var domainType eth2p0.DomainType
 	copy(domainType[:], req.Domain)
@@ -98,11 +101,13 @@ func (s *BeaconNodeValidatorServer) DomainData(ctx context.Context, req *prysmpb
 	return &prysmpb.DomainResponse{SignatureDomain: resp[:]}, nil
 }
 
+// DomainData 🎉 can proxy to valiatgorAPI (copy simple scheduler logic)
 func (s *BeaconNodeValidatorServer) WaitForChainStart(e *empty.Empty, server prysmpb.BeaconNodeValidator_WaitForChainStartServer) error {
 	scheduler.WaitChainStart(server.Context(), s.eth2Cl, clockwork.NewRealClock())
 	return nil
 }
 
+// WaitForActivation 🕰 requires scheduler epoch events, 🎉 but can then just query validatorAPI.
 func (s *BeaconNodeValidatorServer) WaitForActivation(req *prysmpb.ValidatorActivationRequest, server prysmpb.BeaconNodeValidator_WaitForActivationServer) error {
 	done := make(chan struct{})
 	s.subscribeEpoch(func(epoch eth2p0.Epoch) {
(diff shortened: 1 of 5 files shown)


core/validatorapi/grpc/validator.go
```diff
@@ -0,0 +1,512 @@
...
+
+// ValidatorStatus 🎉 can proxy to valiatgorAPI, 🙈 but will not return some data either not used in prysm or just used for logging.
+func (s *BeaconNodeValidatorServer) ValidatorStatus(ctx context.Context, req *prysmpb.ValidatorStatusRequest) (*prysmpb.ValidatorStatusResponse, error) {
+	resp, _ := s.valsProvider.ValidatorsByPubKey(ctx, "head", toGroupPubkeys([][]byte{req.PublicKey}))
+	return &prysmpb.ValidatorStatusResponse{
+		Status:                    prysmpb.ValidatorStatus(resp[0].Status),
+		Eth1DepositBlockNumber:    0, // TODO(corver): We do not have access to this, but it seems unused.
+		DepositInclusionSlot:      0, // TODO(corver): We do not have access to this, but it seems unused.
+		ActivationEpoch:           pb.Epoch(resp[0].Validator.ActivationEpoch),
+		PositionInActivationQueue: 0, // TODO(corver): We do not have access to this, it is only used for logging.
```
dB2510, line 145: this data can be queried from an execution node which is what a beacon node does, since we are connected to beacon node and not connected to execution node, we cannot get this data.

core/validatorapi/grpc/validator.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// GetBeaconBlock 🎉 can proxy to valiatgorAPI, 🔧 but requires access to config flag BuilderAPI.
+func (s *BeaconNodeValidatorServer) GetBeaconBlock(ctx context.Context, req *prysmpb.BlockRequest) (*prysmpb.GenericBeaconBlock, error) {
+	var randao eth2p0.BLSSignature
+	copy(randao[:], req.RandaoReveal)
+
+	if s.builderAPI {
+		block, _ := s.eth2Cl.BlindedBeaconBlockProposal(ctx, eth2p0.Slot(req.Slot), randao, req.Graffiti)
+		return convertBlindedBlock(block), nil
```
dB2510, line 162: this is interesting, why prysm would have generic beacon block object? why not deal with separate types?

core/validatorapi/grpc/validator.go
```diff
@@ -0,0 +1,512 @@
...
+// SubmitAggregateSelectionProof can use integrated but probably 🆙 requires upgrade along with SubscribeCommitteeSubnets.
+func (s *BeaconNodeValidatorServer) SubmitAggregateSelectionProof(ctx context.Context, req *prysmpb.AggregateSelectionRequest) (*prysmpb.AggregateSelectionResponse, error) {
+	// TODO(corver): This is an interesting alternative to our v2 API change.
+	//  The VC provides the partial selection proof and the BN returns the complete aggregation and proof
+	//  for the VC to sign. The VC basically doesn't need to be any logic.
+
+	// Assuming we already completed `core.DutyPrepareAggregator` for this slot.
+	aggSig, _ := s.aggSigDB.Await(ctx, core.NewPrepareAggregatorDuty(int64(req.Slot)), toGroupPubkey(req.PublicKey))
+
+	// TODO(corver): We will need to add a different API to dutyDB to query the DutyAggregator data by pubkey instead of attestation hash.
```
dB2510, line 240: another alternative can be to query attestation and then use its data root to query aggregate attestation, but this will require 2 queries to DutyDB.

core/validatorapi/grpc/validator.go
```diff
@@ -0,0 +1,512 @@
...
+		Signature: eth2p0.BLSSignature{},
+	}
+
+	_ = s.eth2Cl.SubmitVoluntaryExit(ctx, signedExit)
+
+	root, _ := signedExit.Message.HashTreeRoot()
+	return &prysmpb.ProposeExitResponse{ExitRoot: root[:]}, nil
+}
+
+// SubscribeCommitteeSubnets 🆙 requires upgrade along with SubscribeCommitteeSubnets.
```
dB2510, line 276: what does this mean?

core/validatorapi/grpc/validator.go
```diff
@@ -0,0 +1,512 @@
...
+	prepareSubmitter   eth2client.ProposalPreparationsSubmitter
+	valsProvider       eth2client.ValidatorsProvider
+	eventsProvider     eth2client.EventsProvider
+	blockProvider      eth2client.SignedBeaconBlockProvider
+
+	eth2Cl     eth2wrap.Client
+	builderAPI bool
+
+	mu      sync.Mutex
+	streams map[int64]func(epoch eth2p0.Epoch)
```
xenowits, line 61: // map[slot][epoch] ??

core/validatorapi/grpc/validator.go
```diff
@@ -0,0 +1,512 @@
...
+func (s *BeaconNodeValidatorServer) GetDuties(ctx context.Context, req *prysmpb.DutiesRequest) (*prysmpb.DutiesResponse, error) {
+	pubkeys := toGroupPubkeys(req.PublicKeys)
+
+	current := s.getDuties(ctx, eth2p0.Epoch(req.Epoch), pubkeys)
+	next := s.getDuties(ctx, eth2p0.Epoch(req.Epoch+1), pubkeys)
+
+	return &prysmpb.DutiesResponse{
+		CurrentEpochDuties: current,
+		NextEpochDuties:    next,
+	}, nil
```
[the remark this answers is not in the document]
xenowits: i think that field (`Duties`) is deprecated

core/validatorapi/grpc/validator.go
corverroos: they have separate types, they just use a generic endpoint and wrapper

core/validatorapi/grpc/validator.go
corverroos: sure

core/validatorapi/grpc/validator.go
corverroos: these are callbacks that take an epoch as param

core/validatorapi/grpc/validator.go
corverroos: typo, should be "similar to beaconAPI upgrade proposals"

corverroos:
Closing since this is pseudo code

PR #1258, opened 2022-10-12
cmd: remove deprecated `--beacon-node-endpoint` flag

xenowits:
Removes deprecated `--beacon-node-endpoint` flag.

category: misc
ticket: none

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -27,7 +27,6 @@ import (
 	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/app/featureset"
 	"github.com/obolnetwork/charon/app/log"
-	"github.com/obolnetwork/charon/app/z"
 	"github.com/obolnetwork/charon/p2p"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -77,16 +74,8 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
...
-		if beaconNodeAddr != "" {
-			log.Warn(ctx, "Deprecated flag 'beacon-node-endpoint' used, please use new flag 'beacon-node-endpoints'", nil)
-			config.BeaconNodeAddrs = []string{beaconNodeAddr}
-		} else if beaconNodeAddr != "" && len(config.BeaconNodeAddrs) > 0 {
-			log.Warn(ctx, "Deprecated flag 'beacon-node-endpoint' ignored since new flag 'beacon-node-endpoints' takes precedence",
-				nil, z.Str("beacon-node-endpoint", beaconNodeAddr), z.Any("beacon-node-endpoints", config.BeaconNodeAddrs))
-		} else if len(config.BeaconNodeAddrs) == 0 && !config.SimnetBMock {
+		if len(config.BeaconNodeAddrs) == 0 && !config.SimnetBMock {
 			return errors.New("either flag 'beacon-node-endpoints' or flag 'simnet-beacon-mock=true' must be specified")
-			// TODO(corver): Use MarkFlagsRequiredTogether once beacon-node-endpoint is removed.
```
corverroos, line 89: why not do this?

corverroos (approved):

cmd/run.go
xenowits: `cmd.MarkFlagsRequiredTogether("beacon-node-endpoints", "simnet-beacon-mock")` will only error if both the flags are not provided together. However, i found out that `BeaconNodeAddrs` is evaluated to `[]` not nil, so the function doesn't panic since it understands that the flag WAS provided

xenowits: Also, explicitly returning the error is "more" clean as we can specify an error message

Issue #1257, opened 2022-10-12
Refactor prioritise protocol


corverroos:
## Problem to be solved

Prioritise needs to support arbitrary priorities, see [first use-case design doc](https://docs.google.com/document/d/1oZm3FbvVXNOvrSroL9slzkB0HrI0hBaCnq914tXKXs0/edit#).

We should also remove the metadata supported by PeerInfo protocol since it isn't required to be part of prioritise anymore.

## Proposed solution

Refactor protobuf types.

### Out of Scope

There are no existing use-cases of prioritise to main backwards compatibility, so can break it.

PR #1271, opened 2022-10-12
core/priority: refactor priority protocol

corverroos:
Refactors the priority protocol:
- Removes metadata from proto messages since not required anymore
- Makes all types generic `anypb.Any` to support arbitrary use-cases.
- Refactor static msgProvider to provide when triggered.

category: refactor
ticket: #1257


core/priority/calculate.go
```diff
@@ -118,29 +166,41 @@ func validateMsgs(msgs []*pbv1.PriorityMsg) error {
...
+		} else if !proto.Equal(inst, instance) {
+			return errors.New("mismatching instances")
 		}
 		if dedupPeers(msg.PeerId) {
 			return errors.New("duplicate peer")
 		}
-		dedupTopics := newDeduper[string]() // Peers may not provide duplicate topics.
+		dedupTopics := newDeduper[[32]byte]() // Peers may not provide duplicate topics.
 		for _, topic := range msg.Topics {
-			if dedupTopics(msg.PeerId) {
```
corverroos, line 135: bug!

core/priority/calculate.go
```diff
@@ -37,28 +41,40 @@ func calculateResult(msgs []*pbv1.PriorityMsg, minRequired int) (*pbv1.PriorityR
...
 	// Calculate cluster wide resulting priorities by topic.
 	var topicResults []*pbv1.PriorityTopicResult
-	for topic, prioritySet := range prioritySetsByTopic {
+	for _, proposals := range proposalsByTopic {
 		// Calculate overall score for all priorities in the topic
 		// which effectively orders by count then by overall priority.
 		var (
-			scores        = make(map[string]int)
-			allPriorities []string
+			scores        = make(map[[32]byte]int)
```
xenowits, line 62: can we give `[32]byte` a type? will make it more readable

core/priority/prioritiser.go
```diff
@@ -33,13 +34,29 @@ import (
...
+// Instance identifies an instance of the priority protocol.
+type Instance proto.Message
+
+// Topic groups priorities in an instance.
+type Topic proto.Message
+
+// Priority is one of many grouped by a Topic being prioritised in an Instance.
+type Priority proto.Message
+
+// instanceKey is the hash of an Instance proto.Message and is used to index instanceData. See hashProto.
```
xenowits, line 46:
```suggestion
// instanceKey is the hash of an Instance and is used to index instanceData. See hashProto.
```

corverroos: think the original is fine

core/priority/calculate.go
corverroos: yeah not sure, I actually removed the `instanceKey` type, reverting to just `[32]byte`. I think defining types is mostly useful for the scope of the type is very large. In this case, it is pretty small.

Issue #1194, opened 2022-09-28
Log charon version/githash in `charon dkg`


xenowits:
## Problem to be solved

Log charon `version` and `githash` in `charon dkg` and `charon create dkg` commands.

## Proposed solution
Add the following code to `dkg/dkg.go` and `cmd/createdkg.go`.

```go
	log.Info(ctx, "Charon DKG starting",
		z.Str("version", version.Version),
		z.Str("git_commit_hash", gitHash),
		z.Str("git_commit_time", gitTimestamp),
	)
```

### Out of Scope

None.

PR #1276, opened 2022-10-12
version: log charon version info on startup

xenowits:
Adds a common function to log charon version information on startup.

category: feature
ticket: #1194

### DIFF
--- a/app/version/version.go
+++ b/app/version/version.go
@@ -15,7 +15,13 @@
 
 package version
 
-import "runtime/debug"
+import (
+	"context"
+	"runtime/debug"
+
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
 
 // Version is the release version of the codebase.
 // Usually overridden by tag names when building binaries.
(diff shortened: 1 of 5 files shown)


dkg/dkg.go
```diff
@@ -68,6 +69,8 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	version.LogCharonInfo(ctx, "Charon DKG starting")
```
xenowits, line 72: i couldn't up with a better name, so suggestions welcome

app/version/version.go
```diff
@@ -40,3 +46,13 @@ func GitCommit() (hash string, timestamp string) {
 
 	return hash, timestamp
 }
+
+// LogCharonInfo logs charon version information along-with the provided message.
+func LogCharonInfo(ctx context.Context, msg string) {
```
xenowits, line 51: i couldn't up with a better name, so suggestions welcome

corverroos: `version.LogInfo`

corverroos: "Charon" is superfluous, since everything in this codebase is charon...

corverroos (approved):

Issue #1272, opened 2022-10-12
Log format not applied to all logs


corverroos:
## Problem to be solved

When configuring charon with a non-default logging format, e,g. `--log-format=logfmt`, some initial logs are in the default format.

This is because we only init the logging format in `app.Run`, while we log stuff before that in the `cmd` package.

## Proposed solution

Init logging as soon as log flags are parsed.

PR #1277, opened 2022-10-13
cmd: add init logger in printFlags

dB2510:
Inits logging as soon as log flags are parsed in printFlags function.

category: bug
ticket: #1272

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -146,9 +146,19 @@ func titledHelp(cmd *cobra.Command) {
 }
 
 // printFlags INFO logs all the given flags in alphabetical order.
-func printFlags(ctx context.Context, flags *pflag.FlagSet) error {
+func printFlags(ctx context.Context, flags *pflag.FlagSet) {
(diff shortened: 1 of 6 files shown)


cmd/cmd.go
```diff
@@ -146,13 +146,28 @@ func titledHelp(cmd *cobra.Command) {
...
 
 // printFlags INFO logs all the given flags in alphabetical order.
-func printFlags(ctx context.Context, flags *pflag.FlagSet) {
+func printFlags(ctx context.Context, flags *pflag.FlagSet) error {
 	ctx = log.WithTopic(ctx, "cmd")
 
+	logLevel := flags.Lookup("log-level")
+	logFmt := flags.Lookup("log-format")
+
+	if logLevel != nil && logFmt != nil {
```
corverroos, line 155: best not do two things at once, call this manually before `printFlags`

corverroos (approved):

cmd/cmd.go
dB2510: sure

PR #1287, opened 2022-10-14
app/log: add pretty field to structured loggers

corverroos:
Adds a `pretty` to structured loggers to improve signal to noise for loki grafana logs.
Also pad console log messages for better signal to noise.

category: refactor
ticket: none


corverroos:
image

app/log/config.go
```diff
@@ -72,101 +80,175 @@ func InitLogger(config Config) error {
 		return err
 	}
 
+	writer, _, _ := zap.Open("stderr")
```
xenowits, line 83: why ignoring this error?

xenowits (approved):

app/log/config.go
corverroos: copy/paste, thanks fixed

PR #1288, opened 2022-10-14
p2p: zero connection type metrics

corverroos:
Zero connection type metrics so combing them in prometheus queries is simpler.

Update simnet dash.

category: misc
ticket: none


corverroos:
image

xenowits (approved):

PR #1291, opened 2022-10-14
p2p: fix connection metrics

corverroos:
Exclude relays from `p2p_peer_connection_types` since relays are not peers and this resulted in relays appearing on the peer dashboard. 

Had to jump through some nasty hoops :(

category: refactor
ticket: none

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -212,31 +212,43 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 // setupP2P returns a started libp2p tcp node and a shutdown function.
 func setupP2P(ctx context.Context, key *ecdsa.PrivateKey, p2pConf p2p.Config, peers []p2p.Peer, lockHashHex string) (host.Host, func(), error) {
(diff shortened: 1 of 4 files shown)


dkg/dkg.go
```diff
@@ -212,32 +212,44 @@ func Run(ctx context.Context, conf Config) (err error) {
...
 
 	udpNode, err := p2p.NewUDPNode(ctx, p2pConf, localEnode, key, bootnodes)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "")
+		return nil, nil, err
 	}
 
 	relays := p2p.NewRelays(p2pConf, bootnodes)
 
-	tcpNode, err := p2p.NewTCPNode(ctx, p2pConf, key, p2p.NewOpenGater())
```
corverroos, line 236: DKG should not use an open connection gater, fix it.

dB2510 (approved):

dkg/dkg.go
xenowits: true

Issue #1261, opened 2022-10-12
Add integration test for DutySyncMessage


dB2510:
## Problem to be solved

We need an integration test on simnet to ensure end to end flow for DutySyncMessage.

## Proposed solution

- Add support for DutySyncMessage to validatormock.
- Add an integration test to app/simnet_test.go for DutySyncMessage using simnet beaconmock and validatormock.

PR #1293, opened 2022-10-15
testutil/validatormock: integrate DutySyncMessage

dB2510:
Integrates DutySyncMessage by adding support to validatormock and adds a simnet_test.

category: feature
ticket: #1261

### DIFF
--- a/testutil/beaconmock/options.go
+++ b/testutil/beaconmock/options.go
@@ -372,10 +372,6 @@ func WithNoAttesterDuties() Option {
 func WithSyncCommitteeDuties() Option {
 	return func(mock *Mock) {
 		mock.SyncCommitteeDutiesFunc = func(ctx context.Context, epoch eth2p0.Epoch, indices []eth2p0.ValidatorIndex) ([]*eth2v1.SyncCommitteeDuty, error) {
-			if epoch%3 != 0 {
-				return nil, nil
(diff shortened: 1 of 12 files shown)


testutil/beaconmock/server.go
```diff
@@ -51,7 +52,7 @@ type HTTPMock interface {
 	eth2client.SlotDurationProvider
 	eth2client.SlotsPerEpochProvider
 	eth2client.SpecProvider
-	eth2client.SyncCommitteeDutiesProvider
+	eth2client.SyncCommitteeSubscriptionsSubmitter
```
corverroos, line 55: no other submitters here... probably not required

testutil/beaconmock/options.go
```diff
@@ -367,6 +367,37 @@ func WithNoAttesterDuties() Option {
 	}
 }
 
+func WithSyncCommitteeDuties() Option {
```
corverroos, line 370: add a comment what this does. Suggest just turning this duty on always if called?

corverroos (approved):

testutil/beaconmock/options.go
dB2510: for simnet?
like adding it in app.go at the time of instantiating beaconmock

PR #1296, opened 2022-10-16
core: fix beacon committee subscription marshalling

corverroos:
Fixes bug in beacon committee subscription json marshalling that didn't include committee length.

category: bug
ticket: none

### DIFF
--- a/core/parsigdb/memory_internal_test.go
+++ b/core/parsigdb/memory_internal_test.go
@@ -30,10 +30,9 @@ import (
 
 func TestCalculateOutput(t *testing.T) {
 	tests := []struct {
-		name     string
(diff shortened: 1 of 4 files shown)


core/parsigdb/memory_internal_test.go
```diff
@@ -24,36 +24,39 @@ import (
...
 
 func TestCalculateOutput(t *testing.T) {
 	tests := []struct {
-		name   string
-		roots  []int
-		output []int
+		name     string
+		input    []int
+		output   []int
+		provider int
```
dB2510, line 36: this field isn't used i think

dB2510 (approved):

PR #1299, opened 2022-10-17
app/peerinfo: reset peer info metrics

corverroos:
Reset previous peerinfo metrics when peer githash or version changes.
Also do basic sanity check on values (since peers can send arbitrary values). 

category: bug
ticket: none


corverroos:
Problem is that previous metrics where never cleared, e.g.
image

dB2510 (approved):

Issue #999, opened 2022-08-19
Distinguish between "not readyz reasons"


corverroos:
## Problem to be solved

Currently, the `app_monitoring_readyz` only distinguishes between ready (1) and not ready (0). 

It is however very helpful to users to know whether their beacon node is down vs peers are down as different actions are required.

## Proposed solution

Distinguish between different values for `app_monitoring_readyz`:
 - 0: unset
 - 1: ready!
 - 2: beacon node down
 - 3: insufficient peers

### Out of Scope

We might need to add some kind of "beacon node score" metric which indicates how good your beacon node is, since that is the main thing affecting performance. We can do something with bn_error_count and bn_latency and create a score out of 100?

PR #1302, opened 2022-10-17
app: instrument not ready reasons

corverroos:
Instruments not ready reasons for improved UX. 
Also fix a "concurrent map write" panic in peerinfo.

category: feature 
ticket: #999

### DIFF
--- a/app/metrics.go
+++ b/app/metrics.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	// readyzReady indicates the readyz is returning 200s and the node is operational.
+	// readyzReady indicates that readyz returns 200s and the node is operational.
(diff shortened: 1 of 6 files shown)


app/metrics.go
```diff
@@ -23,6 +23,18 @@ import (
 	"github.com/obolnetwork/charon/eth2util"
 )
 
+const (
+	// readyzReady indicates the readyz is returning 200s and the node is operational.
```
xenowits, line 27:
```suggestion
	// readyzReady indicates that readyz is returns 200s and the node is operational.
```
simple vs continuous tenses

xenowits (approved):
LGTM🚀

PR #1305, opened 2022-10-18
cmd: support insecure keys with create cluster

corverroos:
Adds support for insecure keystoreV4 files to `charon create cluster` to enable testing of large validator sets in `compose` as this speeds up encryption and decryption from 1s per file to a few milliseconds.

category: feature 
ticket: none

### DIFF
--- a/eth2util/keystore/keystore.go
+++ b/eth2util/keystore/keystore.go
@@ -37,7 +37,9 @@ import (
 	"github.com/obolnetwork/charon/tbls/tblsconv"
 )
 
-const insecurePower = 4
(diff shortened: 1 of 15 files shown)


eth2util/keystore/keystore.go
```diff
@@ -36,16 +37,37 @@ import (
 	"github.com/obolnetwork/charon/tbls/tblsconv"
 )
 
+const insecurePower = 4
```
xenowits, line 40: what is this variable for?

xenowits (approved):

eth2util/keystore/keystore.go
corverroos: added comment

PR #1309, opened 2022-10-18
core: refactor duty prepare aggregator

corverroos:
Refactors duty prepare aggregator to align with new design.

category: refactor
ticket: none


core/validatorapi/validatorapi.go
```diff
@@ -698,21 +685,6 @@ func (c Component) SubmitAggregateAttestations(ctx context.Context, aggregateAnd
...
-
-		sub, ok := s.(core.SignedBeaconCommitteeSubscription)
-		if !ok {
-			return errors.New("slot signature not found")
-		}
-
-		// Override selection proof with slot signature from subscription.
-		agg.Message.SelectionProof = sub.SlotSignature
-
 		// Verify Signature.
```
corverroos, line 688: this wasn't required actually.

testutil/beaconmock/options.go
```diff
@@ -242,35 +243,31 @@ func WithDeterministicAttesterDuties(factor int) Option {
 					continue
 				}
 
+				// Assign aggregation duties: even validator indexes in even epochs, odd validator indexes in odd epochs.
+				const notAgg = uint64(math.MaxInt16) // No attesters in such a large committee will ever be aggregator.
+				const isAgg = uint64(1)              // All attesters in such a small committee is aggregator.
+
+				commLength := notAgg
```
corverroos, line 250: using committeelength in attester duty to control who is aggregator

core/bcast/bcast.go
```diff
@@ -146,39 +144,8 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, pubkey core.
...
-		}}
-
-		err = b.eth2Cl.SubmitBeaconCommitteeSubscriptions(ctx, subs)
-		if err == nil {
-			log.Info(ctx, "Successfully submitted beacon committee subscription to beacon node",
-				z.Any("delay", b.delayFunc(duty.Slot)))
-		}
-
-		return err
+		// Beacon committee selections are only applicable to DVT, not broadcasted to beacon chain
```
dB2510, line 147: beacon node needs to support this endpoint in future? Or is it just for DVT?

core/signeddata.go
```diff
@@ -39,7 +38,7 @@ var (
 	_ SignedData = VersionedSignedBlindedBeaconBlock{}
 	_ SignedData = VersionedSignedValidatorRegistration{}
 	_ SignedData = SignedRandao{}
-	_ SignedData = SignedBeaconCommitteeSubscription{}
+	_ SignedData = BeaconCommitteeSelection{}
```
dB2510, line 41: suggest keeping this as `SignedBeaconCommitteeSelection` to be consistent with other Signed types, we can also have `BeaconCommitteeSelection` as an unsigned type to be used in fetcher or wherever required. This way it will be more explicit which is unsigned and which is signed even internally they are same types.

dB2510 (approved):

xenowits (approved):

core/bcast/bcast.go
corverroos: Beacon node could support it if they want by just returning the request. We could also support upstream DVTs if we want, but for now we do not.

core/signeddata.go
corverroos: Not sure about this, as there isn't such a thing as an unsigned selection.

PR #1314, opened 2022-10-19
testutil/compose: handle signals better

corverroos:
Fix issues in `compose` by handling signals better in both compose and in main app.
Also tune and rename simnet dash to "charon overview dash" to handle the datasources beter.

category: refactor
ticket: none


xenowits (approved):
Looks good!

Issue #1265, opened 2022-10-12
Add support for DutySyncContribution to scheduler


dB2510:
## Problem to be solved

We need add support for `DutySyncContribution` to scheduler. It is scheduled by scheduler ⅔ into each slot for all validators with sync committee duties for that slot.

## Proposed solution

- Implement DutyDefinition for `eth2v1.SyncCommitteeDuty`.
- Add `DutySyncContribution` to scheduler/offset.go with ⅔ as offset.
- Resolve sync committee duties by calling https://ethereum.github.io/beacon-APIs/#/Validator/getSyncCommitteeDuties. This adds DutyDefinition for DutySyncContribution.

PR #1336, opened 2022-10-21
core/scheduler: adds support for sync committee duties

xenowits:
Adds support for sync committee duties to the `scheduler` component.

category: feature
ticket: https://github.com/ObolNetwork/charon/issues/1265

### DIFF
--- a/core/dutydefinition.go
+++ b/core/dutydefinition.go
@@ -76,10 +76,13 @@ func (d ProposerDefinition) MarshalJSON() ([]byte, error) {
 	return d.ProposerDuty.MarshalJSON()
 }
 
-func NewSyncCommitteDefinition(duty *eth2v1.SyncCommitteeDuty) DutyDefinition {
+// NewSyncCommitteeDefinition is a convenience function that returns a new SyncCommitteeDefinition.
+func NewSyncCommitteeDefinition(duty *eth2v1.SyncCommitteeDuty) DutyDefinition {
 	return SyncCommitteeDefinition{SyncCommitteeDuty: *duty}
 }
 
+// SyncCommitteeDefinition defines a sync committee duty. It implements DutyDefinition.
+// Note the slight rename from Duty to Definition to avoid overloading the term Duty.
 type SyncCommitteeDefinition struct {
 	eth2v1.SyncCommitteeDuty
 }
@@ -91,7 +94,7 @@ func (s SyncCommitteeDefinition) Clone() (DutyDefinition, error) {
 		return nil, errors.Wrap(err, "clone sync committee definition")
 	}
 
-	return NewSyncCommitteDefinition(duty), nil
+	return NewSyncCommitteeDefinition(duty), nil
 }
 
 func (s SyncCommitteeDefinition) MarshalJSON() ([]byte, error) {
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 	"time"
 
+	v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/jonboulle/clockwork"
 	"github.com/stretchr/testify/require"
@@ -265,6 +266,11 @@ func (s *Scheduler) resolveDuties(ctx context.Context, slot core.Slot) error {
 		return err
 	}
 
+	err = s.resolveSyncCommDuties(ctx, slot, vals)
+	if err != nil {
+		return err
+	}
+
 	s.setResolvedEpoch(uint64(slot.Epoch()))
 
 	return nil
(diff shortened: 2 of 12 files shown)


core/dutydefinition.go
```diff
@@ -76,22 +76,24 @@ func (d ProposerDefinition) MarshalJSON() ([]byte, error) {
...
-	return SyncCommitteeDefinition{SyncCommitteeDuty: *duty}
-}
-
+// SyncCommitteeDefinition defines a sync committee duty. It implements DutyDefinition.
+// Note the slight rename from Duty to Definition to avoid overloading the term Duty.
 type SyncCommitteeDefinition struct {
 	eth2v1.SyncCommitteeDuty
 }
 
+func NewSyncCommitteeDefinition(duty *eth2v1.SyncCommitteeDuty) DutyDefinition {
```
corverroos, line 85: types should be next to the methods. Constructors should be before the type.

core/scheduler/scheduler.go
```diff
@@ -169,13 +169,39 @@ func (s *Scheduler) GetDutyDefinition(ctx context.Context, duty core.Duty) (core
 
 // scheduleSlot resolves upcoming duties and triggers resolved duties for the slot.
 func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
-	if s.getResolvedEpoch() != uint64(slot.Epoch()) {
-		err := s.resolveDuties(ctx, slot)
+	vals, err := resolveActiveValidators(ctx, s.eth2Cl, s.pubkeys, slot.Slot)
```
corverroos, line 172: rather revert this refactor, keep it simple, stick to the current pattern.

core/scheduler/scheduler.go
```diff
@@ -213,11 +239,13 @@ func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
 		}()
 	}
 
+	// Resolve sync committee duties for the next epoch if the current slot is second last in the current epoch.
+	if slot.LastNInEpoch(2) {
```
corverroos, line 243: Rather not introduce this complexity, keep to existing logic and patterns. We can improve the logic in a separate PR to do lookaheads, for v1, just do current slot/epoch.

core/scheduler/scheduler.go
```diff
@@ -374,6 +390,50 @@ func (s *Scheduler) resolveProDuties(ctx context.Context, slot core.Slot, vals v
...
+		if slot.FirstInEpoch() {
+			startSlot = slot.Previous()
+		}
+
+		for sl := startSlot; sl.Slot <= endSlot.Slot; sl = sl.Next() {
+			// Schedule sync committee contribution aggregation.
+			duty := core.NewSyncContributionDuty(sl.Slot)
+			s.setDutyDefinition(duty, pubkey, core.NewSyncCommitteeDefinition(syncCommDuty))
+
+			log.Info(ctx, "Resolved sync committee duty",
```
corverroos, line 425: this will span the logs, rather just log once per epoch

core/scheduler/scheduler.go
```diff
@@ -374,6 +380,66 @@ func (s *Scheduler) resolveProDuties(ctx context.Context, slot core.Slot, vals v
...
+		subscriptions = append(subscriptions, &v1.SyncCommitteeSubscription{
+			ValidatorIndex:       vIdx,
+			SyncCommitteeIndices: syncCommDuty.ValidatorSyncCommitteeIndices,
+			UntilEpoch:           eth2p0.Epoch(slot.NextEpoch()), // Renew subscription again next epoch.
+		})
+	}
+
+	if len(subscriptions) > 0 {
+		// Subscribe to sync committee subnets.
+		err = s.eth2Cl.SubmitSyncCommitteeSubscriptions(ctx, subscriptions)
```
xenowits, line 434: `fetcher` needs this since it queries BN to get sync comm contribution (`/eth/v1/validator/sync_committee_contribution`)

core/scheduler/scheduler.go
```diff
@@ -374,6 +380,66 @@ func (s *Scheduler) resolveProDuties(ctx context.Context, slot core.Slot, vals v
...
+		)
+
+		subscriptions = append(subscriptions, &v1.SyncCommitteeSubscription{
+			ValidatorIndex:       vIdx,
+			SyncCommitteeIndices: syncCommDuty.ValidatorSyncCommitteeIndices,
+			UntilEpoch:           eth2p0.Epoch(slot.NextEpoch()), // Renew subscription again next epoch.
+		})
+	}
+
+	if len(subscriptions) > 0 {
```
corverroos, line 432: scheduler schedules, scheduler doesn't subscribe to subnets not does it do any POST-like calls. This isn't in the design doc I think. Please remove this.

core/types.go
```diff
@@ -436,11 +436,39 @@ func (s Slot) Next() Slot {
...
+	}
+}
+
 // Epoch returns the epoch of the slot.
 func (s Slot) Epoch() int64 {
 	return s.Slot / s.SlotsPerEpoch
 }
 
+// NextEpoch returns the next epoch of the slot.
+func (s Slot) NextEpoch() int64 {
```
corverroos, line 455: rather remove this, uses can inline this logic if required.

core/types.go
```diff
@@ -436,11 +436,39 @@ func (s Slot) Next() Slot {
...
+	return s.Epoch() + 1
+}
+
+// LastSlotInEpoch returns the last slot in the epoch.
+func (s Slot) LastSlotInEpoch() Slot {
+	lastSlot := s.SlotsPerEpoch*s.NextEpoch() - 1
```
corverroos, line 461: this is a bug, you subtract 1 from an epoch, then assign it to a slot variable. Please remember to add tests for logic you add to avoid bugs.

corverroos: Rather remove this. The existing API supports it already:
```
epoch := slot.Epoch()
for s := slot; s.Epoch() == epoch; s = s.Next() {
  ...
}
```

testutil/beaconmock/options.go
```diff
@@ -268,6 +268,39 @@ func WithDeterministicAttesterDuties(factor int) Option {
 	}
 }
 
+// WithDeterministicSyncCommDuties configures the mock to provide deterministic sync committee duties based on provided arguments and config.
+// Note that it depends on ValidatorsFunc being populated, e.g. via WithValidatorSet.
+func WithDeterministicSyncCommDuties() Option {
```
corverroos, line 273: remove this, we already have `WithSyncCommitteeDuties`, can remove it to `WithDeterministicSyncCommDuties`

core/types.go
xenowits: > this is a bug, you subtract 1 from an epoch, then assign it to a slot variable. Please remember to add tests for logic you add to avoid bugs.

nope, i don't think it is a bug. we don't subtract 1 from an epoch, we do: `first slot in the next epoch - 1`. Although, it seems we can already iterate using the method you suggested

testutil/beaconmock/options.go
xenowits: oh, right, i see it now. thanks!

core/scheduler/scheduler.go
xenowits: sure yeah, got it

corverroos (approved):

PR #1339, opened 2022-10-22
core: add label to deadliner

corverroos:
Aid in debugging when warning occurs.
```
WARN app-start  Deadliner output channel full
```

category: misc
ticket: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -358,8 +358,8 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 		return err
 	}
 
-	deadlinerFunc := func() core.Deadliner {
-		return core.NewDeadliner(ctx, deadlineFunc)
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -413,7 +413,7 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 		return err
 	}
 
-	cons, startCons, err := newConsensus(conf, lock, tcpNode, p2pKey, sender, nodeIdx, deadlinerFunc())
+	cons, startCons, err := newConsensus(conf, lock, tcpNode, p2pKey, sender, nodeIdx, deadlinerFunc("consensus"))
```
xenowits, line 416: can we have these label strings as variables?

xenowits (approved):

app/app.go
corverroos: not required I think

Issue #1342, opened 2022-10-24
Track inconsistent partial signature data


corverroos:
## Problem to be solved

We need to track the number of duties that contain inconsistent partial signed data. This is expected for SyncCommitteeMessage, but may also occur for other duty types.

## Proposed solution

- Track partial signatures in tracker.
- Distinguish between insufficient and inconsistent partial signatures when duty failed in parsigdb.
- Log and instrument when multiple partial signed datas detected.

PR #1343, opened 2022-10-24
core/tracker: detect inconsistent partial signed data

corverroos:
Track inconsistent partial signature data in tracker.

- Distinguish between insufficient and inconsistent partial signatures when duty failed in parsigdb.
- Log and instrument when multiple partial signed datas detected.

category: feature
ticket: #1342


core/tracker/tracker.go
```diff
@@ -559,105 +636,131 @@ func (t *Tracker) ConsensusEvent(ctx context.Context, duty core.Duty, data core.
...
+	} else {
+		log.Warn(ctx, "Inconsistent partial signed data", nil,
+			z.Any("data", parsigMsgs))
+	}
+}
+
+// expectInconsistentParSigs returns true if the duty type is expected to sometimes
+// produce inconsistent partial signed data.
+func expectInconsistentParSigs(duty core.DutyType) bool {
+	return duty == core.DutySyncMessage || duty == core.DutySyncContribution
```
xenowits, line 765: shouldn't this be
```suggestion
	return duty == core.DutySyncMessage || duty == core.DutyPrepareSyncContribution
```
??

core/tracker/tracker.go
```diff
@@ -287,15 +327,49 @@ func analyseDutyFailed(duty core.Duty, allEvents map[core.Duty][]event) (bool, c
 	return true, comp, msg
 }
 
+// analyseParSigs returns a mapping partial signed data messages by peers (share index).
```
xenowits, line 330:
```suggestion
// analyseParSigs returns a mapping of partial signed data messages by peers (share index).
```

core/tracker/tracker.go
```diff
@@ -287,15 +327,49 @@ func analyseDutyFailed(duty core.Duty, allEvents map[core.Duty][]event) (bool, c
...
+func analyseParSigs(duty core.Duty, allEvents map[core.Duty][]event) parsigsByMsg {
```
xenowits, line 331: also curious as to why this function needs all the events and not just the events for the duty?
will this work?
```suggestion
func analyseParSigs(duty core.Duty, events []event) parsigsByMsg {
```

xenowits (approved):

core/tracker/tracker.go
corverroos: no, `SyncCommitteeSelection` should always be consistent, whlie `ContributionAndProof` contains the beaconblockroot that VC pick themselves.

core/tracker/tracker.go
corverroos: indeed, good catch

core/tracker/tracker.go
xenowits: true

Issue #1347, opened 2022-10-25
Aggregate attestation fails with inconsistent data


corverroos:
# 🐞 Bug Report

### Description

Seeing sporadic attestation aggregation failures in parsigdb due to `insufficient partial signatures received, minimum required threshold not reached`. But seeing the logs, threshold partial signatures were submitted. This is unexpected.

<img width="1430" alt="image" src="https://user-images.githubusercontent.com/29249923/197701823-ccab406f-9907-49d5-b5ca-ff162180f0a3.png">

This PR should highlight the issue further https://github.com/ObolNetwork/charon/pull/1343.

corverroos:
The inconsistency is in selection proof.....
![image](https://user-images.githubusercontent.com/29249923/198023982-8854d149-b786-4bf7-bf5b-261135e5e76f.png)

Issue #920, opened 2022-08-05
Extract signing root function for all eth2 types


corverroos:
## Problem to be solved

We sign eth2 structs in validatormock and in many tests. Signing requires calculating the domain wrapped signature root of the struct. We currently do this manually in many places. It is time to DRY.

## Proposed solution

Create a function  in `eth2util/signing` that calculates signing roots for all unsigned eth2 types, following the example of `VerifySignedData`.

`func UnsignedRoot(ctx context.Context, eth2Cl Eth2Provider, unsigned interface{}) ([32]byte, error) {`

Refactor all manual instances to use this code.

PR #1349, opened 2022-10-25
core: add message root to signed data interface

corverroos:
Add explicit unsigned data message root function to signed data as a better way to ensure consistent unsigned data in parsigdb. Relying on json serialisation to check for consistent signed data is brittle.

category: refactor
ticket: #1347


dB2510 (approved):
I was also thinking about the same in line with https://github.com/ObolNetwork/charon/issues/920. This will simplify verification may also help in refactoring verification logic into single function.

xenowits (approved):

Issue #473, opened 2022-05-03
Lodestar integration issues (placeholder)


dB2510:
## Problem to be solved

At the time of integration with lodestar some issues have been identified and informed to chainsafe team. 
* Lodestar doesn't support custom network configs with ACTIVE_PRESET as gnosis or something else. They have a hard requirement of having ACTIVE_PRESET as mainnet.
* When integrating with lodestar on prater with infura's beacon node it has been found that lodestar asserts the config obtained from infura with its local hardcoded one. Refer the image below: 
![Screenshot 2022-05-01 at 11 13 45 AM](https://user-images.githubusercontent.com/37813203/166429809-79a464b2-3f41-428d-8059-79c25c775cf9.png)
* The fields found that caused error cannot be modified with undefined since string is expected in config.

## Proposed solution
* Remove the config assertion on lodestar: https://github.com/ChainSafe/lodestar/blob/a3cb49517057037e963b303c5a1b0d81b621c292/packages/validator/src/validator.ts#L144

dB2510:
tracking this ticket here: https://github.com/ObolNetwork/charon-distributed-validator-cluster/issues/23 
closing this as it is repetitive

PR #1353, opened 2022-10-26
core/dutydb: add attestation data for zero committee index

dB2510:
Adds attestation data for zero committee index and the given slot.

category: misc
ticket: #473

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,14 @@ and the beacon node via the official [Eth Beacon Node REST API](https://ethereum
 Charon supports any upstream beacon node that serves the Beacon API.
 Charon aims to support any downstream standalone validator client that consumes the Beacon API.
 
-| Client                                             | Beacon Node | Validator Client | Notes                                           |
-| -------------------------------------------------- | :---------: | :--------------: | ----------------------------------------------- |
-| [Teku](https://github.com/ConsenSys/teku)          |     ✅      |        ✅        | Fully supported                                 |
-| [Lighthouse](https://github.com/sigp/lighthouse)   |     ✅      |        ✅        | Fully supported                                 |
-| [Lodestar](https://github.com/ChainSafe/lodestar)  |     ✅      |       \*️⃣        | Validator client requires Loadestar beacon node |
-| [Vouch](https://github.com/attestantio/vouch)      |     \*️⃣     |        ✅        | Only validator client provided                  |
-| [Prysm](https://github.com/prysmaticlabs/prysm)    |     ✅      |        🛑        | Validator client requires gRPC API              |
-| [Nimbus](https://github.com/status-im/nimbus-eth2) |     ✅      |       \*️⃣        | No standalone validator client provided         |
+| Client                                             | Beacon Node | Validator Client | Notes                                   |
+| -------------------------------------------------- | :---------: |:----------------:|-----------------------------------------|
+| [Teku](https://github.com/ConsenSys/teku)          |     ✅      |        ✅         | Fully supported                         |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -65,7 +65,7 @@ The following table outlines which clients have produced which duties on a publi
 
 | Duty \ Client                        |                      Teku                       |                    Lighthouse                     | Lodestar | Nimbus | Vouch | Prysm |
 | ------------------------------------ | :---------------------------------------------: | :-----------------------------------------------: | :------: | :----: | :---: | :---: |
-| _Attestation_                        |                       ✅                        |                        ✅                         |    🚧    |   🚧   |  🚧   |  🚧   |
+| _Attestation_                        |                       ✅                        |                        ✅                         |    ✅    |   🚧   |  🚧   |  🚧   |
 | _Attestation Aggregation_            |                       🚧                        |                        🚧                         |    🚧    |   🚧   |  🚧   |  🚧   |
 | _Block Proposal_                     |                       ✅                        |                        ✅                         |    🚧    |   🚧   |  🚧   |  🚧   |
```
dB2510, line 70: haven't updated block proposal part since it isn't tested with lodestar.
if it works on simnet can we mark it as ✅ ?

core/dutydb/memory.go
```diff
@@ -316,6 +316,22 @@ func (db *MemDB) storeAttestationUnsafe(pubkey core.PubKey, unsignedData core.Un
...
+	// This is done in order to support clients like lodestar who call GET attestation data endpoint with 0 committee_index,
+	// to optimize network calls, since attestation data object does not depends on committee_index.
+	aKeyZeroComm := attKey{
+		Slot:    int64(attData.Data.Slot),
+		CommIdx: 0,
+	}
+
+	if value, ok := db.attDuties[aKeyZeroComm]; ok {
+		if value.String() != attData.Data.String() {
+			return errors.New("clashing attestation data", z.Any("key", aKeyZeroComm))
```
corverroos, line 329: make this a warning rather with a `TODO: return error once we are sure attestation data does not vary with committee index`

corverroos (approved):
please add a test

README.md
corverroos: yeah we should decide whether we are going to add all these clients to the compose tests, we should probably.

dB2510: sure, i'll add lodestar to compose tests alongside lighthouse, teku and vmock in another PR.

corverroos:
actually, I would prefer to follow a more defensive approach.

Instead of duplicating all attestation datas in the set to committee index 0, rather only duplicate the lowest committee index to 0 (if it isn't zero itself).

corverroos (changes requested):
- TL;DR: lodestar is not spec compliant, which probably results in us not supporting them
 - Lodestar's optimisation to use a single attestation data for all committees is a probable blocker ito supporting it.
 - The beacon spec provides NO guarantees that attestation datas for different committees MUST be identical. 
 - In the event that charon stores different attestation datas in the dutydb (for the same slot), then lodestar will basically attest with a SINGLE attestation data while other VCs will attest using the different datas, resulting in aggregation failure. 
 - It might be that that in 99% of the time, all attestation datas are actually the same for different committees, it is however only an accidental implementation phenomenon which might change at anypoint due to changes in implementation or race conditions in the code.
 - We are going to monitor whether we can detect inconsistent attestation data for different committees, which will prove that we cannot support lodestar.
 - If we cannot detect inconsistencies, then we are in a tough spot, since we can provisionally support lodestar with the caveat that attestation might fail at any point.

PR #1354, opened 2022-10-26
core/fetcher: call attestation data once

dB2510:
Optimise fetchAttesterData to call GET attestation data endpoint once.

category: misc
ticket: none


corverroos (changes requested):
mmm, not aligned with this, rather just add a TODO. We need to confirm with Jim or someone that this is a good aproach.

dB2510:
Holding off this change until we confirm from someone about this optimisation.

PR #1358, opened 2022-10-26
core/fetcher: monitor inconsistent attestation data

dB2510:
Adds a counter metric to monitor inconsistent attestation data obtained from beacon node for the same slot.

category: misc
ticket: none

### DIFF
--- a/core/fetcher/fetcher.go
+++ b/core/fetcher/fetcher.go
@@ -149,9 +149,14 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot int64, defSet core
 		resp[pubkey] = attData
 
 		// Store Attestation data root excluding committee index.
-		data := *eth2AttData
-		data.Index = 0
-		root, err := data.HashTreeRoot()
+		clone, err := attData.Clone()
+		if err != nil {
+			return nil, err
+		}
+
+		data := clone.(core.AttestationData)
+		data.Data.Index = 0
+		root, err := data.Data.HashTreeRoot()
 		if err != nil {
 			return nil, err
 		}
--- a/core/fetcher/fetcher.go
+++ b/core/fetcher/fetcher.go
@@ -120,6 +120,7 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot int64, defSet core
 	// We may have multiple validators in the same committee, use the same attestation data in that case.
 	dataByCommIdx := make(map[eth2p0.CommitteeIndex]*eth2p0.AttestationData)
 
+	attDataRoots := make(map[eth2p0.Root]bool)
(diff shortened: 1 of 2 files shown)


core/fetcher/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/obolnetwork/charon/app/promauto"
+)
+
+var inconsistentAttDataCounter = promauto.NewCounter(prometheus.CounterOpts{
+	Namespace: "core",
+	Subsystem: "fetcher",
+	Name:      "inconsistent_att_data_total",
+	Help:      "Total number of times attestation data found different for the same slot excluding committee_index.",
```
corverroos, line 28: `Total number of inconsistent attestation data detected. Note this is expected.`

core/fetcher/fetcher.go
```diff
@@ -146,6 +147,19 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot int64, defSet core
 		}
 
 		resp[pubkey] = attData
+
+		// Store Attestation data root excluding committee index.
+		eth2AttData.Index = 0
```
corverroos, line 152: are you sure we are not modifying `attData` at the same time? There are pointers involved here... Please ensure there is a test that ensures output index != 0.

corverroos (approved):

core/fetcher/fetcher.go
dB2510: We are not modifying attData since attData is an unsigned implementation where we don't have point to attestation data.

dB2510: Also in https://github.com/ObolNetwork/charon/blob/19dd5b72466206885880e7d1f39662610dd20dff/core/fetcher/fetcher_test.go#L83
we assert committee index to be equal to the input provided.

core/fetcher/fetcher.go
```diff
@@ -146,6 +147,20 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot int64, defSet core
 		}
 
 		resp[pubkey] = attData
+
+		// Store Attestation data root excluding committee index.
+		data := *eth2AttData
```
dB2510, line 152: @corverroos using value of `eth2AttData` now rather than modifying pointer.

core/fetcher/fetcher.go
corverroos: thanks

core/fetcher/fetcher.go
corverroos: what about `attData.Clone()`?

corverroos:
Also please add a TODO to remove this after the data has been collected.

PR #1359, opened 2022-10-26
core/validatorapi: verify attestation aggregation selection proof

corverroos:
Verify inner attestation aggregation selection proof, not just the aggregation signature. This should fix the sporadic aggregation failures which I suspect is due to the cluster running both validatormock and teku, and that teku is per-chance submitting aggregation using legacy selection proofs in the same slot. It is also good practice to verify all signatures.

category: bug
ticket: #1343

### DIFF
--- a/core/validatorapi/validatorapi.go
+++ b/core/validatorapi/validatorapi.go
@@ -685,7 +685,19 @@ func (c Component) SubmitAggregateAttestations(ctx context.Context, aggregateAnd
 			return err
 		}
 
-		// Verify Signature.
+		// Verify inner selection proof (outcome of DutyPrepareAggregator).
+		if !c.insecureTest {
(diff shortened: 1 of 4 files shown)


core/validatorapi/validatorapi.go
```diff
@@ -685,7 +685,19 @@ func (c Component) SubmitAggregateAttestations(ctx context.Context, aggregateAnd
 			return err
 		}
 
-		// Verify Signature.
+		// Verify inner selection proof (outcome of DutyPrepareAggregator).
+		if !c.insecureTest {
+			blsPubkey, err := tblsconv.KeyFromETH2(eth2Pubkey) // Use group pubkey, not pubshare.
```
corverroos, line 690: still need to add tests

dB2510:
@corverroos should we test/monitor the aggregations flow by running ONLY validatormock for sometime?

xenowits (approved):

Issue #1360, opened 2022-10-27
Add creator to cluster definition


corverroos:
## Problem to be solved

We want to allow non-participants to create the cluster definition. We also want to make it explicit who created the cluster definition if it is one of the operators (who is the leader). Currently, it is implicitly the first operator.
 
## Proposed solution

Add a `creator` nested top level field to the cluster definition with two fields itself, `address` and `config_signature`.
```
"creator": {
  "address": "0x123",
  "config_signature": "0x567"
}
``` 

The address and config_signature fields are exactly the same as the operator address and config_signature fields. If the creator is also an operator, then the data is duplicated.

These fields will be left empty with `charon create cluster`, same as the operator fields.

This is part of new **draft** version v1.4.

Add the creator config_signature as part of definition signature verification. 

The ssz spec:
```
type Definition struct {
	// UUID is a human-readable random unique identifier. Max 64 chars.
	UUID string `json:"uuid" ssz:"ByteList[64]" config_hash:"0" definition_hash:"0"`

	// Name is a human-readable cosmetic identifier. Max 256 chars.
	Name string `json:"name" ssz:"ByteList[256]" config_hash:"1" definition_hash:"1"`

	// Version is the schema version of this definition. Max 16 chars.
	Version string `json:"version" ssz:"ByteList[16]" config_hash:"2" definition_hash:"2"`

	// Timestamp is the human-readable timestamp of this definition. Max 32 chars.
	// Note that this was added in v1.1.0, so may be empty for older versions.
	Timestamp string `json:"timestamp" ssz:"ByteList[32]" config_hash:"3" definition_hash:"3"`

	// NumValidators is the number of DVs (n*32ETH) to be created in the cluster lock file.
	NumValidators int `json:"num_validators" ssz:"uint64" config_hash:"4" definition_hash:"4"`

	// Threshold required for signature reconstruction. Defaults to safe value for number of nodes/peers.
	Threshold int `json:"threshold" ssz:"uint64" config_hash:"5" definition_hash:"5"`

	// FeeRecipientAddress 20 byte Ethereum address.
	FeeRecipientAddress string `json:"fee_recipient_address,0xhex" ssz:"Bytes20" config_hash:"6" definition_hash:"6"`

	// WithdrawalAddress 20 byte Ethereum address.
	WithdrawalAddress string `json:"withdrawal_address,0xhex" ssz:"Bytes20" config_hash:"7" definition_hash:"7"`

[...]

PR #1362, opened 2022-10-27
cluster: add creator to draft v1.4

corverroos:
Adds cluster creator to new draft definition v1.4.

category: feature
ticket: #1360

### DIFF
--- a/cluster/definition.go
+++ b/cluster/definition.go
@@ -41,6 +41,15 @@ type NodeIdx struct {
 	ShareIdx int
 }
 
+// WithV1x4 provides an option to create a new definition with v1.4 including a creator.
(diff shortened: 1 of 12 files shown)


cluster/definition.go
```diff
@@ -41,6 +41,15 @@ type NodeIdx struct {
 	ShareIdx int
 }
 
+// WithV1x4 provides an option to create a new definition with v1.4 including a creator.
+// TODO(corver): Remove this and add creator to NewDefinition when releasing v1.4.
```
dB2510, line 45: can we have a ticket for this TODO with label as release version, to better track the tickets to be handled before next release?

dB2510 (approved):

cluster/definition.go
corverroos: sure

PR #1367, opened 2022-10-28
app: add network label to all metrics

corverroos:
Adds a network label to metrics so centralised monitoring can easily group and filter and aggregate by network.

Also refactors the "overview" dashboard to filter by cluster name/hash/peer so that it can used "as-is" in an aggregate context.

category: misc
ticket: none


corverroos:
image

xenowits (approved):
Looks good!

PR #1369, opened 2022-10-28
core/scheduler: skip missed slots

corverroos:
Skip missed slots in scheduler to avoid thundering heard due to pause-the-world-events.

category: misc
ticket: none

### DIFF
--- a/core/scheduler/metrics.go
+++ b/core/scheduler/metrics.go
@@ -58,6 +58,13 @@ var (
 		Name:      "validator_balance_gwei",
 		Help:      "Total balance of a validator by public key",
 	}, []string{"pubkey_full", "pubkey"})
+
+	skipCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "core",
+		Subsystem: "scheduler",
+		Name:      "skipped_slots_total",
+		Help:      "The total number times that slots were skipped",
+	})
 )
 
 // instrumentSlot sets the current slot and epoch metrics.
--- a/core/scheduler/metrics.go
+++ b/core/scheduler/metrics.go
@@ -63,7 +63,7 @@ var (
 		Namespace: "core",
 		Subsystem: "scheduler",
 		Name:      "skipped_slots_total",
-		Help:      "The total number times that slots were skipped",
+		Help:      "Total number times slots were skipped",
 	})
 )
 
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -495,36 +495,45 @@ func newSlotTicker(ctx context.Context, eth2Cl eth2wrap.Client, clock clockwork.
 		return nil, err
 	}
 
+	currentSlot := func() core.Slot {
+		chainAge := clock.Since(genesis)
+		slot := int64(chainAge / slotDuration)
+		startTime := genesis.Add(time.Duration(slot) * slotDuration)
+
+		return core.Slot{
+			Slot:          slot,
+			Time:          startTime,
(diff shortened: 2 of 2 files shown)


core/scheduler/metrics.go
```diff
@@ -58,6 +58,13 @@ var (
 		Name:      "validator_balance_gwei",
 		Help:      "Total balance of a validator by public key",
 	}, []string{"pubkey_full", "pubkey"})
+
+	skipCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "core",
+		Subsystem: "scheduler",
+		Name:      "skipped_slots_total",
+		Help:      "The total number times that slots were skipped",
```
xenowits, line 66:
```suggestion
		Help:      "Total number times slots were skipped",
```

core/scheduler/metrics.go
```diff
@@ -58,6 +58,13 @@ var (
 		Name:      "validator_balance_gwei",
 		Help:      "Total balance of a validator by public key",
 	}, []string{"pubkey_full", "pubkey"})
+
+	skipCounter = promauto.NewCounter(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
xenowits: in short, how frequently are slots skipped + by how much are slots skipped. wdyt?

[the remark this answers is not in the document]
corverroos: Just because you "can", doesn't mean you "should". What would be the purpose? what problem do you have that this would be the answer to?  

I added this mostly to ensure the logic and scheduling and timing isn't faulty. So a simple counter is fine. I'll add a TODO to remove this if no weird behaviour is seen.

core/scheduler/scheduler.go
```diff
@@ -495,30 +495,46 @@ func newSlotTicker(ctx context.Context, eth2Cl eth2wrap.Client, clock clockwork.
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-clock.After(slot.Time.Sub(clock.Now())):
 			}
 
-			height++
-			startTime = startTime.Add(slotDuration)
-			delay := startTime.Sub(clock.Now())
+			// Recalculate slot to avoid thundering-heard by skipping slots if missed due
```
xenowits, line 523:
```suggestion
			// Recalculate slot to avoid "thundering herd" problem by skipping slots if missed due
```

core/scheduler/scheduler.go
```diff
@@ -495,30 +495,46 @@ func newSlotTicker(ctx context.Context, eth2Cl eth2wrap.Client, clock clockwork.
...
 		for {
-			resp <- core.Slot{
-				Slot:          height,
-				Time:          startTime,
-				SlotsPerEpoch: int64(slotsPerEpoch),
-				SlotDuration:  slotDuration,
+			select {
+			case <-ctx.Done():
+				return
+			case <-clock.After(slot.Time.Sub(clock.Now())):
```
xenowits, line 520: it'd be good to add tests for `newSlotTicket`

xenowits (approved):

core/scheduler/metrics.go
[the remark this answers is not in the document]
xenowits: yeah, seems like `no_of_slots` doesn't have a solid motivation. Though, my actual thought driving this was to monitor how "big" GC (or any end-of-the-world event) pauses are. But, we have logs for that so no need to add metrics i believe

core/scheduler/scheduler.go
corverroos: we have tests for scheduler

PR #1370, opened 2022-10-28
app: revert invalid forkjoin error

corverroos:
Revert failing on invalid forkversion as it seems to be a common use-case in the wild.

category: bug 
ticket: none


eth2devops (approved):
LGTM

Issue #1326, opened 2022-10-20
eth2wrap returns zero slot duration sporadically on shutdown.


corverroos:
# 🐞 Bug Report

## Description

Charon integration tests sporadically fail with "division by zero". This seems due to a shutdown race inside eth2wrap or go-eth2-client that returns 0 for `SlotDuration` (ie empty `Spec`).

PR #1378, opened 2022-11-02
app/eth2wrap: fix cancelled context race

corverroos:
Fixes a race in `eth2wrap` forkjoin code when context is cancelled before any results are returned. This resulted in `<zero>,nil` being returned which resulted in unexpected behaviour, including "divide by zero".

category: bug
ticket: #1326

### DIFF
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -163,8 +163,9 @@ func provide[O any](ctx context.Context, clients []Client,
 	defer cancel()
 
 	var (
-		nokResp forkjoin.Result[Client, O]
-		zero    O
+		nokResp    forkjoin.Result[Client, O]
+		hasNokResp bool
+		zero       O
 	)
 	for res := range join() {
 		if ctx.Err() != nil {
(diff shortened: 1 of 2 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -163,8 +163,9 @@ func provide[O any](ctx context.Context, clients []Client,
 	defer cancel()
 
 	var (
-		nokResp forkjoin.Result[Client, O]
-		zero    O
+		nokResp    forkjoin.Result[Client, O]
+		hasNokResp bool
+		zero       O
 	)
 	for res := range join() {
```
corverroos, line 170: when the calling context is already/immediately closed, then forkjoin doesn't return any results.

app/eth2wrap/eth2wrap.go
```diff
@@ -173,9 +174,16 @@ func provide[O any](ctx context.Context, clients []Client,
 			return res.Output, nil
 		} else {
 			nokResp = res
+			hasNokResp = true
 		}
 	}
 
+	if ctx.Err() != nil {
+		return zero, ctx.Err()
+	} else if !hasNokResp {
```
corverroos, line 183: this is just a sanity check

dB2510 (approved):

Issue #1380, opened 2022-11-02
Charon clusters can agree on supported versions


OisinKyne:
## Problem to be solved

We introduce a breaking change. We tell people to update. For a period of time, 2 clients will be on the old version and 2 clients will be on the new version, so we will be offline. Step 1 to solve this is to agree on what versions are supported by the cluster.

## Proposed solution

Use prioritise to choose a version that ever speaks. 

### Out of Scope

Enabling/Disabling features based on prioritised versions.

PR #1385, opened 2022-11-03
core/infosync: implement infosync priority protocol use case

corverroos:
Adds the `infosync` component that prioritised cluster supported versions. Also add the `priority.Component` that wraps the prioritiser to hide protobuf types and do signing.

category: feature
ticket: #1380

### DIFF
--- a/core/consensus/component.go
+++ b/core/consensus/component.go
@@ -23,7 +23,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"go.opentelemetry.io/otel/trace"
@@ -50,12 +49,12 @@ func New(tcpNode host.Host, sender *p2p.Sender, peers []p2p.Peer, p2pKey *ecdsa.
 	// Extract peer pubkeys.
 	keys := make(map[int64]*ecdsa.PublicKey)
 	for i, p := range peers {
-		var pk enode.Secp256k1
-		if err := p.ENR.Load(&pk); err != nil {
-			return nil, errors.Wrap(err, "load pubkey")
+		pk, err := p.PublicKey()
+		if err != nil {
+			return nil, err
 		}
-		epk := ecdsa.PublicKey(pk)
-		keys[int64(i)] = &epk
+
+		keys[int64(i)] = pk
 	}
 
 	c := &Component{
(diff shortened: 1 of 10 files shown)


corverroos:
Will add tests in next PR
Will wire in PR after that

core/infosync/infosync.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+	"github.com/obolnetwork/charon/core"
+	"github.com/obolnetwork/charon/core/priority"
+)
+
+// New return a new infosync component.
```
xenowits, line 28:
```suggestion
// New returns a new infosync component.
```

core/priority/component.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return err
+	}
+
+	return c.prioritiser.Prioritise(ctx, msg)
+}
+
+// signMsg returns a copy of the proto message with a populated signature signed by the provided private key.
+func signMsg(msg *pbv1.PriorityMsg, privkey *ecdsa.PrivateKey) (*pbv1.PriorityMsg, error) {
+	clone := proto.Clone(msg).(*pbv1.PriorityMsg)
```
xenowits, line 146: what if msg is nil? `clone.Signature` would panic

core/types.go
```diff
@@ -72,13 +73,14 @@ func (d DutyType) String() string {
 		DutySyncMessage:             "sync_message",
 		DutyPrepareSyncContribution: "prepare_sync_contribution",
 		DutySyncContribution:        "sync_contribution",
+		DutyInfoSync:                "info_sync",
```
xenowits, line 76: what is this duty for?

xenowits (approved):

core/consensus/component.go
```diff
@@ -213,7 +212,7 @@ func (c *Component) Propose(ctx context.Context, duty core.Duty, data core.Unsig
 
 // ProposePriority participants in a consensus instance proposing the provided priority message.
```
dB2510, line 213:
```suggestion
// ProposePriority participants in a consensus instance proposing the provided priority result.
```

dB2510 (approved):

core/priority/component.go
corverroos: nil checks are not required in golang, one assumes valid data is being passed in 99% of cases

core/types.go
corverroos: we prioritise "cluster-agreed-upon-supported-version" every epoch. This is used to decide to upgrade or not.

Issue #1268, opened 2022-10-12
Integrate DutySyncContribution


dB2510:
## Problem to be solved

To test end-to-end flow we need to add integration tests on simnet for `DutySyncContribution`.

## Proposed solution

- Add support for `DutySyncContribution` to validatormock.
- Add an integration test in app/simnet_test with simnet beaconmock and validatormock for `DutySyncContribution`.

PR #1388, opened 2022-11-06
core: integrate `DutySyncContribution`

xenowits:
Integrate `DutySyncContribution` into core workflow. Also, test on simnet.

category: feature 
ticket: #1268

### DIFF
--- a/testutil/validatormock/attest.go
+++ b/testutil/validatormock/attest.go
@@ -377,7 +377,7 @@ func aggregate(ctx context.Context, eth2Cl eth2wrap.Client, signFunc SignFunc, s
 
 		val, ok := vals[selection.ValidatorIndex]
 		if !ok {
-			return false, errors.New("missing validator index")
+			return false, errors.New("missing validator index", z.U64("vidx", uint64(selection.ValidatorIndex)))
 		}
 
 		proofSig, err := signFunc(val.Validator.PublicKey, sigData[:])
--- a/testutil/validatormock/attest.go
+++ b/testutil/validatormock/attest.go
@@ -80,6 +80,7 @@ func (a *SlotAttester) Slot() eth2p0.Slot {
 // - Fetches attester attDuties for the slot (this could be cached at start of epoch).
 // - Prepares aggregation attDuties for slot attesters.
 // It panics if called more than once.
+// TODO(xenowits): Fix off-chance race condition bug when Prepare panics if called for the same slot multiple times.
 func (a *SlotAttester) Prepare(ctx context.Context) error {
 	var err error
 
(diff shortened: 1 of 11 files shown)


testutil/validatormock/attest.go
```diff
@@ -80,6 +80,7 @@ func (a *SlotAttester) Slot() eth2p0.Slot {
 // - Fetches attester attDuties for the slot (this could be cached at start of epoch).
 // - Prepares aggregation attDuties for slot attesters.
 // It panics if called more than once.
+// TODO(xenowits): Fix off-chance race condition bug when Prepare panics if called for the same slot multiple times.
```
corverroos, line 83: this should never happen... it wasn't designed to be called twice

testutil/validatormock/synccomm.go
```diff
@@ -144,19 +146,31 @@ func (s *SyncCommMember) PrepareSlot(ctx context.Context, slot eth2p0.Slot) erro
...
+	if err != nil {
+		return err
+	}
+
+	err = submitSyncMessages(ctx, s.eth2Cl, slot, blockRoot, s.signFunc, s.duties)
+	if err != nil {
+		return err
+	}
+
+	s.blockRoots[slot] = blockRoot
```
corverroos, line 163: you need to add a synchronisation channel so Aggregate waits for this to be populated

testutil/validatormock/attest.go
corverroos: So the comment should be: TODO: figure out why this is sometimes called twice (which causes panics)

testutil/validatormock/synccomm.go
```diff
@@ -316,3 +325,72 @@ func submitSyncMessage(ctx context.Context, eth2Cl eth2wrap.Client, slot eth2p0.
...
+
+	epoch, err := eth2util.EpochFromSlot(ctx, eth2Cl, slot)
+	if err != nil {
+		return false, err
+	}
+
+	var signedContribAndProofs []*altair.SignedContributionAndProof
+	for _, selection := range selections {
+		// Check if the validator is an aggregator.
+		ok, err := eth2exp.IsSyncCommAggregator(ctx, eth2Cl, selection.SelectionProof)
```
corverroos, line 343: rather follow SlotAttester pattern where `IsAttAggregator` is called in the prepare/selection step, so only cache selections that are actually aggregators

corverroos: because you will return true below even when no validators are aggregators from this slot. And `SubmitSyncCommitteeContributions` will probably error if no validators are aggregators for this slot.

corverroos (approved):

testutil/validatormock/attest.go
xenowits: created a ticket here: https://github.com/ObolNetwork/charon/issues/1389 to solve this. I'll fix this in the next PR

testutil/validatormock/synccomm.go
xenowits: yeah, missed that

PR #1393, opened 2022-11-07
app: ignore failed infosyncs

corverroos:
When a priority protocol is triggered and not all peers are ready it will complete with an empty result. Just ignore that for now. This fixes the flapping test.

category: test
ticket: none

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -466,6 +466,10 @@ func (a *priorityAsserter) Callback(t *testing.T, i int) func(ctx context.Contex
 	t.Helper()
 
 	return func(ctx context.Context, duty core.Duty, results []priority.TopicResult) error {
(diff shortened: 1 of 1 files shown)


xenowits (approved):

app/app_test.go
```diff
@@ -466,6 +466,10 @@ func (a *priorityAsserter) Callback(t *testing.T, i int) func(ctx context.Contex
 	t.Helper()
 
 	return func(ctx context.Context, duty core.Duty, results []priority.TopicResult) error {
+		if len(results) == 0 {
+			// Some but not all peers participated, ignore.
+			return nil
+		}
 		if !assert.Len(t, results, 1) {
```
xenowits, line 473: but why didn't this if catch it?

PR #1409, opened 2022-11-09
core/validatorapi: batch validator queries

corverroos:
Another reason why `validators_by_pubkey` has the highest rate of all BN endpoints is that we do not batch multiple validators in the same query. Since caching introduces problems, this should also reduce load with adding complexity.

category: refactor
ticket: none

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -268,17 +268,10 @@ func getValidators(p eth2client.ValidatorsProvider) handlerFunc {
 	return func(ctx context.Context, params map[string]string, query url.Values, body []byte) (interface{}, error) {
 		stateID := params["state_id"]
 
-		var resp []v1Validator
-		for _, id := range getValidatorIDs(query) {
-			val, ok, err := getValidatorByID(ctx, p, stateID, id)
-			if err != nil {
-				return nil, err
-			} else if ok {
-				resp = append(resp, v1Validator(*val))
-			}
-		}
-
-		if len(resp) == 0 {
+		resp, err := getValidatorsByID(ctx, p, stateID, getValidatorIDs(query)...)
+		if err != nil {
+			return nil, err
+		} else if len(resp) == 0 {
 			return nil, apiError{
 				StatusCode: http.StatusNotFound,
 				Message:    "NotFound",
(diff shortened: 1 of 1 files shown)


core/validatorapi/router.go
```diff
@@ -1039,46 +1034,59 @@ func getValidatorIDs(query url.Values) []string {
...
 
-	vIdx, err := strconv.ParseUint(id, 10, 64)
-	if err != nil {
-		return nil, false, errors.Wrap(err, "parse validator index")
+	var vIdxs []eth2p0.ValidatorIndex
+	for _, id := range ids {
+		vIdx, err := strconv.ParseUint(id, 10, 64)
+		if err != nil {
+			return nil, errors.Wrap(err, "parse validator index")
+		}
```
xenowits, line 1080: nit: add newline

core/validatorapi/router.go
```diff
@@ -1039,46 +1034,59 @@ func getValidatorIDs(query url.Values) []string {
...
+			vals = append(vals, v1Validator(*v))
 		}
 
-		temp, err := p.ValidatorsByPubKey(ctx, stateID, []eth2p0.BLSPubKey{eth2Pubkey})
-		if err != nil {
-			return nil, false, err
+		return vals
+	}
+
+	if strings.HasPrefix(ids[0], "0x") {
```
xenowits, line 1052: why do we assume that all `ids` are either pubkeys or validator indexes?

corverroos: because it is

xenowits: For ex, ids: ["1", "0xbef...", "0xaac..", "4"]

corverroos: that is not a valid usecase we intend to support

core/validatorapi/router.go
xenowits: sure

core/validatorapi/router.go
xenowits: i'm thinking in terms of how the endpoint is itself structured - https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidators.

`
Information will be returned for all indices or public key that match known validators. If an index or public key does not match any known validator, no information will be returned but this will not cause an error. There are no guarantees for the returned data in terms of ordering; both the index and public key are returned for each validator, and can be used to confirm for which inputs a response has been returned.
`

but, it's upto us how we want to structure it on our end

corverroos: we cannot handle a mix, since goeth2 doesn't support a mix

xenowits (approved):

Issue #1406, opened 2022-11-09
Replace loki docker plugin with built-in loki spout


corverroos:
## Problem to be solved

We currently rely on 12factor app functionality to get structured logs into loki. We therefore use loki-docker-plugin to collect docker logs and push to loki. 

The loki-docker-plugin has a big problem however, it causes containers to sporadically hang on shutdown. https://github.com/grafana/loki/issues/2361

It is also bad UX since users need to install things on the host.

## Considered solutions

Problem with 12factor apps is that they are meant for the cloud, not for local docker-compose.  Docker-compose is "poor-mans-cloud". For cloud, we shouldn't use local loki client or loki driver, as k8s and fancy cloud stuff already do log forwarding and aggregation (probably using promtail with k8s service discovery). 

So for local docker-compose we have the following options (k8s is not a problem):
 - 1️⃣ : loki-docker-driver
    - Pros: Easy to install and works all OSes. Can get logs of all containers.
    - Cons: Charon cannot do graceful shutdown. User need to manually install the driver.
 - 2️⃣ : promtail container
    - Pros: No manual installation.  Can get logs of all containers.
    - Cons: Need to mount host docker socket or host docker folders. We'll need customisation for Windows. Permissions might be a problem
 - 3️⃣ : built-in loki client (sends logs to stdout AND via http to loki)
    - Pros: Works very similar to jaeger-client. Full control. Everything works. Can group logs by peer index/name. Can disable for k8s.
    - Cons: Only charon logs will be sent to loki, no geth or lighthouse, they will remain with docker. 

Further notes:
 - We need to get away from loki-docker-driver 1️⃣ for docker-compose deployments since it causes containers to sporadically hang forever (on macOS, and on linux)
 - The side-car promtail container solution 2️⃣ is theoretically the best (#12factorapps) but requires additional config and host mounted resources which is complicated by OS variance and permission issues.
- The "opt-in" built-in loki client 3️⃣ acts as a "T-spout" causing normal syserr logging (in pretty console format) AND sends structured logs to a loki  endpoint. This avoids additional containers and allows best of both worlds ito console logs for humands and structured logs for machines. This will be disabled by default. If this sucks or is hard to maintain, we can always revert to option 2️⃣ (no lock-in). 
 - centralised log collection of logs is a completely different discussion, wether that happens via 2️⃣  or 3️⃣  is the same thing
 - we are only talking about how to do local structured logging to loki when discussion 2️⃣  or 3️⃣
[...]

PR #1425, opened 2022-11-10
app/log: integrate optional loki logging spout

corverroos:
Adds an optional loki-client as log destination for local docker-compose deployments.

category: feature 
ticket: #1406


app/log/config.go
```diff
@@ -38,18 +40,30 @@ const (
...
+	Error(string, ...zap.Field)
+}
+
 var (
-	logger = newDefaultLogger()
 	initMu sync.Mutex
+	// logger is the global logger.
+	logger zapLogger = newDefaultLogger()
+	// stopFuncs are the global logger stop functions.
+	stopFuncs []func(context.Context) = nil
```
xenowits, line 56: why explicitly initializing `stopFuncs` to nil? isn't that the default?

dB2510 (approved):

app/log/loki/client.go
```diff
@@ -46,28 +45,22 @@ const (
...
-type Client struct {
-	input     chan string
-	quit      chan struct{}
-	done      chan struct{}
-	service   string
-	endpoint  string
-	batchWait time.Duration
-	batchMax  int
-}
+// logFunc abstract logging, since this is a logger itself.
```
xenowits, line 48:
```suggestion
// logFunc abstracts logging, since this is a logger itself.
```

app/log/loki/client.go
```diff
@@ -116,15 +122,22 @@ func (c *Client) Run() {
 			if batch.Age() < c.batchWait {
 				continue
 			}
+
 			// Do not send if we are backing off
-			if retries > 0 && expbackoff.Backoff(backoffConfig, retries) > 0 {
-				continue
+			if retries > 0 {
+				nextTry := triedAt.Add(expbackoff.Backoff(backoffConfig, retries))
```
xenowits, line 128: what if `triedAt` is still it's zero value and not initialized?

xenowits (approved):

PR #1429, opened 2022-11-11
core/priority: refactor to fix deadlock

corverroos:
Refactor priority main processing goroutine to fix deadlock when own message is received after and before others.

category: bug
ticket: none


core/priority/prioritiser.go
```diff
@@ -392,28 +392,31 @@ func hashProto(msg proto.Message) ([32]byte, error) {
...
-		if !other.Own {
+	// Send own to any awaiting peers
+	for _, recv := range msgs {
+		if recv.Response == nil {
 			continue
 		}
-		recv.Response <- other.Msg
+
+		recv.Response <- own // Response channel has capacity of 1.
+		recv.Response = nil  // Mark channel as responded..
```
dB2510, line 420: why not `close(recv.Response)` to "Mark channel as responded"?

dB2510 (approved):

xenowits (approved):

core/priority/prioritiser.go
corverroos: I refactored the whole thing, should be more obvious now.

PR #1434, opened 2022-11-14
core/infosync: add metrics

corverroos:
Adds metrics to infosync in order to track results. 

Also refactor duty deadline by moving it to prioritise instead of infosync so users of prioritise do not need to worry about that.

category: refactor
ticket: #1421

### DIFF
--- a/core/infosync/metrics.go
+++ b/core/infosync/metrics.go
@@ -27,7 +27,7 @@ const (
 	resultEmpty = "empty"
 	// resultOwn indicates that resulting cluster-wide priorities match this node's own priorities.
 	resultOwn = "own"
-	// resultOther indicates that resulting cluster-wide priorities is different from this node's own priorities.
+	// resultOther indicates that resulting cluster-wide priorities are different from this node's own priorities.
 	resultOther = "other"
 )
 
@@ -35,5 +35,5 @@ var completeTotal = promauto.NewCounterVec(prometheus.CounterOpts{
 	Namespace: "core",
 	Subsystem: "infosync",
 	Name:      "complete_total",
-	Help:      "Total number of infosync instances completed by result; empty, own, other",
+	Help:      "Total number of infosync instances completed by result: empty, own or other",
 }, []string{"result"})
--- a/core/infosync/metrics.go
+++ b/core/infosync/metrics.go
@@ -31,7 +31,7 @@ const (
 	resultOther = "other"
 )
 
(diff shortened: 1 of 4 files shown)


core/infosync/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/obolnetwork/charon/app/promauto"
+)
+
+const (
+	// resultEmpty indicates that insufficient peers exchanged the same priorities so cluster-wide priorities
+	// could not be calculated.
+	resultEmpty = "empty"
+	// resultOwn indicates that resulting cluster-wide priorities match this node's own priorities.
+	resultOwn = "own"
+	// resultOther indicates that resulting cluster-wide priorities is different from this node's own priorities.
```
xenowits, line 30:
```suggestion
	// resultOther indicates that resulting cluster-wide priorities are different from this node's own priorities.
```

core/infosync/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	resultOther = "other"
+)
+
+var completeTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "core",
+	Subsystem: "infosync",
+	Name:      "complete_total",
+	Help:      "Total number of infosync instances completed by result; empty, own, other",
```
xenowits, line 38: nit:
```suggestion
	Help:      "Total number of infosync instances completed by result: empty, own or other",
```

core/infosync/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	// resultEmpty indicates that insufficient peers exchanged the same priorities so cluster-wide priorities
+	// could not be calculated.
+	resultEmpty = "empty"
+	// resultOwn indicates that resulting cluster-wide priorities match this node's own priorities.
+	resultOwn = "own"
+	// resultOther indicates that resulting cluster-wide priorities is different from this node's own priorities.
+	resultOther = "other"
+)
+
+var completeTotal = promauto.NewCounterVec(prometheus.CounterOpts{
```
xenowits, line 34: nit: how about `completedInfosyncsCounter` ?

xenowits (approved):

core/infosync/metrics.go
corverroos: adding infosync to the name would be "stuttering" since we already in the insosync package. I'll switch to counter suffix.

PR #1437, opened 2022-11-14
core/validatorapi: swallow unexpected validator registrations

dB2510:
Swallow unexpected validator registration based provided builderAPI flag.

category: misc
ticket: none

### DIFF
--- a/app/simnet_test.go
+++ b/app/simnet_test.go
@@ -52,7 +52,7 @@ import (
 )
 
 //go:generate go test . -integration -v
-var integration = flag.Bool("integration", true, "Enable docker based integration test")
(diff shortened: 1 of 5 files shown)


app/simnet_test.go
```diff
@@ -52,7 +52,7 @@ import (
 )
 
 //go:generate go test . -integration -v
-var integration = flag.Bool("integration", false, "Enable docker based integration test")
```
corverroos, line 55: false

corverroos (approved):

app/simnet_test.go
dB2510: sorry pushed accidently

PR #1440, opened 2022-11-14
core/parsigdb: avoid aggregating and submitting again

dB2510:
Avoids signature aggregation and broadcast if threshold is already reached.

category: misc
ticket: none

### DIFF
--- a/core/parsigdb/memory.go
+++ b/core/parsigdb/memory.go
@@ -87,7 +87,6 @@ func (db *MemDB) StoreInternal(ctx context.Context, duty core.Duty, signedSet co
 
 // StoreExternal stores an externally received partially signed duty data set.
 func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet core.ParSignedDataSet) error {
-	thresholdReached := make(map[core.PubKey]bool)
 	for pubkey, sig := range signedSet {
 		sigs, ok, err := db.store(key{Duty: duty, PubKey: pubkey}, sig)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


dB2510:
before
Screenshot 2022-11-14 at 9 37 31 PM

xenowits (approved):
LGTM!

core/parsigdb/memory.go
```diff
@@ -87,6 +87,7 @@ func (db *MemDB) StoreInternal(ctx context.Context, duty core.Duty, signedSet co
 
 // StoreExternal stores an externally received partially signed duty data set.
 func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet core.ParSignedDataSet) error {
+	thresholdReached := make(map[core.PubKey]bool)
```
corverroos, line 90: this doesn't solve the problem you have in the logs, please write tests to verify your logic is correct. Note that you have two peers submitting/storing psigs sequentially.

corverroos (changes requested):

core/parsigdb/memory.go
dB2510: that is the problem, first is reaching the threshold which then proceeds with signature aggregation and then othen peer after that exceeds the threshold which also proceeds with signature aggregation and then bcast.

i have reverted my change and added a test to demonstrate the problem.

dB2510: Added test and corrected implementation

core/parsigdb/memory.go
```diff
@@ -165,6 +168,21 @@ func (db *MemDB) store(k key, value core.ParSignedData) ([]core.ParSignedData, b
 	return append([]core.ParSignedData(nil), db.entries[k]...), true, nil
 }
 
+// isThresholdReached checks if threshold is reached or not for given pubkey.
+func (db *MemDB) isThresholdReached(pubkey core.PubKey) bool {
```
corverroos, line 172: this is incorrect and will prevent 99% of duties to succeed

corverroos (changes requested):

core/parsigdb/memory.go
dB2510: updated

corverroos (approved):

PR #1447, opened 2022-11-16
core/scheduler: add label for validator status

xenowits:
Adds label for validator status (e.g., `active`, `pending`, `exited` and so on) to balanceGauge.

category:  misc
ticket: none

### DIFF
--- a/core/scheduler/metrics.go
+++ b/core/scheduler/metrics.go
@@ -56,7 +56,7 @@ var (
 		Namespace: "core",
 		Subsystem: "scheduler",
 		Name:      "validator_balance_gwei",
-		Help:      "Total balance of a validator by public key",
+		Help:      "Total balance of a validator by public key by status",
 	}, []string{"pubkey_full", "pubkey", "status"})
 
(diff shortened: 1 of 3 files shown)


xenowits:
**BEFORE**
Screenshot 2022-11-16 at 1 42 35 PM
**AFTER**
Screenshot 2022-11-16 at 1 32 53 PM

core/scheduler/metrics.go
```diff
@@ -57,7 +57,7 @@ var (
 		Subsystem: "scheduler",
 		Name:      "validator_balance_gwei",
 		Help:      "Total balance of a validator by public key",
-	}, []string{"pubkey_full", "pubkey"})
+	}, []string{"pubkey_full", "pubkey", "status"})
```
xenowits, line 60: note that this increases cardinality since the label "status" can take on atmost 10 values

corverroos (approved):

core/scheduler/metrics.go
corverroos: `Total balance of a validator by public key by status`

Issue #1450, opened 2022-11-16
Sniff QBFT messages


corverroos:
## Problem to be solved

We are seeing strange QBFT logs. We also want to be ready to debug future issues in QBFT. QBFT is deterministic so given all the messages that a node received, we can reproduce any issue or bug deterministically.

We do have logging, but that is not granular enough to really understand or debug the 4*n*round messages during each qbft instance.

## Proposed solution

Implement a QBFT message sniffer that is exposed as a http endpoint, similar to `debug/pprof/heap`.

- Buffer all qbft instance messages
- Provide these messages to the monitorinapi when the instance is complete
- Keep a buffer of 2MB of these sniffed messages in a fifo queue
- serialised these as protobufs to the user when they query `debug/qbft`

PR #1452, opened 2022-11-16
core/consensus: sniff qbft messages

corverroos:
Sniff recent qbft consensus instance messages and serve them as gzip protos from 2MB fifo buffer from `/debug/qbft`.

category: feature
ticket: #1450


core/corepb/v1/consensus.proto
```diff
@@ -38,3 +39,18 @@ message ConsensusMsg {
...
+
+message SniffedConsensusMsgs {
+  repeated SniffedConsensusMsg msgs = 1;
+}
+
+message SniffedConsensusSets {
+  repeated SniffedConsensusMsgs sets = 1;
+}
+
+
```
xenowits, line 56: remove newline

app/qbftdebug.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const maxQBFTDebugger = 2 * (1 << 20) // 2 MB
+
+// qbftDebugger buffers up to 2MB worth of sniffed qbft messages in a fifo buffer serving them as a gzipped
+// *pbv1.SniffedConsensusSets protobuf on request.
+type qbftDebugger struct {
+	mu        sync.Mutex
+	totalSize int
+	sets      []*pbv1.SniffedConsensusMsgs
```
dB2510, line 38: why not using `pbv1.SniffedConsensusSets` directly here?

dB2510 (approved):
LGTM

app/qbftdebug.go
corverroos: because then one needs to initialise it, slices work without initialisation

core/corepb/v1/consensus.proto
corverroos: done

Issue #992, opened 2022-08-18
Make compose tests more robust


corverroos:
## Problem to be solved

Compose integration tests only fail if the docker containers exit unexpectedly. Failing tests due to alerts thresholds being triggered due to missed duties or high error rate doesn't seem to work (the tests always pass). Also, debugging the output of compose tests is very hard due to the large amount of output that breaks the github UI.
 
## Proposed solution

Break each compose test into its own github action step/job/task?
Or write each test output to a file and make that downloadable from github action results/artifacts.
Figure out why alerts are not being triggered
Figure out why prometheus cannot be polled sporadically.

dB2510:
The purpose of this ticket has been served. We have 2 tickets now related to compose tests:
- https://github.com/ObolNetwork/charon/issues/860
- https://github.com/ObolNetwork/charon/issues/1471

Things that were fixed as part of this ticket:
- Upload docker-compose logs for each test as github action artefact.
- Fix prometheus alerts by separating docker compose build step from docker-compose up.
- Found out issues related to `very_large` smoke test that it wasn't starting in the first place since it takes time in lock step and while building and starting containers which crosses default timeout of 15s.
- Created #1471 for `very_large`.
- Retention time of action artefacts is 3 days to avoid storage pricing issues.

PR #1466, opened 2022-11-21
testutil/compose: handle docker-compose build seperately

dB2510:
Handle `docker-compose build` command separately before starting alerts collector. This step takes longer at first and takes almost no time to build again. This might be the reason why `default_alpha` test in smoke tests seldom completes. Also increased alert timeout for `very_large` test.

category: test
ticket: #992

### DIFF
--- a/testutil/compose/alert.go
+++ b/testutil/compose/alert.go
@@ -41,11 +41,11 @@ func startAlertCollector(ctx context.Context, dir string) chan string {
 		)
 		defer close(resp)
 
+		const iterSleep = time.Second * 2
+
 		// Time required to wait for prometheus container to start.
 		time.Sleep(time.Second * 10)
-		for ctx.Err() == nil {
-			time.Sleep(time.Second * 2)
-
+		for ; ctx.Err() == nil; time.Sleep(iterSleep) { // Sleep for iterSleep before next iteration.
 			cmd := exec.CommandContext(ctx, "docker-compose", "exec", "-T", "curl", "curl", "-s", "http://prometheus:9090/api/v1/rules?type=alert")
 			cmd.Dir = dir
 			out, err := cmd.CombinedOutput()
--- a/testutil/compose/alert.go
+++ b/testutil/compose/alert.go
@@ -40,8 +40,11 @@ func startAlertCollector(ctx context.Context, dir string) chan string {
 			dedup   = make(map[string]bool)
 		)
 		defer close(resp)
+
+		// Time required to wait for prometheus container to start.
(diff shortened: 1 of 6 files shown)


testutil/compose/alert.go
```diff
@@ -40,8 +40,11 @@ func startAlertCollector(ctx context.Context, dir string) chan string {
 			dedup   = make(map[string]bool)
 		)
 		defer close(resp)
+
+		// Time required to wait for prometheus container to start.
+		time.Sleep(time.Second * 10)
 		for ctx.Err() == nil {
-			time.Sleep(time.Second * 5)
+			time.Sleep(time.Second * 2)
```
dB2510, line 47: this will wait for 10 seconds first to let prometheus container to start and then poll alerts every 2 seconds

testutil/compose/auto.go
```diff
@@ -125,6 +125,11 @@ func Auto(ctx context.Context, conf AutoConfig) error {
 		defer cancel()
 	}
 
+	// Build docker-compose services before executing docker-compose up.
+	if err = execBuild(ctx, conf.Dir); err != nil {
```
dB2510, line 129: Found out from logs that this steps longer for the first test which is included in alerts timeout

testutil/compose/auto.go
```diff
@@ -149,8 +154,7 @@ func Auto(ctx context.Context, conf AutoConfig) error {
 		}
 	}
 	if !alertSuccess {
-		log.Error(ctx, "Alerts couldn't be polled", nil)
-		return nil // TODO(corver): Fix this and error
+		return errors.New("alerts couldn't be polled, containers offline")
```
corverroos, line 157: containers offline isn't correct. Sure it might be the most common reason, but it isn't the only reason.

testutil/compose/smoke/smoke_test.go
```diff
@@ -59,26 +60,30 @@ func TestSmoke(t *testing.T) {
 				conf.KeyGen = compose.KeyGenCreate
 				conf.FeatureSet = "alpha"
 			},
+			Timeout: time.Second * 45,
```
corverroos, line 63: suggest using a `defaultTimeout` variable if Timeout is empty

testutil/compose/alert.go
corverroos, line 47: rather sleep at the end of the loop (so you check for context closed directly after sleeping instead of doing logic with a possible cancelled context)

corverroos (approved):

Issue #1445, opened 2022-11-15
Periodic clock offset spikes


dB2510:
# 🐞 Bug Report

### Description

We observed there are periodic spikes corresponding to `app_peerinfo_clock_offset_seconds` metric. There seems to be a problem between commits `55d22dd` and `1a0d6bd`.

### Has this worked before in a previous version?

Yes, the previous version in which this bug was not present was: `55d22dd`

## 🔬 Minimal Reproduction

Run the following promQL query:
```
max(round(app_peerinfo_clock_offset_seconds{cluster_name="$cluster_name",cluster_hash="$cluster_hash",cluster_peer="$cluster_peer"},0.1)) by (peer)  
```

## 🔥 Error

Screenshots of 3 hour, 6 hour and 24 hour window on 15 Nov, 2022:
<img width="1307" alt="Screenshot 2022-11-15 at 11 16 38 PM" src="https://user-images.githubusercontent.com/37813203/201990106-fded973c-7e9b-4a30-a3b8-2f39ef6120dd.png">
<img width="1314" alt="Screenshot 2022-11-15 at 11 16 06 PM" src="https://user-images.githubusercontent.com/37813203/201990076-25fdbe3a-826c-4a5f-971a-249652ba0e95.png">
<img width="1318" alt="Screenshot 2022-11-15 at 11 16 20 PM" src="https://user-images.githubusercontent.com/37813203/201990097-f565bd18-16a3-4800-b5a5-b6a5c31fbf4f.png">

thomasheremans:
Might want to add some logging first

corverroos:
This seems to be due to dial delay on relay connections (delay in opening new relay connections) .
 - This only happens on relay connections (which obivously take a longer time to setup/dial than direct connections)
 - It is correlated with 2min relay connection timeout (not every 2min but every n*2min).
 - The clockDiff is positive and equivalent to ping RTT, so it means the delay in the request leg, not the response.
 - To test, we can change the bootnode connection duration limit to 30min and see if the spike frequence also decreases accordingly.
 
 Actions:
  - I can't seem to find a way to instrument dial latency unfortunately since libp2p doesn't allow providing a custom dialer.
  - We should use a mean/avg value for clockDiff, not a single instant value.
  - We should increase mainnet bootnode connection limit to 1h to avoid issues and delays relating to opening relay connections..   
 
 
[...]

PR #1474, opened 2022-11-22
app/peerinfo: fix clock offset calculation

corverroos:
Refactor `p2p.SendReceiveFunc` to support explicit RTT callback via functional options. This solves the issue that external wrapping calculation includes opening connection delay.

category: bug
ticket: #1445

### DIFF
--- a/p2p/sender.go
+++ b/p2p/sender.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"io"
 	"sync"
+	"time"
 
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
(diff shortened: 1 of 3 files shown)


p2p/sender.go
```diff
@@ -147,24 +149,54 @@ func withRelayRetry(fn func() error) error {
 	return err
 }
 
+type sendRecvOpts struct {
```
xenowits, line 152: add godoc

xenowits (approved):

dB2510 (approved):
LGTM

p2p/sender.go
corverroos: I'm not such a big of comment like `sendRecvOpts defines SendReceive options` since the signal to noise isn't great. And this is unexported, so not sure it is worth it.

Issue #1366, opened 2022-10-28
Upgrade to `go1.19.2 `


xenowits:
## Problem to be solved

Running `govulncheck ./...` (https://go.dev/blog/vuln) on the charon codebase reported some vulnerabilities. These vulnerabilities are present in the dependencies used by charon namely `net/http`, `net/url`, `regexp/syntax`, `os/exec` and `syscall` in the go version (`go1.19`) charon uses. Fortunately, these are fixed in the minor revisions of go (`go1.19.1`, `go1.19.2`, `go1.19.3`) by the go team.

## Proposed solution

Bump go version to [go1.19.3](https://go.dev/doc/devel/release#go1.19.minor) in `go.mod` and github workflow files (`.github/workflows/`). Test if everything works.

### Out of Scope

None.

PR #1481, opened 2022-11-24
github: update workflows to go1.19.3

xenowits:
Updates github workflows to `go1.19.3`

category: misc
ticket: #1366

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/setup-python@v2
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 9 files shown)


corverroos (approved):
Need to update go.mod as well I think and `check-go-version` pre-commit

xenowits:
We can't update go version to a patch release in `go.mod`. Ref: https://go.dev/ref/mod#go-mod-file-go

Screenshot 2022-11-24 at 1 46 26 PM

corverroos:
@xenowits we also need to upgrade the Dockerfile so the built binaries actually use the new version: 
```
FROM golang:1.19-alpine AS builder
```

PR #1485, opened 2022-11-25
core/infosync: prioritise p2p  protocols

corverroos:
Prioritise p2p protocols and make cluster-wise results available for querying by slot. This allows smooth upgrading to new protocols. Adds a `mock_alpha` protocol for ad-hoc testing in prod. Bumps prioritise to stable feature.

category: feature
ticket: #1421


core/infosync/infosync.go
```diff
@@ -18,55 +18,135 @@ package infosync
...
+	return c
 }
 
 type Component struct {
 	prioritiser *priority.Component
 	versions    []string
+	protocols   []protocol.ID
+
+	mu      sync.Mutex
+	results []result
```
dB2510, line 83: do we need to trim this results?

dB2510 (approved):

core/infosync/infosync.go
```diff
@@ -18,55 +18,135 @@ package infosync
...
-const topicVersion = "version"
+const (
+	topicVersion  = "version"
+	topicProtocol = "protocol"
+)
 
 // New returns a new infosync component.
-func New(prioritiser *priority.Component, versions []string) *Component {
+func New(prioritiser *priority.Component, versions []string, protocols []protocol.ID) *Component {
+	// Add a mock alpha protocol if alpha features enabled iot to test infosync in prod.
```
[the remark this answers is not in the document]
corverroos: https://www.abbreviations.com/abbreviation/in+order+to

core/infosync/infosync.go
corverroos: don't think so, unless the config changes constantly which it shouldn't

core/infosync/infosync.go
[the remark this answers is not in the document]
dB2510: use full form please 🙏

core/infosync/infosync.go
corverroos: limiting to 100 results

PR #1489, opened 2022-11-29
scheduler: clear previous status when validator status changes

xenowits:
Clears previous status when validator status changes. For ex: when a validator progresses from `pending_queued` to `active_ongoing`, we currently see two values, one for each status. This PR fixes this by removing the older statuses. In this example, it removes status `pending_queued`.

category: refactor
ticket: https://github.com/ObolNetwork/charon-distributed-validator-node/issues/139

### DIFF
--- a/core/scheduler/metrics.go
+++ b/core/scheduler/metrics.go
@@ -23,7 +23,7 @@ import (
 	"github.com/obolnetwork/charon/core"
 )
 
-// metricSubmitter submits validator balance metric.
+// metricSubmitter submits validator balance and status metrics.
 type metricSubmitter func(pubkey core.PubKey, totalBal eth2p0.Gwei, status string)
 
 var (
@@ -59,7 +59,14 @@ var (
 		Namespace: "core",
 		Subsystem: "scheduler",
 		Name:      "validator_balance_gwei",
-		Help:      "Total balance of a validator by public key by status",
+		Help:      "Total balance of a validator by public key",
+	}, []string{"pubkey_full", "pubkey"})
+
+	statusGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "core",
+		Subsystem: "scheduler",
+		Name:      "validator_status",
+		Help:      "Constant gauge with label set to current status of the validator by pubkey",
 	}, []string{"pubkey_full", "pubkey", "status"})
 
 	skipCounter = promauto.NewCounter(prometheus.CounterOpts{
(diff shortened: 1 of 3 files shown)


core/scheduler/metrics.go
```diff
@@ -78,7 +81,16 @@ func instrumentDuty(duty core.Duty, defSet core.DutyDefinitionSet) {
...
-	balanceGauge.WithLabelValues(string(pubkey), pubkey.String(), status).Set(float64(totalBal))
+// newMetricSubmitter returns a function that sets validator balance metric.
+func newMetricSubmitter() func(pubkey core.PubKey, totalBal eth2p0.Gwei, status string) {
+	prevStatus := make(map[core.PubKey]string)
+
+	return func(pubkey core.PubKey, totalBal eth2p0.Gwei, status string) {
+		balanceGauge.WithLabelValues(string(pubkey), pubkey.String(), status).Set(float64(totalBal))
+
+		if prev, ok := prevStatus[pubkey]; ok && prev != status { // Validator status changed
+			balanceGauge.WithLabelValues(string(pubkey), pubkey.String(), prev).Set(0)
```
corverroos, line 92: This will still result in the same problem, just with value==0ETH. 

The real problem is that we use the balance metric for two things. So instead of having two metrics that each do one thing, we have a dual purpose metric and this isn't correct.

The proper solution is to split the metric into two:
 - balance metic
 - constant value gauge with label=status.

corverroos (changes requested):

core/scheduler/metrics.go
xenowits: You're correct! Let me add in your suggestions

core/scheduler/metrics.go
```diff
@@ -78,7 +88,25 @@ func instrumentDuty(duty core.Duty, defSet core.DutyDefinitionSet) {
...
+
+	return func(pubkey core.PubKey, totalBal eth2p0.Gwei, status string) {
+		balanceGauge.WithLabelValues(string(pubkey), pubkey.String()).Set(float64(totalBal))
+		statusGauge.WithLabelValues(string(pubkey), pubkey.String(), status).Set(1)
+
+		if prev, ok := prevStatus[pubkey]; ok && prev != status { // Validator status changed
+			statusGauge.WithLabelValues(string(pubkey), pubkey.String(), prev).Set(0)
+		}
+		if prev, ok := prevBal[pubkey]; ok && prev != totalBal { // Validator balance changed
+			balanceGauge.WithLabelValues(string(pubkey), pubkey.String()).Set(0)
```
corverroos, line 106: think a bit about what this is doing.... it isn't good.

xenowits: yeah sorry, i was fixing this

xenowits: I also need to fix the grafana panels, doing that

core/scheduler/metrics.go
```diff
@@ -56,7 +59,14 @@ var (
...
 		Name:      "validator_balance_gwei",
-		Help:      "Total balance of a validator by public key by status",
+		Help:      "Total balance of a validator by public key",
+	}, []string{"pubkey_full", "pubkey"})
+
+	statusGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "core",
+		Subsystem: "scheduler",
+		Name:      "validator_status",
+		Help:      "Constant gauge with label set to current status of the validator by pubkey",
```
corverroos, line 69: Gauge with validator pubkey and status as labels, value=1 is current status, value=0 is previous.

corverroos: not really "constant"...

xenowits:
Screenshot 2022-11-29 at 10 27 11 PM

corverroos (approved):

Issue #712, opened 2022-06-14
Investigate dropping discv5 for pure libp2p stack


corverroos:
## Problem to be solved

We forsee that cluster networking will be a massive pain point in the future. We are currently assuming all nodes will directly connect to all other nodes in the cluster. This assumes all nodes are publicly accessible and routable to all other nodes. This might however not be trivially achievable in practice.

libp2p has support for UPNP, it also has other features like circuit relay and might even have other handy feature we might want to take advantage of the future.

Unfortunately, these futures all use multiaddrs for routing. We however use discv5 for peer discovery, which uses ENRs for routing and address reslution. ENR are compatible with multiaddrs, but multiaddrs are not compatible with ENRs. So will probably run into incompatability issues between discv5 and advanced libp2p features.

## Proposed solution

Investigate the impact of swapping discv5 out for a native libp2p peer discovery. What are the draw back and what are the advantages?

Create an implementation document

### Out of Scope

No implementation need to be done yet. We just need more motivation for switching to pure libp2p stack.

PR #1496, opened 2022-11-30
p2p: add alpha relay discovery feature

corverroos:
Adds a new alpha feature called `relay_discovery` which disables discv5, relying purely on relay connections to be upgraded to direct connections via libp2p Indentify protocol that resolves peer direct addresses.

Facts:
 - We know which peers we want to connect to (from lock file)
 - We know all the nodes in a cluster will use the same bootnode(s)/relay(s) (either Obol-hosted or cluster-hosted).
 - We know all the nodes always reserve circuit relays on all configured bootnode/relays.

Connection logic
 - Use direct connection if possible
 - But if direct connection not possible
     - either because peer address unknown 
     - or because known addresses do not work
 - The try to connect to peer via bootnode relay
 - Once connected, libp2p's identify protocol will detect the peers direct address
 - Rinse and repeat

This doesn't actually require any code, it is mostly about removing code (discv5) and relying on existing code and built-in libp2p magic.

Pros:
 - No new libraries or complex code
 - Removes existing libraries and code (discv5) 
 - No steep learning curve of KAD-DHT.
 - Can always add KAD-DHT later (once we go for the whitelisted bootnode network approach where it is unknown at which relays to find your peers).

Cons:
 - No actual peer discovery.
 - Requires either a cluster-wide known bootnode which must be a relay to connect to peer
 - Or requires known direct address of peer (via lock ENRs or possible new config flag --peer-addrs).
 - So use-case of using a bootnode for peer discovery and then only direct connections not supported anymore.
 - Ie. --bootnode-relay=true flag is required 

category: feature
ticket: #712

### DIFF
--- a/app/featureset/featureset.go
+++ b/app/featureset/featureset.go
@@ -46,10 +46,10 @@ const (
 	MockAlpha Feature = "mock_alpha"
 
 	// RelayDiscovery enables relay peer discovery and disables discv5:
-	//   - If a direct connection to a peer is not possible then try to connect to it via all the provided bootnodes/relays
+	//   - If a direct connection to a peer is not possible then try to connect to it via all the provided bootnodes/relays.
 	//   - Direct connections are either not possible since no addresses are known or the addresses do not work.
 	//   - When connected via relay, libp2p's identify protocol detects the remote peer's addresses.
-	//   - Those are added to the peer store so will try to use them.
+	//   - Those are added to the peer store so libp2p will try to use them.
 	RelayDiscovery Feature = "relay_discovery"
 )
 
--- a/app/featureset/featureset.go
+++ b/app/featureset/featureset.go
@@ -44,14 +44,22 @@ const (
 	Priority Feature = "priority"
 	// MockAlpha is a mock feature in alpha status for testing.
 	MockAlpha Feature = "mock_alpha"
+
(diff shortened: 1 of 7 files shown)


app/featureset/featureset.go
```diff
@@ -44,14 +44,22 @@ const (
 	Priority Feature = "priority"
 	// MockAlpha is a mock feature in alpha status for testing.
 	MockAlpha Feature = "mock_alpha"
+
+	// RelayDiscovery enables relay peer discovery and disables discv5:
+	//   - If a direct connection to a peer is not possible then try to connect to it via all the provided bootnodes/relays
```
xenowits, line 49:
```suggestion
	//   - If a direct connection to a peer is not possible then try to connect to it via all the provided bootnodes/relays.
```

app/featureset/featureset.go
```diff
@@ -44,14 +44,22 @@ const (
...
+	//   - Direct connections are either not possible since no addresses are known or the addresses do not work.
+	//   - When connected via relay, libp2p's identify protocol detects the remote peer's addresses.
+	//   - Those are added to the peer store so will try to use them.
```
xenowits, line 52:
```suggestion
	//   - Those are added to the peer store so libp2p will try to use them.
```

xenowits (approved):

dB2510 (approved):

Issue #1509, opened 2022-12-05
Add cluster name, hash, peer and network to pushed loki labels


corverroos:
## Problem to be solved

We will be collecting logs in a centralised loki. But we cannot filter logs by cluster name and hash since we only include a single label to loki, `service=charon`.

## Proposed solution

Similar to prometheus labels, add the following labels when pushing logs, `cluster_hash, cluster_name, cluster_peer, cluster_network`.

PR #1511, opened 2022-12-05
log/loki: lazy loading of cluster labels

corverroos:
Adds lazy loaded cluster labels when pushing logs to loki so we can filter and query by node/cluster/network. 

category: feature
ticket: #1509

### DIFF
--- a/app/log/loki/client_test.go
+++ b/app/log/loki/client_test.go
@@ -33,7 +33,11 @@ import (
 )
 
 func TestLoki(t *testing.T) {
-	const label = "test"
+	const (
+		serviceLabel    = "test"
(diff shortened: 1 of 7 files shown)


app/log/loki/client_test.go
```diff
@@ -33,7 +33,11 @@ import (
 )
 
 func TestLoki(t *testing.T) {
```
dB2510, line 35: suggest: add golden json to this test

dB2510 (approved):

app/log/loki/client_test.go
corverroos: little tricky to do this actually

Issue #1540, opened 2022-12-12
Create promrated utility to report validator metrics


LukeHackett12:
## Problem to be solved

Currently we only have our own charon metrics being reported to prometheus. We would like to be able to see external performance metrics from rated in our dashboard also.

## Proposed solution

[Design doc](https://docs.google.com/document/d/18iTnXNKCaYTJhwRxkuaJvmmVgecmT0H6Nl7FOguDFUY/edit#)

Create a utility tool in testutils. This will get all clusters that currently make their metrics available to our prometheus instance, and intermittently scrape the rated data for each pubkey. This data will then be labelled with the appropriate cluster information and be made available for prometheus polling.

PR #1548, opened 2022-12-13
testutil/promrated: add initial basic looping and read in config

LukeHackett12:
Start of https://github.com/ObolNetwork/charon/issues/1540. Per corvers suggestion taking incremental steps instead of a large code dump. This starts by reading the necessary config values, then runs on a loop every 10 mins with some logging. After this I will expose some metrics to prometheus then start some of the more complex logic for fetching clusters and polling rated.

category: misc


testutil/promrated/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+package main
+
+import (
+	"github.com/spf13/viper"
+)
+
+type Config struct {
+	RatedApiEndpoint string
+	Network          string
```
corverroos, line 24: network shouldn't be hardcoded, you should use the network in the prometheus labels

testutil/promrated/config.go
```diff
@@ -0,0 +1,45 @@
...
+	PromAuth         string
+	PromListenPort   string
```
corverroos, line 26: see bootnode.go, rather call this `MonitoringAddr`

testutil/promrated/config.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func initConfig(cfgFile string, cfg *Config) error {
```
corverroos, line 29: suggest using `cmd/initializeConfig`

testutil/promrated/config.go
```diff
@@ -0,0 +1,45 @@
...
+	viper.SetConfigFile(cfgFile)
+
+	viper.AutomaticEnv()
+	err := viper.ReadInConfig()
+	if err != nil {
+		return err
+	}
+
+	cfg.RatedApiEndpoint = viper.GetString("rated.apiEndpoint")
+	cfg.Network = viper.GetString("rated.network")
```
corverroos, line 39: suggest using actual flags, see cmd package for example

testutil/promrated/main.go
```diff
@@ -0,0 +1,52 @@
...
+	"time"
+
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+var cfg Config
+
+func main() {
+	ctx := context.Background()
```
corverroos, line 32: see cmd package, you should catch sigterm and cancel the context

corverroos (changes requested):

testutil/promrated/promrated/main.go
```diff
@@ -0,0 +1,58 @@
...
+func newRootCmd(runFunc func(context.Context, promrated.Config) error) *cobra.Command {
+	var config promrated.Config
+
+	root := &cobra.Command{
+		Use:   "promrated",
+		Short: "Starts a promrated server",
+		Long:  `Starts a promrated server that polls rated and makes metrics available to prometheus`,
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runFunc(cmd.Context(), config)
```
corverroos, line 45: if you want to support ENV vars and config files (in addition to --flags) you'll need to export and call `cmd.initializeConfig`, but adding that later when actually required is best.

corverroos (approved):

PR #1549, opened 2022-12-13
app: increase sniffed buffer even more

corverroos:
Increase to temporary 2GB buffer to ensure we capture the proposer messages. RAM metrics indicate up to 500MB might be possible.

category: test
ticket: none

### DIFF
--- a/app/qbftdebug.go
+++ b/app/qbftdebug.go
@@ -29,7 +29,7 @@ import (
 	pbv1 "github.com/obolnetwork/charon/core/corepb/v1"
 )
 
-const maxQBFTDebugger = 200 * (1 << 20) // 200MB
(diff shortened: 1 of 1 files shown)


app/qbftdebug.go
```diff
@@ -29,7 +29,7 @@ import (
 	pbv1 "github.com/obolnetwork/charon/core/corepb/v1"
 )
 
-const maxQBFTDebugger = 20 * (1 << 20) // 20MB
+const maxQBFTDebugger = 2000 * (1 << 20) // 2GB // TODO(corver): revert to under 50MB.
```
xenowits, line 32: shouldn't this be `2 * (1 << 30)` ?

PR #1551, opened 2022-12-13
dkg/sync: add disconnect log to server

dB2510:
Add peer disconnected log to server.

category: misc
ticket: none

### DIFF
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -85,10 +85,10 @@ func (c *Client) Run(ctx context.Context) error {
 		relayBroke, connBroke, err := c.sendMsgs(ctx, stream)
 		c.clearConnected()
 		if relayBroke {
-			log.Debug(ctx, "Relay connection dropped, reconnecting")
+			log.Warn(ctx, "Relay connection dropped, reconnecting", nil, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
 			continue // Always reconnect on relay circuit recycling.
 		} else if connBroke && c.reconnect {
-			log.Debug(ctx, "Connection dropped, reconnecting")
+			log.Warn(ctx, "Connection dropped, reconnecting", nil, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
 			continue // Only reconnect for connection breaks in reconnect state.
 		} else if err != nil {
 			return err
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -85,10 +85,10 @@ func (c *Client) Run(ctx context.Context) error {
 		relayBroke, connBroke, err := c.sendMsgs(ctx, stream)
 		c.clearConnected()
 		if relayBroke {
-			log.Warn(ctx, "Relay connection dropped, reconnecting", nil, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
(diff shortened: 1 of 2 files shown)


dkg/sync/server.go
```diff
@@ -230,7 +230,9 @@ func (s *Server) Start(ctx context.Context) {
 
 		ctx = log.WithCtx(ctx, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
 		err := s.handleStream(ctx, stream)
-		if err != nil {
+		if isRelayError(err) {
+			log.Warn(ctx, "Peer disconnected", nil)
+		} else if err != nil {
 			log.Warn(ctx, "Error serving sync protocol", err)
 		}
```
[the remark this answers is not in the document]
corverroos: or maybe rather:
```
		if isRelayError(err) { // Relay errors are expected
			log.Debug(ctx, "Error serving sync protocol", nil)
		} else if err != nil {
			log.Warn(ctx, "Error serving sync protocol", err)
		}
```

xenowits (approved):

dkg/sync/client.go
```diff
@@ -85,10 +85,10 @@ func (c *Client) Run(ctx context.Context) error {
 		relayBroke, connBroke, err := c.sendMsgs(ctx, stream)
 		c.clearConnected()
 		if relayBroke {
-			log.Debug(ctx, "Relay connection dropped, reconnecting")
+			log.Warn(ctx, "Relay connection dropped, reconnecting", nil, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
```
corverroos, line 88: so relay connections breaking is expected (it happens every 2min), we mostly ignore relay errors. Please revert this change.

dkg/sync/client.go
```diff
@@ -85,10 +85,10 @@ func (c *Client) Run(ctx context.Context) error {
...
 			continue // Always reconnect on relay circuit recycling.
 		} else if connBroke && c.reconnect {
-			log.Debug(ctx, "Connection dropped, reconnecting")
+			log.Warn(ctx, "Connection dropped, reconnecting", nil, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
```
corverroos, line 91: peer name also added to ctx above. Not sure the stack traces of WARN level is useful, maybe just info level? Suggest: `log.Info(ctx, "Disconnected from peer")`

corverroos (changes requested):

dkg/sync/client.go
dB2510: peer name added in ctx in local peer and not remote peer. Here's the screenshot with current WARN logs:
Screenshot 2022-12-14 at 12 19 10 PM

dB2510: New logs:

corverroos: when is the `WARN: Error serving sync protocol` logged?
when is the `ERRO: Fatal error...` logged?

corverroos: The aim is to make it look nice:
 - Peer disconnecting isn't such a big problem, a simple "peer disconnected" info log is fine.
 - Relay connections are recycled every 2min, that shouldn't result in any logs (only simple debug) 
 - Warnings and Errors should only happen when things are crashing

corverroos (approved):

PR #1553, opened 2022-12-13
testutil/promrated: add health listener and monitoring endpoint

LukeHackett12:
Start serving prom metrics at the specified address along with a health endpoint to use for a liveness check.

category: misc

### DIFF
--- a/testutil/promrated/prom/prometheus.go
+++ b/testutil/promrated/prom/prometheus.go
@@ -13,32 +13,29 @@
 // You should have received a copy of the GNU General Public License along with
 // this program.  If not, see <http://www.gnu.org/licenses/>.
 
-// Command promrated grabs rated stats for all monitored charon clusters
-
-package promrated
+package prom
 
 import (
 	"context"
-	"time"
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/obolnetwork/charon/app/log"
-	"github.com/obolnetwork/charon/app/z"
-	"github.com/obolnetwork/charon/testutil/promrated/prom"
 )
 
-type Config struct {
-	RatedAPIEndpoint string
-	PromAuth         string
-	MonitoringAddr   string
-}
+func ListenAndServe(ctx context.Context, addr string) {
+	// Healthz handler used in Kubernetes set-ups to automatically restart
+	// the container in case something goes off.
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
 
-func Run(ctx context.Context, config Config) error {
-	go prom.ListenAndServe(ctx, config.MonitoringAddr)
+	// Prometheus handler to expose metrics to prometheus.
+	http.Handle("/metrics", promhttp.Handler())
 
-	for {
-		log.Info(ctx, "Promrated looping.", z.Str("endpoint", config.RatedAPIEndpoint))
-
-		sleepFor := time.Minute * time.Duration(10)
-		time.Sleep(sleepFor)
+	err := http.ListenAndServe(addr, nil)
+	if err != nil {
+		log.Error(ctx, "Failed to serve Prom Metrics", err)
 	}
 }
--- a/testutil/promrated/prometheus.go
+++ b/testutil/promrated/prometheus.go
@@ -24,8 +24,8 @@ import (
 	"github.com/obolnetwork/charon/app/errors"
 )
 
-// listenAndServe creates a liveness endpoint and serves metrics to prometheus.
-func listenAndServe(addr string) error {
+// serveMonitoring creates a liveness endpoint and serves metrics to prometheus.
+func serveMonitoring(addr string) error {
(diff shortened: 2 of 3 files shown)


testutil/promrated/prom/prometheus.go
```diff
@@ -0,0 +1,41 @@
...
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package prom
```
corverroos, line 16: new package isn't required, less packages are better in general

testutil/promrated/prom/prometheus.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/obolnetwork/charon/app/log"
+)
+
+func ListenAndServe(ctx context.Context, addr string) {
+	// Healthz handler used in Kubernetes set-ups to automatically restart
+	// the container in case something goes off.
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
corverroos, line 30: `healthz` is a deprecated pattern, see https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health. Follow pattern of `livez` and `readyz` rather, same as charon monitoringapi.

testutil/promrated/prom/prometheus.go
```diff
@@ -0,0 +1,41 @@
...
+		w.WriteHeader(http.StatusOK)
+	})
+
+	// Prometheus handler to expose metrics to prometheus.
+	http.Handle("/metrics", promhttp.Handler())
+
+	err := http.ListenAndServe(addr, nil)
```
corverroos, line 37: this should crash the program if it errors (e.g. if ports fails to bind). Suggest keeping it simple, see how we do this in bootnode

testutil/promrated/prom/prometheus.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/obolnetwork/charon/app/log"
+)
+
+func ListenAndServe(ctx context.Context, addr string) {
```
corverroos, line 27: always add godoc, especially if the function does something weird, like in this case, it blocks forever

corverroos (changes requested):

testutil/promrated/prom/prometheus.go
```diff
@@ -0,0 +1,41 @@
...
+	// Healthz handler used in Kubernetes set-ups to automatically restart
+	// the container in case something goes off.
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+
+	// Prometheus handler to expose metrics to prometheus.
+	http.Handle("/metrics", promhttp.Handler())
```
corverroos, line 35: also, best not use globals, create explicit http server instance.

testutil/promrated/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/obolnetwork/charon/app/log"
+)
+
+// ListenAndServe creates a liveness endpoint and serves metrics to prometheus.
+func ListenAndServe(ctx context.Context, addr string) {
```
corverroos, line 29: unexport everything by default, only export if really required and the API is justtified.

testutil/promrated/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+
+		// Copied from net/http/pprof/pprof.go
+		server := http.Server{Addr: addr, Handler: mux, ReadHeaderTimeout: time.Second}
+		serverErr <- server.ListenAndServe()
+	}()
+
+	for {
+		select {
+		case err := <-serverErr:
+			panic(err)
```
corverroos, line 50: panics are to be avoided. Suggest returning the http server from this function and then having your main goroutine in your main function do a single "run everything and wait for shutdown".

testutil/promrated/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/obolnetwork/charon/app/log"
+)
+
+// ListenAndServe creates a liveness endpoint and serves metrics to prometheus.
+func ListenAndServe(ctx context.Context, addr string) {
+	serverErr := make(chan error, 2)
```
corverroos, line 30: the buffer 2 was used since there was two servers, in your case you only have 1

corverroos (approved):

Issue #1555, opened 2022-12-14
Remove deprecated consensus wire protocol fields


corverroos:
## Problem to be solved

The value fields for qbft consensus wire messages were refactored to any protos in #1214, this was released in v0.11. The previous values has not been removed yet (for backwards compatibility), but this is causing double the amount network bandwith which is a problem for synthetic blocks. And it is time to compete the upgrade and drop support for old versions.

Note this will remove support for v0.10 and before, so v0.13 will only be compatible with v0.11 and v0.12
 
## Proposed solution

Remove support for deprecated fields.

PR #1556, opened 2022-12-14
core/consensus: fix wire compatibility

corverroos:
Reverts "reserved" fields since they are required to calculate correct hash when verifying signatures. Instead just stop using fields. 

category: bug
ticket: #1555


core/corepb/v1/consensus.proto
```diff
@@ -13,14 +13,28 @@ message QBFTMsg {
   core.corepb.v1.Duty            duty                  = 2;
   int64                          peer_idx              = 3;
   int64                          round                 = 4;
-  reserved                                               5; // Deprecated value
+  core.corepb.v1.UnsignedDataSet value_legacy          = 5 [deprecated = true]; // Prefer value.
```
corverroos, line 16: these fields need to be here, since they are included when calculating signature hashes. Removing them results in server and client calculating different hashes.

dB2510 (approved):

Issue #1490, opened 2022-11-29
Add --definition-file flag to create cluster command that downloads the partial definition and uses it.


HananINouman:
## Problem to be solved

We want to use the solo flow in the launchpad.

## Proposed solution

We have a cluster definition format that supports creator only files, now we need to make this format pullable on the `create cluster` command. 

We should also refactor the outputs of the create cluster command to spit out a folder per node. 
/node0
/node1
/node2
/node3

Each folder should contain a .charon/ folder with the format ingestable by the CDVN repo

thomasheremans:
Need to create separate ticket for the folder restructuring

PR #1558, opened 2022-12-14
cmd: add `--definition-file` flag to create cluster command

xenowits:
Refactors `charon create cluster` command to fetch partial (solo-flow) cluster definition and use it for creating cluster outputs.

category: feature
ticket: #1490


cmd/createcluster_internal_test.go
```diff
@@ -184,3 +213,39 @@ func TestValidNetwork(t *testing.T) {
 	err = validateClusterConfig(ctx, conf)
 	require.Error(t, err, "zero address")
 }
+
+// randomPartialDefinition returns a test cluster definition with version set to v1.4.0.
+func randomPartialDefinition(t *testing.T) cluster.Definition {
```
corverroos, line 218: Rather use existing `cluster.NewForT` and strip fields

cmd/createcluster_internal_test.go
```diff
@@ -108,6 +136,7 @@ func testCreateCluster(t *testing.T, conf clusterConfig) {
 
 	dir, err := os.MkdirTemp("", "")
 	require.NoError(t, err)
+	defer os.RemoveAll(dir)
```
corverroos, line 139: no need to delete temp dirs, you might not have permissions either

cmd/createcluster.go
```diff
@@ -91,6 +94,7 @@ func newCreateClusterCmd(runFunc func(context.Context, io.Writer, clusterConfig)
 func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.StringVar(&config.Name, "name", "", "The cluster name")
 	flags.StringVar(&config.ClusterDir, "cluster-dir", ".charon/cluster", "The target folder to create the cluster in.")
+	flags.StringVar(&config.DefFile, "definition-file", "", "The path to the file or an HTTP URL to fetch cluster definition.")
```
corverroos, line 97: this is not a definition file, it is a partial definition, suggest not supporting local files yet (since they should never actually exist). Also not sure we want to leak the term "partial definition" to users, since it isn't really a well thought out thing. 

Suggest `--config-url: Optional HTTP URL to fetch configuration overriding all other configuration flags.`

cmd/createcluster.go
```diff
@@ -121,6 +125,28 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
...
+
+		if err = def.VerifyPartialSignatures(); err != nil {
+			return err
+		}
+		if err = def.VerifyPartialHashes(); err != nil {
+			return err
+		}
+
+		log.Warn(ctx, "Overriding cluster config with values from fetched definition since `definition-file` flag present", nil)
+		newConf, err := overrideClusterConfig(conf, def)
```
corverroos, line 143: we can't just pick a few things from the partial definition, we need to use the whole thing and only append operators. Since the creator signature needs to remain valid and this depends on the config_hash remaining the same. Please ensure this via tests

corverroos (changes requested):

cmd/createcluster.go
corverroos, line 97: `--definition-file: Optional path to a cluster definition file or an HTTP URL. This overrides all other configuration flags.`

cmd/createcluster.go
```diff
@@ -121,6 +125,28 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return errors.Wrap(err, "mkdir")
 	}
 
+	if conf.DefFile != "" {
+		def, err := loadDefinition(ctx, conf.DefFile)
+		if err != nil {
+			return err
+		}
+		def.DefinitionHash = nil
```
corverroos, line 133: Either clear both operators and definition hash, or clear neither. I would say a more conservative approach is better by expecting valid data that doesn't need to be fixed.

cmd/createcluster_internal_test.go
xenowits: yeah there maybe a permissions issue, good point

cmd/createcluster.go
xenowits: I'm leaning towards clearing both. Reason being that we need to clear `DefinitionHash` since a serialized definition file always contains a `definition-hash`.

xenowits: Cool, let's not clear neither since an invalid "partial" definition will be flagged by the `VerifyPartialHashes()` function

cmd/createcluster.go
xenowits: How can the `config_hash` remain the same? Since partial definition doesn't contain any operator, but here we add operators which changes the config_hash. How do we then update the `creator_config_signature` since the original signature will get outdated? The creator needs to sign the updated definition again, is that the case?

corverroos: `config_hash` only contains operator addresses, which isn't present in this case, so the config_hash MUST remain the same. The `creator_config_signature` MUST remain valid.

cluster/test_cluster.go
```diff
@@ -169,3 +170,34 @@ func getAddrFunc(seed int) func(*testing.T) *net.TCPAddr {
 		}
 	}
 }
+
+// MarshalPartialDefinition returns a json marshaller which marshals partial definitions, i.e., definitions with empty operators list
+// and empty definition-hash.
+func MarshalPartialDefinition(t *testing.T) func(def Definition) ([]byte, error) {
```
corverroos, line 176: this is only used another test package, suggest moving it there. Also why return a function, just marshal it and return the bytes.

xenowits: `definitionJSONv1x4` is private

xenowits:
Closing this in favour of https://github.com/ObolNetwork/charon/pull/1615

PR #1562, opened 2022-12-14
testutils/promrated: add prom query to collect pubkey to cluster information

LukeHackett12:
Adding in prometheus query to get all the necessary cluster information and associated private keys.

category: misc

### DIFF
--- a/testutil/promrated/prometheus.go
+++ b/testutil/promrated/prometheus.go
@@ -31,7 +31,7 @@ import (
 )
 
 const (
-	promQuery = "group by (cluster_name, cluster_hash, cluster_peer, pubkey_full) (core_scheduler_validator_balance_gwei)"
+	promQuery = "group by (cluster_name, cluster_hash, cluster_network, pubkey_full) (core_scheduler_validator_balance_gwei)"
 )
 
 type validator struct {
@@ -39,23 +39,6 @@ type validator struct {
 	ClusterName    string `json:"cluster_name"`
 	ClusterHash    string `json:"cluster_hash"`
 	ClusterNetwork string `json:"cluster_network"`
-	ClusterPeer    string `json:"cluster_peer"`
-}
-
-type promMetric struct {
-	validator
-}
-
-type promResult struct {
-	Metric promMetric `json:"metric"`
-}
-
-type promData struct {
-	Result []promResult `json:"result"`
-}
-
-type promResponse struct {
-	Data promData `json:"data"`
 }
 
 // serveMonitoring creates a liveness endpoint and serves metrics to prometheus.
@@ -70,7 +53,11 @@ func serveMonitoring(addr string) error {
 		promhttp.Handler(),
 	)
 
-	server := http.Server{Addr: addr, Handler: mux, ReadHeaderTimeout: time.Second}
+	server := http.Server{
+		Addr:              addr,
+		Handler:           mux,
+		ReadHeaderTimeout: time.Second,
+	}
 
 	return errors.Wrap(server.ListenAndServe(), "failed to serve prometheus metrics")
 }
@@ -80,14 +67,13 @@ func getValidators(ctx context.Context, promEndpoint string, promAuth string) ([
 	client := new(http.Client)
 
 	url, err := url.Parse(promEndpoint)
+	if err != nil {
+		return nil, errors.Wrap(err, "parse prometheus endpoint")
+	}
 
 	query := url.Query()
 	query.Add("query", promQuery)
 	url.RawQuery = query.Encode()
-
-	if err != nil {
-		return nil, errors.Wrap(err, "parsing url")
-	}
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


testutil/promrated/prometheus.go
```diff
@@ -16,14 +16,54 @@
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
 )
 
+const (
+	promEndpoint = "https://vm.monitoring.gcp.obol.tech"
+	promQuery    = "group%20by%20%28cluster_name%2C%20ccluster_hash%2C%20cluster_peer%2C%20cluster_network%2C%20pubkey_full%29%20%28core_scheduler_validator_balance_gwei%29"
```
corverroos, line 36: write normal text, http libraries should handle escaping

testutil/promrated/prometheus.go
```diff
@@ -16,14 +16,54 @@
...
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
 )
 
+const (
+	promEndpoint = "https://vm.monitoring.gcp.obol.tech"
```
corverroos, line 35: best avoid globals, make this a flag with a default value

testutil/promrated/prometheus.go
```diff
@@ -16,14 +16,54 @@
...
+	promQuery    = "group%20by%20%28cluster_name%2C%20ccluster_hash%2C%20cluster_peer%2C%20cluster_network%2C%20pubkey_full%29%20%28core_scheduler_validator_balance_gwei%29"
+)
+
+type ClusterInfo struct {
```
corverroos, line 39: always unexport names by default.

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
 	return errors.Wrap(server.ListenAndServe(), "failed to serve prometheus metrics")
 }
 
+func getPubkeyToClusterInfo(ctx context.Context, promAuth string) (map[string]prometheus.Labels, error) {
+	res, err := getPromClusters(ctx, promAuth)
+	if err != nil {
+		return nil, err
+	}
+
+	if res.StatusCode == 200 {
```
corverroos, line 90: avoid nesting https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_return_early_rather_than_nesting_deeply

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
...
+
+		return keyToClusterLabels, nil
+	}
+
+	return nil, errors.New("error processing prom metrics", z.Str("url", res.Request.RequestURI))
+}
+
+func getPromClusters(ctx context.Context, promAuth string) (*http.Response, error) {
+	client := new(http.Client)
+	url := fmt.Sprintf("%s/query?query=%s", promEndpoint, promQuery)
```
corverroos, line 118: use the `url` package to build the url (avoid unstructured untyped logic)

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		log.Error(ctx, "HTTP Request malformed for prom query", err)
```
corverroos, line 122: don't log and return errors, rather just return errors and log at the root (the loop or processor that is doing the scheduling).

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
...
+		return nil, errors.Wrap(err, "error creating prom metrics query")
```
corverroos, line 123: wrapped error messages should be concise, otherwise it makes the final error hard to read and full of noise, suggest: `errors.Wrap(err, "new prometheus request")`

corverroos: also no need to add "failure" or "error" when wrapping, the original error usually contains that, and it is clear from the fact that you are logging an error that it is an error.

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
...
+func readPromJSON(ctx context.Context, res *http.Response) (*PromResponse, error) {
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		log.Error(ctx, "Failed to read body", err)
+		return nil, errors.Wrap(err, "failed to read body")
+	}
+
+	var result *PromResponse
+	if err := json.Unmarshal(body, &result); err != nil {
+		log.Error(ctx, "Failed to deserialize json", err)
```
corverroos, line 145: always return errors

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
...
+	res, err := client.Do(req)
+	if err != nil {
+		log.Error(ctx, "HTTP Request failed for prom query", err)
+		return nil, errors.Wrap(err, "error requesting prom metrics")
+	}
+
+	return res, nil
+}
+
+func readPromJSON(ctx context.Context, res *http.Response) (*PromResponse, error) {
```
corverroos, line 136: this  is two function calls and pretty common go code, can inline. Also you need to close the request body.

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
...
+			}
+		}
+
+		return keyToClusterLabels, nil
+	}
+
+	return nil, errors.New("error processing prom metrics", z.Str("url", res.Request.RequestURI))
+}
+
+func getPromClusters(ctx context.Context, promAuth string) (*http.Response, error) {
```
corverroos, line 116: returning a http response isn't a good abstraction, rather colocate http wrangling stuff and return proper business logic types

testutil/promrated/prometheus.go
```diff
@@ -41,6 +81,73 @@ func serveMonitoring(addr string) error {
 	return errors.Wrap(server.ListenAndServe(), "failed to serve prometheus metrics")
 }
 
+func getPubkeyToClusterInfo(ctx context.Context, promAuth string) (map[string]prometheus.Labels, error) {
```
corverroos, line 84: `getPubkeyToClusterInfo` name can improve, what about `getClusters` or `getClustersFromProm`

corverroos: also add godoc please

testutil/promrated/prometheus.go
```diff
@@ -16,14 +16,54 @@
...
+type PromMetric struct {
+	*ClusterInfo
+	PubKey string `json:"pubkey_full"`
+}
+
+type PromResult struct {
+	Metric PromMetric `json:"metric"`
+}
+
+type PromData struct {
```
corverroos, line 55: no need for all these types, you can just define a single nested type since it is only used for unmarshalling. Also, only include fields that you are actually using.
```
type promResponse struct {
  Data struct {
      Result []struct{
           ...
      } `json:result`
  }  `json:"data"`
}
```

testutil/promrated/prometheus.go
```diff
@@ -16,14 +16,54 @@
...
+
+type ClusterInfo struct {
+	ClusterName    string `json:"cluster_name"`
+	ClusterHash    string `json:"cluster_hash"`
+	ClusterNetwork string `json:"cluster_network"`
+	ClusterPeer    string `json:"cluster_peer"`
+}
+
+type PromMetric struct {
+	*ClusterInfo
```
corverroos, line 47: avoid pointers

testutil/promrated/prometheus.go
corverroos: rather return a proper type instead of map of strings. Also remember that clusters contain multiple validators. Suggest:
```
type Validator struct {
  PubkeyHex string
  ClusterName string
  ClusterHash string
  ClusterPeer string
}
```

corverroos (changes requested):

testutil/promrated/prometheus.go
LukeHackett12: > `getPubkeyToClusterInfo` name can improve, what about `getClusters` or `getClustersFromProm`

Yeah this naming is what I was used to for map types, can change it

corverroos (approved):

PR #1565, opened 2022-12-15
cluster: add `Partial()` method to definition

xenowits:
Adds `Partial()` method to `Definition` to determine if a definition is "partial". Also, don't set definition-hash when marshalling partial definition.

category: feature
ticket: none


corverroos (changes requested):
Partial definitions are never written/marshalled, and I'm not a fan of implicit detecting of partial vs non-partial and then writing the definition hash or now. I'd prefer an explicit solution if it really was required, which is isn't.

xenowits:
Closing this in favour of a rather simple solution: add a test marshaller which sets `DefinitionHash` and `Operators` to nil.

Issue #1501, opened 2022-12-01
Detect all validators queried by VC


corverroos:
## Problem to be solved

We need to highlight when a VC isn't connected to charon or when a VC isn't configured with all private shares.

## Proposed solution

Wire a monitorAPI callback into validatorAPI that is called with the validator pubkey when either of the `get_validators_by_id/pubkey` is called. This is assuming the VC calls this endpoint at least once an epoch for all pubkeys it is configured with.

- In monitoring API, keep two maps, `map[pubkey]bool`, current and previous, current initiated with all pubkeys (group pubkey, not pubshare), previous is empty.
- When the callback is called, delete the pubkey from current
- Every epoch period (or 10min), move current to previous and create new current (initated with all pubkeys)
- When readyz is called, if previous is not empty, failed with "VC not configured"

PR #1566, opened 2022-12-15
app/monitoringapi: detect all validators queried by VC

dB2510:
Configures monitoringapi to detect all validators queried by VC.

category: feature
ticket: #1501

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -219,17 +219,37 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	var pubkeys []core.PubKey
+	for _, dv := range lock.Validators {
+		pk, err := dv.PublicKey()
+		if err != nil {
+			return err
+		}
+
+		pubkey, err := tblsconv.KeyToCore(pk)
+		if err != nil {
+			return err
+		}
+		pubkeys = append(pubkeys, pubkey)
+	}
+
 	sender := new(p2p.Sender)
 
 	wirePeerInfo(life, tcpNode, peerIDs, lock.LockHash, sender)
 
 	qbftDebug := newQBFTDebugger()
 
+	// Non-blocking channel to send seen public keys from validatorapi to monitoringapi.
+	seenPubkeys := make(chan core.PubKey, 1)
+	seenPubkeysFunc := func(pubkey core.PubKey) {
+		seenPubkeys <- pubkey
+	}
+
 	wireMonitoringAPI(ctx, life, conf.MonitoringAddr, localEnode, tcpNode, eth2Cl, peerIDs,
-		promRegistry, qbftDebug)
+		promRegistry, qbftDebug, pubkeys, seenPubkeys)
 
 	err = wireCoreWorkflow(ctx, life, conf, lock, nodeIdx, tcpNode, p2pKey, eth2Cl,
-		peerIDs, sender, qbftDebug.AddInstance)
+		peerIDs, sender, qbftDebug.AddInstance, seenPubkeysFunc)
 	if err != nil {
 		return err
 	}
@@ -322,7 +342,7 @@ func wireP2P(ctx context.Context, life *lifecycle.Manager, conf Config,
 func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 	lock cluster.Lock, nodeIdx cluster.NodeIdx, tcpNode host.Host, p2pKey *ecdsa.PrivateKey,
 	eth2Cl eth2wrap.Client, peerIDs []peer.ID, sender *p2p.Sender,
-	qbftSniffer func(*pbv1.SniffedConsensusInstance),
+	qbftSniffer func(*pbv1.SniffedConsensusInstance), seenPubkeysFunc func(key core.PubKey),
 ) error {
 	// Convert and prep public keys and public shares
 	var (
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -219,17 +219,37 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		pubkey, err := tblsconv.KeyToCore(pk)
+		if err != nil {
+			return err
+		}
+		pubkeys = append(pubkeys, pubkey)
+	}
```
corverroos, line 234: extract this as a function

app/app.go
```diff
@@ -219,17 +219,37 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+
 	sender := new(p2p.Sender)
 
 	wirePeerInfo(life, tcpNode, peerIDs, lock.LockHash, sender)
 
 	qbftDebug := newQBFTDebugger()
 
+	// Non-blocking channel to send seen public keys from validatorapi to monitoringapi.
+	seenPubkeys := make(chan core.PubKey, 1)
+	seenPubkeysFunc := func(pubkey core.PubKey) {
```
corverroos, line 244: why convert from channel to callback? Just pass the channel into validatorapi

app/metrics.go
```diff
@@ -32,6 +32,8 @@ const (
 	// readyzInsufficientPeers indicates the readyz is returning 500s since this node isn't connected
 	// to quorum peers via the P2P network.
 	readyzInsufficientPeers = 4
+	// readyzVCNotConfigured indicated the readyz is returning 500s since VC is not configured for this node.
```
corverroos, line 35: suggest differentiating between "no VC (no validators)" vs "VC missing some validators"

app/monitoringapi.go
```diff
@@ -107,11 +112,31 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
 	)
 	go func() {
 		ticker := clock.NewTicker(10 * time.Second)
+		epochTicker := clock.NewTicker(384 * time.Second) // 32 slots * 12 second slot time
```
corverroos, line 115: `32 * 12 * time.Second`

app/monitoringapi.go
```diff
@@ -107,11 +112,31 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
...
+			current[pubkey] = true
+		}
 
 		for {
 			select {
 			case <-ctx.Done():
 				return
+			case <-epochTicker.Chan():
+				// Copy current to previous.
+				previous = make(map[core.PubKey]bool)
```
corverroos, line 130:
```
// Copy current to previous and clear current
previous, current = current, newCurrent()
```

app/monitoringapi.go
```diff
@@ -107,11 +112,31 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
 	)
 	go func() {
 		ticker := clock.NewTicker(10 * time.Second)
+		epochTicker := clock.NewTicker(384 * time.Second) // 32 slots * 12 second slot time
+		current := make(map[core.PubKey]bool)
+		previous := make(map[core.PubKey]bool)
```
corverroos, line 117: suggest:
```
// newCurrent returns a new current map, populated with all the pubkeys.
newCurrent := func() map[core.PubKey]bool{
  current := make(map[core.PubKey]bool)
  for _, pubkey := range pubkeys {
    current[pubkey] = true
  }
}
```

app/monitoringapi_internal_test.go
```diff
@@ -103,13 +117,22 @@ func TestStartChecker(t *testing.T) {
...
+			}
+
+			// Advance clock for first tick.
+			advanceClock(clock, 10*time.Second)
+
+			// Advance clock for first epoch tick.
+			advanceClock(clock, 384*time.Second)
+
+			// Wait for epoch tick to happen before last 10s periodic tick.
+			time.Sleep(100 * time.Millisecond)
```
corverroos, line 134: no sleep in tests, especially not if you have deterministic clocks.

app/monitoringapi_internal_test.go
```diff
@@ -129,3 +152,11 @@ func TestStartChecker(t *testing.T) {
 		})
 	}
 }
+
+func advanceClock(clock clockwork.FakeClock, duration time.Duration) {
+	// We wrap the Advance() calls with blockers to make sure that the ticker
+	// can go to sleep and produce ticks without time passing in parallel.
+	clock.BlockUntil(1)
```
corverroos, line 159: note you added another ticker, so there are more things blocking than previously

corverroos (changes requested):

app/monitoringapi.go
dB2510: this is incorrect, here previous will POINT to current.
https://go.dev/play/p/XsmNEqLRwNH

app/app.go
dB2510: i converted it to callback so that i don't need to worry about passing channel to validatorapi in tests explicitly

app/monitoringapi.go
corverroos: no it is not: https://go.dev/play/p/9GruOUVMEHF

app/app.go
corverroos: not sure why that is a problem, you can support nil in validatorapi
```
if seenPubkeys != nil {
  seenPubkeys <- pubkey
}
```

app/monitoringapi.go
dB2510: yeah right, because current is also reinitialised

app/app.go
```diff
@@ -225,11 +225,19 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 	qbftDebug := newQBFTDebugger()
 
+	// Non-blocking channel to send seen public keys from validatorapi to monitoringapi.
+	seenPubkeys := make(chan core.PubKey, 1)
```
corverroos, line 229: unbuffered is probably better

corverroos (approved):

app/app.go
dB2510: sure

Issue #1464, opened 2022-11-20
Add support for `capella` blocks


dB2510:
## Problem to be solved

Ethereum is eyeing to introduce withdrawals in next upgrade (shanghai). This introduces new block version named `capella` which introduces some new fields to `BeaconBlockBody` and `ExecutionPayload`. For more details [refer](https://github.com/ethereum/consensus-specs/tree/dev/specs/capella).

## Proposed solution

- Add support for capella versioned beacon block wherever block version is considered across charon codebase.
- Add relevant tests.

corverroos:
What is the timeline on this for goerli and gnosis? ie, by when does need to be in?

dB2510:
> What is the timeline on this for goerli and gnosis? ie, by when does need to be in?

There is no announcement for specific date as when it will be live on goerli or gnosis, but client teams are under development currently. `go-eth2-client` has already added support.
We can track shanghai upgrade planning here: https://github.com/ethereum/pm/issues/450.

PR #1570, opened 2022-12-16
core: add support for capella blocks

xenowits:
Adds support for capella blocks.

category: feature
ticket: #1464

### DIFF
--- a/app/eth2wrap/synthproposer.go
+++ b/app/eth2wrap/synthproposer.go
@@ -25,8 +25,8 @@ import (
 	eth2client "github.com/attestantio/go-eth2-client"
 	"github.com/attestantio/go-eth2-client/api"
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
-	apiv1bellatrix "github.com/attestantio/go-eth2-client/api/v1/bellatrix"
-	apiv1capella "github.com/attestantio/go-eth2-client/api/v1/capella"
+	eth2bellatrix "github.com/attestantio/go-eth2-client/api/v1/bellatrix"
+	eth2capella "github.com/attestantio/go-eth2-client/api/v1/capella"
 	"github.com/attestantio/go-eth2-client/spec"
 	"github.com/attestantio/go-eth2-client/spec/bellatrix"
 	"github.com/attestantio/go-eth2-client/spec/capella"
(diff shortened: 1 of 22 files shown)


core/fetcher/fetcher.go
```diff
@@ -258,13 +258,18 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, defSet core
...
 
-		// Ensure fee recipient is correctly populated in block.
-		if block.Version == spec.DataVersionBellatrix {
-			actual := fmt.Sprintf("%#x", block.Bellatrix.Body.ExecutionPayload.FeeRecipient)
-			if !strings.EqualFold(actual, f.feeRecipientAddress) {
-				log.Warn(ctx, "Proposing block with unexpected fee recipient address", nil,
-					z.Str("expected", f.feeRecipientAddress), z.Str("actual", actual))
-			}
+		// Ensure fee recipient is correctly populated in block (if ever populated).
+		var actualAddr string
```
corverroos, line 262: extract this as a function

core/validatorapi/router.go
```diff
@@ -562,8 +572,19 @@ func proposeBlindedBlock(p eth2client.BlindedBeaconBlockProposalProvider) handle
 
 func submitBlock(p eth2client.BeaconBlockSubmitter) handlerFunc {
 	return func(ctx context.Context, _ map[string]string, _ url.Values, body []byte) (interface{}, error) {
+		capellaBlock := new(capella.SignedBeaconBlock)
```
corverroos, line 575: please make sure we have a test for this, ie, when submitting a bellatrix block it gets marshalled as a bellatrix block, etc.

testutil/beaconmock/options.go
```diff
@@ -453,7 +453,7 @@ func defaultMock(httpMock HTTPMock, httpServer *http.Server, clock clockwork.Clo
 							SyncCommitteeBits:      bitfield.NewBitvector512(),
 							SyncCommitteeSignature: testutil.RandomEth2Signature(),
 						},
-						ExecutionPayload: testutil.RandomExecutionPayLoad(),
+						ExecutionPayload: testutil.RandomBellatrixExecutionPayLoad(),
```
corverroos, line 456: lets return capella blocks from beaconmock be default.

core/validatorapi/router.go
xenowits: I have added a test in `sigagg_test.go`(https://github.com/ObolNetwork/charon/blob/a8c623751c9bb5fa1330b6f90673e350ff71f686/core/sigagg/sigagg_test.go#L262). This tests marshalling & unmarshalling as it calls `block.clone()` which involves both steps.

corverroos: this code isn't tested by that package...

corverroos: we need to test the "submitBlock" function specifically

app/eth2wrap/synthproposer.go
```diff
@@ -118,47 +120,11 @@ func (h *synthWrapper) BlindedBeaconBlockProposal(ctx context.Context, slot eth2
 	block, err := h.syntheticBlock(ctx, slot, vIdx)
 	if err != nil {
 		return nil, err
-	} else if block.Version != spec.DataVersionBellatrix {
+	} else if block.Version == spec.DataVersionPhase0 || block.Version == spec.DataVersionAltair {
```
corverroos, line 123: rather handle this in a `default` case in switch case below

core/fetcher/fetcher.go
```diff
@@ -379,3 +373,25 @@ func (f *Fetcher) fetchContributionData(ctx context.Context, slot int64, defSet
 
 	return resp, nil
 }
+
+// verifyFeeRecipient logs a warning when fee recipient is not correctly populated in the block.
+func verifyFeeRecipient(ctx context.Context, block *spec.VersionedBeaconBlock, feeRecipientAddress string) {
+	// Fee-recipient is not available in forks earlier than bellatrix.
+	if block.Version == spec.DataVersionPhase0 || block.Version == spec.DataVersionAltair {
```
corverroos, line 380: handle this default in switch case

corverroos (approved):

PR #1573, opened 2022-12-16
app: fix lint again

corverroos:
Fixing again.  Golangci-lint really doesn't like commented out code as it thinks it is comments, best is to remove completely.

category: fixbuild
ticket: none


LukeHackett12 (approved):
branch name is a winner

Issue #947, opened 2022-08-09
Improve error message when VC key share index doesn't match charon index


corverroos:
## Problem to be solved

Charon nodes infer their peerIndex (and shareIndex) from the charon-enr-private-key at runtime and the index of that identity in the cluster definition. The VC connected to that Charon node needs to be configured with matching private-key-shares. 

The Nth keyshare needs to loaded into the Nth VC connected to the Nth charon node.

If there is mix up, and the Mth keyshare is used to query the Nth charon node, then the error `unknown public key` is returned. 

This is hard to debug.

## Proposed solution

Improve error messages such that:
 - If the pubshare submitted by the VC belongs to another index in the cluster, then return: `mismatching validator client key share index, Mth key share submitted to Nth Charon node; key_share_index=2, charon_index=1` 
 - If the pubshare submitted isn't in the lock file, then return `unknown validator client key share, not in cluster lock`

thomasheremans:
Will need to add to FAQ as well

xenowits:
The solution works. I connected `vc0 to node1` and `vc1 to node0` and then ran the dv cluster.

**Node0 (peerIdx 1) <-> VC1 (keyshareIdx 2)**
<img width="1456" alt="Screenshot 2023-01-08 at 11 23 22" src="https://user-images.githubusercontent.com/43061995/211182600-02ab0c32-518c-4ee9-8b54-898b14f47cd0.png">

**Node1 (peerIdx 0)  <-> VC0 (keyshareIdx 1)**
<img width="1450" alt="Screenshot 2023-01-08 at 11 23 07" src="https://user-images.githubusercontent.com/43061995/211182598-5274c7a8-ab3b-4e6f-aabf-7f19ac8d573d.png">

Are `1-indexed` field values confusing? I think it's better with 0-indexed

PR #1586, opened 2023-01-04
core/validatorapi: improve error

xenowits:
Improves error message when when VC key share index doesn't match charon index.

category: feature
ticket: #947

### DIFF
--- a/core/validatorapi/validatorapi.go
+++ b/core/validatorapi/validatorapi.go
@@ -55,38 +55,37 @@ func NewComponentInsecure(_ *testing.T, eth2Cl eth2wrap.Client, shareIdx int) (*
 }
 
 // NewComponent returns a new instance of the validator API core workflow component.
-func NewComponent(eth2Cl eth2wrap.Client, pubShareByKey map[*bls_sig.PublicKey]*bls_sig.PublicKey,
-	allPubSharesByKey map[core.PubKey]map[int]*bls_sig.PublicKey, shareIdx int, feeRecipientAddress string, builderAPI bool, seenPubkeys func(core.PubKey),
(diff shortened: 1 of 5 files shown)


core/validatorapi/validatorapi.go
```diff
@@ -56,7 +56,7 @@ func NewComponentInsecure(_ *testing.T, eth2Cl eth2wrap.Client, shareIdx int) (*
 
 // NewComponent returns a new instance of the validator API core workflow component.
 func NewComponent(eth2Cl eth2wrap.Client, pubShareByKey map[*bls_sig.PublicKey]*bls_sig.PublicKey,
```
corverroos, line 58: you are passing in the same data `pubShareByKey` twice but in different formats

core/validatorapi/validatorapi_test.go
```diff
@@ -181,6 +181,7 @@ func TestSubmitAttestations_Verify(t *testing.T) {
 	corePubKey, err := tblsconv.KeyToCore(pubkey)
 	require.NoError(t, err)
 	pubShareByKey := map[*bls_sig.PublicKey]*bls_sig.PublicKey{pubkey: pubkey} // Maps self to self since not tbls
+	allPubSharesByKey := make(map[core.PubKey]map[int]*bls_sig.PublicKey)
```
corverroos, line 184: can just use nil

core/validatorapi/validatorapi.go
xenowits: That's right. Maybe we can replace `pubShareByKey` with `allPubSharesByKey`? Since we can calculate relevant pubkey mappings from the latter

corverroos: yeah, lets try that

corverroos (approved):

PR #1595, opened 2023-01-06
core/dutydb: delete cancelled queries

corverroos:
Delete expired dutydb queries since they may never be resolved which causes a memory leak.

category: bug
ticket: #978


dB2510 (approved):

core/dutydb/memory.go
```diff
@@ -668,28 +708,43 @@ type contribKey struct {
...
 	Key      contribKey
 	Response chan<- *altair.SyncCommitteeContribution
+	Cancel   <-chan struct{}
+}
+
+// cancelled returns true if channel has been closed.
+func cancelled(cancel <-chan struct{}) bool {
+	select {
+	case <-cancel:
+		return true
```
xenowits, line 746:
```go
case _, ok := <-cancel:
   return !ok
```
ok is false if there are no more values to receive and the channel is closed.

Issue #1590, opened 2023-01-05
Add support for p2p external IP and Hostname to relay discovery


corverroos:
## Problem to be solved

the `--p2p-external-ip` and `--p2p-external-hostname` flags are not supported in native libp2p relay discovery. It is only supported in the legacy udp discv5 discovery. 

## Proposed solution

Convert those flags to multiaddrs and add to advertised addresses.

PR #1604, opened 2023-01-09
p2p: relay discovery supports external ip and host

corverroos:
Add support for external IP and external hostname to relay discovery.

category: feature
ticket: #1590


p2p/p2p.go
```diff
@@ -103,6 +107,45 @@ func NewTCPNode(ctx context.Context, cfg Config, key *ecdsa.PrivateKey, connGate
...
+		return nil, err
+	}
+
+	var ports []int
+	for _, addr := range tcpAddrs {
+		ports = append(ports, addr.Port)
+	}
+
+	var resp []ma.Multiaddr
+
```
xenowits, line 123: nit: can remove newline

corverroos: added new line since we append to it in two different blocks

xenowits (approved):

PR #1615, opened 2023-01-10
cmd: add `--definition-file` flag to create cluster command

xenowits:
Refactors `charon create cluster` command to fetch cluster definition from disk/network and use it for creating cluster outputs.

category: feature
ticket: https://github.com/ObolNetwork/charon/issues/1490

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -131,18 +131,11 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 	}
 
 	var (
-		def     cluster.Definition
-		err     error
-		network = conf.Network
+		def cluster.Definition
+		err error
 	)
-
 	if conf.DefFile != "" { // Load definition from DefFile
 		def, err = loadDefinition(ctx, conf.DefFile)
-		if err != nil {
-			return err
-		}
-
-		network, err = eth2util.ForkVersionToNetwork(def.ForkVersion)
 		if err != nil {
 			return err
 		}
@@ -155,7 +148,7 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 
 	numNodes := len(def.Operators)
 	// Validate definition
-	err = validateDef(ctx, conf.InsecureKeys, numNodes, def.Name, def.WithdrawalAddress, network)
+	err = validateDef(ctx, conf.InsecureKeys, def)
 	if err != nil {
 		return err
 	}
@@ -186,13 +179,12 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 	def.Operators = ops
 
 	// Write deposit-data file
-	if err = writeDepositData(def.WithdrawalAddress, network, conf.ClusterDir, numNodes, secrets); err != nil {
+	if err = writeDepositData(def.WithdrawalAddress, conf.ClusterDir, def.ForkVersion, numNodes, secrets); err != nil {
 		return err
 	}
 
 	// Create cluster-lock
-	var lock cluster.Lock
-	lock = cluster.Lock{
+	lock := cluster.Lock{
 		Definition: def,
 		Validators: vals,
 	}
@@ -312,7 +304,12 @@ func getKeys(splitKeys bool, splitKeysDir string, numDVs int) ([]*bls_sig.Secret
 }
 
 // writeDepositData writes deposit data to disk for the DVs for all peers in a cluster.
-func writeDepositData(withdrawalAddr, network, clusterDir string, numNodes int, secrets []*bls_sig.SecretKey) error {
+func writeDepositData(withdrawalAddr, clusterDir string, forkVersion []byte, numNodes int, secrets []*bls_sig.SecretKey) error {
+	network, err := eth2util.ForkVersionToNetwork(forkVersion)
+	if err != nil {
+		return err
+	}
+
 	// Create deposit message signatures
 	msgSigs, err := signDepositDatas(secrets, withdrawalAddr, network)
 	if err != nil {
@@ -515,19 +512,23 @@ func checksumAddr(a string) (string, error) {
 	return common.HexToAddress(a).Hex(), nil
 }
 
-// validateDef returns an error if any of the provided cluster fields are invalid.
-func validateDef(ctx context.Context, insecureKeys bool, numNodes int, clusterName, withdrawalAddr, network string) error {
-	if numNodes < minNodes {
-		return errors.New("insufficient number of nodes (min = 4)", z.Int("num_nodes", numNodes))
+// validateDef returns an error if the provided cluster definition is valid.
+func validateDef(ctx context.Context, insecureKeys bool, def cluster.Definition) error {
+	if len(def.Operators) < minNodes {
+		return errors.New("insufficient number of nodes (min = 4)", z.Int("num_nodes", len(def.Operators)))
 	}
 
+	network, err := eth2util.ForkVersionToNetwork(def.ForkVersion)
+	if err != nil {
+		return err
+	}
 	if insecureKeys && isMainNetwork(network) {
 		return errors.New("insecure keys not supported on mainnet")
 	} else if insecureKeys {
 		log.Warn(ctx, "Insecure keystores configured. ONLY DO THIS DURING TESTING", nil)
 	}
 
-	if clusterName == "" {
+	if def.Name == "" {
 		return errors.New("name not provided")
 	}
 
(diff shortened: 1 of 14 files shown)


xenowits:
Sample output:

Screenshot 2023-01-10 at 16 05 06

dkg/disk.go
```diff
@@ -98,8 +98,8 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
 	return def, nil
 }
 
-// fetchDefinition fetches cluster definition file from a remote URI.
-func fetchDefinition(ctx context.Context, url string) (cluster.Definition, error) {
+// FetchDefinition fetches cluster definition file from a remote URI.
+func FetchDefinition(ctx context.Context, url string) (cluster.Definition, error) {
```
corverroos, line 102: move to `cluster` package

cmd/createcluster_internal_test.go
```diff
@@ -170,19 +226,53 @@ func TestValidNetwork(t *testing.T) {
...
 	require.NoError(t, err)
 
 	conf.Network = "mainnet"
-	err = validateClusterConfig(ctx, conf)
+	err = validateClusterConfig(ctx, conf.InsecureKeys, conf.NumNodes, conf.Name, conf.WithdrawalAddr, conf.Network)
 	require.Error(t, err, "zero address")
 }
+
+// newDefinition returns a new definition with creator field populated.
+func newDefinition(t *testing.T, clusterName string) cluster.Definition {
```
corverroos, line 247: instead of exporting logic for a single test, rather just use the `charon/cluster/examples/cluster-definition-002.json` file as a test fixture directly.

cmd/createcluster_internal_test.go
```diff
@@ -85,6 +100,32 @@ func TestCreateCluster(t *testing.T) {
 				Network:   "goerli",
 			},
 		},
+		{
+			Name: "solo flow definition from disk",
+			Prep: func(t *testing.T, config clusterConfig) clusterConfig {
+				t.Helper()
+
+				// Save definition to disk
+				dir, err := os.MkdirTemp("", "")
```
corverroos, line 109: do this at the start of the test and set the "DeFFile" field like normal config

cmd/createcluster_internal_test.go
```diff
@@ -85,6 +100,32 @@ func TestCreateCluster(t *testing.T) {
...
+				err = os.WriteFile(defPath, defBytes, 0o444)
+				require.NoError(t, err)
+				config.DefFile = defPath
+
+				return config
+			},
+		},
+		{
+			Name: "solo flow definition from network",
+			Prep: func(t *testing.T, config clusterConfig) clusterConfig {
```
corverroos, line 122: use normal Config field

cmd/createcluster.go
```diff
@@ -521,3 +581,66 @@ func aggSign(secrets [][]*bls_sig.SecretKeyShare, message []byte) ([]byte, error
...
+		var err error
+		def, err = dkg.FetchDefinition(ctx, defFile)
+		if err != nil {
+			return cluster.Definition{}, errors.Wrap(err, "read definition")
+		}
+
+		log.Info(ctx, "Cluster definition downloaded from URL", z.Str("URL", defFile),
+			z.Str("definition_hash", fmt.Sprintf("%#x", def.DefinitionHash)))
+
+		return def, nil
```
corverroos, line 601: hashes and sigs not verified in this flow

cmd/createcluster.go
```diff
@@ -521,3 +581,66 @@ func aggSign(secrets [][]*bls_sig.SecretKeyShare, message []byte) ([]byte, error
 
 	return b, nil
 }
+
+// loadDefinition returns the cluster definition from disk or an HTTP URL. It also verifies signatures
+// and hashes before returning the definition.
+func loadDefinition(ctx context.Context, defFile string) (cluster.Definition, error) {
+	var def cluster.Definition
```
corverroos, line 588: why define this if you return early below?

cmd/createcluster_internal_test.go
```diff
@@ -146,6 +187,21 @@ func testCreateCluster(t *testing.T, conf clusterConfig) {
...
 		require.NoError(t, lock.VerifyHashes())
 		require.NoError(t, lock.VerifySignatures())
+
+		if conf.DefFile != "" {
+			var def cluster.Definition
+			def, err = loadDefinition(context.Background(), conf.DefFile)
+			require.NoError(t, err)
+
+			// Config hash and creator should remain the same
+			require.Equal(t, def.ConfigHash, lock.ConfigHash)
```
corverroos, line 197: rather create variable for these above, then no need to load the definition again.

cmd/createcluster.go
```diff
@@ -127,48 +131,102 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		conf.Network = eth2util.Goerli.Name
 	}
 
-	if err := validateClusterConfig(ctx, conf); err != nil {
-		return err
+	var (
+		def            cluster.Definition
+		threshold      = conf.Threshold
```
corverroos, line 136: no need for these variables, either load def from DefFile or populate it from conf.

cmd/createcluster.go
```diff
@@ -127,48 +131,102 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
...
+			return err
+		}
 	}
 
-	safeThreshold := cluster.Threshold(conf.NumNodes)
-	if conf.Threshold == 0 {
-		conf.Threshold = safeThreshold
-	} else if conf.Threshold != safeThreshold {
-		log.Warn(ctx, "Non standard `--threshold` flag provided, this will affect cluster safety", nil, z.Int("threshold", conf.Threshold), z.Int("safe_threshold", safeThreshold))
+	err := validateClusterConfig(ctx, conf.InsecureKeys, numNodes, clusterName, withdrawalAddr, network)
```
corverroos, line 164: refactor to validateDef

corverroos (changes requested):

dkg/disk.go
xenowits: ok, moving it to `cluster/helpers.go`

cmd/createcluster_internal_test.go
xenowits: `charon/cluster/examples/cluster-definition-002.json` file is not a valid definition file as it only has 2 operators and will fail in `validateConfig`

cmd/createcluster.go
xenowits: I updated it now so that definition downloaded from both flows, ie, disk and network, are verified before returned

cmd/createcluster_internal_test.go
xenowits: yeah makes sense as we already create a definition which can be passed to `testCluster` and used here

cmd/createcluster_internal_test.go
xenowits: done

cmd/createcluster.go
```diff
@@ -127,48 +130,87 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
...
 	}
 
-	safeThreshold := cluster.Threshold(conf.NumNodes)
-	if conf.Threshold == 0 {
-		conf.Threshold = safeThreshold
-	} else if conf.Threshold != safeThreshold {
-		log.Warn(ctx, "Non standard `--threshold` flag provided, this will affect cluster safety", nil, z.Int("threshold", conf.Threshold), z.Int("safe_threshold", safeThreshold))
+	numNodes := len(def.Operators)
+	// Validate definition
+	err = validateDef(ctx, conf.InsecureKeys, numNodes, def.Name, def.WithdrawalAddress, network)
```
corverroos, line 158: you are not passing in a definition, so this shouldn't be called `validateDef`

cmd/createcluster.go
```diff
@@ -127,48 +130,87 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
...
+	def.Operators = ops
 
-	if err = writeLock(conf, dvs, peers, shareSets); err != nil {
+	// Write deposit-data file
+	if err = writeDepositData(def.WithdrawalAddress, network, conf.ClusterDir, numNodes, secrets); err != nil {
+		return err
+	}
+
+	// Create cluster-lock
+	var lock cluster.Lock
```
corverroos, line 194: remove this var, kist do `lock := cluster.Lock{...`

corverroos (approved):

PR #1616, opened 2023-01-10
core/scheduler: trim duties after 3 epochs

corverroos:
Fixes issue introduced by #1596 that trimmed data after 5 slots which caused inclusion delay calculation to stop working (it requires duties of 32 ago). Instead just delete duties after 3 epochs (avoid races and off-by-one issues of 2 epochs).

category: bug
ticket: none

### DIFF
--- a/app/monitoringapi_internal_test.go
+++ b/app/monitoringapi_internal_test.go
@@ -144,12 +144,12 @@ func TestStartChecker(t *testing.T) {
 			if tt.err != nil {
 				require.Eventually(t, func() bool {
 					err = readyErrFunc()
-					if !errors.Is(err, errReadyUninitialised) {
-						require.EqualError(t, err, tt.err.Error())
-						return true
+					if !errors.Is(err, tt.err) {
+						t.Logf("Ignoring unexpected error, got=%v, want=%v", err, tt.err)
+						return false
 					}
 
-					return false
+					return true
 				}, time.Second, 100*time.Millisecond)
 			} else {
 				require.Eventually(t, func() bool {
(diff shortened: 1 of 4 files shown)


app/monitoringapi_internal_test.go
```diff
@@ -144,12 +144,12 @@ func TestStartChecker(t *testing.T) {
 			if tt.err != nil {
 				require.Eventually(t, func() bool {
 					err = readyErrFunc()
-					if !errors.Is(err, errReadyUninitialised) {
```
corverroos, line 147: @dB2510, this test was flapping so I fixed it

core/scheduler/scheduler.go
```diff
@@ -75,11 +77,11 @@ type Scheduler struct {
 	pubkeys         []core.PubKey
 	quit            chan struct{}
 	clock           clockwork.Clock
-	deadliner       core.Deadliner
 	delayFunc       delayFunc
 	metricSubmitter metricSubmitter
-	resolvedEpoch   uint64
+	resolvedEpoch   int64
```
corverroos, line 82: aligned with other epochs using int64

core/scheduler/scheduler.go
```diff
@@ -277,7 +280,8 @@ func (s *Scheduler) resolveDuties(ctx context.Context, slot core.Slot) error {
 		return err
 	}
 
-	s.setResolvedEpoch(uint64(slot.Epoch()))
+	s.setResolvedEpoch(slot.Epoch())
+	s.trimDuties(slot.Epoch() - trimEpochOffset)
```
corverroos, line 284: trim when resolved so only need to check `resolvedEpoch` to know when things were trimmed

dB2510 (approved):
LGTM

Issue #1503, opened 2022-12-02
Add `--keymanager-address` as a flag to `create cluster`


OisinKyne:
## Problem to be Solved

We want to create a cluster in a centralised manner, and rather than pushing the created BLS private key shares and their passwords to disk; we want to push the private keys into our existing private key infrastructure via the [keymanager api](https://ethereum.github.io/keymanager-APIs/). 

## Proposed Solution

Add a flag `--keymanager-address` that takes a well formed URI (preferably able to handle TLS, basic auth, and query paramaters containing auth codes) and sends a POST request to [this endpoint](https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager/importKeystores). It should have a 10 second time out and a max retrys of 5, and should log the error status code it receives and that it is retrying. If it fails we could either:

- Write the private keys to disk the existing way
- Because `create cluster` is the centralised approach, this is the only process running, we could just abort and cancel the entire process

I think the latter is a safer more 'single-purpose' approach if we fail to write the keys versus inadvertently falling back to writing private keys on disk (that may be an ephemeral volume, or something compromised by hackers that can't get into the remote signer but can DoS it enough to force private keys onto disk). 

We should consider taking the opportunity to change how we output `--cluster-dir` folder structure at the same time. If all of the validator keys are in a remote signer, we only need to output a deposit file, cluster lock and enrs for all of the nodes. Maybe we could use a flat folder structure in `.charon/` with the enrs named like `charon-enr-private-key-0`

## Out of scope

Can we accept a list of keymanagers the same length as `num-operators`, and in that case we put private keys for each `nodeX` folder into a consecutive remote signer? Seems too fancy and should be a future feature request from a user.

PR #1622, opened 2023-01-12
cmd: add `keymanager-address` flag to create cluster command

xenowits:
Adds `--keymanager-address` flag to create cluster command. Also adds tests.

category: feature
ticket: #1503


eth2util/keystore/keystore.go
```diff
@@ -62,6 +62,39 @@ func StoreKeys(secrets []*bls_sig.SecretKey, dir string) error {
 	return storeKeysInternal(secrets, dir, "keystore-%d.json")
 }
 
+// keymanagerReq represents the keymanager API request body for POST request. Refer: https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager/importKeystores
+type keymanagerReq struct {
+	Keystores          []keystore `json:"keystores"`
+	Passwords          []string   `json:"passwords"`
+	SlashingProtection string     `json:"slashing_protection"` // https://eips.ethereum.org/EIPS/eip-3076
```
xenowits, line 69: still need to figure out how to use it

xenowits:
Sample output:

Screenshot 2023-01-12 at 13 31 30
Screenshot 2023-01-12 at 13 31 46

eth2util/keystore/keystore.go
xenowits: This fields is not relevant to `create cluster` as we are creating new validators which don't have a signing history as they aren't even activated yet.

**More on EIP-3076:**
A standard format for transferring a key’s signing history allows validators to easily switch between clients without the risk of signing conflicting messages. While a [common keystore format](https://eips.ethereum.org/EIPS/eip-2335) provides part of the solution, it does not contain any information about a key’s signing history. For a validator moving their keys from client A to client B, this could lead to scenarios in which client B inadvertently signs a message that conflicts with an earlier message signed with client A. The interchange format described here provides a solution to this problem.

cmd/createcluster.go
```diff
@@ -202,7 +214,7 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		writeWarning(w)
 	}
 
-	writeOutput(w, conf.SplitKeys, conf.ClusterDir, numNodes)
+	writeOutput(w, conf.SplitKeys, conf.ClusterDir, numNodes, len(conf.KeymanagerAddrs) != 0)
```
corverroos, line 217: suggest creating a variable above: `keysToDisk := len(conf.KeymanagerAddrs) == 0`

cmd/createcluster.go
```diff
@@ -172,12 +174,22 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
...
 		return err
 	}
 	def.Operators = ops
 
+	if len(conf.KeymanagerAddrs) != 0 { // Push keys to keymanager
+		if err = saveKeysToKeymanager(ctx, conf.KeymanagerAddrs, numNodes, shareSets); err != nil {
+			return err
+		}
+	} else { // Or else write keys to disk
+		if err = saveKeysToDisk(numNodes, conf.ClusterDir, conf.InsecureKeys, shareSets); err != nil {
```
corverroos, line 188: `write` is better than `save`

cmd/createcluster.go
```diff
@@ -94,6 +96,7 @@ func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.StringVar(&config.Name, "name", "", "The cluster name")
 	flags.StringVar(&config.ClusterDir, "cluster-dir", ".charon/cluster", "The target folder to create the cluster in.")
 	flags.StringVar(&config.DefFile, "definition-file", "", "Optional path to a cluster definition file or an HTTP URL. This overrides all other configuration flags.")
+	flags.StringSliceVar(&config.KeymanagerAddrs, "keymanager-addresses", nil, "Comma separated list of keymanager URLs to push validator keys to.")
```
corverroos, line 99: add comment that an address per node is required

cmd/createcluster.go
```diff
@@ -386,43 +398,94 @@ func getValidators(dvs []tbls.TSS) ([]cluster.DistValidator, error) {
 	return vals, nil
 }
 
-// getOperators returns a list of operators from the provided secret keyshares.
-func getOperators(numNodes int, clusterDir string, insecureKeys bool, shareSets [][]*bls_sig.SecretKeyShare) ([]cluster.Operator, error) {
-	var peers []p2p.Peer
+// saveKeysToKeyManager saves validator keys to the provided keymanager addresses.
+func saveKeysToKeymanager(ctx context.Context, addrs []string, numNodes int, shareSets [][]*bls_sig.SecretKeyShare) error {
+	if len(addrs) != numNodes {
+		return errors.New("insufficient no of keymanager addresses", z.Int("expected", numNodes), z.Int("got", len(addrs)))
```
corverroos, line 404: fail fast, do config validation at the start of the process.

cmd/createcluster.go
```diff
@@ -386,43 +398,94 @@ func getValidators(dvs []tbls.TSS) ([]cluster.DistValidator, error) {
...
+		addr := addrs[i]
+		grp.Go(func() error {
+			for retries > 0 {
+				err := keystore.PostKeysToKeymanager(ctx, addr, secrets)
+				if err == nil {
+					log.Debug(ctx, "Pushed keys to keymanager", z.Str("addr", addr))
+					return nil
+				}
+
+				retries--
```
corverroos, line 429: this variable isn't being a lock but it is being updated concurrently. Suggest removing retries. Rather just check if the keymanagers are accessible at the start, if not fail, otherwise just write and fail.

eth2util/keystore/keystore.go
```diff
@@ -62,6 +67,72 @@ func StoreKeys(secrets []*bls_sig.SecretKey, dir string) error {
...
+	res, err := json.Marshal(req)
+	if err != nil {
+		return nil, errors.New("marshal keymanager request body")
+	}
+
+	return res, nil
+}
+
+// PostKeysToKeymanager pushes the secrets to the provided keymanager address. The HTTP request times out after 2s.
+func PostKeysToKeymanager(ctx context.Context, addr string, secrets []*bls_sig.SecretKey) error {
```
corverroos, line 103: libraries shouldn't do IO, libaries should do logic and data, rather return the request type and let the user do IO.

corverroos (changes requested):

cmd/createcluster.go
```diff
@@ -386,43 +401,128 @@ func getValidators(dvs []tbls.TSS) ([]cluster.DistValidator, error) {
...
+}
+
+// saveKeysToKeyManager saves validator keys to the provided keymanager addresses.
+func saveKeysToKeymanager(ctx context.Context, addrs []string, numNodes int, shareSets [][]*bls_sig.SecretKeyShare) error {
+	if len(addrs) != numNodes {
+		return errors.New("insufficient no of keymanager addresses", z.Int("expected", numNodes), z.Int("got", len(addrs)))
+	}
+
+	grp := new(errgroup.Group)
+	retriesCount := 5
```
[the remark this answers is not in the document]
xenowits: okay cool

cmd/createcluster.go
```diff
@@ -386,43 +401,128 @@ func getValidators(dvs []tbls.TSS) ([]cluster.DistValidator, error) {
...
+			}
+			secrets = append(secrets, secret)
 		}
 
-		peers = append(peers, peer)
+		retries := retriesCount
+		addr := addrs[i]
+		grp.Go(func() error {
+			for retries > 0 {
+				err := postKeymanager(ctx, addr, secrets)
```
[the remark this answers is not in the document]
xenowits: Okay, let me log an error when an error occurs and a debug log when the post is successful

eth2util/keystore/keystore.go
xenowits: By library, you mean the `keystore` package here ?

cmd/createcluster.go
corverroos: you do this check twice in the same function, "len(conf.KeymanagerAddrs) != 0", rather create a local variable in the function and use that variable

cmd/createcluster.go
```diff
@@ -386,43 +400,86 @@ func getValidators(dvs []tbls.TSS) ([]cluster.DistValidator, error) {
 	return vals, nil
 }
 
-// getOperators returns a list of operators from the provided secret keyshares.
-func getOperators(numNodes int, clusterDir string, insecureKeys bool, shareSets [][]*bls_sig.SecretKeyShare) ([]cluster.Operator, error) {
-	var peers []p2p.Peer
+// saveKeysToKeyManager writes validator keys to the provided keymanager addresses.
```
corverroos, line 403: comment not correct

cmd/createcluster.go
```diff
@@ -386,43 +400,86 @@ func getValidators(dvs []tbls.TSS) ([]cluster.DistValidator, error) {
...
+			if err != nil {
+				return err
+			}
+			secrets = append(secrets, secret)
 		}
 
-		peers = append(peers, peer)
+		addr := addrs[i]
+		eg.Go(func() error {
+			err := keystore.PostKeysToKeymanager(ctx, addr, secrets)
```
corverroos, line 418: is this is slow call? why even do it concurrently? feels like a premature optimisation, rather just do sequentially.

eth2util/keystore/keystore.go
corverroos: yes

eth2util/keystore/keystore.go
```diff
@@ -62,6 +69,93 @@ func StoreKeys(secrets []*bls_sig.SecretKey, dir string) error {
...
+
+	if resp.StatusCode/100 != 2 {
+		return errors.New("failed posting keys", z.Int("status", resp.StatusCode), z.Str("body", string(data)))
+	}
+
+	return nil
+}
+
+// pingAddress returns an error if the provided address is not reachable.
+func pingAddress(addr string) error {
```
corverroos, line 144: this should not be in the keystore package...

eth2util/keystore/keystore.go
```diff
@@ -62,6 +69,93 @@ func StoreKeys(secrets []*bls_sig.SecretKey, dir string) error {
...
+	u, err := url.Parse(addr)
+	if err != nil {
+		return errors.Wrap(err, "parse address")
+	}
+
+	conn, err := net.DialTimeout("tcp", u.Host, 2*time.Second)
```
corverroos, line 150: always use context when doing IO

eth2util/keystore/keystore.go
xenowits: moved to createcluster.go (`cmd` package)

cmd/createcluster.go
xenowits: done

cmd/createcluster.go
```diff
@@ -628,3 +727,23 @@ func safeThreshold(ctx context.Context, numNodes, threshold int) int {
 
 	return threshold
 }
+
+// pingAddress returns an error if the provided HTTP address is not reachable.
+func pingAddress(ctx context.Context, addr string) error {
```
corverroos, line 732: `verifyConnection`

eth2util/keystore/keystore.go
```diff
@@ -62,6 +62,33 @@ func StoreKeys(secrets []*bls_sig.SecretKey, dir string) error {
...
+// KeymanagerReqBody constructs a KeymanagerReq using the provided secrets and returns it.
+func KeymanagerReqBody(secrets []*bls_sig.SecretKey) (KeymanagerReq, error) {
+	var resp KeymanagerReq
+	for _, secret := range secrets {
+		password, err := randomHex32()
+		if err != nil {
+			return KeymanagerReq{}, err
+		}
+
+		store, err := encrypt(secret, password, rand.Reader)
```
corverroos, line 80: On second thought, rather just export Encrypt, and do this API stuff in cmd package, this shouldn't be part of this package.

corverroos (approved):

eth2util/keystore/keystore.go
xenowits: Agreed, will do this refactor in a separate PR

Issue #1478, opened 2022-11-23
Incorrect failed duty reason when deadline reached - tracker v2


corverroos:
## Problem to be solved

Failed duty reasons doesn't support duties that expire (reach their deadline). It assumes something in the component errored, while it was the duty context. This results in correct but concerning logs, e.g.:

```
19:13:55.908 [35mDEBG[0m [32mvapi [0m Partial signed data stored {"count": 6, "pubkey": "b56_ec9", "duty": "6030218/attester", "vapi_endpoint": "submit_attestations"}
19:13:55.908 [35mDEBG[0m [32mparsigex [0m Partial signed data stored {"count": 7, "pubkey": "b56_ec9", "duty": "6030218/attester", "peer": "frantic-road"}
19:13:57.504 [35mDEBG[0m [32mparsigex [0m Threshold aggregated partial signatures {"duty": "6030218/attester", "peer": "lazy-room"}
19:13:57.507 [31mERRO[0m [32mparsigex [0m Subscribe error: context deadline exceeded {"duty": "6030218/attester", "peer": "lazy-room"}
19:14:20.003 [33mWARN[0m [32mtracker [0m Duty failed {"step": "sig_aggregation", "reason": "bug: threshold aggregation of partial signatures failed due to inconsistent signed data", "duty": "6030218/attester"}
```

## Proposed solution

Refactor tracker into tracker2 which will receive events by "wrapping component inputs" rather than "listening to component outputs".  This way if component returns any error then it can be captured and analysed by tracker2. Make sure to add some time duration to analyser and deleter of tracker2 so that it does not clash with duty expired events.

corverroos:
Best way to do this is probably to refactor tracker events from being "listen to component outputs" to "wrapping component inputs" since then it can detect the errors that components return and check what error it is.

thomasheremans:
Need to create further tickets

PR #1623, opened 2023-01-12
core/tracker: wrap components' input to send events to tracker

dB2510:
It is an early PR to propose solution to wrap core component inputs to improve detection of failed duties.

category: feature
ticket: #1478

### DIFF
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -173,6 +173,11 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
+// Tracker sends core component events for further analysis and instrumentation.
+type Tracker interface {
+	SendEvent(context.Context, Duty, string, PubKey, error, ParSignedData) error
+}
+
 // wireFuncs defines the core workflow components as a list of input and output functions
 // instead as interfaces, since functions are easier to wrap than interfaces.
 type wireFuncs struct {
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -175,7 +175,8 @@ type Broadcaster interface {
 
 // Tracker sends core component events for further analysis and instrumentation.
 type Tracker interface {
-	SendEvent(context.Context, Duty, string, PubKey, error, ParSignedData) error
+	// Fetcher sends fetcher component's events to tracker.
+	Fetcher(context.Context, Duty, PubKey, error)
 }
 
 // wireFuncs defines the core workflow components as a list of input and output functions
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -175,8 +175,20 @@ type Broadcaster interface {
 
 // Tracker sends core component events for further analysis and instrumentation.
 type Tracker interface {
-	// Fetcher sends fetcher component's events to tracker.
-	Fetcher(context.Context, Duty, PubKey, error)
+	// Fetcher sends Fetcher component's events to tracker.
+	Fetcher(context.Context, Duty, DutyDefinitionSet, error)
+
+	// DutyDBStore sends DutyDB component's store events to tracker.
+	DutyDBStore(context.Context, Duty, UnsignedDataSet, error)
+
+	// ParSigDBStoreInternal sends ParSigDB component's store internal events to tracker.
+	ParSigDBStoreInternal(context.Context, Duty, ParSignedDataSet, error)
+
+	// ParSigDBStoreExternal sends ParSigDB component's store external events to tracker.
+	ParSigDBStoreExternal(context.Context, Duty, ParSignedDataSet, error)
+
+	// BroadcasterBroadcast sends Broadcaster component's broadcast events to tracker.
+	BroadcasterBroadcast(context.Context, Duty, PubKey, SignedData, error)
 }
 
 // wireFuncs defines the core workflow components as a list of input and output functions
--- a/core/tracker/tracker.go
+++ b/core/tracker/tracker.go
@@ -33,11 +33,14 @@ const (
 	scheduler              // Duty scheduled with definition
 	fetcher                // Duty data fetched
 	consensus              // Duty data consensus reached
+	dutyDBStore            // Duty data stored in DutyDB
 	validatorAPI           // Partial signed data from local VC submitted to vapi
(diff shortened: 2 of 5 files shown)


core/tracking.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+// WithTracking wraps component input functions to support tracking of core components.
+func WithTracking(tracker Tracker) WireOption {
+	return func(w *wireFuncs) {
+		clone := *w
+
+		w.FetcherFetch = func(ctx context.Context, duty Duty, set DutyDefinitionSet) error {
```
dB2510, line 30: other component inputs can be added here

core/tracking.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"context"
+
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+// WithTracking wraps component input functions to support tracking of core components.
+func WithTracking(tracker Tracker) WireOption {
```
dB2510, line 26: This isn't integrated with core workflow

core/interfaces.go
```diff
@@ -173,6 +173,11 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
+// Tracker sends core component events for further analysis and instrumentation.
+type Tracker interface {
+	SendEvent(context.Context, Duty, string, PubKey, error, ParSignedData) error
```
corverroos, line 178: We'll need interface with methods for each event type, cannot use a common interface for all.

core/tracking.go
```diff
@@ -0,0 +1,44 @@
...
+	return func(w *wireFuncs) {
+		clone := *w
+
+		w.FetcherFetch = func(ctx context.Context, duty Duty, set DutyDefinitionSet) error {
+			fetchErr := clone.FetcherFetch(ctx, duty, set)
+			defer func() {
```
corverroos, line 32: no point in using defer, just call it after the clone

core/tracking.go
```diff
@@ -0,0 +1,44 @@
...
+				for pubkey := range set {
+					err := tracker.SendEvent(ctx, duty, "fetcher", pubkey, fetchErr, ParSignedData{})
```
corverroos, line 34: should probably change this not to return an error

core/tracking.go
corverroos, line 26: component input wrapping order is very important, what should the order be between retryer, tracing and tracking?

core/interfaces.go
corverroos: error should be last, and do not return an error as well

core/tracking.go
dB2510: after tracing and before retryer:
```
core.Wire(sched, fetch, cons, dutyDB, vapi,
		parSigDB, parSigEx, sigAgg, aggSigDB, broadcaster,
		core.WithTracing(),
		core.WithTracking(tracker),
		core.WithAsyncRetry(retryer),
	)
```

core/interfaces.go
dB2510: Why do we need methods for each type?
Since tracker events mostly takes duty, component name and pubkey for failed duty analysis. And it takes ParSignedData from parsigdb and parsigex components to analyse participation.

dB2510: Understood!

core/tracking.go
corverroos: retryer is async so it always return nil error immediately, so it has to "after" it. But retryer might retry, so it will result in multiple events being added to tracker, one for each retry.

core/tracking.go
```diff
@@ -0,0 +1,36 @@
...
+
+// WithTracking wraps component input functions to support tracking of core components.
+func WithTracking(tracker Tracker) WireOption {
+	return func(w *wireFuncs) {
+		clone := *w
+
+		w.FetcherFetch = func(ctx context.Context, duty Duty, set DutyDefinitionSet) error {
+			fetchErr := clone.FetcherFetch(ctx, duty, set)
+			for pubkey := range set {
+				tracker.Fetcher(ctx, duty, pubkey, fetchErr)
```
corverroos, line 30: what not pass in the whole set?

core/tracking.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+)
+
+// WithTracking wraps component input functions to support tracking of core components.
+func WithTracking(tracker Tracker) WireOption {
+	return func(w *wireFuncs) {
+		clone := *w
+
+		w.FetcherFetch = func(ctx context.Context, duty Duty, set DutyDefinitionSet) error {
+			fetchErr := clone.FetcherFetch(ctx, duty, set)
```
corverroos, line 28: just call this `err`

core/tracking.go
dB2510: maybe we can dedup events added to tracker by component/step per duty?

corverroos: not sure how that is going to work?

core/interfaces.go
```diff
@@ -173,6 +173,24 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
+// Tracker sends core component events for further analysis and instrumentation.
+type Tracker interface {
+	// Fetcher sends Fetcher component's events to tracker.
+	Fetcher(context.Context, Duty, DutyDefinitionSet, error)
```
corverroos, line 179: shouldn't this be FetcherFetch?

core/tracker/tracker.go
```diff
@@ -449,7 +461,7 @@ func analyseFetcherFailed(duty core.Duty, allEvents map[core.Duty][]event) (bool
 		}
 
 		// Sync contribution duties will fail if no sync message in DutyDB.
-		syncMsgFailed, syncMsgStep := dutyFailedStep(allEvents[core.NewSyncMessageDuty(duty.Slot)])
+		syncMsgFailed, syncMsgStep, _ := dutyFailedStep(allEvents[core.NewSyncMessageDuty(duty.Slot)])
```
corverroos, line 464: ignore errors is bad

Issue #1608, opened 2023-01-09
Port ENR package


corverroos:
## Problem to be solved

We want to remove dependency on GPL licensing. We are removing discv5, that leaves only the ENR package in geth.

## Proposed solution

Re-write only the required functionality of the ENR package into `charon/app/enr`. 

The functionality required is:
 - Parse previously generated ENRs and extract peer ID.
 - Encode valid ENR only containing a peer ID.

### Out of Scope

We will not include the ability to encode or decode IP addresses or ports from an ENR. Nor timestamps or sequence numbers.

PR #1624, opened 2023-01-12
eth2util/rlp: implement basic rlp encoding

corverroos:
Implements basic rlp encoding/decoding of simple byte slices and lists of bytes slices from the spec.

category: feature
ticket: #1608


eth2util/rlp/rlp_test.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+		{
+			input:  []byte(""),
+			output: []byte{0x80},
+		},
+		{
+			input:  nil,
+			output: []byte{0x80},
+		},
+		// Not sure how to represent "the integer 0" resulting in "[]byte{0x80}"
```
xenowits, line 48: This might help?..

Screenshot 2023-01-12 at 19 56 45

xenowits, line 48: This might help?..

Screenshot 2023-01-12 at 19 56 45

eth2util/rlp/rlp.go
```diff
@@ -0,0 +1,192 @@
...
+
+	length, err := encodeLength(len(item), 0x80)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(length, item...), nil
+}
+
+// decodeLength returns the length of the RLP encoding prefix (offset) and the length of the encoded byte slice.
```
xenowits, line 117: any specific reason to have named return values?

xenowits (approved):
LGTM 🚀

eth2util/rlp/rlp.go
corverroos: since it isn't obvious from the types what is returned

e.g., what is the first in and what is the second int?
```
func decodeLength(item []byte) (int, int, error)
```

PR #1636, opened 2023-01-16
cmd: update default relay address

corverroos:
Update default relay address.

category: refactor
ticket: none


xenowits (approved):

OisinKyne (approved):
Not happy that the internal vars are called UDPBootnode but I'm okay with it for now to test migration, we can clean the rest up when we fully remove discv5 references.... I hope.

Issue #1632, opened 2023-01-16
Relay to serve multiaddrs instead of ENR


corverroos:
## Problem to be solved

As we are moving away from discv5, we should refactor the relay to serve multiaddrs instead of an ENR. This makes is more native libp2p based.

## Proposed solution

Serve the `tcphost.Addrs()` as json array at `/` along with current ENR in relays. 
Charon to support reading ENR and multiaddrs when detecting bootnode.

### Out of Scope

Can remove support for ENRs at a later point.

PR #1646, opened 2023-01-17
cmd/relay: serve multiaddrs from http address

corverroos:
Relay serves multiaddrs from http-address server.

category: feature
ticket: #1632

### DIFF
--- a/cmd/relay/relay_internal_test.go
+++ b/cmd/relay/relay_internal_test.go
@@ -17,10 +17,16 @@ package relay
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
 	"os"
 	"testing"
+	"time"
 
+	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/p2p"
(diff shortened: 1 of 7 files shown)


p2p/p2p.go
```diff
@@ -41,7 +41,7 @@ import (
 )
 
 // NewTCPNode returns a started tcp-based libp2p host.
-func NewTCPNode(ctx context.Context, cfg Config, key *ecdsa.PrivateKey, connGater ConnGater, opts ...libp2p.Option,
+func NewTCPNode(ctx context.Context, cfg Config, key *ecdsa.PrivateKey, connGater ConnGater, advertise bool, opts ...libp2p.Option,
```
xenowits, line 44: what is the `advertise` parameter used for? maybe add to the godoc of this function

xenowits (approved):

p2p/p2p.go
corverroos: it will be removed soon

cmd/relay/relay_internal_test.go
```diff
@@ -72,3 +78,60 @@ func TestRunBootnodeAutoP2P(t *testing.T) {
 	testutil.SkipIfBindErr(t, err)
 	require.NoError(t, err)
 }
+
+func TestServeMultiAddrs(t *testing.T) {
```
dB2510, line 82: can we have `testutil.RequireGoldenJSON(...)` for this test for json response of http call?

Issue #1625, opened 2023-01-13
DKG smoke test shutdown race condition


corverroos:
## Problem to be solved

When DKG instances shutdown as part of compose, there is a race where the first DKG completes, which causes the container the exit, which causes the docker-compose to stop the other container, which can result in other DKG not completing properly by saving the artefacts to disk, see [example](https://github.com/ObolNetwork/charon/actions/runs/3908724620). 

## Proposed solution

- Docker-compose shouldn't terminate all containers when the first exits with 0, but rather wait for manual "down" command.
- Add a hidden "wait for something signal" before exitting the DKG...?

corverroos:
[dkg.log](https://github.com/ObolNetwork/charon/files/10408684/dkg.log)

PR #1647, opened 2023-01-17
testutil/compose: fix dkg shutdown race

corverroos:
Sleep for 2 seconds after successful DKG completion to allow other DKGs to complete as well before exitting causing docker-compose to stop all other containers. This fixes shutdown race in compose DKG tests.

category: test
ticket: #1625

### DIFF
--- a/testutil/compose/lock.go
+++ b/testutil/compose/lock.go
@@ -66,7 +66,11 @@ func Lock(ctx context.Context, dir string, conf Config) (TmplData, error) {
 
 		var nodes []TmplNode
 		for i := 0; i < conf.NumNodes; i++ {
-			n := TmplNode{EnvVars: newNodeEnvs(i, conf, "")}
+			n := TmplNode{
+				EnvVars:    newNodeEnvs(i, conf, ""),
+				Entrypoint: "sh",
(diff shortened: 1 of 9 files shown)


testutil/compose/lock.go
```diff
@@ -66,20 +66,24 @@ func Lock(ctx context.Context, dir string, conf Config) (TmplData, error) {
 
 		var nodes []TmplNode
 		for i := 0; i < conf.NumNodes; i++ {
-			n := TmplNode{EnvVars: newNodeEnvs(i, conf, "")}
+			n := TmplNode{
+				EnvVars:    newNodeEnvs(i, conf, ""),
+				Entrypoint: "sh",
+				Command:    fmt.Sprintf("[-c,'%s %s && sleep 2']", conf.entrypoint(), cmdDKG), // Sleep after completion to allow other nodes to finish
```
xenowits, line 72:
```suggestion
				Command:    fmt.Sprintf("[-c,'%s %s && sleep 2']", conf.entrypoint(), cmdDKG), // Sleep for 2s after completion to allow other nodes to finish
```

xenowits (approved):

testutil/compose/lock.go
corverroos: I prefer not to add details like that so that the risk of the comment becoming out of data/out of sync is lower.

And remember: comments should say why the code does something, not what the is doing.

Issue #1502, opened 2022-12-02
Support Keymanager API


OisinKyne:
## Problem to be solved

Obol creates BLS private key shares during the `charon dkg` command and the `charon create cluster` commands. 

Right now we export them as password protected EIP-2335 format files. This mimics the eth2depositcli approach. 

Some/going on many operators run PKI that supports the Ethereum Keymanager APIs, web3signer in particular. It is commonly used in client switching tools like on dappnode and other launchers. 

It would automate an important process if we could automatically push BLS private keys straight into a web3 signer or validator client running the keymanager api once they are generated, we could potentially even never persist them to disk if people wanted

## Proposed solution

This is the epic that will track adding support for the [keymanager API](https://github.com/ethereum/keymanager-APIs).

- [ ] Add `--keymanager-address` as a flag to `create cluster`
- [ ] Add `--keymanager-address` as a flag to `charon dkg`
- [ ] Document keymanager configuration on the docs site

A design decision that will need to be made (and well error tested for) is what happens when writing to an endpoint fails? We can retry like 5 times maybe before failing outright, but do we fallback to writing to disk? To we check/require access for both in the pre-flight checks? 

### Out of Scope

Out of scope is implementing this in any customer facing template repo. The goal is to use this with dappnode and avado and others in their integration efforts.

PR #1657, opened 2023-01-19
keymanager: api design

xenowits:
Introduces a new package `eth2util/keymanager` for keymanager api functionalities.

category: feature
ticket: https://github.com/ObolNetwork/charon/issues/1502

### DIFF
--- a/eth2util/keymanager/keymanager.go
+++ b/eth2util/keymanager/keymanager.go
@@ -13,6 +13,7 @@
 // You should have received a copy of the GNU General Public License along with
 // this program.  If not, see <http://www.gnu.org/licenses/>.
 
+// Package keymanager provides ETH2 keymanager API (https://ethereum.github.io/keymanager-APIs/) functionalities.
 package keymanager
 
 import (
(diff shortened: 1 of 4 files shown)


eth2util/keymanager/keymanager.go
```diff
@@ -0,0 +1,125 @@
...
+
+// New returns a new Keymanager.
+func New(url string) Keymanager {
+	return Keymanager{
+		baseURL: url,
+	}
+}
+
+// Keymanager is the REST client for keymanager API requests.
+type Keymanager struct {
```
corverroos, line 42: `Client`

eth2util/keymanager/keymanager.go
```diff
@@ -0,0 +1,125 @@
...
+	baseURL string // Base keymanager URL
+}
+
+// ImportKeystores pushes the keystores to keymanager.
+// See https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager/importKeystores.
+func (km Keymanager) ImportKeystores(ctx context.Context, keystores []keystore.Keystore, passwords []string) error {
```
corverroos, line 48: verify len(keystores) == len(passwords)

corverroos (approved):
nice, add tests here as well please

eth2util/keymanager/keymanager.go
```diff
@@ -0,0 +1,131 @@
...
+
+// New returns a new Client.
+func New(url string) Client {
+	return Client{
+		baseURL: url,
+	}
+}
+
+// Client is the REST client for keymanager API requests.
+type Client struct {
```
dB2510, line 42: Why are we having a struct with only one field? why can we not simple pass baseURL as a function parameter and have pure functions?

xenowits: We want to introduce a REST API client for [keymanager APIs](https://ethereum.github.io/keymanager-APIs/) just like we have go-eth2-client for ethereum beacon APIs.

PR #1666, opened 2023-01-19
core/tracker2: add tracker2 package

dB2510:
Adds refactored tracker logic as tracker2 package in `core/tracker` package.

category: refactor
ticket: #1478

### DIFF
--- a/core/tracker/tracker2/metrics.go
+++ b/core/tracker/tracker2/metrics.go
@@ -56,11 +56,4 @@ var (
 		Name:      "inconsistent_parsigs_total",
 		Help:      "Total number of duties that contained inconsistent partial signed data by duty type",
 	}, []string{"duty"})
-
-	inclusionDelay = promauto.NewGauge(prometheus.GaugeOpts{
-		Namespace: "core",
-		Subsystem: "tracker",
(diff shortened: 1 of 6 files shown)


core/tracker/tracker2/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+package tracker2
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/obolnetwork/charon/app/promauto"
+)
+
+var (
+	participationGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
corverroos, line 25: can create duplicate metrics, suggest only moving metrics here once they are "enabled" in v2.

corverroos (approved):

core/tracker/tracker2/metrics.go
dB2510: Added `tracker2` in metrics naming to avoid duplication, v2 will be behind a featureflag until we are confident. These metrics were needed since I copy pasted most of the code and metrics was also part of it.
Will address this in my next PR to wire and add tests.

Issue #1504, opened 2022-12-02
Add `--keymanager-address` as a flag to `charon dkg`


OisinKyne:
## Problem to be Solved

We want to create a DV cluster with a DKG, and rather than pushing the created BLS private key shares and their passwords to disk; we want to push the private keys into our existing private key infrastructure via the [keymanager api](https://ethereum.github.io/keymanager-APIs/). 

## Proposed Solution

Add a flag `--keymanager-address` that takes a well formed URI (preferably able to handle TLS, basic auth, and query paramaters containing auth codes) and sends a POST request to [this endpoint](https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager/importKeystores). It should have a 10 second time out and a max retrys of 5, and should log the error status code it receives and that it is retrying to console in a legible manner. 

When the DKG process starts with this flag on, in the pre-reach out phase of starting up, we should try and GET request the corresponding keymanager read endpoint and assert that we get a 200 response. If we get 404 or an unexpected error we should fail before we get to the `Connecting to peers` stage. 

As discussed in the #1503 ticket, deciding what to do if and when a keymanager write fails post DKG is tricky, in this context we can't abort the whole thing. At the same time, I don't think it is safe to fall back and write it to disk as plan B, because that is weird and hard to test for and dual purpose, and if I was an attacker with access to host machines but not the secure PKI, that would be how I try and exfiltrate keys. That said, I think we just exit the process with a non zero return code to indicate there was an error in the command, and hopefully by inferring node health by the presence of GET requests to the validator API on the charon side, we alert very quickly that something went wrong when the nodes start up too.

## Out of scope

Deploying this in a public facing template repo.

PR #1671, opened 2023-01-23
dkg: add keymanager support

xenowits:
Adds support for keymanager API to `charon dkg` command.

category: feature
ticket: #1504

### DIFF
--- a/cmd/dkg.go
+++ b/cmd/dkg.go
@@ -57,7 +57,7 @@ this command at the same time.`,
 }
 
 func bindKeymanagerAddrFlag(flags *pflag.FlagSet, addr *string) {
-	flags.StringVar(addr, "keymanager-address", "", "The keymanager URL to push validator keyshares to.")
+	flags.StringVar(addr, "keymanager-address", "", "The keymanager URL to import validator keyshares.")
 }
 
 func bindDefDirFlag(flags *pflag.FlagSet, dataDir *string) {
--- a/cmd/dkg.go
+++ b/cmd/dkg.go
@@ -47,12 +47,17 @@ this command at the same time.`,
 	}
 
 	bindDataDirFlag(cmd.Flags(), &config.DataDir)
+	bindKeymanagerAddrFlag(cmd.Flags(), &config.KeymanagerAddr)
 	bindDefDirFlag(cmd.Flags(), &config.DefFile)
 	bindNoVerifyFlag(cmd.Flags(), &config.NoVerify)
 	bindP2PFlags(cmd, &config.P2P)
 	bindLogFlags(cmd.Flags(), &config.Log)
 
 	return cmd
+}
+
+func bindKeymanagerAddrFlag(flags *pflag.FlagSet, addr *string) {
+	flags.StringVar(addr, "keymanager-address", "", "The keymanager URL to push validator keyshares to.")
 }
 
 func bindDefDirFlag(flags *pflag.FlagSet, dataDir *string) {
--- a/dkg/disk.go
+++ b/dkg/disk.go
@@ -17,6 +17,7 @@ package dkg
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"net/url"
@@ -33,8 +34,10 @@ import (
 	"github.com/obolnetwork/charon/cluster"
 	"github.com/obolnetwork/charon/core"
 	"github.com/obolnetwork/charon/eth2util/deposit"
+	"github.com/obolnetwork/charon/eth2util/keymanager"
 	"github.com/obolnetwork/charon/eth2util/keystore"
 	"github.com/obolnetwork/charon/tbls/tblsconv"
+	"github.com/obolnetwork/charon/testutil"
 )
 
 // loadDefinition returns the cluster definition from disk or an HTTP URL. It returns the test definition if configured.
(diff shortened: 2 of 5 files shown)


dkg/disk.go
```diff
@@ -95,8 +98,43 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
...
-func writeKeystores(datadir string, shares []share) error {
+// writeKeysToKeymanager writes validator private keyshares for the node to the provided keymanager address.
+func writeKeysToKeymanager(ctx context.Context, keymanagerURL string, shares []share) error {
+	var (
+		keystores []keystore.Keystore
+		passwords []string
+	)
+
+	for _, s := range shares {
+		password, err := testutil.RandomHex32()
```
corverroos, line 109: can't use testutil in production code, also make sure you not using math.Random, but crypto.Random.

dkg/dkg.go
```diff
@@ -73,6 +75,14 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	// Check if keymanager address is reachable.
+	if conf.KeymanagerAddr != "" {
+		cl := keymanager.New(conf.KeymanagerAddr)
+		if err = cl.VerifyConnection(ctx); err != nil {
+			return err
```
corverroos, line 82: what does this error look like to users? please paste a screenshot

dkg/dkg.go
```diff
@@ -191,10 +201,17 @@ func Run(ctx context.Context, conf Config) (err error) {
 	// Write keystores, deposit data and cluster lock files after exchange of partial signatures in order
 	// to prevent partial data writes in case of peer connection lost
 
-	if err := writeKeystores(conf.DataDir, shares); err != nil {
-		return err
+	if conf.KeymanagerAddr != "" { // Save to keymanager
+		if err = writeKeysToKeymanager(ctx, conf.KeymanagerAddr, shares); err != nil {
```
corverroos, line 205: what happens if keys have already been pushed in a previous DKG attempt?

dkg/dkg.go
```diff
@@ -191,10 +201,17 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+			return err
+		}
+		log.Debug(ctx, "Saved keyshares to keymanager")
```
corverroos, line 208: add URL ass param, suggest "import" as verb as that is what the keymannager API uses.

cmd/dkg.go
```diff
@@ -55,6 +56,10 @@ this command at the same time.`,
 	return cmd
 }
 
+func bindKeymanagerAddrFlag(flags *pflag.FlagSet, addr *string) {
+	flags.StringVar(addr, "keymanager-address", "", "The keymanager URL to push validator keyshares to.")
```
corverroos, line 60: suggest "import" as verb as that what keymanager uses.

dkg/dkg_test.go
```diff
@@ -405,3 +408,134 @@ func startNewDKG(t *testing.T, parentCtx context.Context, config dkg.Config, dkg
...
+	require.NoError(t, def.VerifySignatures())
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Setup
+	dir, err := os.MkdirTemp("", "")
+	require.NoError(t, err)
+
+	allReceived := make(chan struct{}, 1) // Receives struct{} if all `numNodes` keystores are intercepted by the keymanager server
```
corverroos, line 427: why buffer of 1? unbuffered is default

dkg/dkg_test.go
```diff
@@ -405,3 +408,134 @@ func startNewDKG(t *testing.T, parentCtx context.Context, config dkg.Config, dkg
 
 	return cancel
 }
+
+// TestKeymanager tests the keymanager functionality where the dkg command pushes validator keyshares
+// to a keymanager API url.
+func TestKeymanager(t *testing.T) {
```
corverroos, line 414: this feels like a duplicate test, can't we add this to existing table test?

dkg/dkg_test.go
```diff
@@ -405,3 +408,134 @@ func startNewDKG(t *testing.T, parentCtx context.Context, config dkg.Config, dkg
...
+	Passwords []string       `json:"passwords"`
+}
+
+// noopKeystore is a mock keystore.keystore for use in tests.
+type noopKeystore struct {
+	Crypto map[string]interface{} `json:"crypto"`
+}
+
+// decrypt returns the secret from the encrypted keystore.
+func decrypt(t *testing.T, store noopKeystore, password string) (*bls_sig.SecretKey, error) {
```
corverroos, line 505: this feels like testing the keymanager client, which isn't required, you test that implementation in the keymanager package, this test is for DKG, testing that is submits something is fine.

corverroos (approved):

dkg/disk.go
xenowits: gotcha. The random function's using crypto/rand though it shouldn't be part of the testutil package. Will refactor this out into `eth2util` in a separate PR

dkg/dkg.go
xenowits: Screenshot 2023-01-24 at 15 29 16

Should we have `cannot ping keymanager address` in error log?

dkg/dkg.go
xenowits: A new DKG attempt creates fresh keys so new keys are pushed to keymanager and previously pushed keys would not be overwritten (are not affected).

dkg/dkg.go
xenowits: Screenshot 2023-01-24 at 16 00 32

dkg/dkg_test.go
xenowits: yeah we can, doing this now

dkg/dkg_test.go
xenowits: cool, removing this. I'm now only ensuring that the request body received by the server is non-empty

dkg/dkg_test.go
xenowits: i initially intended to make it unblocking, but i have updated the tests and made this unbuffered

dkg/dkg.go
corverroos: suggest: `return errors.Wrap(err, "verify keymanager address"`

dkg/dkg_test.go
```diff
@@ -405,3 +440,35 @@ func startNewDKG(t *testing.T, parentCtx context.Context, config dkg.Config, dkg
...
+		mu.Unlock()
+
+		t.Logf("Received keystore: %d/%d\n", cnt, numNodes)
+		if cnt == numNodes {
+			go func() {
+				allReceived <- struct{}{}
+			}()
+		}
+
+		w.WriteHeader(http.StatusOK)
```
corverroos, line 472: no need for this, can just remove

dkg/dkg_test.go
```diff
@@ -90,6 +113,14 @@ func testDKG(t *testing.T, def cluster.Definition, p2pKeys []*ecdsa.PrivateKey)
 		TestDef: &def,
 	}
 
+	allReceived := make(chan struct{}) // Receives struct{} if all `numNodes` keystores are intercepted by the keymanager server
+	if keymanager {
+		srv := httptest.NewServer(newKeymanagerHandler(t, len(def.Operators), allReceived))
```
corverroos, line 118: I would just inline a very simple handler:
```
httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    receives <- struct{}
  }
}
```
and then do below:
```
for expectReceives > 0 {
   <-receives
   expectReceives--
}
```
```

File: dkg/disk.go [hunk 11]
```diff
@@ -199,3 +237,14 @@ func validURI(str string) bool {
 
 	return err == nil && (u.Scheme == "http" || u.Scheme == "https") && u.Host != ""
 }
+
+// randomHex32 returns a random 32 character hex string. It uses crypto/rand.
+func randomHex32() (string, error) {
```
corverroos, line 242: need to do this for createcluster as well

corverroos (approved):

PR #1672, opened 2023-01-23
cmd/relay: serve enr

corverroos:
Reintroduces the ability for the relay to serve its ID and address via an ENR for backwards compatibility to v0.13 charon nodes.

category: bug
ticket: none

### DIFF
--- a/app/monitoringapi_internal_test.go
+++ b/app/monitoringapi_internal_test.go
@@ -136,13 +136,11 @@ func TestStartChecker(t *testing.T) {
 			advanceClock(clock, 10*time.Second, 2)
 
 			// Advance clock for first epoch tick.
-			advanceClock(clock, 384*time.Second, 2)
-
-			// Advance clock for last tick.
-			advanceClock(clock, 10*time.Second, 2)
+			advanceClock(clock, 32*12*time.Second, 2)
 
 			if tt.err != nil {
 				require.Eventually(t, func() bool {
+					advanceClock(clock, 10*time.Second, 2)
 					err = readyErrFunc()
 					if !errors.Is(err, tt.err) {
 						t.Logf("Ignoring unexpected error, got=%v, want=%v", err, tt.err)
@@ -153,6 +151,7 @@ func TestStartChecker(t *testing.T) {
 				}, time.Second, 100*time.Millisecond)
 			} else {
 				require.Eventually(t, func() bool {
+					advanceClock(clock, 12*time.Second, 2)
 					return readyErrFunc() == nil
 				}, time.Second, 100*time.Millisecond)
 			}
--- a/cmd/relay/relay.go
+++ b/cmd/relay/relay.go
@@ -156,9 +156,9 @@ func Run(ctx context.Context, config Config) error {
 	}
 }
 
-// newENRHandler returns a handler that returns the nodes ID and public address encoded as a ENR.
+// newENRHandler returns a handler that returns the node's ID and public address encoded as a ENR.
 func newENRHandler(ctx context.Context, tcpNode host.Host, p2pKey *ecdsa.PrivateKey, config p2p.Config) func(ctx context.Context) ([]byte, error) {
-	// Resolve external hostname peridodically.
+	// Resolve external hostname periodically.
 	var (
 		extHostMu sync.Mutex
 		extHostIP net.IP
@@ -226,7 +226,7 @@ func newENRHandler(ctx context.Context, tcpNode host.Host, p2pKey *ecdsa.Private
 		}
 		tcpAddr, ok := addr.(*net.TCPAddr)
 		if !ok {
-			return nil, errors.New("not a TCP address")
+			return nil, errors.New("invalid TCP address")
 		}
 
 		// Override IP with external IP or external hostname if set.
--- a/cmd/relay/relay.go
+++ b/cmd/relay/relay.go
@@ -17,14 +17,20 @@ package relay
 
 import (
 	"context"
+	"crypto/ecdsa"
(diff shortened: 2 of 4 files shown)


app/monitoringapi_internal_test.go
```diff
@@ -136,13 +136,11 @@ func TestStartChecker(t *testing.T) {
...
 			// Advance clock for first epoch tick.
-			advanceClock(clock, 384*time.Second, 2)
-
-			// Advance clock for last tick.
-			advanceClock(clock, 10*time.Second, 2)
+			advanceClock(clock, 32*12*time.Second, 2)
 
 			if tt.err != nil {
 				require.Eventually(t, func() bool {
+					advanceClock(clock, 10*time.Second, 2)
```
corverroos, line 143: @dB2510 this was flapping. When retrying, you need to advance the clock, otherwise nothing happens and retries do nothing always returning the same result.

cmd/relay/relay.go
```diff
@@ -169,3 +155,132 @@ func Run(ctx context.Context, config Config) error {
 		}
 	}
 }
+
+// newENRHandler returns a handler that returns the nodes ID and public address encoded as a ENR.
+func newENRHandler(ctx context.Context, tcpNode host.Host, p2pKey *ecdsa.PrivateKey, config p2p.Config) func(ctx context.Context) ([]byte, error) {
+	// Resolve external hostname peridodically.
```
dB2510, line 161:
```suggestion
	// Resolve external hostname periodically.
```

cmd/relay/relay.go
```diff
@@ -169,3 +155,132 @@ func Run(ctx context.Context, config Config) error {
 		}
 	}
 }
+
+// newENRHandler returns a handler that returns the nodes ID and public address encoded as a ENR.
```
dB2510, line 159:
```suggestion
// newENRHandler returns a handler that returns the node's ID and public address encoded as a ENR.
```

dB2510 (approved):

cmd/relay/relay.go
```diff
@@ -169,3 +155,132 @@ func Run(ctx context.Context, config Config) error {
...
+		return []byte(r.String()), nil
+	}
+}
+
+// newMultiaddrHandler returns a handler that returns the nodes multiaddrs (as json array).
+func newMultiaddrHandler(tcpNode host.Host) func(ctx context.Context) ([]byte, error) {
+	return func(ctx context.Context) ([]byte, error) {
+		p2pAddr, err := ma.NewMultiaddr(fmt.Sprintf("/p2p/%s", tcpNode.ID()))
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create p2p multiaddr")
```
xenowits, line 254:
```suggestion
			return nil, errors.Wrap(err, "create p2p multiaddr")
```

cmd/relay/relay.go
```diff
@@ -169,3 +155,132 @@ func Run(ctx context.Context, config Config) error {
...
+			iPublic, jPublic := manet.IsPublicAddr(addrs[i]), manet.IsPublicAddr(addrs[j])
+			if jPublic && !iPublic {
+				return true // Only swap if j is public and i is not.
+			}
+
+			return false
+		})
+
+		// Use first address (ip and port).
+		addr, err := manet.ToNetAddr(addrs[0])
```
xenowits, line 223: discuss: why are we using only the first address?

cmd/relay/relay.go
```diff
@@ -169,3 +155,132 @@ func Run(ctx context.Context, config Config) error {
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to convert multiaddr to net addr")
+		}
+		tcpAddr, ok := addr.(*net.TCPAddr)
+		if !ok {
+			return nil, errors.New("not a TCP address")
```
xenowits, line 229: nit:
```suggestion
			return nil, errors.New("invalid TCP address")
```

cmd/relay/relay.go
xenowits, line 161:
```suggestion
	// Resolve external hostname peridodically (every 5 mins).
```

cmd/relay/relay.go
```diff
@@ -169,3 +155,132 @@ func Run(ctx context.Context, config Config) error {
...
+		for {
+			select {
+			case <-onStartup:
+				resolveExtHost()
+			case <-time.After(5 * time.Minute):
+				resolveExtHost()
+			case <-ctx.Done():
+				return
+			}
+		}
```
xenowits, line 194:
```suggestion
		resolveExtHost()

		for {
			select {
			case <-time.After(5 * time.Minute):
				resolveExtHost()
			case <-ctx.Done():
				return
			}
		}
```

xenowits: can we remove the `onStartup` channel like the above?

xenowits (approved):

cmd/relay/relay.go
corverroos: ENRs only supports 1 address

cmd/relay/relay.go
corverroos: I prefer not to put constants in comments, then stuff gets out of sync faster

cmd/relay/relay.go
corverroos: I prefer to have all scheduling in the same for/select loop

cmd/relay/relay.go
corverroos: it is an error, no need to put "failed" or "error" in every wrap.

Issue #1645, opened 2023-01-17
Add multiple withdrawal addresses to cluster-definition


dB2510:
## Problem to be solved

We want to allow our users to add different withdrawal address & fee-recipient address corresponding to each validator public key. Currently, we use `withdrawal_address`  & `fee_recipient_address` fields of cluster-definition for all the validators.

## Proposed solution

Add a `validators` nested top level field to the cluster-definition as an array of objects with two fields, `withdrawal_address`  & `fee_recipient_address`. Also remove current `withdrawal_address`  & `fee_recipient_address` fields in cluster-definition.

```
"validators": [
    {
      "fee_recipient_address": "0x52fdfc072182654f163f5f0f9a621d729566c74d",
      "withdrawal_address": "0x81855ad8681d0d86d1e91e00167939cb6694d2c4"
    },
    {
      "fee_recipient_address": "0xf5059875921e668a5bdf2c7fc4844592d2572bcd",
      "withdrawal_address": "0xa3b539eb1e5849c6077dbb5722f5717a289a266f"
    }
  ]
```
The `withdrawal_address`  & `fee_recipient_address` fields in `validators` are exactly the same as the current `withdrawal_address`  & `fee_recipient_address` fields.

This is part of new **draft** version v1.5.0.

The `validators` field will be included in `config_hash` calculation.

Check `len(definition.Validators) == definition.NumValidators`.

The ssz spec:
```
type Definition struct {
	// UUID is a human-readable random unique identifier. Max 64 chars.
	UUID string `json:"uuid" ssz:"ByteList[64]" config_hash:"0" definition_hash:"0"`

	// Name is a human-readable cosmetic identifier. Max 256 chars.
	Name string `json:"name" ssz:"ByteList[256]" config_hash:"1" definition_hash:"1"`

	// Version is the schema version of this definition. Max 16 chars.
	Version string `json:"version" ssz:"ByteList[16]" config_hash:"2" definition_hash:"2"`

	// Timestamp is the human-readable timestamp of this definition. Max 32 chars.
	// Note that this was added in v1.1.0, so may be empty for older versions.
	Timestamp string `json:"timestamp" ssz:"ByteList[32]" config_hash:"3" definition_hash:"3"`

	// NumValidators is the number of DVs (n*32ETH) to be created in the cluster lock file.
	NumValidators int `json:"num_validators" ssz:"uint64" config_hash:"4" definition_hash:"4"`

[...]

PR #1674, opened 2023-01-24
cluster: create draft 1.5 definition

LukeHackett12:
Adds a new 1.5 cluster definition version with individual fee recipients and withdrawal addresses per validator. This PR is only to add these fields, not use them for now.

category: feature
ticket: #1645


cluster/lock.go
```diff
@@ -93,6 +93,11 @@ func (l *Lock) UnmarshalJSON(data []byte) error {
 		if err != nil {
 			return err
 		}
+	case isAnyVersion(version.Definition.Version, v1_5):
```
dB2510, line 96: no need for this case since you can just add v1_5 to the case above. We haven't changed anything in cluster-lock structure

cluster/version.go
```diff
@@ -21,6 +21,7 @@ const (
 	currentVersion = v1_4
 	dkgAlgo        = "default"
 
+	v1_5 = "v1.5.0"
```
dB2510, line 24:
```suggestion
	v1_5 = "v1.5.0" // Draft
```

cluster/lock.go
corverroos: +1

corverroos (approved):

dB2510 (approved):

Issue #1651, opened 2023-01-18
Use fee-recipient addresses from cluster-definition v1.5.0


dB2510:
## Problem to be solved

We are adding support for multiple fee-recipient addresses as part of our v1.5.0 cluster definition upgrade. For more details refer [this](https://github.com/ObolNetwork/charon/issues/1645) issue. We need to add support for this in our core workflow.

## Proposed solution

We need to add support multiple fee-recipient addresses per validator public at following places:
- Update `setFeeRecipient` function in 
app.go to take multiple fee-recipient addresses as input parameter to submit proposal preparations. https://github.com/ObolNetwork/charon/blob/d7ee2681103c8896216f7a43e77c33c9b7e59eb7/app/app.go#L794
- Update validatorapi to take multiple fee-recipient addresses per validator in `NewComponent(..)` constructor and `TekuProposerConfig` method.
- Add unit tests for each change.

PR #1679, opened 2023-01-25
core: use multiple fee-recipient addresses

dB2510:
Uses multiple fee-recipient addresses from lock introduced in v1.5.

category: feature
ticket: #1651

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -400,7 +400,7 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 	sched.SubscribeSlots(setFeeRecipient(eth2Cl, eth2Pubkeys, feeRecipientFunc))
 	sched.SubscribeSlots(tracker.NewInclDelayFunc(eth2Cl, sched.GetDutyDefinition))
 
-	fetch, err := fetcher.New(eth2Cl, feeRecipientAddrByCorePubkey)
+	fetch, err := fetcher.New(eth2Cl, feeRecipientFunc)
 	if err != nil {
 		return err
 	}
--- a/app/app.go
+++ b/app/app.go
@@ -328,8 +328,8 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 		eth2Pubkeys                  []eth2p0.BLSPubKey
 		pubshares                    []eth2p0.BLSPubKey
 		allPubSharesByKey            = make(map[core.PubKey]map[int]*bls_sig.PublicKey) // map[pubkey]map[shareIdx]pubshare
-		feeRecipientAddsByEth2Pubkey = make(map[eth2p0.BLSPubKey]string)
-		feeRecipientAddsByCorePubkey = make(map[core.PubKey]string)
+		feeRecipientAddrByEth2Pubkey = make(map[eth2p0.BLSPubKey]string)
+		feeRecipientAddrByCorePubkey = make(map[core.PubKey]string)
 	)
 	for i, dv := range lock.Validators {
 		pubkey, err := dv.PublicKey()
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -317,14 +317,21 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
...
 	var (
-		corePubkeys       []core.PubKey
-		eth2Pubkeys       []eth2p0.BLSPubKey
-		pubshares         []eth2p0.BLSPubKey
-		allPubSharesByKey = make(map[core.PubKey]map[int]*bls_sig.PublicKey) // map[pubkey]map[shareIdx]pubshare
+		corePubkeys                  []core.PubKey
+		eth2Pubkeys                  []eth2p0.BLSPubKey
+		pubshares                    []eth2p0.BLSPubKey
+		allPubSharesByKey            = make(map[core.PubKey]map[int]*bls_sig.PublicKey) // map[pubkey]map[shareIdx]pubshare
+		feeRecipientAddsByEth2Pubkey = make(map[eth2p0.BLSPubKey]string)
```
xenowits, line 331:
```suggestion
		feeRecipientAddrsByEth2Pubkey = make(map[eth2p0.BLSPubKey]string)
```
similarly for `feeRecipientAddsByEth2Pubkey`

app/app.go
```diff
@@ -772,7 +775,7 @@ func wireTracing(life *lifecycle.Manager, conf Config) error {
 
 // setFeeRecipient returns a slot subscriber for scheduler which calls prepare_beacon_proposer endpoint at start of each epoch.
 // TODO(dhruv): move this somewhere else once more use-cases like this becomes clear.
-func setFeeRecipient(eth2Cl eth2wrap.Client, pubkeys []eth2p0.BLSPubKey, feeRecipient string) func(ctx context.Context, slot core.Slot) error {
+func setFeeRecipient(eth2Cl eth2wrap.Client, pubkeys []eth2p0.BLSPubKey, feeRecipientByPubkey map[eth2p0.BLSPubKey]string) func(ctx context.Context, slot core.Slot) error {
```
corverroos, line 778: you are passing in pubkeys twice, only pass it in once.

core/validatorapi/teku.go
```diff
@@ -43,10 +43,15 @@ type TekuProposerConfigProvider interface {
...
 
 func (c Component) TekuProposerConfig(ctx context.Context) (TekuProposerConfigResponse, error) {
+	var defaultFeeRecipientAddr string
+	for _, addr := range c.feeRecipientAddressByPubkey {
+		defaultFeeRecipientAddr = addr
+		break
+	}
 	resp := TekuProposerConfigResponse{
 		Proposers: make(map[string]TekuProposerConfig),
 		Default: TekuProposerConfig{ // Default doesn't make sense, disable for now.
```
corverroos, line 53: as the comment says, "Default doesn't make sense, disable for now", rather just disable and use zero values for the rest.

core/validatorapi/validatorapi.go
```diff
@@ -58,7 +58,7 @@ func NewComponentInsecure(_ *testing.T, eth2Cl eth2wrap.Client, shareIdx int) (*
 //
 //nolint:gocognit
 func NewComponent(eth2Cl eth2wrap.Client, allPubSharesByKey map[core.PubKey]map[int]*bls_sig.PublicKey,
-	shareIdx int, feeRecipientAddress string, builderAPI bool, seenPubkeys func(core.PubKey),
+	shareIdx int, feeRecipientAddressByPubkey map[core.PubKey]string, builderAPI bool, seenPubkeys func(core.PubKey),
```
corverroos, line 61: same here, rather not pass in pubkeys twice, wouldn't it be better to use a single map with both shares and feerecipients? The other option is pass in a callback for `feeRecipientFunc func(core.Pubkey)string`

dB2510: Thanks for pointing out, went with `feeRecipientFunc func(core.Pubkey)string`

core/validatorapi/teku.go
```diff
@@ -36,7 +36,10 @@ type TekuBuilder struct {
 	Overrides map[string]string `json:"registration_overrides,omitempty"`
 }
 
-const gasLimit = 30000000
+const (
+	gasLimit    = 30000000
+	zeroAddress = "0x0000000000000000000000000000000000000000"
```
corverroos, line 41: zero values mean golang zero values, so do not populate the `Default`field. Just make sure Teku validation doesn't fail in this case.

dB2510: yup, tried that earlier unfortunately teku doesn't like that:
Screenshot 2023-01-25 at 4 23 41 PM

corverroos (approved):

PR #1686, opened 2023-01-25
eth2util: remove randomHex from testutil

xenowits:
Remove randomHex from testutil and add to eth2util package.

category: refactor
ticket: none

### DIFF
--- a/eth2util/hash.go
+++ b/eth2util/hash.go
@@ -16,6 +16,9 @@
 package eth2util
 
 import (
+	"crypto/rand"
+	"encoding/hex"
+
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	ssz "github.com/ferranbt/fastssz"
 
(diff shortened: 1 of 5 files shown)


eth2util/hash.go
```diff
@@ -40,3 +43,14 @@ func SlotHashRoot(slot eth2p0.Slot) ([32]byte, error) {
 
 	return hash, nil
 }
+
+// RandomHex32 returns a random 32 character hex string. It uses crypto/rand.
+func RandomHex32() (string, error) {
```
xenowits, line 48: This is used by `cmd`, `dkg` and `keymanager` packages. Although, it is not specifically "eth2" related so suggestions welcome on where to move this to

corverroos: can just duplicate, it is pretty standard stuff

corverroos: I would increase the length of these passwords to 32bits. 64 chars

corverroos (approved):

Issue #1658, opened 2023-01-19
Design a BLS12-381 library abstraction for Charon


gsora:
## Problem to be solved

One of the immediate result of the team discussion regarding a Kryptology replacement for Charon (design doc [here](https://docs.google.com/document/d/1oh90q-QMitxIp0c-e8ZcLjhZTKvdU0OufRhxJqHhun8/edit#)) is the need for an abstraction on top of the chosen BLS12-381 library.

## Proposed solution

Design and implement an abstraction in the form of Go struct and/or interfaces that generalizes the operations that Charon needs in terms of BLS12-381 key handling and usage.

### Out of Scope

No effective replacement will be implemented, this issue will mostly reflect on API design and testing.

Another ticket will be created to signal and discuss replacement in Charon.

gsora:
This is the first iteration of the abstraction design: https://github.com/ObolNetwork/charon/commit/aa5904768d55b2c8bc1ab8d8e936c018b33a9810

It's a _very_ rough draft, intended for gathering team's feedback.

I tried capturing the spirit of the current API, while moving to interface-based "primitive" types (private and public keys, signatures).

I also added an implementation of it on top of Herumi, so we can gather the real ergonomics of this abstraction on a real-world example.

Please let me know what you think!

gsora:
This is the latest iteration of the design, built after team's feedback on Discord: https://github.com/ObolNetwork/charon/tree/gsora/bls-abstraction/tbls/taketwo

The `taketwo` package is a placeholder, its content will be moved to the `tbls` package once we all agree on the design.

This iteration is much leaner --- only dealing with a thin abstraction over byte slices --- and allows for quick iteration in case we need to swap BLS engine again in the future.

I added a generic test suite that tests the expected behavior from the point of view of the interface, and a fuzzing target that randomly select an implementation each time an interface method gets called --- this is useful to test compatibility among any implemented.

Let me know what you think!

PR #1692, opened 2023-01-26
tbls: library for BLS12-381 abstraction

gsora:
This PR adds a subpackage in `tbls` called `v2`, which is a new abstraction for BLS12-381 cryptography.

It contains two concrete implementations — one using [Herumi](https://github.com/herumi/bls-eth-go-binary) and another based on [Kryptology](https://github.com/coinbase/kryptology), added for backwards compatibility while we work out the migration.

A complete test suite is provided, alongside a fuzzing target to test cross-compatibility between implementation.

Preliminary tests show substantial performance improvements. 

category: refactor
ticket: https://github.com/ObolNetwork/charon/issues/1658


gsora:
Fixing linting issues as we speak.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/ferranbt/fastssz v0.1.2
 	github.com/golang/snappy v0.0.4
 	github.com/gorilla/mux v1.8.0
+	github.com/herumi/bls-eth-go-binary v1.28.1
```
corverroos, line 13: might need to remove any references of "CGO_ENABLED=0"

tbls/v2/herumi/herumi.go
```diff
@@ -0,0 +1,232 @@
...
+	"sync"
+
+	"github.com/herumi/bls-eth-go-binary/bls"
+
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/z"
+	v2 "github.com/obolnetwork/charon/tbls/v2"
+)
+
+var initializationOnce = sync.Once{}
```
corverroos, line 30: nit, short variables FTW: `initOnce`

tbls/v2/herumi/herumi.go
```diff
@@ -0,0 +1,232 @@
...
+
+// Herumi is an Implementation with Herumi-specific inner logic.
+type Herumi struct{}
+
+func (Herumi) GenerateSecretKey() (v2.PrivateKey, error) {
+	var p bls.SecretKey
+	p.SetByCSPRNG()
+
+	// Commenting here once, this syntax will appear often:
+	// here I'm converting ret to a pointer to instance of v2.PrivateKey, which is
```
corverroos, line 58: no "ret" here...

tbls/v2/tbls.go
```diff
@@ -0,0 +1,97 @@
...
+	Verify(compressedPublicKey PublicKey, data []byte, signature Signature) error
+
+	// Sign signs data with the provided private key, and returns the resulting signature.
+	// This function works on both shares of private keys, and complete private keys.
+	Sign(privateKey PrivateKey, data []byte) (Signature, error)
+}
+
+// SetImplementation sets newImpl as the package backing implementation.
+func SetImplementation(newImpl Implementation) {
+	implLock.Lock()
```
corverroos, line 66: now that I see it, lock only makes sense if we also use it when accessing/reading the global variable in funcs below. Lets rather remove the lock and add a comment: `this isn't threadsafe and should be called on startup before using any functions below`...?

corverroos (approved):

tbls/v2/tbls.go
corverroos: sure no problem

Issue #1689, opened 2023-01-25
Cluster definition hash incorrect for empty addresses or signatures


corverroos:
## Problem to be solved

The current (and previous) versions of the charon definition (<=v1.4) calculates the incorrect SSZ hash for empty Bytes20 (eth addresses) and Bytes65 (k1 signatures). 

> `BytesN` type fields should ALWAYS be of length N. So an empty address should be 20 bytes of just 0s, similarly an empty signature should be 56 bytes of just 0s. Charon uses empty arrays instead which is incorrect.

This is causing an incompatibility with the launchpad, specifically the solo slow that includes empty operator addresses. The JS implementation is doing it correctly. That means the launchpad generated solo flow definitions are not compatible with charon v1.4. 
 
We cannot however fix this in charon v1.4, since it would break any lock files out there with empty addresses or signatures, specifically lock files created by the `create cluster` cli. 
 
The proper solution is to fix this in charon, so as part of the upcoming v1.5 that will be released as part of charon v0.14.

## Proposed solution

If the launchpad needs to support solo flow with charon v0.13 (so cluster definition v1.4), then we either need to:
 - fix this in charon now v1.4, thereby throwing v1.4 `create cluster` users under the bus (requiring them to use `--no-verify`).
 - or better, the launchpad can implement a workaround to achieve the same "buggy" behavior for the solo flow (and fix it in their v1.5 later).
 
 Charon v1.5 should however include the fix. Charon v1.3 must still be compatible with buggy logic.

PR #1695, opened 2023-01-26
cluster: fix SSZ BytesN issue

corverroos:
Fixes the SSZ issue that didn't ensure `BytesN` are always of length N. This is only included in v1.5 onwards. 

Also removes the lock.DistributedValidator.FeeRecipient field completely, also for previous versions as this isn't used at all and we do not need to support it for older versions.

category: bug
ticket: #1689

### DIFF
--- a/cluster/ssz.go
+++ b/cluster/ssz.go
@@ -480,7 +480,7 @@ func hashLockV1x5(l Lock, hh ssz.HashWalker) error {
 		return err
 	}
 
-	// Field (1) 'ValidatorAddresses' CompositeList[65536]
+	// Field (1) 'Validators' CompositeList[65536]
 	{
 		subIndx := hh.Index()
 		num := uint64(len(l.Validators))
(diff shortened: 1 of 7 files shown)


cluster/ssz.go
```diff
@@ -529,7 +593,7 @@ func hashValidatorLegacy(v DistValidator, hh ssz.HashWalker) error {
 	}
 
 	// Field (2) 'FeeRecipientAddress'
-	hh.PutBytes([]byte(to0xHex(v.FeeRecipientAddress)))
+	hh.PutBytes(nil)
```
dB2510, line 596: can remove this as it is v1.5

cluster/ssz.go
```diff
@@ -460,8 +466,39 @@ func hashLockV1x3orLater(l Lock, hh ssz.HashWalker) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Field (0) 'Definition' Composite
+	if err := defHashFunc(l.Definition, hh, false); err != nil {
+		return err
+	}
+
+	// Field (1) 'ValidatorAddresses' CompositeList[65536]
```
dB2510, line 483:
```suggestion
	// Field (1) 'Validators' CompositeList[65536]
```

dB2510 (approved):

Issue #1637, opened 2023-01-16
Warn if a relay without https is specified


OisinKyne:
## Problem to be solved

As we harden charon, we need to be less permissive with unsafe configurations, we may even treat mainnet differently to testnets. One thing that's become apparent is that we are hitting our old bootnode over http. This could be easily abused. 

## Proposed solution

On the platform side, we will make the relay available over https. On charon side for now the only requirement is to warn for every relay provided that uses http not https. After that, we can in future consider making http a dealbreaker for mainnet configurations, but not yet. 

## Edge Cases

HTTP 301 304 Redirects when requesting the unprotected endpoint ideally can be respected (is this unsafe, should we avoid redirects entirely?)

### Out of Scope

Making relays able to terminate TLS directly at themselves

PR #1704, opened 2023-01-27
cmd: add warning if http relay is provided

xenowits:
Logs a warning if a relay with `http` scheme is provided instead of `https`.

category: feature
ticket: #1637

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -138,12 +138,10 @@ func bindP2PFlags(cmd *cobra.Command, config *p2p.Config) {
 			u, err := url.Parse(relay)
 			if err != nil {
 				return errors.Wrap(err, "parse relay address", z.Str("relay", relay))
-			} else if u.Host == "" {
-				return errors.New("invalid host", z.Str("relay", relay))
 			}
 
 			if u.Scheme == "http" {
-				log.Warn(ctx, "Relay uses HTTP which is unsafe, relays should use HTTPS instead.", nil, z.Str("relay", relay))
+				log.Warn(ctx, "Insecure relay address provided, not HTTPS", nil, z.Str("address", relay))
 			}
 		}
 
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"context"
+	"net/url"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -129,6 +131,24 @@ func bindP2PFlags(cmd *cobra.Command, config *p2p.Config) {
...
+		for _, relay := range config.Relays {
+			u, err := url.Parse(relay)
+			if err != nil {
+				return errors.Wrap(err, "parse relay address", z.Str("relay", relay))
+			} else if u.Host == "" {
+				return errors.New("invalid host", z.Str("relay", relay))
+			}
+
+			if u.Scheme == "http" {
+				log.Warn(ctx, "Relay uses HTTP which is unsafe, relays should use HTTPS instead.", nil, z.Str("relay", relay))
```
xenowits, line 146: suggest appropriate warning text

xenowits:
Sample output:

Screenshot 2023-01-27 at 12 43 59

cmd/run.go
corverroos: `log.Warn(ctx, "Insecure relay address provided, not HTTPS", nil, z.Str("address", relay))`

Logs should not end in punctuation

cmd/run.go
```diff
@@ -129,6 +131,24 @@ func bindP2PFlags(cmd *cobra.Command, config *p2p.Config) {
...
 	cmd.Flags().StringVar(&config.Denylist, "p2p-denylist", "", "Comma-separated list of CIDR subnets for disallowing certain peer connections. Example: 192.168.0.0/16 would disallow connections to peers on your local network. The default is to accept all connections.")
 	cmd.Flags().BoolVar(&config.DisableReuseport, "p2p-disable-reuseport", false, "Disables TCP port reuse for outgoing libp2p connections.")
+
+	wrapPreRunE(cmd, func(cmd *cobra.Command, args []string) error {
+		ctx := log.WithTopic(cmd.Context(), "cmd")
+		for _, relay := range config.Relays {
+			u, err := url.Parse(relay)
+			if err != nil {
+				return errors.Wrap(err, "parse relay address", z.Str("relay", relay))
+			} else if u.Host == "" {
```
corverroos, line 141: remove this check, or rather, why only do this check?

corverroos (approved):

cmd/run.go
xenowits: Added this check to ensure the relay url is valid, but we can remove it here since it will be handled by the main run program

PR #1705, opened 2023-01-27
app/log: fix loki caller

corverroos:
Fixes the resulting log `caller` field when using the multilogger (when pushing logs to loki).

Previously all loki logs had the caller: `log/log.go:82`.

category: bug
ticket: none

### DIFF
--- a/app/log/config.go
+++ b/app/log/config.go
@@ -35,9 +35,10 @@ import (
 )
 
 const (
-	padLength = 40
-	keyStack  = "stacktrace"
-	keyTopic  = "topic"
+	defaultCallerSkip = 1
+	padLength         = 40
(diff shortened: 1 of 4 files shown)


app/log/config.go
```diff
@@ -229,7 +235,7 @@ func newStructuredLogger(format string, level zapcore.Level, ws zapcore.WriteSyn
 	return zap.New(
 		zapcore.NewCore(structured, ws, zap.NewAtomicLevelAt(level)),
 		zap.WithCaller(true),
-		zap.AddCallerSkip(1),
```
corverroos, line 232: had to make this dynamic based on whether logger are wrapper by multilogger or not.

xenowits (approved):

PR #1709, opened 2023-01-27
core/scheduler: mark epoch as resolve if no validators

corverroos:
Marks scheduler epoch as resolved when no validators are active. This mitigates errors on new clusters:
```
Emit scheduled slot event: epoch not resolved yet
```

category: bug
ticket: none

### DIFF
--- a/core/scheduler/scheduler_test.go
+++ b/core/scheduler/scheduler_test.go
@@ -383,6 +383,36 @@ func TestScheduler_GetDuty(t *testing.T) {
 	require.NoError(t, sched.Run())
 }
 
+//go:generate go test . -run=TestNoActive -count=100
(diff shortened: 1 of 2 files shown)


core/scheduler/scheduler_test.go
```diff
@@ -383,6 +383,36 @@ func TestScheduler_GetDuty(t *testing.T) {
 	require.NoError(t, sched.Run())
 }
 
+//go:generate go test . -run=TestNoActive -count=100
```
dB2510, line 386: can remove this, i assume this was used to test if this flaps

dB2510 (approved):

xenowits (approved):

Issue #1698, opened 2023-01-27
Port geth eip-712 package


corverroos:
## Problem to be solved

One last feature from the geth dependency to remove is the eip-712 functionality: `github.com/ethereum/go-ethereum/signer/core/apitypes`

## Proposed solution

Port code to eth2util.

### Out of Scope

If there is anything to highlight as out of scope for this issue, please outline it here.

PR #1710, opened 2023-01-30
gomod: remove geth dependency

corverroos:
Finally removes all traces of go-ethereum and the dependency.

category: refactor
ticket: #1698


dB2510 (approved):
🚀

PR #1719, opened 2023-01-30
app: fail startup if BN on wrong fork/network

corverroos:
Fail startup if BN connected to wrong network, as it is identical to not being connected to a BN. 

category: feature
ticket: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -18,6 +18,7 @@
 package app
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"net/http"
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -680,6 +681,22 @@ func newETH2Client(ctx context.Context, conf Config, life *lifecycle.Manager,
 		eth2Cl = eth2wrap.WithSyntheticDuties(eth2Cl, pubkeys)
 	}
 
+	// Check BN chain/network.
```
xenowits, line 684: can move this BN check to before eth2Cl is initialized

xenowits (approved):

app/app.go
corverroos: mmmm, you need the eth2Cl for the check, so this isn't possible?

PR #1737, opened 2023-01-31
Dockerfile: bring in `ca-certificates`

gsora:
Bring in `ca-certificates` in Docker image, to allow connection to TLS-enabled servers.

category: bug 
ticket: none

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Container for building Go binary.
 FROM golang:1.19.5-bullseye AS builder
 # Install dependencies
-RUN apt-get update && apt-get install -y build-essential git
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Container for building Go binary.
 FROM golang:1.19.5-bullseye AS builder
 # Install dependencies
-RUN apt-get update && apt-get install -y build-essential git
+RUN apt-get update && apt-get install -y build-essential git ca-certificates
```
corverroos, line 4: do we need it here as well?

corverroos (approved):

Dockerfile
gsora: probably not, the `golang` base image will have it embedded, removing it

Issue #1738, opened 2023-01-31
Promrated needs authentication token and is failing silently


LukeHackett12:
# 🐞 Bug Report

### Description

Promrated has stopped working as they made a breaking change to the endpoint we use to need auth. We didn't get notified as it fails silently.

### Has this worked before in a previous version?

 Yes, the previous version in which this bug was not present was: ....

## 🔬 Minimal Reproduction

run promrated

## 🔥 Error

<pre><code>

31mERRO[0m Getting validator statistics: not ok http response {"validator": "0xaa17636b71b5829e11e796df1dd24bf949b612139ce84eee9688f25d32b04c749b57d406f119c1eac21e848c2d16abb9", "body": "{\"detail\":\"Not authenticated\"}"}
</code></pre>

LukeHackett12:
Opened PR to panic so we will be alerted, still waiting for auth token off rated

PR #1739, opened 2023-01-31
testutil/promrated: add metric to count errors from rated api

LukeHackett12:
Rated recently added authentication and this has been failing silently as the error was just being logged

category: bug
ticket: #1738

### DIFF
--- a/testutil/promrated/rated.go
+++ b/testutil/promrated/rated.go
@@ -42,7 +42,7 @@ type validatorEffectivenessData struct {
 
 // getValidationStatistics queries rated for a pubkey and returns rated data about the pubkey
 // See https://api.rated.network/docs#/default/get_effectiveness_v0_eth_validators__validator_index_or_pubkey__effectiveness_get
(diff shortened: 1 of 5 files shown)


testutil/promrated/rated.go
```diff
@@ -90,6 +90,8 @@ func getValidationStatistics(ctx context.Context, ratedEndpoint string, validato
 			backoff()
 
 			continue
+		} else if res.StatusCode == http.StatusUnauthorized {
+			panic("unauthorized")
```
dB2510, line 94: suggest handle this error rather than panic

LukeHackett12: How would you handle it? The intention behind the panic is to start a crash loop so we can know that it isn't working. Nothing will change if we repeatedly try do this.

corverroos: crashloop is not a way to do monitoring, prefer logs + metrics and then grafana alerts

LukeHackett12: OK I added a metric for counting rated api errors if that works better 👍🏼

OisinKyne (approved):

Issue #1652, opened 2023-01-18
Prioritise block proposal strategy


corverroos:
## Problem to be solved

Block proposals can either be:
 - normal beacon blocks
 - normal beacon blocks with synthetic blocks
 - blinded blocks
 - blinded blocks with synthetic blocks
 
 All nodes in the cluster should use the same strategy
 
## Proposed solution

Prioritise block proposal strategies

### Out of Scope

If there is anything to highlight as out of scope for this issue, please outline it here.

PR #1751, opened 2023-02-01
core: introduce mutable builder enabled config

corverroos:
Introduces `core.BuilderEnabled` that abstracts `app.mutableConfig` which wraps `infosync.Component` such that whether builderAPI is enabled is dynamically based on prioritise results.

category: feature
ticket: #1652

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -1,0 +1,19 @@
+// Copyright © 2022 Obol Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option)
+// any later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+// BuilderEnabled determines whether the builderAPI is enabled for the provided slot.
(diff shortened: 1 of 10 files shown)


core/config.go
```diff
@@ -0,0 +1,19 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package core
+
+// BuilderEnabled determines whether the builderAPI is enabled for the provided slot.
+type BuilderEnabled func(slot int64) bool
```
xenowits, line 19: is builderAPI enabled per slot? I didn't understand the relation between these two

core/validatorapi/validatorapi.go
```diff
@@ -553,8 +553,17 @@ func (c Component) submitRegistration(ctx context.Context, registration *eth2api
 
 // SubmitValidatorRegistrations receives the partially signed validator (builder) registration.
 func (c Component) SubmitValidatorRegistrations(ctx context.Context, registrations []*eth2api.VersionedSignedValidatorRegistration) error {
+	if len(registrations) == 0 {
+		return nil // Nothing to do
+	}
+
+	slot, err := c.slotFromTimestamp(ctx, time.Now())
```
xenowits, line 560: what happens when the machine's time is not in sync with actual time?

xenowits (approved):

core/config.go
corverroos: yes, prioritise enables/disables config based on slot, so from slot x to slot y it is enabled, otherwise disabled.

core/validatorapi/validatorapi.go
corverroos: there is no such thing as actual time in distributed system

Issue #1744, opened 2023-01-31
Refactor `tbls` usage to `tbls/v2` package


gsora:
## Problem to be solved

Right now the charon codebase is hardwired to work with `tbls` and Kryptology.

`tbls/v2` provides a BLS12-381 implementation-agnostic interface that must be used in place of `tbls` + Kryptology.

By means of `tbls/v2` we'll be able to support `herumi_bls` feature flag - described in #1743 - and conduct testing of this new implementation.

The following packages are involved in the refactoring:

- [ ] `app`
- [ ] `cluster`
- [ ] `cmd`
- [ ] `core`
- [ ] `core/parsigex`
- [ ] `core/sigagg`
- [ ] `core/validatorapi`
- [ ] `dkg`
- [ ] `docs`
- [ ] `eth2util/deposit`
- [ ] `eth2util/eth2exp`
- [ ] `eth2util/keymanager`
- [ ] `eth2util/keystore`
- [ ] `eth2util/signing`
- [ ] `testutil`
- [ ] `testutil/combine`
- [ ] `testutil/validatormock`

PR #1754, opened 2023-02-01
core/sigagg: refactor to use tbls/v2

gsora:
This PR moves `core/sigagg` package to use `tbls/v2` package, which allows us to leverage Herumi BLS.

category: refactor
ticket: #1744 
feature_flag: herumi_bls


corverroos (approved):

gsora:
(fun fact, as of now this PR's test are a hybrid between `tbls` and `tbls/v2` package, and it works!)

Issue #1312, opened 2022-10-19
Charon cannot start if any beacon node is down


corverroos:
# 🐞 Bug Report

## Problem to solve

Charon cannot start if any of the configured beacon-node-endpoints are not responding. So even though some beacon nodes are up and available, charon cannot start if only a single of the BNs are down. 

This is due to go-eth2-client that fails to initialise if the BN is not accessible.

## Proposed Solution

Wrap go-eth2-client instances in a `LazyClient` type that tries to initialise the underlying client on each API call if not initialised yet.

OisinKyne:
Is this long fixed? @corverroos

corverroos:
this isn't fixed no

PR #1760, opened 2023-02-02
app/eth2wrap: add lazy init

corverroos:
Wraps eth2clients in a "lazy loader", this allows charon to start even if one or more beacon nodes are down (as long as one is up and available).

category: feature
ticket: #1312

### DIFF
--- a/app/eth2wrap/eth2wrap_test.go
+++ b/app/eth2wrap/eth2wrap_test.go
@@ -21,6 +21,8 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httputil"
+	"net/url"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


app/eth2wrap/eth2wrap_test.go
```diff
@@ -276,3 +288,53 @@ func TestBlockAttestations(t *testing.T) {
...
+	}
+
+	eth2Cl, err := eth2wrap.NewMultiHTTP(ctx, time.Second, addresses...)
+	require.NoError(t, err)
+
+	_, err = eth2Cl.SlotDuration(ctx)
+	require.NoError(t, err)
+}
+
+func TestLazy(t *testing.T) {
```
dB2510, line 309: can we also have a testcase in this test where we have 3 addresses with following testcases:
- All disabled.
- One enabled and two disabled
- Two enabled and one disabled

This can replace above `TestOneDown` test

dB2510 (approved):

app/eth2wrap/lazy.go
```diff
@@ -0,0 +1,118 @@
...
+	var err error
+	l.client, err = l.provider()
+
+	return l.client, err
+}
+
+func (l *lazy) Name() string {
+	cl, err := l.getClient()
+	if err != nil {
+		return ""
```
xenowits, line 69: why not return an error here?

xenowits (approved):

app/eth2wrap/lazy.go
corverroos: This implements the `Client` interface which doesn't return an error, so we can't return an error

app/eth2wrap/eth2wrap_test.go
corverroos: not sure this add a lot of value additional value, but you are welcome to add it if you want

Issue #1612, opened 2023-01-10
Refactor VC readyz status


corverroos:
## Problem to be solved

The current VC readyz statuses are:
- `vc missing some validators` when not all validators are configured in the VC
- `vc not configured` when no validators are configured in the VC

The fact that this is only a distinction between "some" validators configured and "no" validators configured isn't clear and isn't very important. The signal to noise isn't great.

Instead, the more common VC issues would be:
 - `vc not connected`: no vc connected the charon, no vapi calls
 - `vc missing validators`: not all validators configured, either missing all or missing some.

## Proposed solution

 - rename `vc missing some validators` to `vc missing validators`
 - remove `vc not configured`
 - replace with new `vc not connected` status that checks for any vapi calls (not just seen pubkeys).
 
 Update grafana panels accordingly

PR #1761, opened 2023-02-02
app: refactor VC readyz checks

corverroos:
Changes VC readyz statuses:
 - `vc not connected` indicates no vapi calls
 - `vc missing validators` indicates some or all validators not configured in VC 

category: refactor
ticket:  #1612

### DIFF
--- a/app/monitoringapi.go
+++ b/app/monitoringapi.go
@@ -42,8 +42,8 @@ var (
 	errReadyInsufficientPeers = errors.New("quorum peers not connected")
 	errReadyBeaconNodeSyncing = errors.New("beacon node not synced")
 	errReadyBeaconNodeDown    = errors.New("beacon node down")
-	errReadyVCNotConfigured   = errors.New("vc not configured")
-	errReadyVCMissingVals     = errors.New("vc missing some validators")
+	errReadyVCNotConnected    = errors.New("vc not connected")
+	errReadyVCMissingVals     = errors.New("vc missing validators")
 )
 
 // wireMonitoringAPI constructs the monitoring API and registers it with the life cycle manager.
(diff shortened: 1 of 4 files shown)


app/monitoringapi.go
```diff
@@ -111,28 +113,22 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
...
-		current := newCurrent()
-
 		for {
 			select {
 			case <-ctx.Done():
 				return
 			case <-epochTicker.Chan():
 				// Copy current to previous and clear current.
-				previous, current = current, newCurrent()
+				prevPKs, currPKs = currPKs, make(map[core.PubKey]bool)
```
xenowits, line 130:
```suggestion
				prevPubkeys, currPubkeys = currPubkeys, make(map[core.PubKey]bool)
```

xenowits: same for other places as well

xenowits (approved):

app/monitoringapi.go
```diff
@@ -111,28 +113,22 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
...
-
-		// newCurrent returns a new current map, populated with all the pubkeys.
-		newCurrent := func() map[core.PubKey]bool {
-			current := make(map[core.PubKey]bool)
-			for _, pubkey := range pubkeys {
-				current[pubkey] = true
-			}
-
-			return current
+		currVAPICount := 0
```
dB2510, line 116: nit: this vapi count can be boolean too

dB2510 (approved):
Let's test this on our performance cluster

app/monitoringapi.go
corverroos: I feel `PKs` is a clear enough

app/monitoringapi.go
corverroos: sure, but a count is more representative

Issue #1747, opened 2023-02-01
Implement `NodeVersion` for charon


xenowits:
## Problem to be solved

Currently, when a VC queries charon for the [node version](https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion), charon simply proxies this request to the upstream beacon node. So, the VC gets the node version of the connected beacon node instead. 

Some VCs (ex: [vouch](https://github.com/attestantio/vouch)) may want to know if they are talking to a beacon node or a DVT client such as charon.

## Proposed solution

Implement `/eth/v1/node/version` to return this:
```json
{
  "data": {
    "version": "obolnetwork/charon/${CHARON_VERSION}-{short-githash}/${GO_ARCH}_${GO_OS}"
  }
}
```
Example response:
```json
{
  "data": {
    "version": "obolnetwork/charon/v0.13.0-ec8f9be/arm64-darwin"
  }
}
```

### Out of Scope

None.

PR #1765, opened 2023-02-03
validatorapi: add node version

xenowits:
Adds node version endpoint to validatorapi.

category: feature
ticket: #1747

### DIFF
--- a/core/validatorapi/eth2types.go
+++ b/core/validatorapi/eth2types.go
@@ -171,9 +171,7 @@ type syncCommitteeContributionResponse struct {
 // nodeVersionResponse defines the response to the node version endpoint.
 // See: https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion
 type nodeVersionResponse struct {
-	Data nodeVersionInner `json:"data"`
-}
-
-type nodeVersionInner struct {
-	Version string `json:"version"`
+	Data struct {
+		Version string `json:"version"`
+	} `json:"data"`
 }
--- a/core/validatorapi/eth2types.go
+++ b/core/validatorapi/eth2types.go
@@ -167,3 +167,13 @@ type syncCommitteeDutiesResponse struct {
 type syncCommitteeContributionResponse struct {
 	Data *altair.SyncCommitteeContribution `json:"data"`
 }
+
+// nodeVersionResponse defines the response to the node version endpoint.
+// See: https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion
+type nodeVersionResponse struct {
+	Data nodeVersionInner `json:"data"`
+}
+
+type nodeVersionInner struct {
+	Version string `json:"version"`
+}
--- a/core/validatorapi/router_internal_test.go
+++ b/core/validatorapi/router_internal_test.go
@@ -271,21 +271,21 @@ func TestRawRouter(t *testing.T) {
 
 //nolint:maintidx // This function is a test of tests, so analysed as "complex".
 func TestRouter(t *testing.T) {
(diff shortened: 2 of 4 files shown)


xenowits:
Sample output running latest charon with compose:
Screenshot 2023-02-03 at 19 56 32

core/validatorapi/eth2types.go
```diff
@@ -167,3 +167,13 @@ type syncCommitteeDutiesResponse struct {
...
 	Data *altair.SyncCommitteeContribution `json:"data"`
 }
+
+// nodeVersionResponse defines the response to the node version endpoint.
+// See: https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion
+type nodeVersionResponse struct {
+	Data nodeVersionInner `json:"data"`
+}
+
+type nodeVersionInner struct {
```
gsora, line 177: Is this struct needed? Can't we merge it into `nodeVersionResponse`?

The struct is small anyway, and only used to return one value.

gsora (approved):

core/validatorapi/eth2types.go
dB2510: +1
Something like:
```
type nodeVersionResponse struct {
	Data struct {
		Version string `json:"version"`
	} `json:"data"`
}
```

dB2510 (approved):

core/validatorapi/router_internal_test.go
```diff
@@ -962,6 +937,15 @@ func testRawRouter(t *testing.T, handler testHandler, callback func(context.Cont
 	callback(context.Background(), server.URL)
 }
 
+// newTestHandler returns a new testHandler with NodeVersionFunc set.
+func newTestHandler() testHandler {
+	return testHandler{
+		NodeVersionFunc: func(ctx context.Context) (string, error) {
```
corverroos, line 943: why is this required?

core/validatorapi/validatorapi.go
```diff
@@ -1024,6 +1026,13 @@ func (c Component) ValidatorsByPubKey(ctx context.Context, stateID string, pubsh
 	return c.convertValidators(valMap)
 }
 
+// NodeVersion returns the current version of charon.
+func (Component) NodeVersion(context.Context) (string, error) {
+	commitSHA, _ := version.GitCommit()
+
+	return fmt.Sprintf("obolnetwork/charon/%s-%s/%s-%s", version.Version, commitSHA, runtime.GOARCH, runtime.GOOS), nil
```
corverroos, line 1033: is `runtime.GOARCH` and `runtime.GOOS` present in the container?

core/validatorapi/router_internal_test.go
xenowits: Since this endpoint is required for go-eth2-client to initialize - https://github.com/attestantio/go-eth2-client/blob/master/http/service.go#L142

core/validatorapi/validatorapi.go
xenowits: Yes it should be. I checked it running latest charon on my PC:
Screenshot 2023-02-06 at 13 05 45

core/validatorapi/router_internal_test.go
corverroos: Suggest adding it a method on `testHandler` and removing the `/eth/v1/node/version` endpoint from `newBeaconHandler` since it isn't used anymore.
```
func (h testHandler) NodeVersion(ctx) string, err {
  if h.NodeVersionFunc != nil {
    return h.NodeVersionFunc(ctx)
  }
  // Default to 
  return "mock_version", nil
}
```

PR #1769, opened 2023-02-07
tbls/v2: add AggregatePublicKeys method

gsora:
Used to aggregate multiple public keys into a single one.

category: feature 
ticket: none


gsora:
Not needed, since this functionality would only be used in one place where we can use FastVerifyAggregate.

Issue #1492, opened 2022-11-29
Create a `--publish` flag on the `dkg` and `create cluster` commands that pushes the produced lockfile to the obol api


OisinKyne:
## Problem to be solved

We want to give a status page to people that shows the health of a cluster. We are creating a resource in the obol api to store pubkeys and their associated details. https://github.com/ObolNetwork/obol-api/issues/13

## Proposed solution

Add an optional flag to 
- [ ] `charon dkg`
- [ ]  `charon create cluster`

Called `--publish` (open to counterproposals), that will push the signed lock file to the obol API. The dkg/create command shouldn't fail outright if this POST request fails, but it should log a `warn` to indicate that publishing failed and that a cluster status page would not be available. 

### In Scope

Updating the docs and the charon cli output to reflect this new flag
Consulting with legal to ensure we have waivers/explainers set up that allow us to record this information if they give it to us.

corverroos:
Another alternative in a simple bash script in cdvn repo, `./publish_lock_file_to_obol.sh`:
 - Dedicated action that any single operator can perform
 - Nice dedicated to print waiver
 - No need to complicate and include in DKG. Less things to worry about during DKG.
 - Retry on failure is handled by user
 - Can be performed at any time, not limited to making the decision at DKG.

Battenfield:
This `--publish` flag will help out with the UX tremendously as it requires no extra steps on the user's behalf.

Alternatively, we can also have a manual upload option in the launchpad, similar to the dkg. Where users can drag and drop their lock file.

PR #1781, opened 2023-02-08
dkg: add --publish flag to upload lock file

LukeHackett12:
Adds publish flag to dkg and create cluster commands to upload the lock file generated to out api. This introduces the new concept of our launchpad api, so added a configurable address with prod as default. 

category: feature
ticket: #1492

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -41,6 +41,7 @@ import (
 	"github.com/obolnetwork/charon/eth2util/enr"
 	"github.com/obolnetwork/charon/eth2util/keymanager"
 	"github.com/obolnetwork/charon/eth2util/keystore"
+	"github.com/obolnetwork/charon/launchpad"
 	"github.com/obolnetwork/charon/p2p"
 	"github.com/obolnetwork/charon/tbls"
 	"github.com/obolnetwork/charon/tbls/tblsconv"
@@ -70,6 +71,9 @@ type clusterConfig struct {
 	SplitKeysDir string
 
 	InsecureKeys bool
+
+	LaunchpadAPIAddr string
+	Publish          bool
 }
 
 func newCreateClusterCmd(runFunc func(context.Context, io.Writer, clusterConfig) error) *cobra.Command {
@@ -105,6 +109,8 @@ func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.IntVar(&config.NumDVs, "num-validators", 1, "The number of distributed validators needed in the cluster.")
 	flags.BoolVar(&config.SplitKeys, "split-existing-keys", false, "Split an existing validator's private key into a set of distributed validator private key shares. Does not re-create deposit data for this key.")
 	flags.StringVar(&config.SplitKeysDir, "split-keys-dir", "", "Directory containing keys to split. Expects keys in keystore-*.json and passwords in keystore-*.txt. Requires --split-existing-keys.")
+	flags.StringVar(&config.LaunchpadAPIAddr, "launchpad-api-address", "https://obol-api-prod-green.gcp.obol.tech", "The launchpad api URL.")
+	flags.BoolVar(&config.Publish, "publish", false, "Publish lock file to Obol DVT launchpad.")
 }
 
 func bindInsecureFlags(flags *pflag.FlagSet, insecureKeys *bool) {
@@ -205,10 +211,19 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return err
 	}
 
-	// Write cluster-lock file
+	if conf.Publish {
+		cl := launchpad.New(conf.LaunchpadAPIAddr)
+		if err = cl.PublishLock(ctx, lock); err != nil {
+			log.Warn(ctx, "Publishing lock file", err)
+		} else {
+			log.Debug(ctx, "Published lock file to api")
+		}
+	}
+
 	if err = writeLock(lock, conf.ClusterDir, numNodes, shareSets); err != nil {
 		return err
 	}
+	log.Debug(ctx, "Saved lock file to disk")
 
 	if conf.SplitKeys {
 		writeWarning(w)
--- a/cmd/dkg.go
+++ b/cmd/dkg.go
@@ -52,6 +52,8 @@ this command at the same time.`,
 	bindNoVerifyFlag(cmd.Flags(), &config.NoVerify)
 	bindP2PFlags(cmd, &config.P2P)
 	bindLogFlags(cmd.Flags(), &config.Log)
+	bindLaunchpadAPIAddrFlag(cmd.Flags(), &config.LaunchpadAPIAddr)
+	bindPublishFlag(cmd.Flags(), &config.Publish)
 
 	return cmd
 }
@@ -66,4 +68,12 @@ func bindDefDirFlag(flags *pflag.FlagSet, dataDir *string) {
 
 func bindDataDirFlag(flags *pflag.FlagSet, dataDir *string) {
 	flags.StringVar(dataDir, "data-dir", ".charon", "The directory where charon will store all its internal data")
+}
+
+func bindLaunchpadAPIAddrFlag(flags *pflag.FlagSet, addr *string) {
+	flags.StringVar(addr, "launchpad-api-address", "https://obol-api-prod-green.gcp.obol.tech", "The launchpad api URL.")
+}
+
+func bindPublishFlag(flags *pflag.FlagSet, publish *bool) {
+	flags.BoolVar(publish, "publish", false, "Publish lock file to Obol DVT launchpad.")
 }
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -36,6 +36,7 @@ import (
 	"github.com/obolnetwork/charon/eth2util"
 	"github.com/obolnetwork/charon/eth2util/deposit"
 	"github.com/obolnetwork/charon/eth2util/keymanager"
+	"github.com/obolnetwork/charon/launchpad"
 	"github.com/obolnetwork/charon/p2p"
 	tblsv2 "github.com/obolnetwork/charon/tbls/v2"
 	tblsconv2 "github.com/obolnetwork/charon/tbls/v2/tblsconv"
@@ -48,6 +49,9 @@ type Config struct {
 	DataDir        string
 	P2P            p2p.Config
 	Log            log.Config
+
+	LaunchpadAPIAddr string
+	Publish          bool
 
 	TestDef          *cluster.Definition
 	TestSyncCallback func(connected int, id peer.ID)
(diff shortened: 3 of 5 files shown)


cmd/createcluster.go
```diff
@@ -70,6 +71,9 @@ type clusterConfig struct {
 	SplitKeysDir string
 
 	InsecureKeys bool
+
+	LaunchpadAPIAddr string
```
xenowits, line 75: isn't this `ObolAPIAddr` now?

launchpad/api.go
```diff
@@ -0,0 +1,98 @@
...
+// PublishLock posts the lockfile to obol-api.
+func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
+	if lock.Version != "v1.5.0" {
+		return errors.New("version not supported", z.Str("version", lock.Version))
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	addr, err := url.JoinPath(c.baseURL, "/lock")
```
xenowits, line 53:
```suggestion
	addr, err := url.JoinPath(c.baseURL, "lock")
```

launchpad/api.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Client is the REST client for launchpad API requests.
+type Client struct {
+	baseURL string // Base launchpad URL
+}
+
+// PublishLock posts the lockfile to obol-api.
+func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
+	if lock.Version != "v1.5.0" {
```
xenowits, line 46: what happens to future lock versions, say `v1.6.0`? does this check need to be updated then?

launchpad/api.go
```diff
@@ -0,0 +1,98 @@
...
+		baseURL: url,
+	}
+}
+
+// Client is the REST client for launchpad API requests.
+type Client struct {
+	baseURL string // Base launchpad URL
+}
+
+// PublishLock posts the lockfile to obol-api.
```
xenowits, line 44: Need a consistent naming across charon, ie, either `obol-api` or `launchpad-api`. @corverroos what do you suggest?

cmd/createcluster.go
```diff
@@ -105,6 +109,8 @@ func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.IntVar(&config.NumDVs, "num-validators", 1, "The number of distributed validators needed in the cluster.")
 	flags.BoolVar(&config.SplitKeys, "split-existing-keys", false, "Split an existing validator's private key into a set of distributed validator private key shares. Does not re-create deposit data for this key.")
 	flags.StringVar(&config.SplitKeysDir, "split-keys-dir", "", "Directory containing keys to split. Expects keys in keystore-*.json and passwords in keystore-*.txt. Requires --split-existing-keys.")
+	flags.StringVar(&config.LaunchpadAPIAddr, "launchpad-api-address", "https://obol-api-prod-green.gcp.obol.tech", "The launchpad api URL.")
+	flags.BoolVar(&config.Publish, "publish", false, "Publish lock file to Obol DVT launchpad.")
```
xenowits, line 113: Or, we could infer whether to publish from `config.LaunchpadAPIAddr != ""`..

dkg/dkg.go
```diff
@@ -210,6 +214,15 @@ func Run(ctx context.Context, conf Config) (err error) {
 		log.Debug(ctx, "Saved keyshares to disk")
 	}
 
+	if conf.Publish {
+		cl := launchpad.New(conf.LaunchpadAPIAddr)
+		if err = cl.PublishLock(ctx, lock); err != nil {
+			log.Warn(ctx, "Publishing lock file", err)
```
xenowits, line 220:
```suggestion
			log.Warn(ctx, "Couldn't publish lock file to obol API", err)
```

launchpad/api.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of  MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package launchpad
```
corverroos, line 16: I'd say it is too early to create abstractions. We follow AHA https://en.wikipedia.org/wiki/Don%27t_repeat_yourself, so only create abstractions if really required. A single HTTP post doesn't warrent an abstraction yet.

corverroos: Suggest just copy pasting `postHTTP` into both `cmd` and `dkg` packages

cmd/createcluster.go
```diff
@@ -105,6 +109,8 @@ func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.IntVar(&config.NumDVs, "num-validators", 1, "The number of distributed validators needed in the cluster.")
 	flags.BoolVar(&config.SplitKeys, "split-existing-keys", false, "Split an existing validator's private key into a set of distributed validator private key shares. Does not re-create deposit data for this key.")
 	flags.StringVar(&config.SplitKeysDir, "split-keys-dir", "", "Directory containing keys to split. Expects keys in keystore-*.json and passwords in keystore-*.txt. Requires --split-existing-keys.")
+	flags.StringVar(&config.LaunchpadAPIAddr, "launchpad-api-address", "https://obol-api-prod-green.gcp.obol.tech", "The launchpad api URL.")
```
corverroos, line 112: Launchpad is the website, I think the API should be called `obol-api-address`

cmd/createcluster.go
corverroos: yeah indeed, not sure we should two flags, one to enable and one for the address, maybe just the address is better as v1, ie. `--publish=https://obol-api.tech`

cmd/dkg.go
```diff
@@ -52,6 +52,8 @@ this command at the same time.`,
 	bindNoVerifyFlag(cmd.Flags(), &config.NoVerify)
 	bindP2PFlags(cmd, &config.P2P)
 	bindLogFlags(cmd.Flags(), &config.Log)
+	bindLaunchpadAPIAddrFlag(cmd.Flags(), &config.LaunchpadAPIAddr)
+	bindPublishFlag(cmd.Flags(), &config.Publish)
```
corverroos, line 56: these are related, can bind together `bindPublishFlags`

dkg/dkg.go
```diff
@@ -210,6 +214,15 @@ func Run(ctx context.Context, conf Config) (err error) {
 		log.Debug(ctx, "Saved keyshares to disk")
 	}
 
+	if conf.Publish {
+		cl := launchpad.New(conf.LaunchpadAPIAddr)
+		if err = cl.PublishLock(ctx, lock); err != nil {
```
corverroos, line 219: see `keymanager.VerifyConnection` we need to ensure BEFORE we start DKG that the API is available to fail fast.

launchpad/api.go
corverroos: Or rather, I do think `obolapi` should be a top level package, it is not critical and it is trivial... so suggest `app/obolapi` rather

launchpad/api.go
corverroos: suggest not checking version, the server can check if it supports it and error if it doesn't

launchpad/api.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return errors.Wrap(err, "invalid address")
+	}
+
+	url, err := url.Parse(addr)
+	if err != nil {
+		return errors.Wrap(err, "invalid endpoint")
+	}
+
+	b, err := json.MarshalIndent(lock, "", " ")
```
corverroos, line 63: `MarshalIndent` is for humans, not computers

launchpad/api.go
corverroos: +1 for `obol-api`

dkg/dkg.go
```diff
@@ -210,6 +214,15 @@ func Run(ctx context.Context, conf Config) (err error) {
 		log.Debug(ctx, "Saved keyshares to disk")
 	}
 
+	if conf.Publish {
```
corverroos, line 217: need to test that this is call correctly

cmd/createcluster.go
```diff
@@ -206,6 +212,15 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 	}
 
 	// Write cluster-lock file
+	if conf.Publish {
+		cl := launchpad.New(conf.LaunchpadAPIAddr)
```
corverroos, line 216: need to test that this is call correctly

dkg/dkg.go
LukeHackett12: It is non blocking on failure so added a verify connection in right before but not sure if it serves much of a purpose?

cmd/createcluster.go
LukeHackett12: As discussed in thread, went for --publish-address

cmd/createcluster.go
LukeHackett12: Discussed in discord also, Oisin keen to keep it two separate flags for now with one defaulting. Shortened to the nice URL

cmd/createcluster.go
LukeHackett12: Going for PublishAddr in line with the flag

dkg/dkg.go
```diff
@@ -635,3 +645,19 @@ func dvsFromShares(shares []share) []cluster.DistValidator {
...
+	cl := obolapi.New(publish)
+	if err := cl.VerifyConnection(ctx); err != nil {
+		return err
+	}
+
+	if err := cl.PublishLock(ctx, lock); err != nil {
+		return err
+	}
+
+	log.Debug(ctx, "Published lock file to api")
```
corverroos, line 660: this can be info level, suggest `Published lock file, z.Str("addr", publish_addr)`

dkg/dkg.go
```diff
@@ -635,3 +645,19 @@ func dvsFromShares(shares []share) []cluster.DistValidator {
 
 	return dvs
 }
+
+// writeLockToAPI posts the lock file to obol-api.
+func writeLockToAPI(ctx context.Context, publish string, lock cluster.Lock) error {
```
corverroos, line 650: `publishAddr`

dkg/dkg_test.go
```diff
@@ -114,18 +120,31 @@ func testDKG(t *testing.T, def cluster.Definition, dir string, p2pKeys []*k1.Pri
...
-				allReceived <- struct{}{}
+				allReceivedKeystores <- struct{}{}
 			}()
 		}))
 		defer srv.Close()
 
 		conf.KeymanagerAddr = srv.URL
 	}
 
+	receivedLockfile := make(chan bool) // Receives string for lockfile intercepted by the obol-api server
```
corverroos, line 135: if you don't care about the type of the channel, use `struct{}`

corverroos (approved):

PR #1785, opened 2023-02-09
app/eth2wrap: select best client

corverroos:
Fix issue that always proxied vapi requests to the first beacon node, instead detect and select "best" beacon node.

category: bug
ticket: none


app/eth2wrap/eth2wrap.go
```diff
@@ -105,20 +109,32 @@ func NewMultiHTTP(ctx context.Context, timeout time.Duration, addresses ...strin
...
+	}
+}
+
 // multi implements Client by wrapping multiple clients, calling them in parallel
 // and returning the first successful response.
 // It also adds prometheus metrics and error wrapping.
 type multi struct {
-	clients []Client
+	clients  []Client
+	selector *bestSelector
```
xenowits, line 124: update godoc for multi as it now adds selector functionality as well

app/eth2wrap/eth2wrap.go
```diff
@@ -305,3 +331,53 @@ func wrapError(ctx context.Context, err error, label string) error {
...
+	period time.Duration
+	counts []int
+}
+
+// Best returns the best index or 0 if no counts.
+func (s *bestSelector) Best() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	var resp, count int
```
xenowits, line 358: what is the purpose of the `count` variable? seems like it's never updated

app/eth2wrap/eth2wrap.go
```diff
@@ -305,3 +331,53 @@ func wrapError(ctx context.Context, err error, label string) error {
...
+	for i, c := range s.counts {
+		if c > count {
+			resp = i
```
xenowits, line 361:
```suggestion
			resp = i
			count++
```

app/eth2wrap/eth2wrap.go
xenowits: umm..i think i now understand what you're trying to do: `get the array index with the greatest value`

app/eth2wrap/eth2wrap.go
```diff
@@ -305,3 +331,53 @@ func wrapError(ctx context.Context, err error, label string) error {
...
 	return errors.Wrap(err, "beacon api "+label, z.Str("label", label))
 }
+
+// newBestSelector returns a new bestSelector.
+func newBestSelector(n int, period time.Duration) *bestSelector {
+	return &bestSelector{
+		n:      n,
+		period: period,
+		counts: make([]int, n),
+	}
```
xenowits, line 341: nit since `Increment()` already does this
```suggestion
	    start: time.Now(),
	}
```

xenowits (approved):

app/eth2wrap/eth2wrap.go
corverroos: sure, but better safe than sorry. You could technically do stuff with current time that will cause a panic if not initialised.

Issue #1471, opened 2022-11-22
Fix `very_large` smoke test


dB2510:
## Problem to be solved

`very_large` test in testutil/compose/smoke/smoke_test.go takes a lot of time to create keys and then run docker-compose up. It produces some errors like: 
- Subscriber error: clashing public key. [DutyDB]
- Validator api 5xx response: invalid signature. [Validatorapi]
- Peer exchanged invalid partial signature: invalid signature: getting slots per epoch: beacon api slots_per_epoch: context deadline exceeded. [Parsigex]
- Validator mock error: mock proposal failed: vmock beacon block proposal: failed to request beacon block proposal: http get: Get "http://0.0.0.0:3600/eth/v2/validator/blocks/22951728?randao_reveal=0x8e81785cd1b35dae7c65f4b7e642675eb93542f33529d27d17000b89a11cf1b15fa9816976828b98cbc6a831eaf9c818073191a4b1c64c35d0e8608184fe02b67ee8484a945b3b1cd96de2c7c4fe1d26a41f88b0e31a946a0da8eac0a794080d&graffiti=": EOF. [Validatormock]

## Proposed solution

- These errors can be related to beaconmock configuration for ex: returning deterministic attester duties and 1s slot times.
- Figure out ideal configuration for beaconmock to fix above mentioned errors with `very_large` smoke test.
- Find out right configuration for `very_large` (like, nodes, threshold, num_validators) such that it is well under smoke_tests timeout of 20m.
- Error logs can be found in smoke_test github action artefacts, for ex: https://github.com/ObolNetwork/charon/actions/runs/3516395550

PR #1797, opened 2023-02-11
compose: fix very large smoke test

dB2510:
Enables very large smoke test with 7 of 10 cluster and 100 validators.

category: test
ticket: #1471

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -90,6 +90,7 @@ type Config struct {
 	SimnetBMock             bool
 	SimnetVMock             bool
 	SimnetValidatorKeysDir  string
+	SimnetSlotDuration      time.Duration
 	SyntheticBlockProposals bool
 	BuilderAPI              bool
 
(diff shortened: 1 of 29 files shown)


app/app.go
```diff
@@ -688,7 +688,7 @@ func newETH2Client(ctx context.Context, conf Config, life *lifecycle.Manager,
 	if conf.SimnetBMock { // Configure the beacon mock.
 		const dutyFactor = 100 // Duty factor spreads duties deterministically in an epoch.
 		opts := []beaconmock.Option{
-			beaconmock.WithSlotDuration(time.Second),
+			beaconmock.WithSlotDuration(time.Second * 6),
```
corverroos, line 691: why this?

testutil/compose/smoke/smoke_test.go
```diff
@@ -84,15 +84,16 @@ func TestSmoke(t *testing.T) {
...
 				conf.NumNodes = 10
 				conf.Threshold = 7
-				conf.NumValidators = 15
+				conf.NumValidators = 100
 				conf.InsecureKeys = true
 				conf.KeyGen = compose.KeyGenCreate
+				conf.VCs = []compose.VCType{compose.VCMock}
 			},
-			Timeout: time.Second * 120,
+			Timeout: time.Minute * 5,
```
corverroos, line 96: 5min is very long

app/app.go
dB2510: I tested with slot times 3s and 6s. And found out this large cluster is stable with 6s slot times. Whereas in 3s slot times, I observed 70% of duties were failing and those that were passing had broadcast delay of around 9s.
With 6s slot times the cluster had broadcast delays of around 5.5s to 6.5s. This is only for simnet/compose run and not for simnet_test.

corverroos: I would prefer 1s for compose by default, we can do 6s only for this test?

dB2510: Done!

testutil/compose/smoke/smoke_test.go
dB2510: updated to 2 mins

corverroos (approved):

Issue #1790, opened 2023-02-10
Issue with p2p connection type metric


corverroos:
# 🐞 Bug Report

### Description

The `p2p_peer_connection_types` metric seems to have a bug.

It is varying greatly and includes negative numbers which should be impossible. 

See the core team cluster example. It might be related to multiple relays we added in latest main.

![image](https://user-images.githubusercontent.com/29249923/218093249-42581d0e-af3c-482d-bc37-43ba4d05f016.png)

PR #1798, opened 2023-02-13
p2p: fix connection metrics

corverroos:
Refactor `p2p_peer_connection_types` metric, calculating it from scratch periodically instead of stateful approach.

category: bug
ticket: #1790

### DIFF
--- a/p2p/p2p.go
+++ b/p2p/p2p.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/libp2p/go-libp2p"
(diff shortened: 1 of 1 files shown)


p2p/p2p.go
```diff
@@ -196,16 +197,17 @@ func RegisterConnectionLogger(ctx context.Context, tcpNode host.Host, peerIDs []
...
 
 	var (
-		peers      = make(map[peer.ID]bool)
-		events     = make(chan logEvent)
-		typeCounts = make(map[connType]int)
+		peers  = make(map[peer.ID]bool)
+		events = make(chan logEvent)
+		ticker = time.NewTicker(time.Second * 30)
 	)
+	defer ticker.Stop()
```
gsora, line 210: Won't this line stop the ticker at the end of this function scope?

I believe this should be moved around like 255.

corverroos: good catch!

gsora (approved):

Issue #1311, opened 2022-10-18
Make it easier to recombine keys


OisinKyne:
## Problem to be solved

Potential customers expressed their concern that they can't bail from a DV fast. This has been intentional for us so far, but in practice it makes sense to hear customers say "we need a way to safely unwind a DV and go with a normal validator if something goes wrong". 

## Proposed solution

We have a variety of options for implementing this, and I think they have different tradeoffs. 

- We could do something slick and automated and involve the network
- We can do something simple and basic and offline

I am a proponent of the latter, which is basically bringing in the [`combine.go`](https://github.com/ObolNetwork/charon/blob/main/testutil/combine/main.go#L37) program into the main charon cli interface. It puts the onus on assembling the private keys in one place on the operators, and leaves nothing to the imagination with respect to this being a centralising option. 

My suggestion is we add a command (`charon combine`) 
With short description: `Combines private key shares into a single private key for a DV.` 

And long description `Combines private key shares into a single private key for a DV. Warning, running this private key in a validator alongside the original distributed validator will result in slashing.`

I think the default input dir could be `.`. Which means that the command would scan the current folder for `.charon`'s

Also in scope of this ticket is documenting this process in the docs site with good [`tree`](https://formulae.brew.sh/formula/tree) outlines of the folder hierarchy. 

Could also invest in lots of good error handling to help someone get the folder structure into something that works for the command. Unhelpful file parsing errors here will be an unpleasant UX for the probably distressed user. 

### Example input

If we want to combine a 5 of 7, one operator must gather a threshold of .charon folders, and must put them alongside one another. This command will traverse this folder (going 1 deep to find lock files and enrs), and will combine the private keys and output them to a folder of keystores and passwords at the end if it can find everything it needs. 

`charon combine --data-dir=example_combine`

Folder structure:
```
 tree -a example_combine
example_combine
├── .charon-cgvhb
│   ├── charon-enr-private-key
│   ├── cluster-lock.json
│   ├── deposit-data.json
│   └── validator_keys
│       ├── keystore-0.json
│       └── keystore-0.txt
├── .charon-dsf
│   ├── charon-enr-private-key
│   ├── cluster-definition.json
│   ├── cluster-lock.json
│   ├── deposit-data.json
│   └── validator_keys
│       ├── keystore-0.json
│       └── keystore-0.txt
├── .charon-egdfg
│   ├── charon-enr-private-key
│   ├── cluster-definition.json
│   ├── cluster-lock.json
[...]

PR #1799, opened 2023-02-13
cmd: add `combine` command

gsora:
This commit adds the `combine` command.

It accepts an input directory path containing all the private key shares of a DV, an output directory path, and a lockfile path.

While the output directory can be nonexistent (will be created automatically), input and lockfile paths must exist.

The result of this process is a keystore file containing the combined private key of the distributed validator.

category: feature
ticket: #1311


cmd/combine.go
```diff
@@ -0,0 +1,63 @@
...
+		lockfile  string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "combine",
+		Short: "Combines private key shares into a single private key for a distributed validator.",
+		Long:  "Combines private key shares into a single private key for a distributed validator.\nWarning: running the resulting private key in a validator alongside the original distributed validator will result in slashing.",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runFunc(lockfile, inputDir, outputDir)
```
corverroos, line 40: can pass `cmd.Context()` to the runFunc

gsora: Addressed

cmd/combine.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func newCombineFunc(ctx context.Context, lockfile, inputDir, outputDir string) error {
+	return combine.Combine(ctx, lockfile, inputDir, outputDir)
+}
+
+func bindCombineFlags(flags *pflag.FlagSet, inputDir, outputDir, lockfile *string) {
+	flags.StringVar(inputDir, "keyfile-dir", "./validator-keys", `Directory containing all the "keyfile-N.json" and "keyfile-N.txt" files`)
+	flags.StringVar(outputDir, "out-dir", "./recombined-validator-key", "Directory where to save the resulting private key")
+	flags.StringVar(lockfile, "lockfile-path", "./lock.json", "Charon lock.json file")
```
corverroos, line 61: suggest aligning with `run` flag: `lock-file`

cmd/combine.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return cmd
+}
+
+func newCombineFunc(ctx context.Context, lockfile, inputDir, outputDir string) error {
+	return combine.Combine(ctx, lockfile, inputDir, outputDir)
+}
+
+func bindCombineFlags(flags *pflag.FlagSet, inputDir, outputDir, lockfile *string) {
+	flags.StringVar(inputDir, "keyfile-dir", "./validator-keys", `Directory containing all the "keyfile-N.json" and "keyfile-N.txt" files`)
```
corverroos, line 59: I think it is `keystore` not `keyfile`

corverroos (approved):

cmd/combine.go
OisinKyne: How does this work for e.g. a 10 validator, 4 node cluster? Do you put 40 files in a flat folder? or 4 subfolders? do we do any traversing/exploring subdirectories?

dB2510, line 61:
```suggestion
	flags.StringVar(lockfile, "lock-file", "./cluster-lock.json", "The path to the cluster lock file defining distributed validator cluster.")
```

dB2510 (approved):

cmd/combine.go
gsora: I believe this command should be require some work from the user: to recombine all the keys, it's assumed some work is needed to arrange them in the correct order.

If we want to automate some part of this, we could (as you suggested) let the user provide the `validator_keys` directories renamed with the public key share, and do some directory traversing to recombine.

I still believe though that each validator recombination needs a separate invocation of the `combine` command.

OisinKyne: Mmmm similar to exits, I believe in practice we will see people mostly combine/exit all validators in a cluster or none. I think partial will be a rare edge case. With the scale of enterprise deployments, I don't think a separate invocation per key is scalable, everyone will DIY their own shell script for-loop. 

I think @Battenfield should user test some options, but I think 'people will send their `.charon` folder to the combiner, the combiner will dump all `n` folders into one directory, having to rename each of them something other than .charon (awkward but feels unavoidable), and from there I think the should be able to run `charon combine --data-dir ./this-folder-ive-prepared`, from where we should be able to traverse it, assuming that each subdir is a .charon folder, and we can find `./this-folder-ive-prepared/*/cluster-lock.json` in each of them, and even a `./this-folder-ive-prepared/*/validator_keys/*` path that we can use to find hopefully all of the keystores we expect. Assuming we can find everything we need, get or create `output-dir` and spit out the combined keystores.

gsora: > everyone will DIY their own shell script for-loop

Correct, I implied users will do something similar... I don't see it as being completely unfeasible, and on the flipside depending on deployments needs it could actually be more comfortable.

I imagine a K8S deployment where validator keys are secrets, you have to script your way around the secrets API to mount directories in pods and have `combine` work with them anyway.

Regardless, I'm open to changes on this.

cmd/combine.go
gsora: Addressed.

cmd/combine.go
OisinKyne: Yeah, you make me also wonder should we add a `keymanager-address` flag like the DKG and create cluster command, that can push the combined keys into a web3 keymanager. 

I think humour me on this one and we go with this `combine` command attempting to combine every private key in the lockfile and warns/(errors and needs to be flag overridden) if it can't discover a threshold of the private keys and passwords it needs to complete.

gsora: @corverroos @OisinKyne considering this process yields standard private keys, I think it's better to write a file with the private key bytes to disk instead of our own keystore files: if a group of operator wants to recombine their private key(s), most probably they want to stop use charon, so writing keystore files seems a bit pointless.

Thoughts?

gsora:
Removing the `do-not-merge` tag since the code is ready for prime time.

@OisinKyne I added some docs on what the `combine` command expects from the user in order to work on the function doc itself, I'll replicate them in the obol-docs as soon as I have an ACK on the UX.

cmd/combine.go
OisinKyne: @gsora the intention behind writing a keystore is to align with an ethereum standard here https://eips.ethereum.org/EIPS/eip-2335 This is a format supported by all  clients. The idea is they would import these keystores into the clients as if they were exported from the eth2-staking-cli. Our `.charon/validator_keys` folder also matches that structure (albeit with key *shares* not 'normal' keys)

gsora: Ok, I’ll write keystore files then. 

The rest of the UX is fine for you?

cmd/combine.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/obolnetwork/charon/combine"
+)
+
+func newCombineCmd(runFunc func(ctx context.Context, inputDir string, force bool) error) *cobra.Command {
+	var (
+		inputDir string
```
dB2510, line 29: nit: change name of this variable to `keystoresDir`

combine/combine.go
```diff
@@ -0,0 +1,171 @@
...
+	lfPath := filepath.Join(inputDir, "cluster-lock.json")
+	b, err := os.Open(lfPath)
+	if err != nil {
+		return errors.Wrap(err, "read lock file")
+	}
+
+	var lock cluster.Lock
+	if err := json.NewDecoder(b).Decode(&lock); err != nil {
+		return errors.Wrap(err, "unmarshal lock file")
+	}
```
dB2510, line 52: you can do what we do here: https://github.com/ObolNetwork/charon/blob/bf0f4694d0e4a505105d2175138b010146464252/app/disk.go#L29

combine/combine.go
```diff
@@ -0,0 +1,171 @@
...
+	tblsv2 "github.com/obolnetwork/charon/tbls/v2"
+)
+
+// Combine combines validator keys contained in inputDir, and writes the original BLS12-381 private keys.
+// Combine is validator-aware: it'll recombine all the validator keys listed in the "Validator" field of the lock file.
+// To do so, the user must prepare inputDir as follows:
+//   - place the lock file in input dir, named as "cluster-lock.json"
+//   - create one directory for each operator, named after their ENR
+//   - place in each of those directories the content of the "validator_keys" directory, contained in their Charon runtime
+//     directory
```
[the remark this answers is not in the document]
gsora: I'm not sure it would look the best in godoc.

I would add it in the web documentation.

combine/combine.go
gsora: Implemented!

gsora:
@OisinKyne instead of creating a single `validator_keys` output directory, I create one for each validator named after its public key.

I implemented what you suggested wrt the input directory tree (aka it's compatible with `.charon` directory scheme).

Is that fine?

combine/combine.go
```diff
@@ -0,0 +1,226 @@
...
+func loadLockfile(dir string) (cluster.Lock, []string, error) {
+	root, err := os.ReadDir(dir)
+	if err != nil {
+		return cluster.Lock{}, nil, errors.Wrap(err, "can't read directory")
+	}
+
+	lfFound := false
+	lastLockfileHash := [32]byte{}
+	lfContent := []byte{}
+	possibleValKeysDir := []string{}
```
corverroos, line 170: nit: use `var` style when declaring but not initialising variables: https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_use_a_consistent_declaration_style

corverroos:
```
var (
  lfFound bool
  lastLockFileHash [32]byte
  ...
)
```

OisinKyne:
> @OisinKyne instead of creating a single `validator_keys` output directory, I create one for each validator named after its public key.
> 
> I implemented what you suggested wrt the input directory tree (aka it's compatible with `.charon` directory scheme).
> 
> Is that fine?

Have you tried importing private keys in that format into validator clients? I don't believe that structure would be recognised. See:
https://lighthouse-book.sigmaprime.io/validator-import-launchpad.html#instructions
https://chainsafe.github.io/lodestar/usage/validator-management/#option-1-import-keys-to-lodestars-keystores-folder
https://docs.teku.consensys.net/Reference/CLI/CLI-Syntax#validator-keys
https://docs.prylabs.network/docs/wallet/nondeterministic#import-validator-accounts


Second thing I want to check is that the format `charon create cluster` outputs can immediately work with `charon combine ...` will do now

corverroos (approved):

OisinKyne:
Screenshot 2023-02-22 at 12 47 20
Okay this is way old, was under the impression this was changed, will make tickets for this

gsora:
> I think only one validator_keys folder, files starting with keystore- and .json/.txt for file and password are all important conventions.

Oh I see!

Ok, I'll move the json and txt file down under a validator_keys directory, then I'd say we're good to go with the merge.

combine/combine.go
gsora: Yup, will refactor.

Old habits die hard.

combine/combine.go
```diff
@@ -0,0 +1,226 @@
...
+	var lock cluster.Lock
+	if err := json.Unmarshal(lfContent, &lock); err != nil {
+		return cluster.Lock{}, nil, errors.Wrap(err, "unmarshal lock file")
+	}
+
+	if err := lock.VerifyHashes(); err != nil {
+		return cluster.Lock{}, nil, errors.Wrap(err, "cluster lock hash verification failed")
+	}
+
+	if err := lock.VerifySignatures(); err != nil {
```
OisinKyne, line 224: Should we allow `--no-verify` flags on this command @corverroos ? Imagined use case would be trying to recombine a really old cluster in a modern (>v1.0) charon version, and not being able to get it to verifySignatures correctly? (Or we can just pinky promise to always support old signatures/verification for anything that could possibly be in prod?)

Happy to skip this if not a concern

OisinKyne (approved):

OisinKyne (approved):
Just want a slight change to the output format, a couple of tweaks to the explainer text. 

For now, I've suggested a different default folder path, I don't like what I've suggested as `.charon/cluster`, but that is where the default create goes, so I think we should put the default combine there too, and fix both shortly together.

combine/combine.go
```diff
@@ -0,0 +1,226 @@
...
+			return errors.New("refusing to overwrite existing private key", z.Int("validator_number", idx), z.Str("path", ksPath))
+		}
+
+		if err := keystore.StoreKeys([]tblsv2.PrivateKey{secret}, outPath); err != nil {
+			return errors.Wrap(err, "cannot store keystore", z.Int("validator_number", idx))
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
gsora: Hmm, the function will error out if one of the validator keys can't be combined but the artifacts of the other keys will remain.

cmd/combine.go
```diff
@@ -0,0 +1,59 @@
...
+
+	return cmd
+}
+
+func newCombineFunc(ctx context.Context, keystoresDir string, force bool) error {
+	return combine.Combine(ctx, keystoresDir, force)
+}
+
+func bindCombineFlags(flags *pflag.FlagSet, keystoresDir *string, force *bool) {
+	flags.StringVar(keystoresDir, "keystores-dir", "./", `Directory containing all the keystore files organized by ENR, and the lock file.`)
```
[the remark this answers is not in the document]
gsora: I would push for `./.charon` to keep it aligned with all the other cli commands.

combine/combine.go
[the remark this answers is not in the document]
gsora: I'm merging this PR, let's address this behavior in a separate one if needs a change.

combine/combine.go
gsora: I'm merging this PR, let's address this behavior in a separate one if needs a change.

PR #1805, opened 2023-02-14
cmd/relay: fix connection limits

corverroos:
Fixes issue with relay limits not being applied correctly getting stuck at 256 connections.

category: bug
ticket: none


dB2510 (approved):

app/app_test.go
```diff
@@ -398,6 +400,46 @@ func TestInfoSync(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestRelayConnections(t *testing.T) {
```
LukeHackett12, line 403: Could be good to add a negative case to show it failed before this

corverroos: ok, added 1025th connection which should error

PR #1810, opened 2023-02-15
cmd: route libp2p to charon logger

corverroos:
Fixes issue with libp2p using its own logger. Instead, it now logs via the configured charon logger (incl pushing to loki if enabled). 

category: refactor
ticket: none


app/log/config.go
```diff
@@ -327,20 +336,21 @@ func (e consoleEncoder) EncodeEntry(ent zapcore.Entry, fields []zap.Field) (*buf
...
-			ent.LoggerName = topic
+			ent.LoggerName = f.String
 
 			continue
 		}
 
 		filtered = append(filtered, f)
 	}
 
+	ent.LoggerName = (ent.LoggerName + "          ")[:10] // Align topic/LoggerName spacing.
```
corverroos, line 347: pad and color `LoggerName` not just topic, libp2p topics are used.

dB2510 (approved):

cmd/relay/relay.go
```diff
@@ -52,7 +52,7 @@ type Config struct {
 	AutoP2PKey     bool
 	MaxResPerPeer  int
 	MaxConns       int
-	RelayLogLevel  string
+	RelayLogLevel  string // TODO(corver): Rename to LibP2PLogLevel.
```
xenowits, line 55: why did you defer this renaming refactor?

corverroos: since it includes a flag rename and we have old relays still running, want to avoid two different flags in production

Issue #1775, opened 2023-02-08
Include deposit data in the lockfile


OisinKyne:
# 🎯 Problem to be solved

A client wants to be a DV creator via the API. The create a definition file. The operators use the launchpad operator flow to sign and do the dkg. They publish the definition file. The creator can see the definition file, and the validators it would like to activate, however, it does not have the deposit data to send to the deposit contract. Those files are still sitting on operators devices where the creator can't access.

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Add an optional field to the validators object in the distributed_validators array called `deposit`, that includes a valid deposit message for the corresponding `withdrawal_address` alongside it. 

# 🧪 Tests

- Testcase: Lockfile validation should fail if `distributed_validators[*].deposit.withdrawal_credential  !== "0x010000000000"+ distributed_validators[*].withdrawal_address[2:]`

- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

corverroos:
Add the following field:
```
{
  definition: {...}
  distributed_validators: [
    {
      distributed_public_key: "0x123...",
      public_shares: [....]
      deposit: {
       {
          "pubkey": "80d0...",
          "withdrawal_credentials": "0100000...",
          "amount": 32000000000,
          "signature": "ab757712...",
          "deposit_message_root": "b4376b...",
          "deposit_data_root": "d402b324...",
          "fork_version": "00001020",
          "network_name": "goerli",
          "deposit_cli_version": "2.3.0"
        },
      }
    }, ...
  ]
}
```

PR #1813, opened 2023-02-16
cluster: draft v1.6 lock with deposit data

corverroos:
Adds a new draft lock version v1.6 which includes deposit data in the lock's distributed validators.

category: feature
ticket: #1775


cluster/deposit.go
```diff
@@ -0,0 +1,60 @@
...
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cluster
+
+// DepositData defines the deposit data to activate a validator.
+// https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#depositdata
+type DepositData struct {
```
dB2510, line 20: q: any reason behind not going with `eth2p0.DepositData` type?

xenowits (approved):

cluster/distvalidator.go
```diff
@@ -52,13 +55,21 @@ type distValidatorJSONv1x1 struct {
...
 	PubKey              ethHex   `json:"distributed_public_key"`
 	PubShares           []ethHex `json:"public_shares,omitempty"`
 	FeeRecipientAddress ethHex   `json:"fee_recipient_address,omitempty"`
 }
 
+// distValidatorJSONv1x6 is the json formatter of DistValidator for versions v1.6.0 or later.
+type distValidatorJSONv1x6 struct {
+	PubKey              ethHex          `json:"distributed_public_key"`
+	PubShares           []ethHex        `json:"public_shares,omitempty"`
+	FeeRecipientAddress ethHex          `json:"fee_recipient_address,omitempty"`
```
dB2510, line 69: i think we should remove this field `FeeRecipientAddress` from here since it is not part of `DistValidator` and we don't use it

dB2510 (approved):

cluster/deposit.go
corverroos: I prefer not to couple our cluster definition with 3rd-party types. It should be exactly the same as `eth2p0.DepositData`.

cluster/distvalidator.go
corverroos: yes indeed, should be removed, it isn't used.

Issue #1403, opened 2022-11-09
Add vouch to compose


dB2510:
## Problem to be solved

We currently support teku and lighthouse as part of our compose tool to run a test cluster on simnet. We have started supporting vouch recently. We need to add vouch beside teku and lighthouse in compose to be included in smoke tests.

## Proposed solution

Follow [charon-distributed-validator-cluster](https://github.com/ObolNetwork/charon-distributed-validator-cluster/blob/main/docker-compose.yml) repo to add support to `testutil/compose`. Also add a separate smoke test in `testutil/compose/compose/smoke_internal_test` for vouch.

dB2510:
Closed by #1814

PR #1814, opened 2023-02-16
compose: add vouch to compose

dB2510:
Adds vouch to compose and also made some changes to beaconmock which were needed by vouch specifically.
Note: Vouch doesn't support block proposals because of RANDAO check in go-eth2-client.

category: feature
ticket: #1403

### DIFF
--- a/testutil/beaconmock/headproducer.go
+++ b/testutil/beaconmock/headproducer.go
@@ -49,9 +49,9 @@ type headProducer struct {
 	quit   chan struct{}
 
 	// Mutable state
-	streamsByTopic map[string][]string
 	mu             sync.Mutex
 	currentHead    *eth2v1.HeadEvent
+	streamsByTopic map[string][]string
 }
 
 // Start starts the internal slot ticker that updates head.
@@ -96,6 +96,20 @@ func (p *headProducer) setCurrentHead(currentHead *eth2v1.HeadEvent) {
 	p.currentHead = currentHead
 }
 
+func (p *headProducer) getStreamIDs(topic string) []string {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	return p.streamsByTopic[topic]
+}
+
+func (p *headProducer) setStreamIDs(topic string, streamID string) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.streamsByTopic[topic] = append(p.streamsByTopic[topic], streamID)
+}
+
 // updateHead updates current head based on provided slot.
 func (p *headProducer) updateHead(slot eth2p0.Slot) {
 	currentHead := pseudoRandomHeadEvent(slot)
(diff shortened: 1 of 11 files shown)


testutil/beaconmock/headproducer.go
```diff
@@ -176,7 +197,28 @@ func (p *headProducer) handleEvents(w http.ResponseWriter, r *http.Request) {
 	query := r.URL.Query()
 	query.Set("stream", sseStreamID) // Add sseStreamID for sse server to serve events on.
 	r.URL.RawQuery = query.Encode()
-	p.server.ServeHTTP(w, r)
+
+	for _, topic := range query["topics"] {
```
corverroos, line 201: this will break if multiple topics requested

dB2510: yeah initially i tried with single server that handles different topics but on the vouch side it has 2 different handlers for head and block events and it panics if head event handler gets block event and vice versa: https://github.com/attestantio/vouch/blob/9b8114dc0374d7e5bb6b01ce5f519213bfe30d37/services/cache/standard/service.go#L77

Here's the head event handler: https://github.com/attestantio/vouch/blob/9b8114dc0374d7e5bb6b01ce5f519213bfe30d37/services/cache/standard/events.go#L39

Here's the block event handler: https://github.com/attestantio/vouch/blob/9b8114dc0374d7e5bb6b01ce5f519213bfe30d37/services/cache/standard/events.go#L27

testutil/beaconmock/headproducer.go
```diff
@@ -95,20 +96,55 @@ func (p *headProducer) setCurrentHead(currentHead *eth2v1.HeadEvent) {
...
-		Event: []byte("head"),
-		Data:  data,
-	})
+	blockData, err := json.Marshal(currentBlock)
+	if err != nil {
+		panic(err) // This should never happen and this is test code sorry ;)
+	}
+
+	// Publish head events.
+	for _, streamID := range p.getStreamIDs("head") {
```
corverroos, line 134: make these constants `topicHead` and `topicBlock`

testutil/beaconmock/headproducer.go
```diff
@@ -173,9 +209,33 @@ func (p *headProducer) handleGetBlockRoot(w http.ResponseWriter, r *http.Request
...
+	p.server.CreateStream(streamID)
+
 	query := r.URL.Query()
-	query.Set("stream", sseStreamID) // Add sseStreamID for sse server to serve events on.
+	query.Set("stream", streamID) // Add sseStreamID for sse server to serve events on.
 	r.URL.RawQuery = query.Encode()
+
+	for _, topic := range query["topics"] {
+		if topic != "head" && topic != "block" {
+			log.Warn(context.Background(), "Unknown topic requested", nil, z.Str("topic", topic))
```
corverroos, line 222: Unsupported not unknown

corverroos (approved):

testutil/beaconmock/headproducer.go
```diff
@@ -173,9 +209,33 @@ func (p *headProducer) handleGetBlockRoot(w http.ResponseWriter, r *http.Request
 
 // handleEvents is a http handler to handle "/eth/v1/events".
 func (p *headProducer) handleEvents(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
dB2510: Done!

corverroos (approved):

PR #1817, opened 2023-02-17
*: update go version to 1.20.1

dB2510:
Updates go version to 1.20.1. Also updates versions of go-libp2p to v0.25.1 and golangci-lint to v1.51.1.

category: misc
ticket: none


dB2510:
First we need to upgrade go-libp2p version and then upgrade go version. Created a ticket for the same: #1821

PR #1823, opened 2023-02-20
gomod: upgrade libp2p 

corverroos:
Upgrades libp2p to v0.25.1

category: refactor
ticket: none


dB2510 (approved):
why v0.23.4?

p2p/errors_internal_test.go
```diff
@@ -102,3 +103,26 @@ func TestDialErrMsgs(t *testing.T) {
 	require.Contains(t, msgs[badAddr.String()], "dial backoff")
 	require.Contains(t, msgs[hostBAddr.String()], "dial backoff")
 }
+
+// closedGater is a connection gater that blocks all connections.
+type closedGater struct{}
+
+func (closedGater) InterceptPeerDial(peer.ID) bool {
```
dB2510, line 110: okay so what we were doing earlier returned true even with nil, parameters which is why this test was failing

dB2510 (approved):

PR #1830, opened 2023-02-22
core: fix duty expired context tracker2 

dB2510:
Fixes duty expired context in tracker2. This happens when duty is already expired and retryer calls tracker2 with a context which has timed out resulting in closing of ctx.Done() channel. Due to this some events don't get registered in tracker2:
https://github.com/ObolNetwork/charon/blob/eeab76e1f388af478052734eea61803df8f5f38d/core/tracker/tracker2/tracker.go#L741

category: bug
ticket: #1478 
feature_flag: tracker_v2

### DIFF
--- a/core/tracking.go
+++ b/core/tracking.go
@@ -20,61 +20,61 @@ import (
 )
 
 // WithTracking wraps component input functions to support tracking of core components.
-func WithTracking(tracker Tracker) WireOption {
+func WithTracking(parentCtx context.Context, tracker Tracker) WireOption {
 	return func(w *wireFuncs) {
 		clone := *w
 
 		w.FetcherFetch = func(ctx context.Context, duty Duty, set DutyDefinitionSet) error {
 			err := clone.FetcherFetch(ctx, duty, set)
-			tracker.FetcherFetched(ctx, duty, set, err)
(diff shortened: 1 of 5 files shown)


dB2510:
This caused inconsistencies in tracker2 and tracker Duty failed logs:
Screenshot 2023-02-22 at 2 33 26 PM

core/tracking.go
```diff
@@ -20,61 +20,61 @@ import (
...
 // WithTracking wraps component input functions to support tracking of core components.
-func WithTracking(tracker Tracker) WireOption {
+func WithTracking(parentCtx context.Context, tracker Tracker) WireOption {
 	return func(w *wireFuncs) {
 		clone := *w
 
 		w.FetcherFetch = func(ctx context.Context, duty Duty, set DutyDefinitionSet) error {
 			err := clone.FetcherFetch(ctx, duty, set)
-			tracker.FetcherFetched(ctx, duty, set, err)
+			tracker.FetcherFetched(parentCtx, duty, set, err)
```
corverroos, line 29: rather remove the context from the tracker interface methods. Since it is a "post-fact" thing, ctx isn't required.

dB2510: Done!

corverroos (approved):

Issue #1552, opened 2022-12-13
Reduce QBFT throughput 


corverroos:
## Problem to be solved

QBFT currently uses a lot of bandwidth (network bytes sent and received). So much so, that memory spikes to 1GB when doing block proposals of 300KB. This is due to a lot of duplication in the protocol and wire formats.

## Proposed solution

Incorporate these optimisations:
 - use SSZ compression for proposed values (instead of json), saves 50%. 
 - Use nested messages with signature wrapper types to avoid copying protos when verifying signatures, saves 50%.
 - QBFT messages should only contain hashes, only include uniq map of hashes to original values in a wrapper type, saves 90%. 
 - Avoid all in-memory duplication of original values. Only buffer hash based QBFT messages. 
 
This is a big protocol level refactor, implement this as QBFT v2 and use prioritise to enable/disable.

PR #1838, opened 2023-02-28
core/consensus: add support for pointer values on the wire

corverroos:
Upgrades the QBFT wire protocol (without breaking backwards compatibility) by adding the whole value to the `ConsensusMsg` envelope and hashes of those values to the `QBFTMsg` itself. This will allow us to remove the values from the QBFT messages, which will reduce network bandwidth usage by a lot (90% for large values; beacon blocks).

This increases the network bandwidth for next release v0.14, but once we remove support for v0.13, it will reduce it by a lot.

category: feature
ticket: #1552


core/consensus/msg_internal_test.go
```diff
@@ -73,150 +69,28 @@ func TestSigning(t *testing.T) {
 	require.False(t, ok)
 }
 
-func TestBackwardsCompatibility(t *testing.T) {
```
corverroos, line 76: this tests 2 updates ago, we already removed support, can remove now.

core/corepb/v1/consensus.proto
```diff
@@ -13,31 +13,20 @@ message QBFTMsg {
...
+  bytes                          value_hash            = 11;
+  bytes                          prepared_value_hash   = 12;
 }
 
 message ConsensusMsg {
-  QBFTMsg          msg           = 1;
-  repeated QBFTMsg justification = 2;
+  QBFTMsg                      msg           = 1;
+  repeated QBFTMsg             justification = 2;
+  repeated google.protobuf.Any values        = 3;
```
corverroos, line 29: so the values are added here and can therefore be removed from `QBFTMsg` above once we remove support for v0.13.

core/consensus/component.go
```diff
@@ -545,3 +568,22 @@ func newRoundTimer(round int64) (<-chan time.Time, func()) {
 	timer := time.NewTimer(roundStart + (time.Duration(round) * roundIncrease))
 	return timer.C, func() { timer.Stop() }
 }
+
+func valuesByHash(values []*anypb.Any) (map[[32]byte]*anypb.Any, error) {
+	resp := make(map[[32]byte]*anypb.Any)
```
gsora, line 573: Should this function return error if values is empty?

core/corepb/v1/consensus.proto
gsora: Nit: can we add a comment for future reference?

gsora (approved):
General approve because it looks good.

I wonder if we could get rid of the `*anypb.Any` and deal with copies instead.

core/consensus/component.go
corverroos: nope, empty values result in a empty map, this is required for legacy v0.13 support.

core/corepb/v1/consensus.proto
corverroos: sure

corverroos:
Since the `Anys` wrap protobufs, the wrapped objects are still pointers unfortunately.

PR #1866, opened 2023-03-09
core: fix data race in scheduler

xenowits:
Fixes data race in scheduler.go. 

category: fixbuild
ticket: none


core/scheduler/scheduler.go
```diff
@@ -203,8 +203,8 @@ func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
 					return
```
corverroos, line 203: missed the log.Error on line 202

corverroos (approved):

PR #1869, opened 2023-03-09
cmd/createcluster: handle multiple addresses with remote definition

dB2510:
Handle multiple addresses error in `charon create cluster` command for remote definition.

category: bug
ticket: none


cmd/createcluster.go
```diff
@@ -161,6 +151,11 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return err
 	}
 
+	// Create cluster directory at the given location.
+	if err := os.MkdirAll(conf.ClusterDir, 0o755); err != nil {
```
dB2510, line 155: create cluster directory just before writing the files

Issue #1879, opened 2023-03-14
Container healthcheck not working


t3d3r3:
# 🐞 Bug Report

### Description

 Container healthcheck fails as wget is not installed.

### Has this worked before in a previous version?

 
 Yes, the previous version in which this bug was not present was: v0.13.0

## 🔬 Minimal Reproduction

Run using docker compose: https://github.com/ObolNetwork/charon-distributed-validator-node/blob/main/docker-compose.yml
docker status:
```
      Name                       Command                   State                                              Ports                                       
------------------------------------------------------------------------------------------------------------------------------------------------------------
obol-bia_charon_1       /usr/local/bin/charon run        Up (unhealthy)   0.0.0.0:3610->3610/tcp,:::3610->3610/tcp                                          
```

## 🔥 Error

<pre><code>

docker inspect obol-bia_charon_1
[
    {
        "Id": "c3343c6d08336fdb606f65d8baa0fe21f19f7869dcd06eb81cd82bb89e63dc75",
        "Created": "2023-03-14T10:12:35.381398597Z",
        "Path": "/usr/local/bin/charon",
        "Args": [
            "run"
        ],
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 938207,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2023-03-14T10:13:12.018667986Z",
            "FinishedAt": "2023-03-14T10:12:58.922153009Z",
            "Health": {
                "Status": "unhealthy",
                "FailingStreak": 396,
                "Log": [
                    {
                        "Start": "2023-03-14T13:29:44.210431289Z",
                        "End": "2023-03-14T13:29:44.286692087Z",
                        "ExitCode": 127,
                        "Output": "/bin/sh: 1: wget: not found\n"
                    },
                    {
                        "Start": "2023-03-14T13:30:14.311212516Z",
                        "End": "2023-03-14T13:30:14.394728602Z",
                        "ExitCode": 127,
                        "Output": "/bin/sh: 1: wget: not found\n"
                    },
                    {
                        "Start": "2023-03-14T13:30:44.396305226Z",
                        "End": "2023-03-14T13:30:44.493960184Z",
                        "ExitCode": 127,
                        "Output": "/bin/sh: 1: wget: not found\n"
                    },
.....
</code></pre>

## 🌍  Your Environment

**Operating System:**

[...]

PR #1880, opened 2023-03-14
dockerfile: fix container healthcheck

t3d3r3:
Wget is needed for the container healthcheck as defined here: https://github.com/ObolNetwork/charon-distributed-validator-node/blob/main/docker-compose.yml  

Moved to slim base in commit 8c1dcc6bfa2e4ba1f2db9b54dfaf8a953a43defa but slim image does not include wget anymore, needs to be added.

category: bug
ticket: #1879

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN \
 
 # Copy final binary into light stage.
 FROM debian:bullseye-slim
-RUN apt-get update && apt-get install -y ca-certificates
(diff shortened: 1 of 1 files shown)


dB2510:
Hey @t3d3r3 👋 
Thank you for your fix!
Have you tested this fix locally? If not you can do so by running following commands:
```
cd charon
docker build . -t obolnetwork/charon:local
```
Then update charon image tag to `local` in `charon-distributed-validator-node`/docker-compose.yml.

Issue #1895, opened 2023-03-17
DKG must error when peer runs different version


corverroos:
# 🎯 Problem to be solved
The DKG doesn't (and SHOULD not) support backwards and forwards compatibility with others versions. Since it is a once-off command, all peers MUST run the latest version of the app.

This allows us to make breaking changes to the DKG protocol without needing go through the tedious process of "stepwise-migrations" and communicating the "version compatibility matrix" to customers. 
 
# 🛠️ Proposed solution
Add the `version` field to the sync protocol, server should error if it detects mismatching versions.

PR #1901, opened 2023-03-18
dkg/sync: enforce version

corverroos:
Enforce matching peer minor version. DKG only guarantees compatibility with patch versions.

category: bug
ticket: #1895


dkg/sync/server.go
```diff
@@ -179,24 +190,26 @@ func (s *Server) handleStream(ctx context.Context, stream network.Stream) error
...
+		} else if ok, err := pubkey.Verify(s.defHash, msg.HashSignature); err != nil { // Note: libp2p verify does another hash of defHash.
 			return errors.Wrap(err, "verify sig hash")
 		} else if !ok {
 			resp.Error = errInvalidSig
-
-			s.mu.Lock()
-			s.errResponse = true
-			s.mu.Unlock()
-
+			s.setError()
```
dB2510, line 204: can we have this message verification as a separate method to server struct?

app/version/version.go
```diff
@@ -53,3 +55,13 @@ func LogInfo(ctx context.Context, msg string) {
 		z.Str("git_commit_time", gitTimestamp),
 	)
 }
+
+// Minor returns the minor version of the provided version string.
+func Minor(version string) (string, error) {
+	split := strings.Split(version, ".")
+	if len(split) < 2 {
```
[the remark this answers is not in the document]
corverroos: Nope, it supports `v0.1` as input and just returns it.

dkg/sync/server.go
corverroos: doesn't really help much

corverroos: good idea!

xenowits (approved):

dkg/sync/server.go
```diff
@@ -209,6 +214,29 @@ func (s *Server) handleStream(ctx context.Context, stream network.Stream) error
 	}
 }
 
+// validReq returns an error message and false if the request version or definition hash are invalid.
+// Else it returns true or an error.
+func (s *Server) validReq(ctx context.Context, pubkey crypto.PubKey, msg *pb.MsgSync) (string, bool, error) {
```
dB2510, line 219: we are reading `s.version` and `s.defHash` without locking mutex

dB2510 (approved):

dkg/sync/server.go
corverroos: they are immutable fields, no need to lock

PR #1902, opened 2023-03-19
app/genssz: implement ssz hash generator

corverroos:
Implement basic SSZ hash generator that reads struct tags.

category: feature
ticket: none


app/genssz/types.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return size
+}
+
+func (f Field) Size() (int, error) {
+	var intStr string
+	if f.IsByteList() || f.IsCompositeList() {
+		openIdx := strings.Index(f.SSZTag, "[")
+		closeIdx := strings.Index(f.SSZTag, "]")
+		intStr = f.SSZTag[openIdx+1 : closeIdx]
```
gsora, line 77: I suggest checking for presence of opening and closing brackets to avoid puzzling errors like this when the expression isn't well-formed: 

```
Fatal run error: write template: exec template: template: :48:76: executing "" at <.Size>: error calling Size: runtime error: slice bounds out of range [:-1]
```

```suggestion
		openIdx := strings.Index(f.SSZTag, "[")
		closeIdx := strings.Index(f.SSZTag, "]")

		if openIdx == -1 || closeIdx == -1 {
			return 0, errors.New("field has malformed size tag", z.Str("field_name", f.Name))
		}

		intStr = f.SSZTag[openIdx+1 : closeIdx]

File: app/genssz/testdata/example.go [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+package testdata
+
+//go:generate go install github.com/obolnetwork/charon/app/genssz
+//go:generate genssz
+
+type Foo struct {
+	ByteList []byte  `ssz:"ByteList[32]"`
+	Number   int     `ssz:"uint64"`
+	Bytes4   [4]byte `ssz:"Bytes4"`
+	Bytes2   []byte  `ssz:"Bytes2"`
```
gsora, line 12: In case of SSZ arrays, should the field be specified as a Go array rather than a slice? (see field `Bytes4`).

If yes, maybe we could refrain from specifying the SSZ size and infer it from the Go struct field size.

app/genssz/testdata/example.go
```diff
@@ -0,0 +1,23 @@
...
+	Bar      Bar     `ssz:"Composite"`
+	Quxes    []Qux   `ssz:"CompositeList[256]"`
+}
+
+type Bar struct {
+	Name string `ssz:"ByteList[32]"`
+}
+
+type Qux struct {
+	Number int `ssz:"uint64"`
```
gsora, line 22: To avoid truncation issues I suggest we do stricter checks on the field integer type and its SSZ side.

gsora (changes requested):

app/genssz/testdata/example.go
corverroos: No, BytesN may also be slices.

app/genssz/testdata/example.go
corverroos: We do not support 32bit OSes, so not sure truncation is an issue?

gsora: True.

Another issue though, an SSZ uint64 could not fit in a int

corverroos: sure, but we never do that conversion, we only use ssz for hashing.

gsora (approved):

PR #1905, opened 2023-03-20
cmd/relay: disable resource manager

corverroos:
Disables the relay's libp2p resource manager so it uses all available resources. Note that the circuitv2 relay resources still apply.

category: feature
ticket: none


dB2510 (approved):
can we test this change in a testing environment rather than directly deploying in production relays?

cc: @LukeHackett12

Issue #1906, opened 2023-03-20
lock file post does not add content type to request


LukeHackett12:
# 🐞 Bug Report

### Description

lock file post does not add content type to request

## 🔬 Minimal Reproduction

create cluster with --publish flag

PR #1907, opened 2023-03-20
dkg: add content type to lock file post

LukeHackett12:
Adds content type header to api post so that it can be understood.

category: bug
ticket: #1906


HananINouman (approved):
👍

app/obolapi/api.go
```diff
@@ -60,6 +60,7 @@ func httpPost(ctx context.Context, url *url.URL, b []byte) error {
 	if err != nil {
 		return errors.Wrap(err, "new POST request with ctx")
 	}
+	req.Header.Add("Content-Type", `application/json`)
```
corverroos, line 63: avoid `backtick` strings.

PR #1910, opened 2023-03-20
app/version: bump charon version to v0.14.3

dB2510:
Bumps charon version to v0.14.3

category: misc
ticket: none


corverroos (changes requested):
As mentioned, this should be a "v0.14" branch on v0.14.2 commit, not main.

PR #1915, opened 2023-03-22
dkg: checks that there isn't a "validator_keys" file in the destination directory

gsora:
Also clarify the error when a "validator_keys" directory already exists.

category: bug
ticket: none


gsora:
Closed in favor of #1919.

PR #1924, opened 2023-03-23
dkg/sync: fix shutdown reconnect deadlock

corverroos:
This fixes a DKG race deadlock on shutdown. When one `sync/client` is busy "connecting with reconnect=true", it blocks forever trying to reconnect even though the DKG is shutting down an `DisableReconnect` has been called.

category: bug
ticket: #1925

### DIFF
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -63,9 +63,7 @@ func (c *Client) Run(ctx context.Context) error {
 			return ctx.Err()
 		}
 
-		retry := c.reconnect // Retry connecting if never connected.
-
-		stream, err := c.connect(ctx, retry)
+		stream, err := c.connect(ctx)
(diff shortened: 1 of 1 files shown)


dkg/sync/client.go
```diff
@@ -63,9 +63,7 @@ func (c *Client) Run(ctx context.Context) error {
 			return ctx.Err()
 		}
 
-		retry := c.reconnect // Retry connecting if never connected.
```
corverroos, line 66: the issue is that we copied this variable and passed it into `connect` below which blocked forever retrying even though `DisableReconnect` was called a little after this.

gsora (approved):
Y'all guys are rockstars

dB2510 (approved):

xenowits (approved):

Issue #1867, opened 2023-03-09
Fix race conditions in charon


xenowits:
# 🎯 Problem to be solved

We recently detected some panics in charon code execution in one of our [CI runs](https://github.com/ObolNetwork/charon/actions/runs/4371211493/jobs/7646868260).

If you wish to see possible race condition scenarios, simply run `go test -race ./...` locally in charon. This reports multiple possible race conditions in the whole codebase. Search for `WARNING: DATA RACE` in the test logs.

# 🛠️ Proposed solution

- [ ] Core team consensus on the proposed solution

* Fix all the panic scenarios reported by the `-race` flag.
* Add `race detection` as a github action (`.github/workflows`).

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

# 👐 Additional acceptance criteria

No race conditions should be reported in CI.

# ❌ Out of Scope

None.

PR #1938, opened 2023-03-27
dkg/sync: fix ctx race in server

corverroos:
Fix context concurrent access bug in `dkg/sync.Server.Start`.

category: bug
ticket: #1867


dkg/sync/server.go
```diff
@@ -243,11 +243,7 @@ func (s *Server) validReq(ctx context.Context, pubkey crypto.PubKey, msg *pb.Msg
 // Start registers sync protocol with the libp2p host.
 func (s *Server) Start(ctx context.Context) {
 	s.tcpNode.SetStreamHandler(protocolID, func(stream network.Stream) {
-		if ctx.Err() != nil {
-			return
-		}
-
-		ctx = log.WithCtx(ctx, z.Str("peer", p2p.PeerName(stream.Conn().RemotePeer())))
```
corverroos, line 250: Overwriting ctx here is the problem. Solution is to shadow ctx instead.

dB2510 (approved):

PR #1942, opened 2023-03-27
dkg: fix race condition in keycast

dB2510:
Fixes race condition in keycast for concurrent write of map.

category: bug
ticket: #1867


dkg/keycast.go
```diff
@@ -121,7 +122,11 @@ func leadKeyCast(ctx context.Context, tp kcTransport, def cluster.Definition) ([
 
 	log.Info(ctx, "Node selected as dealer, serving shares...")
 
+	var serveMutex sync.Mutex
```
corverroos, line 125: suggest defining this in variable group above and add comment, `shared server state`

corverroos (approved):

Issue #1884, opened 2023-03-15
Upgrade all libp2p wire protocols to use length-delimited messages


corverroos:
# 🎯 Problem to be solved

All custom charon libp2p wire protocols send a single raw protobuf message on a stream and then closes the write-stream. The other side reads all bytes from the read-stream until it is closed before it unmarshalls the expected protobuf message.

Although simple, this has the unintended side-affect of needing to wait for the stream to be closed before the message can be processed. It seems in some cases, that the close stream message takes much longer to arrive at the other side than the message itself. Most `LibP2P read timeout` errors indicate that a valid protobuf message was already read before timeout.

# 🛠️ Proposed solution

Libp2p has a simple https://github.com/libp2p/go-msgio library that implements length-delimited write/reader allowing messages to processes as soon as they have been received.

This should decrease read timeout errors and possibly even improve network performance in general.

This is however a breaking change so needs to be implemented as a migration.
  - In next release, support both old and new wire formats
  - In release after that, remove support for old wire format
  
Step 1: Refactor `p2p.Send[Receive]`
  - Add support for providing a mapping of `writerProviderByProtocol map[protocol.ID]func(io.WriteCloser) io.WriteCloser` to `p2p.Send` and `p2p.SendReceive`
  - Maybe do this via functional options
  - Add a convenience `WithDelimitedUpgrade(old, new protocol.ID) Option` that uses a noop provider for old and `pbio.NewDelimitedWriter` for new.
  - When opening the stream, map use the writer of the associated protocol
   
Step 2: Support both old and new
  - Given a protocol with v1.0.0
  - Add a second additional version v2.0.0
  - Support sending and receiving with both
  
Step 3:  Only support new
 - Remove support for old.

PR #1945, opened 2023-03-27
p2p: explicitly register fuzzy matched protocols

corverroos:
#1884 (v0.15.0-dev) introduces support for length-delimited libp2p wire protocols using fuzzy protocol matching. This doesn't explicitly register supported protocols which results in v0.14.0 disabling peerinfo and other protocols since it cannot detect supported protocols.

category: bug
ticket: #1948

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -219,7 +219,10 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 	sender := new(p2p.Sender)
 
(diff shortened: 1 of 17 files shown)


dB2510 (dismissed):
LGTM

corverroos:
Not sure this fixes it, need to add tests first

corverroos:
We'll need to delay length delimited wire protocols, first need to remove the naive "p2p.ProtocolSupported".

PR #1950, opened 2023-03-27
core/aggsigdb: refactor concurrent access issue in test

gsora:
Use sync.Map with a thin abstraction for data and keys by duty maps.

category: bug
ticket: #1867


corverroos (changes requested):
aggsigdb has a central goroutine that accesses all state, so sycnronisation not required.

gsora:
Will fix the test instead.

Issue #1956, opened 2023-03-28
Relay `peerinfo` protocol is failing


corverroos:
# 🐞 Bug Report

### Description

Relay `peerinfo` protocol is failing with latest main peers since they error on 0 length requests.

### Proposed Solution

Remove request length check for stream delimited protocols.

PR #1957, opened 2023-03-28
p2p: support zero length request protos

corverroos:
Fix regression introduced by length-delimited upgrades which assumed that legacy-stream-delimited protocols handled requests and responses the same while it didn't. It allowed zero length requests while it didn't allow zero-length responses. This aligns the new implementation with that.

category: bug
ticket: #1956

### DIFF
--- a/p2p/sender.go
+++ b/p2p/sender.go
@@ -143,10 +143,11 @@ func withRelayRetry(fn func() error) error {
 type SendRecvOption func(*sendRecvOpts)
 
 type sendRecvOpts struct {
-	protocols         []protocol.ID // Protocols ordered by higher priority first
-	writersByProtocol map[protocol.ID]func(network.Stream) pbio.Writer
-	readersByProtocol map[protocol.ID]func(network.Stream) pbio.Reader
-	rttCallback       func(time.Duration)
+	protocols             []protocol.ID // Protocols ordered by higher priority first
+	writersByProtocol     map[protocol.ID]func(network.Stream) pbio.Writer
+	reqReadersByProtocol  map[protocol.ID]func(network.Stream) pbio.Reader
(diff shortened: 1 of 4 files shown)


p2p/sender.go
```diff
@@ -175,10 +176,15 @@ func defaultSendRecvOpts(pID protocol.ID) sendRecvOpts {
 	return sendRecvOpts{
 		protocols: []protocol.ID{pID},
 		writersByProtocol: map[protocol.ID]func(s network.Stream) pbio.Writer{
-			pID: func(s network.Stream) pbio.Writer { return legacyReadWriter{s} },
+			pID: func(s network.Stream) pbio.Writer { return legacyReadWriter{stream: s} },
 		},
-		readersByProtocol: map[protocol.ID]func(s network.Stream) pbio.Reader{
-			pID: func(s network.Stream) pbio.Reader { return legacyReadWriter{s} },
+		// Request readers are allowed to read zero bytes, this is because the legacy protocol
```
dB2510, line 181:
```suggestion
		// Request readers are allowed to read zero bytes, in order to support the legacy protocol
```

p2p/sender.go
```diff
@@ -175,10 +176,15 @@ func defaultSendRecvOpts(pID protocol.ID) sendRecvOpts {
...
+		reqReadersByProtocol: map[protocol.ID]func(s network.Stream) pbio.Reader{
+			pID: func(s network.Stream) pbio.Reader { return legacyReadWriter{stream: s, allowZeroRead: true} },
+		},
+		// Response readers are not allowed to read zero bytes, this is because the legacy protocol
```
dB2510, line 185:
```suggestion
		// Response readers are not allowed to read zero bytes, in order to support the legacy protocol
```

dB2510 (approved):

PR #1958, opened 2023-03-29
p2p: support zero length request protos

corverroos:
Fix regression introduced by length-delimited upgrades which assumed that legacy-stream-delimited protocols handled requests and responses the same while it didn't. It allowed zero length requests while it didn't allow zero-length responses. This aligns the new implementation with that.

category: bug
ticket: #1956

### DIFF
--- a/p2p/sender.go
+++ b/p2p/sender.go
@@ -223,8 +223,17 @@ func SendReceive(ctx context.Context, tcpNode host.Host, peerID peer.ID,
 		return errors.Wrap(err, "close write", z.Any("protocol", s.Protocol()))
 	}
 
+	zeroResp := proto.Clone(resp)
+
 	if err = reader.ReadMsg(resp); err != nil {
 		return errors.Wrap(err, "read response", z.Any("protocol", s.Protocol()))
+	}
+
+	// TODO(corver): Remove this once we use length-delimited protocols.
+	//  This was added since legacy stream delimited readers couldn't distinguish between
+	//  no response and a zero response.
+	if proto.Equal(resp, zeroResp) {
+		return errors.New("no or zero response received", z.Any("protocol", s.Protocol()))
 	}
 
 	if err = s.Close(); err != nil {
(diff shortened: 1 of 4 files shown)


p2p/sender.go
```diff
@@ -223,10 +223,19 @@ func SendReceive(ctx context.Context, tcpNode host.Host, peerID peer.ID,
 		return errors.Wrap(err, "close write", z.Any("protocol", s.Protocol()))
 	}
 
+	zeroResp := proto.Clone(resp)
```
dB2510, line 226: what if `resp` is NOT a zero/empty response?

dB2510 (approved):

p2p/sender.go
corverroos: true, that is an edge-case not handled by this code

corverroos: but that would be the code that uses this function. We can look at adding some build-time checks that all passed in responses are zero.

p2p/sender.go
```diff
@@ -223,10 +223,19 @@ func SendReceive(ctx context.Context, tcpNode host.Host, peerID peer.ID,
...
+
 	if err = reader.ReadMsg(resp); err != nil {
```
dB2510, line 228: this will still return the errors that we are seeing in logs:
`Peerinfo failed: read response: peer errored, no response`

how is this allowing zero length responses?

p2p/sender.go
corverroos: I added a `isZeroProto` check to the SendReceive function.

p2p/sender.go
```diff
@@ -288,13 +301,11 @@ func (w legacyReadWriter) WriteMsg(m proto.Message) error {
 func (w legacyReadWriter) ReadMsg(m proto.Message) error {
 	b, err := io.ReadAll(w.stream)
 	if err != nil {
-		return errors.Wrap(err, "read response")
-	} else if len(b) == 0 {
-		return errors.New("peer errored, no response")
+		return errors.Wrap(err, "read proto")
 	}
```
corverroos, line 305: @dB2510 this was the zero length check that was erroring, it has been removed.

p2p/sender.go
corverroos: See `legacyReadWriter ReadMsg`

p2p/sender.go
dB2510: Got it!

xenowits (approved):

dB2510 (approved):

PR #1960, opened 2023-03-29
app: fix flapping integration test

gsora:
When a test is killed by timeout, `t.Cleanup()` functions are not called.

Since we relied on them to fully stop Teku Docker containers, they would stick around and make the CI fail because processes were still running.

This commit introduces the `testutil.EnsureCleanup()` function, which calls a function 100ms before the global test timeout.

By calling `docker kill` this way, we always ensure cleanup on test timeout.

category: bug
ticket: none

### DIFF
--- a/app/simnet_test.go
+++ b/app/simnet_test.go
@@ -479,9 +479,10 @@ func startTeku(t *testing.T, args simnetArgs, node int) simnetArgs {
 	}()
 
 	// Kill the container when done (context cancel is not enough for some reason).
-	t.Cleanup(func() {
+	testutil.EnsureCleanup(t, func() {
 		cancel()
-		_ = exec.Command("docker", "kill", name).Run()
+		t.Log("stopping teku docker container", name)
+		_ = exec.Command("docker", "stop", name).Run()
 	})
 
 	return args
--- a/testutil/timeout.go
+++ b/testutil/timeout.go
@@ -1,0 +1,32 @@
+// Copyright © 2022-2023 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package testutil
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+// EnsureCleanup calls cleanupFunc in one of two cases:
+//   - 100ms before the test deadline is exceeded when it is run with `-timeout`, or
+//   - on test cleanup
+func EnsureCleanup(t *testing.T, cleanupFunc func()) {
+	t.Helper()
+	testDeadline, present := t.Deadline()
+	if !present {
+		t.Cleanup(cleanupFunc)
+		return
+	}
+
+	ctx := context.Background()
+
+	// Adapted from https://github.com/golang/go/issues/24050#issuecomment-1137682781
+	testDeadline = testDeadline.Add(-100 * time.Millisecond)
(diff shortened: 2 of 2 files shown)


app/simnet_test.go
```diff
@@ -479,9 +479,10 @@ func startTeku(t *testing.T, args simnetArgs, node int) simnetArgs {
 	}()
 
 	// Kill the container when done (context cancel is not enough for some reason).
-	t.Cleanup(func() {
+	testutil.EnsureCleanup(t, func() {
 		cancel()
-		_ = exec.Command("docker", "kill", name).Run()
+		t.Log("stopping teku docker container", name)
+		_ = exec.Command("docker", "stop", name).Run()
```
corverroos, line 485: suggest keeping it `kill`, as stop is soft shutdown and we want everything to exit now.

gsora: Correct, was trying with `stop` as well and somehow committed it this way.

Fixing as we speak.

testutil/timeout.go
```diff
@@ -0,0 +1,32 @@
...
+		return
+	}
+
+	ctx := context.Background()
+
+	// Adapted from https://github.com/golang/go/issues/24050#issuecomment-1137682781
+	testDeadline = testDeadline.Add(-100 * time.Millisecond)
+	dctx, cancel := context.WithDeadline(ctx, testDeadline)
+	go func() {
+		<-dctx.Done()
```
corverroos, line 28: not sure the context deadline approach is required since we don't do anything with context, wouldn it be simpler to just do:
```
time.Sleep(testDeadline)
cleanupFunc()
```

dB2510: +1

gsora: Alright, I’ll refactor as soon as I’m back at my keyboard.

testutil/timeout.go
```diff
@@ -0,0 +1,26 @@
...
+	t.Helper()
+	testDeadline, present := t.Deadline()
+	if !present {
+		t.Cleanup(cleanupFunc)
+		return
+	}
+
+	// Adapted from https://github.com/golang/go/issues/24050#issuecomment-1137682781
+	go func() {
+		time.Sleep(time.Until(testDeadline.Add(-100 * time.Millisecond)))
```
corverroos, line 23: Note this will not call cleanup if deadline present but test doesn't time out

corverroos (changes requested):

testutil/timeout.go
gsora: Fixed, cleanupFunc now is always called either in our timeout, or on Cleanup(), and wrapped in a sync.Once.

testutil/timeout.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+)
+
+// EnsureCleanup calls cleanupFunc in one of two cases:
+//   - 100ms before the test deadline is exceeded when it is run with `-timeout`, or
+//   - on test cleanup
+func EnsureCleanup(t *testing.T, cleanupFunc func()) {
+	t.Helper()
+
+	cfOnce := sync.Once{}
```
corverroos, line 17: nit: `var cfOnce sync.Once`

corverroos (approved):

Issue #1732, opened 2023-01-31
Improve BN health check


corverroos:
## Problem to be solved

BN health check didn't detect that a lighthouse BN wasn't healthy (synced) when it couldn't sync due to not having any connectivity to the p2p network. The readyz endpoint reported Healthy. Only once the network issues were resolve, did the node start syncing and did the readyz turn to "BN is syncing". 

See https://discord.com/channels/930779837218562078/1069932708689629234/1069932710849683466

## Proposed solution

Reproduce and look at the value of `NodeSyncing` response:
```
// SyncState is the data regarding the node's synchronization state to the chain.
type SyncState struct {
	// HeadSlot is the head slot of the chain as understood by the node.
	HeadSlot phase0.Slot
	// SyncDistance is the distance between the node's highest synced slot and the head slot.
	SyncDistance phase0.Slot
	// IsOptimistic is true if the node is optimistic.
	IsOptimistic bool
	// IsSyncing is true if the node is syncing.
	IsSyncing bool
}
```

SyncDistance might be high or HeadSlot might be zero.

Once the data has been identified, incorporate this check and new readzy value into app package.

PR #1961, opened 2023-03-29
app: add readyz error when BN can't sync because of zero peers

xenowits:
Adds a new `readyz` error for scenarios when beacon node cannot sync because of zero peers. Without this fix, the grafana healthy panel shows "OK".

`IsSyncing: false` means that the "Beacon node is not syncing" because either
- "It is fully synced" OR
- "Can't sync as it has no peers" 

category: bug 
ticket: https://github.com/ObolNetwork/charon/issues/1732


app/monitoringapi.go
```diff
@@ -123,6 +124,11 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2client
 					notConnectedRounds++
 				}
 
+				bnPeerCount, bnErr := eth2Cl.NodePeerCount(ctx)
+				if bnErr != nil {
+					log.Error(ctx, "Failed to get beacon node peer count", bnErr)
```
corverroos, line 129: suggest WARN

Issue #1964, opened 2023-03-29
Deploy latest promrated on updates without manual change


LukeHackett12:
# 🎯 Problem to be solved

Currently you need to update the image manually in infra repo.

# 🛠️ Proposed solution

Add a tag that will work for deploying latest after merging https://github.com/ObolNetwork/obol-infrastructure/pull/522

PR #1965, opened 2023-03-29
testutil/promrated: set tag var correctly for deployment

LukeHackett12:
Adds a tag variable so that the latest promrated gets deployed without manual update.

category: misc
ticket: #1964

### DIFF
--- a/.github/workflows/build-push-deploy-promrated.yaml
+++ b/.github/workflows/build-push-deploy-promrated.yaml
@@ -50,11 +50,14 @@ jobs:
         build-args: GITHUB_SHA=${{ github.sha }}
         tags: ${{ steps.meta.outputs.tags }}
 
-    - name: Split branch name
-      env:
-        SHA_TAG: ${{ steps.meta.outputs.tags[1] }}
-      id: split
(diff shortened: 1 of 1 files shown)


.github/workflows/build-push-deploy-promrated.yaml
```diff
@@ -49,11 +49,15 @@ jobs:
         push: true
         build-args: GITHUB_SHA=${{ github.sha }}
         tags: ${{ steps.meta.outputs.tags }}
+        
+    - name: Set shortened SHA
+      id: vars
+      run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
```
corverroos, line 55: `Define docker image meta data tags` already does this. Feels like we are hacking here...

corverroos (changes requested):

.github/workflows/build-push-deploy-promrated.yaml
LukeHackett12: I went with what you suggested, however the tages are in the form "name:sha" so there is still an additional step needed. Let me know which you prefer.

corverroos (approved):

PR #1988, opened 2023-03-30
dkg: fix race in sync test

corverroos:
Fixes the following race condition in the sync test that caused it to flap:

- 3 peers are connected to each other (0,1,3)
- stop 1 peer (3)
- start remaining peers (2,3)
- There is a race in peers that remain up (0,1)
  - between detecting 3 went down and 2 came up.
  - If it detects that 2 is up, BEFORE it detects 3 is down, 
  - Then it thinks it is connected to all peers
  - And stops retrying to connect
  - Then it detects 3 is down
  - Then it fails. 

category: test 
ticket: none

### DIFF
--- a/dkg/dkg_test.go
+++ b/dkg/dkg_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gsora (approved):

dB2510 (approved):
🙏

PR #1990, opened 2023-03-31
eth2util/rlp: add fuzz tests

corverroos:
Add fuzz tests to `eth2util/rlp` as identified by sigma-prime audit.

Fixes OBOL-04.

category: bug
ticket: none

### DIFF
--- a/eth2util/rlp/rlp_test.go
+++ b/eth2util/rlp/rlp_test.go
@@ -71,7 +71,11 @@ func FuzzEncodeBytesList(f *testing.F) {
 	data := make([]byte, minPayloadSize)
 	d := merge(prefix, data)
 
-	f.Add(0, d, d, d, d, d, d, d, d, d, d)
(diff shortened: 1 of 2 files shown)


dB2510 (approved):

eth2util/rlp/rlp_test.go
```diff
@@ -58,6 +121,14 @@ func TestBytesList(t *testing.T) {
 	}
 }
 
+func merge(a, b []byte) []byte {
+	resp := make([]byte, len(a)+len(b))
```
gsora, line 125: nit: this can be as simple as `append(a, b...)`

gsora (approved):

eth2util/rlp/rlp_test.go
corverroos: no it can't, since that modifies a's underlying array. I specifically wanted to avoid that.

Issue #2002, opened 2023-04-03
DKG doesn't fail server side if version mismatch is detected 


corverroos:
# 🐞 Bug Report

### Description

DKG doesn't exit if sync server (v0.15.0-dev) detects a version mismatch from v0.14. Only the client errors.

PR #2003, opened 2023-04-03
dkg/sync: fail server on version mismatch

corverroos:
Fail sync server as soon as any error is detected, instead of only erroring after all clients connected.

category: bug
ticket: #2002


xenowits (approved):

dB2510 (approved):
LGTM

PR #2015, opened 2023-04-04
p2p: fix relay conn checker

corverroos:
Fix relay connection check, should by regular ticker, not once off timer.

category: bug
ticket: none

### DIFF
--- a/p2p/relay.go
+++ b/p2p/relay.go
@@ -69,7 +69,7 @@ func NewRelayReserver(tcpNode host.Host, relay *MutablePeer) lifecycle.HookFuncC
 
 			refresh := time.After(refreshDelay)
 
-			timer := time.NewTimer(time.Millisecond * 100)
+			timer := time.NewTicker(time.Millisecond * 500)
 
 			for {
 				select {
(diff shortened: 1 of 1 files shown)


p2p/relay.go
```diff
@@ -69,7 +69,7 @@ func NewRelayReserver(tcpNode host.Host, relay *MutablePeer) lifecycle.HookFuncC
 
 			refresh := time.After(refreshDelay)
 
-			timer := time.NewTimer(time.Millisecond * 100)
+			timer := time.NewTicker(time.Millisecond * 500)
```
dB2510, line 72:
```suggestion
			ticker := time.NewTicker(time.Millisecond * 500)
```

dB2510 (approved):

p2p/relay.go
dB2510: why updated to 500ms?

corverroos: 100ms is very fast, 500ms should be fine. It was 1s originally. We decreased to 100ms to debug.

PR #2016, opened 2023-04-05
testutil/compose: add build-local command

corverroos:
Add a `build-local` command to `compose` to support running new source code without regenerating the whole compose cluster.

category: misc
ticket: none

### DIFF
--- a/testutil/compose/compose/main.go
+++ b/testutil/compose/compose/main.go
@@ -39,6 +39,7 @@ func newRootCmd() *cobra.Command {
 
 	root.AddCommand(newNewCmd())
 	root.AddCommand(newCleanCmd())
+	root.AddCommand(newBuildLocalCmd())
 	root.AddCommand(newAutoCmd())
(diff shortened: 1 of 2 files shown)


testutil/compose/compose/main.go
```diff
@@ -106,6 +107,17 @@ func newAutoCmd() *cobra.Command {
 	return cmd
 }
 
+func newBuildLocalCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "build-local",
+		Short: "Builds the obolnetwork/charon:local docker container from the local source code. Note this requires the CHARON_REPO env var.",
```
dB2510, line 113: `--charon-repo` flag? not necessary though

dB2510 (approved):

xenowits (approved):

testutil/compose/compose/main.go
corverroos: No, env var is much better UX

PR #2025, opened 2023-04-05
app/retry: make sure to not enqueue new jobs when shutting down

gsora:
This is achieved by adding a lock around r.wg.Wait() and r.wg.Add() calls: you don't want to add a new worker if you're waiting for the existing to finish.

Also added a test case that simulates calling Shutdown() and DoAsync() roughly at the same time.

This test usually exhibits a race condition in the first 10-20 runs on my machine (M1 Pro Mac).

category: bug
ticket: none


gsora:
Closing in favor of #2026.

PR #2027, opened 2023-04-05
core/tracker: unlock mutex

dB2510:
Unlock mutex before doing any i/o in core/tracker/incldelay.go.

category: bug
ticket: #2028

### DIFF
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -149,10 +149,12 @@ func (s *Scheduler) GetDutyDefinition(ctx context.Context, duty core.Duty) (core
 
 	epoch := duty.Slot / int64(slotsPerEpoch)
 	if !s.isEpochResolved(epoch) {
-		return nil, errors.New("epoch not resolved yet")
+		return nil, errors.New("epoch not resolved yet",
+			z.Any("duty", duty.String()), z.I64("epoch", epoch))
 	}
 	if s.isEpochTrimmed(epoch) {
-		return nil, errors.New("epoch already trimmed")
+		return nil, errors.New("epoch already trimmed",
+			z.Any("duty", duty.String()), z.I64("epoch", epoch))
 	}
 
 	defSet, ok := s.getDutyDefinitionSet(duty)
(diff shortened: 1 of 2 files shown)


core/scheduler/scheduler.go
```diff
@@ -149,10 +149,12 @@ func (s *Scheduler) GetDutyDefinition(ctx context.Context, duty core.Duty) (core
...
 	epoch := duty.Slot / int64(slotsPerEpoch)
 	if !s.isEpochResolved(epoch) {
-		return nil, errors.New("epoch not resolved yet")
+		return nil, errors.New("epoch not resolved yet",
+			z.Any("duty", duty.String()), z.I64("epoch", epoch))
 	}
 	if s.isEpochTrimmed(epoch) {
-		return nil, errors.New("epoch already trimmed")
+		return nil, errors.New("epoch already trimmed",
+			z.Any("duty", duty.String()), z.I64("epoch", epoch))
```
xenowits, line 157:
```suggestion
			z.Str("duty", duty.String()), z.I64("epoch", epoch))
```

core/tracker/incldelay.go
```diff
@@ -31,20 +31,33 @@ func NewInclDelayFunc(eth2Cl eth2wrap.Client, dutiesFunc dutiesFunc) func(contex
...
+		return dutyStartSlot
+	}
+
+	return func(ctx context.Context, current core.Slot) error {
 		// blockSlot the block we want to instrument.
 		blockSlot := current.Slot - inclDelayLag
 
-		if dutyStartSlot == 0 {
-			dutyStartSlot = current.Slot // Set start slot.
+		if getStartSlot() == 0 {
```
gsora, line 57: What happens if in the span of time between `getStartSlot()` and `setStartSlot()`, `dutyStartSlot` gets changed by another function call?

Is this something that can happen?

Do we care if this happens?

xenowits (approved):

core/tracker/incldelay.go
dB2510: Technically it cannot happen as dutyStartSlot is set when charon is started i.e., when dutyStartSlot is 0.

dB2510: Or we can say it is set when we receive first slot tick event since charon start.

gsora (approved):

core/tracker/incldelay.go
dB2510: Hmmm, we call `tracker.NewInclDelayFunc(eth2Cl, sched.GetDutyDefinition)` in app.go. So if we call `app.Run` multiple times then it will create new set of dutyStartSlot variable for each `app.Run` call.

PR #2029, opened 2023-04-05
app: fix peer count bn query to occur every min instead of every 10s

xenowits:
Fixes peer count beacon node query to occur every min instead of every 10s. 

Also, fixes it to query only once since we query this both in `startReadyzChecker()` and `beaconNodeMetrics()` and set the `beaconNodeVersionGauge` in `startReadyzChecker` instead.

category: misc 
ticket: none

### DIFF
--- a/app/monitoringapi.go
+++ b/app/monitoringapi.go
@@ -108,10 +108,7 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
 		ticker := clock.NewTicker(10 * time.Second)
 		peerCountTicker := clock.NewTicker(1 * time.Minute)
 		epochTicker := clock.NewTicker(32 * 12 * time.Second) // 32 slots * 12 second slot time
-		var (
-			bnPeerCount    int   // Beacon node peer count value which is queried on startup & every minute
-			bnPeerCountErr error // Error if the peer count query wasn't successful
-		)
+		var bnPeerCount *int                                  // Beacon node peer count value which is queried on startup & then every minute
 		currVAPICount := 0
 		prevVAPICount := 1 // Assume connected.
 		currPKs := make(map[core.PubKey]bool)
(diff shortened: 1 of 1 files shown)


app/monitoringapi.go
```diff
@@ -143,7 +147,7 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
 				} else if syncing {
 					err = errReadyBeaconNodeSyncing
 					readyzGauge.Set(readyzBeaconNodeSyncing)
-				} else if bnPeerCount == 0 && bnErr == nil {
+				} else if bnPeerCount == 0 {
```
corverroos, line 150: this will trigger if `NodePeerCount` never ran or if it failed previously

app/monitoringapi.go
```diff
@@ -115,26 +120,38 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
 			prevPKs[pubkey] = true
 		}
 
+		// beaconNodePeerCount queries beacon node peer count and sets the peer count gauge if err is nil.
+		beaconNodePeerCount := func() {
+			bnPeerCount, bnPeerCountErr = eth2Cl.NodePeerCount(ctx)
+			if bnPeerCountErr != nil {
+				log.Error(ctx, "Failed to get beacon node peer count", bnPeerCountErr)
```
corverroos, line 127: suggest warn

app/monitoringapi.go
```diff
@@ -115,26 +120,38 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
...
+
+		onStartup := make(chan struct{}, 1)
+		onStartup <- struct{}{}
+
 		for {
 			select {
 			case <-ctx.Done():
 				return
+			case <-onStartup:
+				beaconNodePeerCount()
```
corverroos, line 141: lets remove this onstartup part, it isn't important

app/monitoringapi.go
```diff
@@ -106,7 +106,12 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
 	)
 	go func() {
 		ticker := clock.NewTicker(10 * time.Second)
+		peerCountTicker := clock.NewTicker(1 * time.Minute)
 		epochTicker := clock.NewTicker(32 * 12 * time.Second) // 32 slots * 12 second slot time
+		var (
+			bnPeerCount    int   // Beacon node peer count value which is queried on startup & every minute
```
corverroos, line 112: slightly simpler, lets use a `bnPeerCount *bool`,  then below use `if bnPeerCount != nil && *bnPeerCount == 0`

xenowits: i suppose you mean `bnPeerCount *int`

app/monitoringapi.go
```diff
@@ -115,6 +117,20 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
...
+		beaconNodePeerCount := func() {
+			peerCount, err := eth2Cl.NodePeerCount(ctx)
+			if err != nil {
+				log.Warn(ctx, "Failed to get beacon node peer count", err)
+				return
+			}
+			if bnPeerCount == nil {
+				bnPeerCount = new(int)
+			}
+			*bnPeerCount = peerCount
```
corverroos, line 130: replace with: `bnPeerCount = &peerCount`

corverroos (approved):

Issue #2031, opened 2023-04-05
Add deployment workflows for relays


LukeHackett12:
# 🎯 Problem to be solved

We need to deploy latest relay to dev and have a prod deployment and revert possible from charon repo

# 🛠️ Proposed solution

Add deployment workflow and deploy dev on build

PR #2032, opened 2023-04-05
github/workflows: add deploy relay workflows

LukeHackett12:
Relies on https://github.com/ObolNetwork/obol-infrastructure/pull/538

Adds deployment of dev relay on merge, and a workflow to deploy two of our prod relays with a chosen tag/short sha

category: feature
ticket: #2031


.github/workflows/relay-0-prod-deploy.yml
```diff
@@ -0,0 +1,33 @@
+name: 'Deploy Relay To Prod'
```
eth2devops, line 1: let's include relay-dev as well

LukeHackett12: Added as a separate workflow 👍🏼

eth2devops (approved):
LGTM

.github/workflows/relay-nonprod-deploy.yml
```diff
@@ -0,0 +1,32 @@
+name: 'Deploy Nonprod Relay'
```
corverroos, line 1: suggest rename `Deploy Dev Relay`

.github/workflows/relay-nonprod-deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  workflow_dispatch:
```
corverroos, line 4: also please rename the deploy files to: `deploy-relay-X.yml`

.github/workflows/relay-nonprod-deploy.yml
LukeHackett12: Yeah so I think it depends on the preference here for if we add more relays in pre production environments, then it may be preferable to have a choice setup like the prod workflow?

corverroos (approved):

PR #2035, opened 2023-04-06
app/eth2wrap: fix lazy client if http times out

corverroos:
Fix eth2wrap lazy client when http requests block/timeout. Also respect context cancel while lazy resolving.

category: bug
ticket: none

### DIFF
--- a/app/eth2wrap/lazy.go
+++ b/app/eth2wrap/lazy.go
@@ -5,6 +5,7 @@ package eth2wrap
 import (
 	"context"
 	"sync"
+	"time"
 
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 
@@ -12,7 +13,7 @@ import (
 )
 
 // newLazy creates a new lazy client.
-func newLazy(provider func() (Client, error)) *lazy {
+func newLazy(provider func(context.Context) (Client, error)) *lazy {
 	return &lazy{
 		provider: provider,
 	}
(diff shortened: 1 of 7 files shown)


app/eth2wrap/lazy.go
```diff
@@ -5,71 +5,96 @@ package eth2wrap
...
+func newLazy(provider func(context.Context) (Client, error)) *lazy {
 	return &lazy{
 		provider: provider,
 	}
 }
 
 // lazy is a client that is created on demand.
 type lazy struct {
-	provider func() (Client, error)
+	providerMu sync.Mutex
```
corverroos, line 24: only one caller will try to resolve the client if it doesn't exit.

app/eth2wrap/lazy.go
```diff
@@ -5,71 +5,96 @@ package eth2wrap
...
+	if cl, ok := l.getClient(); ok {
+		return cl, nil
 	}
-	l.mu.RUnlock()
 
-	// Else create a new client.
-	l.mu.Lock()
-	defer l.mu.Unlock()
+	// Try until we get the provider lock or the context is cancelled.
+	for !l.providerMu.TryLock() {
```
corverroos, line 55: only wait to wait for a lock while respecting context cancel

dB2510 (approved):

dB2510:
@corverroos we need to upgrade to go1.20.3 for govulncheck. Not a blocker though.

PR #2057, opened 2023-04-07
workflows: run govulncheck on releases

gsora:
Run `govulncheck` on releases as well, not just PRs and `main` commits.

category: misc
ticket: none

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - main*
-    tags:
-      - 'v*'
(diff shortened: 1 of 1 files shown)


.github/workflows/govulncheck.yml
```diff
@@ -3,7 +3,9 @@ on:
   pull_request:
   push:
     branches:
-      - main
+      - main*
+    tags:
+      - 'v*'
```
corverroos, line 8: can remove this, as tags should only be on main* branches.

corverroos (approved):

PR #2062, opened 2023-04-07
core/fetcher: dedup aggregate attestations

dB2510:
Dedup aggregate attestations in fetcher for aggregators of same committee. This solves `clashing aggregated attestation` error in dutydb in case of different aggregate attestations for aggregators of same committee.

category: bug
ticket: none

### DIFF
--- a/core/fetcher/fetcher.go
+++ b/core/fetcher/fetcher.go
@@ -169,6 +169,9 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot int64, defSet core
 
 // fetchAggregatorData fetches the attestation aggregation data.
 func (f *Fetcher) fetchAggregatorData(ctx context.Context, slot int64, defSet core.DutyDefinitionSet) (core.UnsignedDataSet, error) {
+	// We may have multiple aggregators in the same committee, use the same aggregated attestation in that case.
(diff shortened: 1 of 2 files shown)


core/fetcher/fetcher.go
```diff
@@ -196,6 +199,11 @@ func (f *Fetcher) fetchAggregatorData(ctx context.Context, slot int64, defSet co
 		}
 		log.Info(ctx, "Resolved attester aggregation duty", z.Any("pubkey", pubkey))
 
+		_, ok = aggAttByCommIdx[attDef.CommitteeIndex]
+		if ok {
+			continue // Skips querying aggregate attestation for aggregators of same committee.
```
corverroos, line 204: please follow `fetchAttesterData` example, this will result in `resp[pubkey]` not being set for this validator

corverroos: please also add a test for this edge-case

corverroos (changes requested):

core/fetcher/fetcher.go
dB2510: Done!
Setting response for this validator also added tests for this case, converted existing tests to table tests.

corverroos (approved):

Issue #1537, opened 2022-12-11
Teku migrated to SSZ block creation by default


corverroos:
## Problem to be solved

Teku has migrated to SSZ serialisation for block proposals, both fetching and submitting blocks (incl blinded).
 - https://github.com/ConsenSys/teku/issues/5400
 - https://github.com/ConsenSys/doc.teku/pull/415

The validatorAPI doesn't support SSZ, it only supports json.

Teku does fallback, but only on the next call it seems, so it seems we might mess blocks due to this.

## Proposed solution

Add support for SSZ content types, both in marshalling and unmarshalling.  
In the mean time, we should disable this feature in teku until supported `--beacon-node-ssz-blocks-enabled`

### Out of Scope

Update cdvn and cdcc repo with config flag when updating to latest teku version.

PR #2063, opened 2023-04-07
core/validatorapi: support ssz encoded request bodies

corverroos:
Enable support for ssz encoding in `validatorapi`.

category: feature
ticket: #1537

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -143,6 +143,7 @@ linters:
     - interfacer
     - ireturn
     - lll # Think about adding this (max line length)
+    - maintidx
     - maligned
     - musttag
     - nonamedreturns
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -957,7 +957,7 @@ func writeError(ctx context.Context, w http.ResponseWriter, endpoint string, err
 	}
 }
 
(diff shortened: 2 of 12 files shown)


.golangci.yml
```diff
@@ -143,6 +143,7 @@ linters:
     - interfacer
     - ireturn
     - lll # Think about adding this (max line length)
+    - maintidx
```
corverroos, line 146: new linter, similar to "complexity" linters that we disable.

xenowits: so the default is `20`, so ideally any function with maintainability index > 20 should raise linter errors

dB2510 (approved):
LGTM!
I suggest to test this locally in compose with teku VCs.

core/validatorapi/router.go
```diff
@@ -947,7 +959,7 @@ func writeError(ctx context.Context, w http.ResponseWriter, endpoint string, err
 
 // unmarshal parses the JSON-encoded request body and stores the result
```
gsora, line 960:
```suggestion
// unmarshal parses body with the appropriate unmarshaler based on the contentType and stores the result
```

gsora (approved):

Issue #2034, opened 2023-04-06
Improve tracker participation metric


dB2510:
# 🎯 Problem to be solved
We currently show peer participation metric as `core_tracker_participation_total` with labels duty type and peer name. It shows count of all the duties in which peer has participated. But if we have multiple duties in a single slot then it fails to show total duties submitted by the peer. 

For example, in our performance cluster (1000 keys) we have attester duties in each slot, but `core_tracker_participation_total` reports 1 for each slot while it is greater than 1 per slot. 

Below is the snapshot of participation in performance cluster in an hour (300 slots) window:
![Screenshot 2023-04-06 at 1 31 59 PM](https://user-images.githubusercontent.com/37813203/230313409-60e3d524-d7c9-41a4-ab8a-817cde76b445.png)

We have submitted 1000 attestations per epoch but this metric only reports 1 attestation per slot.

# 🛠️ Proposed solution

Improve `analyseParticipation` function of core/tracker/tracker.go to return count of duty data submitted by each peer for the given duty. Also calculate count of failed participation and expected participation with metrics `core_tracker_failed_participation_total` and `core_tracker_expected_participation_total`.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2075, opened 2023-04-10
core/tracker: add missed participation counter

corverroos:
Adds a `missed participation by peer` counter so that we can calculate `duty participation % per peer` and so we can highlight missed participations instead of only successful participations.

category: misc
ticket: #2034

### DIFF
--- a/core/tracker/tracker.go
+++ b/core/tracker/tracker.go
@@ -672,8 +672,8 @@ func newParticipationReporter(peers []p2p.Peer) func(context.Context, core.Duty,
 	// Initialise participation metrics to 0 to avoid non-existent metrics issue on startup.
 	for _, duty := range core.AllDutyTypes() {
 		for _, peer := range peers {
-			participationCounter.WithLabelValues(duty.String(), peer.Name).Add(0)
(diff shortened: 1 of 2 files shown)


core/tracker/tracker.go
```diff
@@ -672,8 +672,8 @@ func newParticipationReporter(peers []p2p.Peer) func(context.Context, core.Duty,
 	// Initialise participation metrics to 0 to avoid non-existent metrics issue on startup.
 	for _, duty := range core.AllDutyTypes() {
 		for _, peer := range peers {
-			participationCounter.WithLabelValues(duty.String(), peer.Name).Add(0)
-			participationGauge.WithLabelValues(duty.String(), peer.Name).Set(0)
```
corverroos, line 676: @dB2510 setting `participationGauge` to 0 on startup means all peers start a "not participation" which is incorrect since the node didn't have to participate yet, so it hasn't missed or failed participating.

xenowits (approved):
LGTM. The new `participationSuccess` counter needs to be added to simnet dashboard

Issue #2052, opened 2023-04-06
Lack of Size Checks When Slicing Arrays (OBOL-05)


xenowits:
# 🎯 Problem to be solved

The codebase contains instances where arrays are being sliced without first checking the size of the array. This can result in the program crashing if the array being sliced is of insufficient size. This is particularly problematic in
the context of p2p when receiving data from other connected peers, as malicious peers may intentionally send malformed
data to crash the system.

# 🛠️ Proposed solution

Review all instances where arrays are being sliced to ensure that size checks are being performed prior to slicing.
Specifically, the program should check the size of the array before attempting to slice it, and should handle any errors
that occur gracefully.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests

PR #2077, opened 2023-04-10
*: check slice length before slicing slices

xenowits:
Check length of array/string before slicing.

category: bug 
ticket: #2052

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -151,7 +151,11 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
-	lockHashHex := hex.EncodeToString(lock.LockHash)[:7]
+	lockHashLong := hex.EncodeToString(lock.LockHash)
+	if len(lockHashLong) < 7 {
+		return errors.New("insufficient characters in lockhash", z.Str("lockhash", lockHashLong))
+	}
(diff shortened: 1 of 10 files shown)


app/version/version.go
```diff
@@ -36,6 +36,10 @@ func GitCommit() (hash string, timestamp string) {
 
 	for _, s := range info.Settings {
 		if s.Key == "vcs.revision" {
+			if len(s.Value) < 7 {
```
corverroos, line 39: This isn't critical stuff, rather just use `min(len(x),7]` everywhere: ie. `hash := s.Value[:min(len(v.Value,7)]`

app/app.go
```diff
@@ -923,3 +922,14 @@ func ProposalTypes(builder bool, synthetic bool) []core.ProposalType {
 
 	return resp
 }
+
+// lockHash returns the lock hash in string representation.
+func lockHash(lockHash []byte) string {
```
corverroos, line 927: suggest:
```
// hex7 returns the first 7 (or less) hex chars of the provided bytes.  
func hex7(bytes []byte) string {
   resp := hex.EncodeToString(lockHash)
   if len(resp) <= 7 {
     return resp
   }
   
   return resp[:7]
}
```

PR #2078, opened 2023-04-10
license: revert change to license text

OisinKyne:
Reverts change to the additional use grant clause of the BSL. 

category: misc
ticket: none

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -12,8 +12,8 @@ Licensor:             Obol Labs, Inc.
 Licensed Work:        Charon, the Distributed Validator Client
                       The Licensed Work is © 2022-2023 Obol Labs, Inc.
 
-Additional Use Grant: The use of the Licensed Work by the Licensee is limited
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -12,8 +12,8 @@ Licensor:             Obol Labs, Inc.
 Licensed Work:        Charon, the Distributed Validator Client
                       The Licensed Work is © 2022-2023 Obol Labs, Inc.
 
-Additional Use Grant: The use of the Licensed Work by the Licensee is limited
-                      to one distributed validator on a production network.
+Additional Use Grant: Any uses listed and defined at
+                      charon-license-grants.obol.eth
```
xenowits, line 16: This ENS probably doesn't have any details associated with it?..

OisinKyne: Yeah I need Collin to give me admin access atm

thomasheremans:
Parameters in the BSL should not refer to an address rather should be set out in text

Issue #2076, opened 2023-04-10
RLP Length in Bits Rather Than Bytes (OBOL-13)


xenowits:
# 🎯 Problem to be solved

RLP decoding validates that the length is not more than 64 bits before passing this value to `fromBigEndian()` which expects bytes.

The function [fromBigEndian()](https://github.com/ObolNetwork/charon/blob/main/eth2util/enr/enr.go#L236) expects length to be in units of bytes rather than bits. It is therefore possible to have a length larger than uint64 .

# 🛠️ Proposed solution

For these cases, ensure that length is less than or equal to 8 rather than 64 and encode values in bytes rather than bits.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests

PR #2081, opened 2023-04-11
rlp: fix bug in decodeLength

xenowits:
Fixes bug in RLP `decodeLength()` by ensuring that calculations of length should use units of bytes instead of bits.

category: bug 
ticket: #2076

### DIFF
--- a/eth2util/rlp/rlp.go
+++ b/eth2util/rlp/rlp.go
@@ -125,8 +125,8 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	length = int(prefix - 0xf7)
-	if length <= 0 { // This is impossible based on outer if else checks
-		panic("length not positive (>=1)")
+	if length > 8 || length <= 0 { // This is impossible based on outer if else checks
+		panic("length not in expected range [1,8]")
 	}
 
 	offset = 1 + length
--- a/eth2util/rlp/rlp.go
+++ b/eth2util/rlp/rlp.go
@@ -102,7 +102,7 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 
 	if prefix < 0xc0 {
 		length = int(prefix - 0xb7) // length of the string in bytes in binary form
-		if length > 8 || length < 0 {
+		if length > 8 || length <= 0 {
 			return 0, 0, errors.New("invalid length prefix")
 		}
 
@@ -113,7 +113,7 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 			return 0, 0, err
 		} else if resp < 0 {
 			return 0, 0, errors.New("negative length")
-		} else if offset+resp > len(item) {
+		} else if offset > offset+resp {
 			return 0, 0, errors.New("overflow")
 		}
 
@@ -125,7 +125,7 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	length = int(prefix - 0xf7)
-	if length > 8 || length < 0 {
+	if length > 8 || length <= 0 {
 		return 0, 0, errors.New("invalid length prefix")
 	}
 
@@ -136,7 +136,7 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 		return 0, 0, err
 	} else if resp < 0 {
 		return 0, 0, errors.New("negative length")
-	} else if offset+resp > len(item) {
+	} else if offset > offset+resp {
 		return 0, 0, errors.New("overflow")
 	}
 
--- a/eth2util/rlp/rlp.go
+++ b/eth2util/rlp/rlp.go
@@ -101,9 +101,9 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	if prefix < 0xc0 {
-		length = int(prefix - 0xb7) // length of the string in bytes in binary form
-		if length > 8 || length <= 0 {
-			return 0, 0, errors.New("invalid length prefix")
+		length = int(prefix - 0xb7)    // length of the string in bytes in binary form
+		if length > 8 || length <= 0 { // This is impossible based on outer if else checks
+			panic("length not in expected range [1,8]")
 		}
 
 		offset := 1 + length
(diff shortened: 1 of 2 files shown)


eth2util/rlp/rlp.go
```diff
@@ -101,8 +101,8 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	if prefix < 0xc0 {
-		length = int(prefix - 0xb7)
-		if length > 63 || length < 0 {
+		length = int(prefix - 0xb7) // length of the string in bytes in binary form
+		if length > 8 || length < 0 {
```
xenowits, line 105: Screenshot 2023-04-11 at 15 50 41
https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/

eth2util/rlp/rlp.go
```diff
@@ -113,7 +113,7 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 			return 0, 0, err
 		} else if resp < 0 {
 			return 0, 0, errors.New("negative length")
-		} else if offset > offset+resp {
+		} else if offset+resp > len(item) {
```
corverroos, line 116: not sure about this

eth2util/rlp/rlp.go
corverroos: length == 8 includes negative integers when casting from uint, this should not be possible.

eth2util/rlp/rlp.go
xenowits: isn't `offset > offset+resp` same as `resp < 0`?

eth2util/rlp/rlp.go
xenowits: On second thought, i think `length == 0` is also NOT possible since the range of the first byte is [`0xb8`, 0xbf]. So, the algorithm expects a positive length (>= 1).

Also, from this comment @corverroos , I believe the correct check should be:
```
if length >= 8 || length <= 0 { ...
```

corverroos: sure can do `if length > 8 || length <= 0`

eth2util/rlp/rlp.go
corverroos: you can move the input length check into the `decodeLength`, but then you should remove it from the caller side

gsora: @xenowits why is this check better than `offset > offset+resp`?

xenowits: @gsora there was a confusion around overflowing: `overflow of length of input` (1) vs `overflow of int64` (2). I'm reverting this check since the caller checks (1) and `decodeLength` checks (2).

corverroos (approved):

eth2util/rlp/rlp_internal_test.go
```diff
@@ -0,0 +1,30 @@
...
+		require.Equal(t, 2, offset)
+		require.Equal(t, 1, length)
+	})
+
+	t.Run("prefix absent", func(t *testing.T) {
+		_, _, err := decodeLength(nil)
+		require.ErrorContains(t, err, "input too short")
+	})
+
+	t.Run("prefix provided but length absent", func(t *testing.T) {
```
corverroos, line 25: what about a test case for the code you changed: `invalid length prefix`?

xenowits: missed it, adding it now

xenowits: On further thought, i think the check `length > 8 || length <= 0` is not possible altogether, it should be a panic if it happens. It should therefore be removed IMO.

Reasoning: 
```
Since
0xb8 <= prefix < 0xc0
=> 184 <= prefix < 192 
=> 184-183 <= prefix-0xb7 < 192-183
=> 1 <= length < 9
=> length range [1,8]
```
If we provide a `value < 0xb8` to trigger `length == 0`, it will be caught before the code block by:
```go
    if prefix < 0xb8 {
        return 1, int(prefix - 0x80), nil
    }
```
And if we provide a `value >= 0xc0`, it will skip the code block altogether:
```go
if prefix < 0xc0 { ... }
```

eth2util/rlp/rlp.go
```diff
@@ -101,9 +101,9 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	if prefix < 0xc0 {
-		length = int(prefix - 0xb7)
-		if length > 63 || length < 0 {
-			return 0, 0, errors.New("invalid length prefix")
+		length = int(prefix - 0xb7)    // length of the string in bytes in binary form
+		if length > 8 || length <= 0 { // This is impossible based on outer if else checks
```
xenowits, line 105: Reasoning:
```
Since
0xb8 <= prefix < 0xc0
=> 184 <= prefix < 192 
=> 184-183 <= prefix-0xb7 < 192-183
=> 1 <= length < 9
=> length range [1,8]
```

eth2util/rlp/rlp.go
```diff
@@ -125,8 +125,8 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	length = int(prefix - 0xf7)
-	if length > 63 || length < 0 {
-		return 0, 0, errors.New("invalid length prefix")
+	if length <= 0 { // This is impossible based on outer if else checks
```
xenowits, line 128: Reasoning:
```
0xf8 <= prefix
0xf8-0xf7 <= prefix-0xf7
1 <= length
```

eth2util/rlp/rlp.go
```diff
@@ -125,8 +125,8 @@ func decodeLength(item []byte) (offset int, length int, err error) {
 	}
 
 	length = int(prefix - 0xf7)
-	if length > 63 || length < 0 {
-		return 0, 0, errors.New("invalid length prefix")
+	if length > 8 || length <= 0 { // This is impossible based on outer if else checks
```
xenowits, line 128: Reasoning:
```
0xf8 <= prefix <= 0xff
0xf8-0xf7 <= prefix-0xf7 <= 0xff-0xf7
1 <= length <= 8
```

corverroos (approved):

Issue #1997, opened 2023-04-03
`nil` Pointer References from Protobuf Messages.


dB2510:
## Problem to be solved

Protobuf allows decoding pointer objects to nil . A nil pointer reference occurs when a program attempts to access
or dereference a pointer that has not been initialised, resulting in a runtime panic.

## Proposed Solution

Check page 10 and 11 of sigma prime audit report.

PR #2088, opened 2023-04-11
dkg: add nil checks

dB2510:
Adds nil pointer checks for relevant functions in dkg package.

Refer: OBOL-03

category: misc
ticket: #1997

### DIFF
--- a/dkg/frostp2p.go
+++ b/dkg/frostp2p.go
@@ -32,11 +32,7 @@ var (
 )
 
 // newFrostP2P returns a p2p frost transport implementation.
-func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k1.PrivateKey, threshold int) (*frostP2P, error) {
-	if secret == nil {
-		return nil, errors.New("secret cannot be nil")
-	}
-
+func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k1.PrivateKey, threshold int) *frostP2P {
 	var (
 		round1CastsRecv = make(chan *pb.FrostRound1Casts, len(peers))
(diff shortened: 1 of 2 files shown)


dkg/frostp2p.go
```diff
@@ -32,7 +32,11 @@ var (
 )
 
 // newFrostP2P returns a p2p frost transport implementation.
-func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k1.PrivateKey, threshold int) *frostP2P {
+func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k1.PrivateKey, threshold int) (*frostP2P, error) {
+	if secret == nil {
+		return nil, errors.New("secret cannot be nil")
+	}
```
corverroos, line 38: I'm not worried about this, as the code should ensure not to pass in nil.

dkg/frostp2p.go
```diff
@@ -122,6 +126,10 @@ func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k
 	p2p.RegisterHandler("frost", tcpNode, round1P2PID,
 		func() proto.Message { return new(pb.FrostRound1P2P) },
 		func(ctx context.Context, pID peer.ID, req proto.Message) (proto.Message, bool, error) {
+			if req == nil {
```
corverroos, line 129: same here, the contract of `RegisterHandler` is that it will never pass in nil

dkg/frostp2p.go
dB2510: cool, I'll revert this one

dkg/frostp2p.go
dB2510: will revert this too

corverroos (approved):

Issue #2083, opened 2023-04-11
updated outdated dependencies (OBOL-12)


xenowits:
# 🎯 Problem to be solved

Update the outdated dependencies.

# 🛠️ Proposed solution

Update the following dependencies:
```
1. github.com/bufbuild/buf v1.14.0 (latest version is v1.17.1)
2. github.com/golang/snappy v0.0.4 (latest version is v0.0.5)
3. github.com/gorilla/mux v1.8.0 (latest version is v1.8.1)
4. github.com/spf13/cobra v1.6.1 (latest version is v1.6.3)
5. github.com/spf13/pflag v1.0.5 (latest version is v1.0.8)
6. github.com/spf13/viper v1.15.0 (latest version is v1.16.0)
7. github.com/stretchr/testify v1.8.1 (latest version is v1.8.3)
8. golang.org/x/oauth2 v0.5.0 (latest version is v0.5.2)
9. gopkg.in/cenkalti/backoff.v1 v1.1.0 (latest version is v2.2.1)
```

Describe the solution to be implemented

# 🧪 Tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

xenowits:
The `latest version` recommendations from the audit are wrong. In particular, all of them are "1 patch version" ahead of latest releases, so they are not released yet. 

Closing this since dependabot already tracks the mentioned dependencies

PR #2089, opened 2023-04-12
beaconmock: remove backoff dependency

xenowits:
Remove `gopkg.in/cenkalti/backoff.v1` direct dependency by removing its usage in `headproducer_internal_test.go`. Though it is still present as an indirect dependency since `sse/v2` uses it.

category: refactor
ticket: https://github.com/ObolNetwork/charon/issues/2083

### DIFF
--- a/testutil/beaconmock/headproducer_internal_test.go
+++ b/testutil/beaconmock/headproducer_internal_test.go
@@ -123,6 +123,9 @@ func TestHeadProducer(t *testing.T) {
 }
 
 // Refer https://github.com/cenkalti/backoff/blob/v4/backoff.go#L46 for the following snippet.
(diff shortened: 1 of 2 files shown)


corverroos (approved):

testutil/beaconmock/headproducer_internal_test.go
```diff
@@ -117,3 +121,16 @@ func TestHeadProducer(t *testing.T) {
 		})
 	}
 }
+
+// Refer https://github.com/cenkalti/backoff/blob/v4/backoff.go#L46 for the following snippet.
+
+// Stop indicates that no more retries should be made for use in NextBackOff().
```
corverroos, line 127: add a comment why we do this and not just backoff dependency

PR #2090, opened 2023-04-12
core: add nil checks

dB2510:
Adds nil checks before reading proto messages in core package mentioned in audit report.

Refer: OBOL-03

category: misc
ticket: #1997

### DIFF
--- a/core/proto.go
+++ b/core/proto.go
@@ -31,6 +31,10 @@ func ParSignedDataFromProto(typ DutyType, data *pbv1.ParSignedData) (ParSignedDa
 	//  For now, it is a good way to catch compatibility issues. But we should
 	//  recover panics and return an error before launching mainnet.
 
+	if data == nil {
+		return ParSignedData{}, errors.New("partial signed data proto cannot be nil")
+	}
+
 	var signedData SignedData
 	switch typ {
 	case DutyAttester:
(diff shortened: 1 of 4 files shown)


core/proto.go
```diff
@@ -179,6 +187,10 @@ func UnsignedDataSetToProto(set UnsignedDataSet) (*pbv1.UnsignedDataSet, error)
 
 // UnsignedDataSetFromProto returns the set from a protobuf.
 func UnsignedDataSetFromProto(typ DutyType, set *pbv1.UnsignedDataSet) (UnsignedDataSet, error) {
+	if set == nil {
+		return nil, errors.New("unsigned data set proto cannot be nil")
+	}
+
 	resp := make(UnsignedDataSet)
 	for pubkey, data := range set.Set {
```
corverroos, line 195: what about set.Set?

corverroos (approved):

core/proto.go
dB2510: for range loop over set.Set will take care of that, so no panic. However, I can add explicit check for that too.

dB2510: Done!
Also added some tests.

dB2510:
@corverroos 
A short note: The audit report also mentioned to add check in `core/proto.go::DutyFromProto()`. I didn't added that as we already do nil checks before passing duty proto to this function wherever it is called. Also adding check to this will modify function signature to return error which impacts uses of this function some other places as well. Not worth it IMO. WDYT?

core/proto.go
corverroos: as, set is a map, in that case, probably better to check for len(set.Set) == 0

Issue #1888, opened 2023-03-15
Validate received DKG messages


corverroos:
# 🎯 Problem to be solved

When receiving message frost round messages from peers in `newFrostP2P`, we do not verify the message contents. See original sigma-prime [thread](https://discord.com/channels/849256203614945310/1077519610116579408/1085529996573032508)

# 🛠️ Proposed solution

Check `FrostMsgKey`:
```
Message Keys:
- FrostRound1Msg
  - Each element of Casts []*FrostRound1Cast check Key *FrostMsgKey
  - Each element of P2Ps ShamirShare check Key *FrostMsgKey
- FrostRound2Msg 
  - Each elemeent of Casts []*FrostRound2Cast check Key *FrostMsgKey


For the FrostMsgKey that are listed above validate
- 0 <= ValIdx < numValidators
- 1 <= SourceId <= numNodes
- 1 <= TargetId <= numNodes
- SourceId matches the peer ID (can check frostP2P.peers[sourceId] == s.Conn().RemotePeer()
- TargetId matches our peer ID (this is done already in that we ignore other messages) 
```

Describe the solution to be implemented

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2107, opened 2023-04-17
dkg: verify valIdx in frost dkg messages

xenowits:
Adds check to ensure validator indexes (`ValIdx`) in frost DKG messages are in the range `[0, num_validators)`.

Also refactors out broadcast and p2p callbacks and adds tests for them.

category: feature
ticket: #1888

### DIFF
--- a/dkg/frostp2p.go
+++ b/dkg/frostp2p.go
@@ -38,6 +38,7 @@ func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k
 		round1P2PRecv   = make(chan *pb.FrostRound1P2P, len(peers))
 		round2CastsRecv = make(chan *pb.FrostRound2Casts, len(peers))
 
+		mu               sync.Mutex
 		dedupRound1Casts = make(map[peer.ID]bool)
 		dedupRound1P2P   = make(map[peer.ID]bool)
 		dedupRound2Casts = make(map[peer.ID]bool)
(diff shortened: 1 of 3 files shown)


corverroos (commented):
Can we add a test?

dkg/frostp2p.go
```diff
@@ -32,13 +32,12 @@ var (
...
 
 // newFrostP2P returns a p2p frost transport implementation.
-func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k1.PrivateKey, threshold int) *frostP2P {
+func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k1.PrivateKey, threshold, numVals int) *frostP2P {
 	var (
 		round1CastsRecv = make(chan *pb.FrostRound1Casts, len(peers))
 		round1P2PRecv   = make(chan *pb.FrostRound1P2P, len(peers))
 		round2CastsRecv = make(chan *pb.FrostRound2Casts, len(peers))
 
-		mu               sync.Mutex
```
gsora, line 41: `bcastCallback` and `p2pCallback` now use two different mutexes instead of sharing `mu`.

Is this expected? Will this bite us in the back down the road?

xenowits: yes true, there are two different mutexes now. I think this is fine and AFAICT doesn't introduce a bug.

The mutex for `bcastCallback` guards the dedup maps for broadcast:
```go
dedupRound1Casts = make(map[peer.ID]bool)
dedupRound2Casts = make(map[peer.ID]bool)
```
While the mutex for `p2pCallback` guards the dedup map for P2P:
```go
dedupRound1P2P   = make(map[peer.ID]bool)
```

xenowits: But, this makes me wonder if we should initialize these maps inside the callback itself instead of passing down as parameters..

xenowits: okay, I'm pushing a change to share mutex instead of declaring new mutex instances inside each function. Since it's better for one mutex to guard all 3 maps:
```go
mu               sync.Mutex
dedupRound1Casts = make(map[peer.ID]bool)
dedupRound1P2P   = make(map[peer.ID]bool)
dedupRound2Casts = make(map[peer.ID]bool)
```

gsora: If anything it's better not to alter the execution flow if it worked well up until now, although I believe that giving each closure their own maps + mutexes is still fine.

dkg/frostp2p.go
```diff
@@ -53,111 +48,137 @@ func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, secret *k
...
+		round2CastsRecv: round2CastsRecv,
+	}
+}
 
-				round1CastsRecv <- msg
-			case round2CastID:
-				mu.Lock()
-				defer mu.Unlock()
+// bcastCallback returns a callback for broadcast in round 1 and round 2 of frost protocol.
+func bcastCallback(peers map[peer.ID]cluster.NodeIdx, round1CastsRecv chan *pb.FrostRound1Casts, round2CastsRecv chan *pb.FrostRound2Casts, threshold, numVals int) bcast.Callback {
```
corverroos, line 71: `newBcastCallback`

corverroos (approved):

PR #2109, opened 2023-04-18
*: implement pidfile for run and dkg

### LINKED ISSUE #1918: Mitigate side-effects due to multiple `charon` commands running at the same time
# 🎯 Problem to be solved

Right now if a user were to run multiple `charon` commands, like two `dkg` at the same time, or `dkg` and `run`, the respective processes will be executed until termination even if both read/write the same files.

This could lead to undefined behavior due to individual operating system's quirks in handling concurrent writes to the same file.

# 🛠️ Proposed solution

Mitigate this possibility by writing a PID lock file in the charon data directory.

Charon will read this file and if present and the contained PID isn't the same as theirs, it will exit reporting an error.

At the end of its execution, charon should delete the associated PID file.

# ❌ Out of Scope
Documentation must be written to communicate to the users that in case of charon crashing, the PID file must be deleted manually.

gsora:
Pidfile allows multiple instances of charon to run on the same data directory, effectively avoiding that users run e.g. `dkg` while having `run` running in the background.

Closes #1918.

category: feature
ticket: #1918

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/hex"
 	"net/http"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -31,6 +32,7 @@ import (
 	"github.com/obolnetwork/charon/app/lifecycle"
 	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/app/peerinfo"
+	"github.com/obolnetwork/charon/app/pidfile"
 	"github.com/obolnetwork/charon/app/promauto"
 	"github.com/obolnetwork/charon/app/retry"
 	"github.com/obolnetwork/charon/app/tracer"
@@ -121,6 +123,17 @@ func Run(ctx context.Context, conf Config) (err error) {
 	defer func() {
 		if err != nil {
 			log.Error(ctx, "Fatal run error", err)
+		}
+	}()
+
+	pidfileDeleteFunc, err := pidfile.New(filepath.Dir(conf.LockFile), "run")
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err := pidfileDeleteFunc(); err != nil {
+			log.Error(ctx, "Cannot delete pidfile", err)
 		}
 	}()
 
--- a/app/pidfile/pidfile.go
+++ b/app/pidfile/pidfile.go
@@ -1,0 +1,105 @@
+// Copyright © 2022-2023 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package pidfile
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"sync/atomic"
+
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+const (
+	filename = "charon-pidfile"
+)
+
+// New creates a pidfile called "charon-pidfile" in dataDir, writing contextStr in it.
+// If a pidfile exists already in dataDir New returns an error, a clean-up function otherwise.
+// New also registers a SIGINT signal handler so that it cleans its state if CTRL-C is called.
+func New(dataDir, contextStr string) (func() error, error) {
+	pfPath := filepath.Join(dataDir, filename)
+
+	if _, err := os.Stat(pfPath); err == nil {
+		readCtxStr, err := os.ReadFile(pfPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read pidfile content even if present")
+		}
+
+		return nil, errors.New(
+			"another instance of charon is running on the selected data directory",
+			z.Str("data_directory", dataDir),
+			z.Str("context", string(readCtxStr)),
+		)
+	} else if errors.Is(err, os.ErrNotExist) {
+		return createPidfile(pfPath, contextStr)
+	} else {
+		return nil, errors.Wrap(err, "fatal error while handling pidfile", z.Str("path", pfPath))
+	}
+}
+
+func createPidfile(path, contextStr string) (func() error, error) {
+	f, err := os.Create(path)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot create pidfile", z.Str("path", path))
+	}
+
+	amt, err := f.WriteString(contextStr)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot write context string in pidfile", z.Str("path", path))
+	}
+
+	if amt != len(contextStr) {
+		return nil, errors.New(
+			"could not write entirety of context string in pidfile",
+			z.Str("path", path),
+			z.Int("expected", len(contextStr)),
+			z.Int("got", amt),
+		)
+	}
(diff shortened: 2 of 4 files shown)


app/pidfile/pidfile.go
```diff
@@ -0,0 +1,105 @@
...
+		)
+	}
+
+	alreadyDeleted := atomic.Bool{}
+	alreadyDeleted.Store(false)
+
+	ctx := context.Background()
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
```
corverroos, line 71: Not of fan of this intricate jumping through hoops, feels brittle. In our applications, we already support shutdown via a root `context.Context`. DKG and Run both support soft shutdown. I would just go for a simple API:
```
func RunCmd(...) {
// First ensure this is the only command
clean, err := pidfile.Create(...)
if err != nil {
  return errors.Wrap(err, "cannot create pidfile")
}
defer clean() // Cleanup on exit

... // Do the command
```

app/app.go
```diff
@@ -124,6 +126,17 @@ func Run(ctx context.Context, conf Config) (err error) {
 		}
 	}()
 
+	pidfileDeleteFunc, err := pidfile.New(filepath.Dir(conf.LockFile), "run")
```
corverroos, line 129: can we be less verbose? `cleanPID, err := ...`

app/pidfile/pidfile.go
```diff
@@ -0,0 +1,105 @@
...
+		<-c
+
+		defer os.Exit(0)
+
+		if alreadyDeleted.Load() {
+			log.Debug(ctx, "Pidfile already deleted, not deleting from ctrl-c handler")
+			return
+		}
+
+		alreadyDeleted.Store(true)
```
corverroos, line 82: the delete might fail in which case it wasn't deleted

app/pidfile/pidfile.go
```diff
@@ -0,0 +1,105 @@
...
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		<-c
+
+		defer os.Exit(0)
+
+		if alreadyDeleted.Load() {
+			log.Debug(ctx, "Pidfile already deleted, not deleting from ctrl-c handler")
```
corverroos, line 78: these debug logs feel noisy, lets remove it

corverroos (changes requested):

app/pidfile/pidfile.go
corverroos: basically I do not think we need to also listen to interupts, that is the responsibility of the application, not the library

app/pidfile/pidfile.go
gsora: I'll remove the SIGINT subscribe but just for the sake of clarity: in case deleting the file was not possible at this stage, it means there's something seriously wrong with the system that runs charon.

It's the edge case we were talking about when we originally discussed this: if we're failing here, the user needs to be aware of how to delete the pidfile and kickstart charon again.

This still holds true, because the deletion closure could fail anyway: it's a documentation problem that we will have to solve.

app/pidfile/pidfile.go
```diff
@@ -0,0 +1,67 @@
...
+			z.Str("context", string(readCtxStr)),
+		)
+	} else if errors.Is(err, os.ErrNotExist) {
+		return createPidfile(pfPath, contextStr)
+	} else {
+		return nil, errors.Wrap(err, "fatal error while handling pidfile", z.Str("path", pfPath))
+	}
+}
+
+func createPidfile(path, contextStr string) (func() error, error) {
```
corverroos, line 40: why does create return the cleanup func? Shouldn't create just create and the main function return a cleanup func?

gsora: Remainings from the old approach, I can inline this in the calling func.

gsora: Cleaned.

cmd/run.go
```diff
@@ -127,6 +128,10 @@ func bindFeatureFlags(flags *pflag.FlagSet, config *featureset.Config) {
 	flags.StringVar(&config.MinStatus, "feature-set", "stable", "Minimum feature set to enable by default: alpha, beta, or stable. Warning: modify at own risk.")
 }
 
+func bindPrivkeyLockFlag(flags *pflag.FlagSet, privkeyLockEnabled *bool) {
```
corverroos, line 131: can colocate with `bindPrivKeyFlag` functoin

app/app.go
```diff
@@ -111,6 +113,8 @@ type TestConfig struct {
 	TCPNodeCallback func(host.Host)
 	// LibP2POpts provide test specific libp2p options.
 	LibP2POpts []libp2p.Option
+	// SkipPidfile skips pidfile creation and checks
+	SkipPidfile bool
```
corverroos, line 117: can remove this and use `PrivkeyLockingEnabled`

app/pidfile/pidfile.go
```diff
@@ -0,0 +1,65 @@
...
+package pidfile
+
+import (
+	"os"
+
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+// New creates a pidfile called "charon-pidfile" in dataDir, writing contextStr in it.
```
corverroos, line 12: comment need updating

app/app.go
```diff
@@ -124,6 +128,19 @@ func Run(ctx context.Context, conf Config) (err error) {
 		}
 	}()
 
+	if !conf.TestConfig.SkipPidfile || conf.PrivkeyLockingEnabled {
+		cleanPID, err := pidfile.New(conf.PrivKeyFile+".lock", "run")
```
corverroos, line 132: suggest `charon run` as just `run` might not be very clear on what it means

app/app.go
```diff
@@ -124,6 +128,19 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		defer func() {
+			if err := cleanPID(); err != nil {
+				log.Error(ctx, "Cannot delete pidfile", err)
```
corverroos, line 139: not a pidfile anymore

app/pidfile/pidfile.go
corverroos: suggest rename package to `lockfile` or something

corverroos (approved):

app/privkeylock/privkeylock.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			return nil
+		}, nil
+	} else {
+		return nil, errors.Wrap(err, "fatal error while handling pidfile", z.Str("path", path))
+	}
+}
+
+func createPidfile(path, contextStr string) error {
```
dB2510, line 44: this would be `func createPrivKeyLockFile(path, contextStr string) error ` as this is not a pidfile anymore. Can you also add godoc to this function?

dB2510 (commented):
@gsora there are still multiple mentions of `pidfile` in error messages in privkeylock package

app/privkeylock/privkeylock.go
gsora: Addressed.

PR #2110, opened 2023-04-19
core/consensus: implement simulator test

corverroos:
Implements a basic consensus simulator using a fake clock that is able to check succesful consensus given different network latencies, start times, and round timers. Next step is to expand it into a benchmark to compare different round timer implementations.

category: test
ticket: none

### DIFF
--- a/core/consensus/strategysim_internal_test.go
+++ b/core/consensus/strategysim_internal_test.go
@@ -1,0 +1,358 @@
+// Copyright © 2022-2023 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package consensus
+
+import (
+	"context"
+	"math/rand"
+	"sort"
+	"sync"
+	"testing"
+	"time"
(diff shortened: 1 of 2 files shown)


core/consensus/strategysim_internal_test.go
```diff
@@ -0,0 +1,358 @@
...
+	latencyStdDev  time.Duration
+	latencyPerPeer map[int64]time.Duration
+	startByPeer    map[int64]time.Duration
+	roundTimerFunc func(clockwork.Clock) roundTimer
+	timeout        time.Duration
+}
+
+func testStrategySimulator(t *testing.T, conf ssConfig) []result {
+	t.Helper()
+	random := rand.New(rand.NewSource(int64(conf.seed)))
```
xenowits, line 82: nit: add a newline after `t.Helper()`

xenowits (approved):

Issue #1971, opened 2023-03-29
Nightly test doing long-running DKG


thomasheremans:
# Problem to be solved

We need an integration type tests for long-wait DKG.

# Proposed Solution

Write a script/test that does a long-wait DKG using latest main and public relays. 

Suggest something like:
 - Start a new node every 10min node
 - Each node randomly stops for 2min in every 10min window.
 - DKG should complete within 10min after last node started. 
 
 Run this test once a day (at noon UTC)

xenowits:
Next step is to add a github workflow to ensure this test runs once a day at noon UTC. 

We can use `on.schedule.cron` to schedule a cron task for this. See [codeql-analysis workflow](https://github.com/ObolNetwork/charon/blob/main/.github/workflows/codeql-analysis.yml#L20) for an example.

PR #2111, opened 2023-04-19
integration: long-wait DKG test

xenowits:
Adds an integration test for long-wait DKG.

category: test 
ticket: #1971

### DIFF
--- a/testutil/integration/longwait_dkg_test.go
+++ b/testutil/integration/longwait_dkg_test.go
@@ -23,92 +23,124 @@ import (
 )
 
 func TestLongWaitDKG(t *testing.T) {
-	// skipIfDisabled(t)
+	skipIfDisabled(t)
 
 	const (
-		threshold     = 3
-		numNodes      = 4
-		numVals       = 10
-		sleepDuration = 10 * time.Second
-		relayURL      = "https://0.relay.obol.tech"
+		threshold = 3
+		numNodes  = 4
+		numVals   = 10
+		relayURL  = "https://0.relay.obol.tech"
+		// The time period after which a new node is started.
+		window = 10
+		// The time period when a node is offline. A window is divided into multiple slots each of nodeDownPeriod duration.
+		// A node goes down in one of these slots in a particular window.
+		nodeDownPeriod = 2
 	)
 
-	dir := t.TempDir()
-	startedChan := make(chan struct{})
+	var (
+		eg              errgroup.Group
+		ctx             = log.WithTopic(context.Background(), "longwait")
+		dir             = t.TempDir()
+		def             = testDef(t, dir, threshold, numNodes, numVals)
+		allNodesStarted = make(chan struct{}, numNodes)
+	)
 
-	startDKG := func(ctx context.Context, def cluster.Definition, nodeIdx int) error {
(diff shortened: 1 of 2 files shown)


xenowits:
The integration test flow looks like this:

longwaitdkg

testutil/integration/longwait_dkg_test.go
```diff
@@ -0,0 +1,162 @@
...
+			return mimicDKGNode(ctx, t, def, dir, relayURL, window, nodeDownPeriod, indx, allNodesStarted)
+		})
+	}
+
+	err := eg.Wait()
+	require.NoError(t, err)
+}
+
+// mimicDKGNode mimics the behaviour of a DKG node that randomly stops for sometime before restarting again but finally participates in the DKG.
+func mimicDKGNode(ctx context.Context, t *testing.T, def cluster.Definition, dir, relayURL string, window, nodeDownPeriod, nodeIdx int, allNodesStarted chan struct{}) error {
```
corverroos, line 72: I think we can simplify this

testutil/integration/longwait_dkg_test.go
```diff
@@ -0,0 +1,162 @@
...
+func testDef(t *testing.T, dir string, threshold, numNodes, numVals int) cluster.Definition {
+	t.Helper()
+
+	const frostAlgo = "frost"
+
+	withAlgo := func(algo string) func(*cluster.Definition) {
+		return func(d *cluster.Definition) {
+			d.DKGAlgorithm = algo
+		}
+	}
```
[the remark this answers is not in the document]
xenowits: yeah, that's correct, `frost` is default. I'll remove it

testutil/integration/longwait_dkg_test.go
```diff
@@ -0,0 +1,184 @@
...
+		t.Skip("Long wait test is disabled")
+	}
+
+	const (
+		threshold = 3
+		numNodes  = 4
+		numVals   = 10
+		relayURL  = "https://0.relay.obol.tech"
+		// The time period after which a new node is started.
+		window = 10 * time.Second
```
gsora, line 58: This constant is too small for a real-world test.

Parametrize `window` and `nodeDownPeriod`, or provide sensible defaults and a `quick-run` flag which overrides them to 10s and 2s respectively.

testutil/integration/longwait_dkg_test.go
```diff
@@ -0,0 +1,184 @@
...
+			break
+		}
+
+		if allStarted { // Do the final DKG since all nodes are up now
+			break
+		}
+
+		ctx, cancel := context.WithCancel(ctx)
+		go func(ctx context.Context) {
+			_ = dkg.Run(ctx, dkgConf)
```
gsora, line 135: Just as a safety measure, I would advice to check this error and require that it originates from `ctx` being cancelled.

On top of that to avoid any kind of conflict, maybe after calling `cancel()` wait that this goroutine effectively finished?

testutil/integration/longwait_dkg_test.go
xenowits: Agreed. I'll modify the defaults to be in `minutes` instead. Will also the `quick-run` flag as you suggested.

However, note that I'm refactoring this PR as the current version contains a bug. Will re-request review once I'm ready

Issue #1538, opened 2022-12-12
Extend inclusion delay logic to track failed attestation


dB2510:
## Problem to be solved

To track cluster wide failed attestation, currently we need to use external tools like beaconcha.in or rated APIs. Every peer reports failed attestation from tracker component but that reporting is local to each peer. We need a method to track cluster wide failed attestation.

## Proposed solution

- Extend inclusion delay logic in order to track failed attestations.
- If an attestation isn't included in next 32 slots (1 epoch) then it should be marked as failed/missed. Refer: 
   - https://kb.beaconcha.in/attestation#attestation-scenarios 
   - https://www.youtube.com/watch?v=SPcgevcDqDE 
   - https://www.attestant.io/posts/defining-attestation-effectiveness/
- Add a counter metric: `core_tracker_failed_attestations_total`.
- Add a panel for the above metric to simnet overview dashboard.

corverroos:
What should the initial delay be at startup before we can mark attestations as failed? Since charon only starts tracking blocks from startup (with a delay as well).

thomasheremans:
First Design Doc

PR #2120, opened 2023-04-21
core/tracker: refactor inclusion checker

corverroos:
Refactors the inclusion delay tracker to also detect missed inclusions and to log broadcast delay of included and missed duties. The aim is to find a correlation between broadcast performance and inclusion delay.

category: feature
ticket: #1538


app/app.go
```diff
@@ -398,9 +398,13 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 	feeRecipientFunc := func(pubkey core.PubKey) string {
 		return feeRecipientAddrByCorePubkey[pubkey]
 	}
-
 	sched.SubscribeSlots(setFeeRecipient(eth2Cl, eth2Pubkeys, feeRecipientFunc))
-	sched.SubscribeSlots(tracker.NewInclDelayFunc(eth2Cl, sched.GetDutyDefinition))
+
+	inclDelay, err := tracker.NewInclusionDelay(ctx, eth2Cl, sched.GetDutyDefinition)
+	if err != nil {
+		return err
```
dB2510, line 405: nit: can you wrap this error just for the sake of context?

core/tracker/incldelay.go
```diff
@@ -6,112 +6,211 @@ import (
...
+	dutiesFunc     dutiesFunc
+	instrumentFunc func(inclDelaySlots []int64)
+	logMappingFunc func(ctx context.Context, slot int64, bcastDelay time.Duration, inclDelay int64, isAggregated bool)
+	genesisTime    time.Time
+	slotDuration   time.Duration
+	slotsPerEpoch  int
+	clock          clockwork.Clock
+
+	mu            sync.Mutex
+	bcastDelays   map[int64]map[eth2p0.Root]time.Duration
```
dB2510, line 75: so we are storing broadcast delays for attestations per slot per block root, but we can submit multiple attestations with the same slot and block root. 

This map will only take care of one of those attestation. I think if we are tracking then we should track all of them. Maybe modify the map as `map[int64]map[eth2p0.BLSPubKey]time.Duration` where we can get the pubkey from duty.

dB2510: Also while iterating all the attestation in the block this broadcast might not EXACTLY map with the desired attestation if we have submitted multiple attestations with the same and block root.

core/tracker/inclusion.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		switch sub.Duty.Type {
+		case core.DutyAttester, core.DutyAggregator:
+			_, ok := block.Attestations[sub.AttRoot]
+			if !ok {
+				continue
+			}
+			i.submissions[j].Included = true
+			i.attIncludedFunc(ctx, sub, block)
```
dB2510, line 144: we need to also check the case where our single unaggregated attestation is included as part of aggregated attestation. Here you are checking if our unaggregated attestation is included or not OR aggregated attestation submitted by us is included or not.

AttRoot of an unaggregated attestation will have aggregation bits with only one bit set. While we need to check for the aggregation bits of an aggregated block attestation at the validator committee index like we were doing earlier.

core/tracker/inclusion.go
```diff
@@ -0,0 +1,300 @@
...
+			z.Any("pubkey", sub.Pubkey),
+			z.I64("block_slot", sub.Duty.Slot),
+			z.Any("broadcast_delay", sub.Delay),
+		)
+	default:
+		panic("bug: unexpected type") // Sanity check, this should never happen
+	}
+}
+
+func reportAttInclusion(ctx context.Context, sub submission, block block) {
```
dB2510, line 189: no need to pass full block, just pass `blockSlot`. `AttSlot` can be taken from submission.

dB2510: Also add godoc

core/tracker/inclusion.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+		if !sub.Included {
+			i.missedFunc(ctx, sub)
+		}
+	}
+	i.submissions = remaining
+}
+
+// CheckBlock checks whether the block includes any of the submitted duties.
+func (i *inclusion) CheckBlock(ctx context.Context, block block) {
```
dB2510, line 128: why this is a public method?

core/tracker/inclusion.go
```diff
@@ -0,0 +1,300 @@
...
+			missedFunc:      reportMissed,
+		},
+		eth2Cl:       eth2Cl,
+		genesis:      genesis,
+		slotDuration: slotDuration,
+	}, nil
+}
+
+// InclusionChecker checks whether duties have been included in blocks.
+type InclusionChecker struct {
```
dB2510, line 234: why you needed to wrap inclusion with InclusionChecker? Also some methods on `inclusion` struct are public which can be private as they only called by InclusionChecker methods which are exposed to other packages.

core/tracker/inclusion.go
corverroos: good catch, added "aggregated" logging field

core/tracker/inclusion.go
corverroos: Does need more stuff

core/tracker/inclusion.go
corverroos: it is its API

core/tracker/inclusion.go
corverroos: renamed it to `inclusionChecker` which wrap a `inclusionCore`. The core is decoupled and optimised for easy testing.

core/tracker/incldelay.go
```diff
@@ -1,124 +0,0 @@
...
-				duty, ok := def.(core.AttesterDefinition)
-				if !ok {
-					return errors.New("invalid attester definition")
-				}
-
-				if duty.CommitteeIndex != att.Data.Index {
-					continue // This duty is for another committee
-				}
-
-				if !att.AggregationBits.BitAt(duty.ValidatorCommitteeIndex) {
```
dB2510, line 98: @corverroos you need to add this check in new inclusion checker

corverroos: done

dB2510: I think you haven't pushed that commit. It might be in local commit?

core/tracker/inclusion.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	log.Info(ctx, msg,
+		z.I64("block_slot", blockSlot),
+		z.I64("attestation_slot", attSlot),
+		z.Any("pubkey", sub.Pubkey),
+		z.I64("inclusion_delay", inclDelay),
+		z.Any("broadcast_delay", sub.Delay),
+		z.Int("aggregate_len", len(aggIndices)),
+		z.Bool("aggregated", len(aggIndices) > 1),
```
corverroos, line 229: @dB2510 the aggregated logging field is added here

core/tracker/incldelay.go
corverroos: Cool, should be here now: https://github.com/ObolNetwork/charon/pull/2120/files#r1174812448

core/tracker/inclusion.go
```diff
@@ -0,0 +1,325 @@
...
+		switch sub.Duty.Type {
+		case core.DutyAttester, core.DutyAggregator:
+			_, ok := block.Attestations[sub.AttRoot]
+			if !ok {
+				remaining = append(remaining, sub)
+				continue
+			}
+
+			// TODO(corver): We should probably check if validator included in AggregationBits
+			//  for attester duty by caching scheduler duty data to get validator committee index.
```
dB2510, line 158: cool, i was looking for this. Will push in next PR probably?

core/tracker/inclusion.go
dB2510: Just to clarify, this aggregated logging will tell us that attestation is an aggregate attestation and NOT an unaggregated one. Well here `aggregated: true` only when we are aggregator of that slot AND it is our aggregate attestation.

core/tracker/inclusion.go
```diff
@@ -0,0 +1,325 @@
...
+	for _, sub := range i.submissions {
+		switch sub.Duty.Type {
+		case core.DutyAttester, core.DutyAggregator:
+			_, ok := block.Attestations[sub.AttRoot]
+			if !ok {
+				remaining = append(remaining, sub)
+				continue
+			}
+
+			// TODO(corver): We should probably check if validator included in AggregationBits
```
xenowits, line 157: Yeah. If we wish, we can also understand if the attestation was aggregated or not by checking
```go
if NO_OF_SET_BITS(attestation.aggregation_bits) == 1 { 
  "attestation was NOT aggregated" 
}
```

core/tracker/inclusion.go
corverroos: yeah

dB2510 (approved):

Issue #1396, opened 2022-11-08
Cache validators by pubkey in eth2client


corverroos:
## Problem to be solved

The eth2client endpoint with the highest call rate is `validators_by_pubkey`. We call that in most validator api endpoints. It is also pretty slow. But the data is pretty static, it only changes every epoch as new validators are activated and exited/slashed.

## Proposed solution

Add an eth2wrap option `WithEpochCache` that:
 - caches validators by pubkey
 - subscribes to scheduler events and clears the cache on new epochs

corverroos:
See high call rate, high latency and warns/errors fetching validators by pubkey
<img width="1442" alt="image" src="https://user-images.githubusercontent.com/29249923/200509779-d4d8f280-7245-4457-9d04-e8872264f30f.png">

OisinKyne:
How heavy is this hitting our Obol Performance Cluster? @LukeHackett12

LukeHackett12:
I haven't seen this request popping up in errors afaik, taking a deeper look now.

OisinKyne:
![Image](https://user-images.githubusercontent.com/4981644/212060227-5b061d5f-6c2a-428f-8988-eaca9a55b723.png)

Looking at the performance cluster now, by far the heaviest request is `get_validator`, which, considering this is lighthouse whereas I guess corver's is teku, is probably the same functionality but lighthouse is querying by index and teku is asking by pub key. But this indicates that if we could cache stuff we could massively reduce our BN load at scale.

LukeHackett12:
Ah OK sorry lacked a bit of context didn't know they were equivalent should have guessed.

OisinKyne:
All good, np

gsora:
I believe this was implemented in https://github.com/ObolNetwork/charon/pull/1397

PR #2121, opened 2023-04-23
app/eth2wrap: implement and wire validator cache

corverroos:
This is a second attempt at caching active validators. Instead of caching the `eth2client.ValidatorProvider` endpoints which return full `Validator` objects that are not actually immutable by epoch, rather implement a new set of client endpoint specifically for cached immutable active validators by current epoch that only return the validator index and pubkey which are immutable. The only race condition is the fact that we query this by "head", so could be for another slot that what we think is current, although this was already the case previously, so no new races are introdcued.

This should significantly decrease BN load, especially if using the validatormock.  

category: feature
ticket: #1396

### DIFF
--- a/app/eth2wrap/httpwrap.go
+++ b/app/eth2wrap/httpwrap.go
@@ -37,34 +37,6 @@ type NodePeerCountProvider interface {
 	NodePeerCount(ctx context.Context) (int, error)
 }
 
-type ActiveValidators map[eth2p0.ValidatorIndex]eth2p0.BLSPubKey
-
-// Pubkeys returns a list of active validator pubkeys.
-func (m ActiveValidators) Pubkeys() []eth2p0.BLSPubKey {
-	var pubkeys []eth2p0.BLSPubKey
-	for _, pubkey := range m {
-		pubkeys = append(pubkeys, pubkey)
-	}
-
-	return pubkeys
-}
-
-// Indices returns a list of active validator indices.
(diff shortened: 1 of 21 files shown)


xenowits (approved):
LGTM!

app/eth2wrap/httpwrap.go
```diff
@@ -36,6 +37,34 @@ type NodePeerCountProvider interface {
 	NodePeerCount(ctx context.Context) (int, error)
 }
 
+type ActiveValidators map[eth2p0.ValidatorIndex]eth2p0.BLSPubKey
```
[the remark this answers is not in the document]
gsora: another nit: maybe this type should live in `valcache.go`?

app/eth2wrap/valcache.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		if !val.Status.IsActive() {
+			continue
+		}
+
+		resp[val.Index] = val.Validator.PublicKey
+	}
+
+	c.active = resp
```
dB2510, line 109: I'd suggest to lock mutex around this (when `c.active` is updated) and not above as we are calling `eth2Cl` after locking mutex.

dB2510 (approved):

app/eth2wrap/valcache.go
corverroos: no the point here is to lock while only one call to populate the cache is done

dB2510: Got it 👍

Issue #2122, opened 2023-04-23
verify reconstructed signature in sigagg (OBOL-17)


xenowits:
# 🎯 Problem to be solved

Currently, we don't verify the reconstructed aggregated signature in the `sigAgg` component.

So, within the function `Aggregate()`, if malformed partial signatures have been supplied, the reconstructed signature will be invalid. The invalid signature will be unnoticed and passed to `sigAggDB` .

For more details, see `OBOL-17` of sigp audit.

# 🛠️ Proposed solution

Verify aggregated signature before passing data to subscribers of the `sigAgg` component. Also add tests.

# 🧪 Tests

None.

PR #2123, opened 2023-04-23
core/sigagg: verify aggregated signature

xenowits:
Verify reconstructed aggregated signature.

Fixes OBOL-17 of sigp audit.

category: feature 
ticket: #2122

### DIFF
--- a/core/sigagg/sigagg.go
+++ b/core/sigagg/sigagg.go
@@ -101,8 +101,8 @@ func (a *Aggregator) Aggregate(ctx context.Context, duty core.Duty, pubkey core.
 	return nil
 }
 
-// NewSigAggVerifier returns a signature verification function for aggregated signatures.
-func NewSigAggVerifier(eth2Cl eth2wrap.Client) func(context.Context, core.PubKey, core.SignedData) error {
+// NewVerifier returns a signature verification function for aggregated signatures.
+func NewVerifier(eth2Cl eth2wrap.Client) func(context.Context, core.PubKey, core.SignedData) error {
 	return func(ctx context.Context, pubkey core.PubKey, data core.SignedData) error {
 		tblsPubkey, err := tblsconv.PubkeyFromCore(pubkey)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


core/sigagg/sigagg.go
```diff
@@ -89,3 +95,25 @@ func (a *Aggregator) Aggregate(ctx context.Context, duty core.Duty, pubkey core.
...
+	}
+
+	msgRoot, err := sigData.MessageRoot()
+	if err != nil {
+		return errors.Wrap(err, "message root")
+	}
+
+	aggSig := tbls.Signature(sigData.Signature())
+
+	err = tbls.Verify(tblsPubkey, msgRoot[:], aggSig)
```
corverroos, line 113: this is a eth2 signature, so it should include eth2 domains. I do not want to add a eth2client to sigagg, so I would actually mark this as WONTFIX.

Reasoning:
  - We already validate the partial signatures.
  - So if aggregation succeeds then resulting aggregate signature is also valid. 
  - Additional validation isn't required.

corverroos (changes requested):

core/sigagg/sigagg.go
corverroos: If invalid aggregate signatures are however still possible given valid partial signatures and succesful aggregation, then we should follow the parsigex approach of decoupling validation from the component via a `verifyFunc`.

corverroos: See parsigex how to verify eth2 signatures

xenowits: I would say we err on the side of caution here and still validate aggregated signatures. I'll follow the `verifyFunc` approach of verifying signatures

core/sigagg/sigagg.go
```diff
@@ -89,3 +100,25 @@ func (a *Aggregator) Aggregate(ctx context.Context, duty core.Duty, pubkey core.
 
 	return nil
 }
+
+// NewSigAggVerifier returns a signature verification function for aggregated signatures.
+func NewSigAggVerifier(eth2Cl eth2wrap.Client) func(context.Context, core.PubKey, core.SignedData) error {
```
corverroos, line 105: avoid stutter, suggest `NewEth2Verifier` or just `NewVerifier` since there isn't a second implementation

corverroos (approved):

PR #2124, opened 2023-04-24
core/tracker: check aggbits for inclusion

corverroos:
When checking attestation and aggregation inclusion, check whether our submissions's aggbits are included in the block attestation aggbits. Also fix matching using attestation data root (not attestation root which changes on additional aggregation). 

category: feature 
ticket: #1538

### DIFF
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -160,6 +160,13 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
+// InclusionChecker checks whether duties have been included on-chain.
+// TODO(corver): Merge this with tracker below as a compose multi tracker.
+type InclusionChecker interface {
+	// Submitted is called when a duty has been submitted.
+	Submitted(duty Duty, pubkey PubKey, data SignedData) error
+}
+
 // Tracker sends core component events for further analysis and instrumentation.
 type Tracker interface {
 	// FetcherFetched sends Fetcher component's events to tracker.
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -160,7 +160,7 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
-// InclusionChecker checks whether duties have been included on-chain.
+// InclusionChecker checks whether submitted duties have been included on-chain.
 // TODO(corver): Merge this with tracker below as a compose multi tracker.
 type InclusionChecker interface {
 	// Submitted is called when a duty has been submitted.
--- a/core/tracker/inclusion.go
+++ b/core/tracker/inclusion.go
@@ -28,17 +28,17 @@ const (
 
 // submission represents a duty submitted to the beacon node/chain.
 type submission struct {
-	Duty    core.Duty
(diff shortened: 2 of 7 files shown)


core/interfaces.go
```diff
@@ -160,6 +160,13 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
+// InclusionChecker checks whether duties have been included on-chain.
```
xenowits, line 163: nit:
```suggestion
// InclusionChecker checks whether submitted duties have been included on-chain.
```

core/tracker/inclusion.go
```diff
@@ -311,15 +363,19 @@ func (a *InclusionChecker) checkBlock(ctx context.Context, slot int64) error {
 
 	attsMap := make(map[eth2p0.Root]*eth2p0.Attestation)
 	for _, att := range atts {
-		root, err := att.HashTreeRoot()
+		if att == nil || att.Data == nil {
+			return errors.New("nil attestation")
```
xenowits, line 367: nit
```suggestion
			return errors.New("invalid attestation")
```

core/interfaces.go
```diff
@@ -160,6 +160,13 @@ type Broadcaster interface {
 	Broadcast(context.Context, Duty, PubKey, SignedData) error
 }
 
+// InclusionChecker checks whether submitted duties have been included on-chain.
+// TODO(corver): Merge this with tracker below as a compose multi tracker.
+type InclusionChecker interface {
+	// Submitted is called when a duty has been submitted.
```
dB2510, line 166: nit:
```suggestion
	// Submitted is called when a duty has been submitted to beacon node.
```

testutil/random.go
```diff
@@ -87,7 +87,15 @@ func RandomValidator(t *testing.T) *eth2v1.Validator {
...
-		AggregationBits: RandomBitList(),
+		AggregationBits: RandomBitList(1),
+		Data:            RandomAttestationData(),
+		Signature:       RandomEth2Signature(),
+	}
+}
+
+func RandomAggregateAttestation() *eth2p0.Attestation {
+	return &eth2p0.Attestation{
+		AggregationBits: RandomBitList(64),
```
dB2510, line 98: According to spec: length of AggregationBits should be `MAX_VALIDATORS_PER_COMMITTEE` which is 2048. Setting this 64 might be problematic when testing with VCs on simnet if they check this on their end.

dB2510 (approved):
LGTM 🚀

testutil/random.go
corverroos: lets cross that bridge when/if we get there

core/tracker/inclusion.go
```diff
@@ -126,9 +126,11 @@ func (i *inclusionCore) Trim(ctx context.Context, slot int64) {
 	i.mu.Lock()
 	defer i.mu.Unlock()
 
+	// Trim submissions
 	var remaining []submission
 	for _, sub := range i.submissions {
 		if sub.Duty.Slot > slot {
+			// Keep
```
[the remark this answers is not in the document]
xenowits: not sure about the comment, seems incomplete

PR #2126, opened 2023-04-24
app: refactor lockfile fork version mismatch error

gsora:
Don't wrap old, nil error when a lockfile fork mismatch one is being created.

category: refactor
ticket: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -732,7 +732,7 @@ func newETH2Client(ctx context.Context, conf Config, life *lifecycle.Manager,
 		}
 	}
 	if !ok {
-		return nil, errors.Wrap(err, "lock file fork version not in beacon node fork schedule (probably wrong chain/network)")
+		return nil, errors.New("lock file fork version not in beacon node fork schedule (probably wrong chain/network)")
 	}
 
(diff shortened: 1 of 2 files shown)


dB2510 (approved):

corverroos (approved):

app/app.go
```diff
@@ -732,7 +732,7 @@ func newETH2Client(ctx context.Context, conf Config, life *lifecycle.Manager,
 		}
 	}
 	if !ok {
-		return nil, errors.Wrap(err, "lock file fork version not in beacon node fork schedule (probably wrong chain/network)")
+		return nil, errors.New("lock file fork version not in beacon node fork schedule (probably wrong chain/network)")
```
corverroos, line 735: another occurrence of this in `dutydb/memory.go`

corverroos: Maybe double check these are the only ones?

gsora: > Maybe double check these are the only ones?

How would you do that? Needs a linter?

corverroos: linter sure, or just ad-hoc multi-line regex?

PR #2127, opened 2023-04-24
app/privkeylock: add periodic updates and timestamps

gsora:
Refactor the privkeylock to contain a timestamp which is periodically updated. This mitigates the need to manually delete the file after a short delay. Also re-introduce signal handling and closing of context into the main package.

category: bug
ticket: none


gsora:
@corverroos refactored SIGINT handling to be explicit in `main`, now leaf functions register their shutdown logic alongside a urgency value: higher urgency makes that shutdown function be executed earlier than others.

`main` package calls for shutdown SIGINT registration once, instead of having potentially multiple handlers thorough the codebase.

Marked the PR in draft because it's a delicate topic and I'd love to hear your input.

corverroos:
Yeah, I'd go for something much simpler, just cancel the context from `main.go`:

```
ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
defer cancel()

cobra.CheckErr(newRootCmd().ExecuteContext(ctx))
```

app/app.go
```diff
@@ -132,6 +135,18 @@ func Run(ctx context.Context, conf Config) (err error) {
 			return err
 		}
 
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
corverroos, line 139: no need to do this, a defer should suffice

corverroos (changes requested):

app/app.go
gsora: nvm that should be gone already

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
+	path       string
+
+	clean func() error
+}
+
+// Run updates the lock file pointed by the path once every second.
+// It must be run asynchronously.
+func (h Handle) Run(ctx context.Context) error {
+	tick := time.NewTicker(1 * time.Second)
+
```
corverroos, line 45: probably better to: `defer tick.Stop()` to also cover error cases

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
 	}
 
-	amt, err := f.WriteString(contextStr)
+	lctx := newLockCtx(contextStr)
+	content, err := json.Marshal(lctx)
+	if err != nil {
+		return errors.Wrap(err, "cannot marshal private key lock context")
+	}
+
+	amt, err := f.Write(content)
```
corverroos, line 152: can simplify by using `os.WriteFile` for both create and update

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
+				)
+			}
+		}
+	}
+}
+
+// New creates a private key lock file in path, writing ContextStr in it.
+// If a private key lock file exists at path New returns an error, a Handle otherwise.
+func New(path, contextStr string) (Handle, error) {
 	if _, err := os.Stat(path); err == nil {
```
corverroos, line 80: can simplify by using `os.ReadFile()`

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
-			return nil, errors.Wrap(err, "could not read private key lock file content even if present")
+			return Handle{}, errors.Wrap(err, "could not read private key lock file content even if present")
 		}
 
-		return nil, errors.New(
-			"another instance of charon is running for the selected private key, check if there is another charon instance running on your machine",
-			z.Str("path", path),
-			z.Str("context", string(readCtxStr)),
-		)
+		lctx := privkeyLockCtx{}
```
corverroos, line 86: nit, we mostly use the following style:
```
var content privkeyLockCtx
if err := json.Unmarshal(roCtxFile, &content); err != nil {
			return Handle{}, errors.Wrap(err, "cannot decode private key lock file content")
}
```

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
 )
 
-// New creates a private key lock file in path, writing contextStr in it.
-// If a private key lock file exists at path New returns an error, a clean-up function otherwise.
-func New(path, contextStr string) (func() error, error) {
+var lockfileGracePeriod = func() time.Duration {
+	return 5 * time.Second
+}
+
+type privkeyLockCtx struct {
```
corverroos, line 19: not sure about naming:
 - it stutters with package name
 - Ctx is usually used for context.Context
 
 what about `fileContent` or `metadata`?

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
+}
+
+func newLockCtx(contextStr string) privkeyLockCtx {
+	return privkeyLockCtx{
+		ContextStr: contextStr,
+		Timestamp:  time.Now(),
+	}
+}
+
+// Handle contains the necessary fields to lock the private key file periodically.
```
corverroos, line 31: not sure about name `Handle`, what about `Service` since that is a bit more intuative?

app/privkeylock/privkeylock.go
```diff
@@ -3,65 +3,172 @@
...
+// It also contains the Clean function, which must be called by callers to remove the private key
+// lock once a graceful shutdown phase is reached.
+type Handle struct {
+	contextStr string
```
corverroos, line 35: what about rename this to `command` to be a bit more clean on what this is.

corverroos (changes requested):

app/privkeylock/privkeylock.go
gsora: `metadata` sounds good!.

corverroos (approved):

PR #2161, opened 2023-05-08
core: wire async consensus

corverroos:
Consensus has a blocking API so should be called async, do that via the retryer.

category: misc
ticket: none


xenowits (approved):
LGTM!

PR #2174, opened 2023-05-10
eth2util/signing: add constant genesis domain test

corverroos:
Simple unit test demonstrating that DomainApplicationBuilder uses genesis fork version domain.

category: test
ticket: none

### DIFF
--- a/eth2util/signing/signing_test.go
+++ b/eth2util/signing/signing_test.go
@@ -67,3 +67,44 @@ func TestVerifyRegistrationReference(t *testing.T) {
 	err = signing.Verify(context.Background(), bmock, signing.DomainApplicationBuilder, 0, sigRoot, eth2p0.BLSSignature(sig), pubkey)
 	require.NoError(t, err)
 }
+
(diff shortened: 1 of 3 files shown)


dB2510 (approved):

eth2util/signing/signing_test.go
```diff
@@ -67,3 +67,44 @@ func TestVerifyRegistrationReference(t *testing.T) {
...
+		}
+
+		domain, err := signing.GetDomain(context.Background(), eth2Cl, signing.DomainApplicationBuilder, 0) // Always use epoch 0 for DomainApplicationBuilder.
+		require.NoError(t, err)
+
+		return domain
+	}
+
+	// Assert genesis domain is used for any fork schedule.
+	expect := eth2p0.Domain{
```
xenowits, line 99: how did we get this domain value?

corverroos: I just ran the test, the point no the value, but that is doesn't change

Issue #1886, opened 2023-03-15
Introduce cluster state data structure


corverroos:
# 🎯 Problem to be solved

Once we have concluded the #1846 design, we need to implement the new cluster state data structure in the cluster package. The first mutations to support should be `legacy_lock` and `add_validator`. 
 
# 🛠️ Proposed solution

Note this should be backwards compatible with legacy lock file. So first add an interface that both Lock and ClusterState implements.

`add_validators` mutation:
 - assume only a DKG executed but all peers with same parameters.
 - so assume the signed mutation signature is similar to the signature_aggregate field, an aggregate of all BLS key share signatures of mutation hash.

PR #2175, opened 2023-05-10
cluster/state: implement legacy lock mutation

corverroos:
Implement the first skeleton and structure of the `LegacyLock` mutation. This is heavy based on previous [PoC](https://github.com/corverroos/clusterstate/blob/main/v7).

category: feature
ticket: #1886


cluster/state/helpers.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/obolnetwork/charon/app/k1util"
+)
+
+// hashRoot hashes a ssz root hasher object.
+func hashRoot(hasher rootHasher) ([32]byte, error) {
+	hw := ssz.DefaultHasherPool.Get()
+	defer ssz.DefaultHasherPool.Put(hw)
+
+	if err := hasher.HashTreeRootWith(hw); err != nil {
+		return [32]byte{}, err
```
gsora, line 19: nit: needs `errors.Wrap()`

gsora (approved):

cluster/state/mutation.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	TypeUnknown       MutationType = ""
+	TypeLegacyLock    MutationType = "dv/legacy_lock/v0.0.1"
+	TypeNodeApproval  MutationType = "dv/node_approval/v0.0.1"
+	TypeNodeApprovals MutationType = "dv/node_approvals/v0.0.1"
+)
+
+var mutationDefs = map[MutationType]struct {
+	DataType      any
```
dB2510, line 32: why this `DataType` is any?

dB2510 (approved):

cluster/state/helpers.go
```diff
@@ -0,0 +1,68 @@
...
+
+	resp, err := hw.HashRoot()
+	if err != nil {
+		return [32]byte{}, errors.Wrap(err, "hash root")
+	}
+
+	return resp, nil
+}
+
+// verifyEmptySig verifies that the signed mutation isn't signed.
```
xenowits, line 30: nit
```suggestion
// verifyEmptySig returns an error if the signed mutation isn't signed.
```

cluster/state/types.go
```diff
@@ -0,0 +1,50 @@
...
+	rootHasher
+	json.Marshaler
+}
+
+// Mutation mutates the cluster state.
+type Mutation struct {
+	// Parent is the hash of the parent mutation.
+	Parent [32]byte `ssz:"Bytes32"`
+	// Type is the type of mutation.
+	Type MutationType `ssz:"ByteList[64]"` // TODO(corver): Make this a numbered enum maybe (instead of a string)?.
```
xenowits, line 33: yeah, numbered enum makes sense. What benefit does having string `MutationType` provides?..

xenowits (approved):

cluster/state/types.go
corverroos: string makes it human readable in the json, otherwise it will be a number in the json.

cluster/state/mutation.go
corverroos: thanks, removed this field, not needed I think.

cluster/state/helpers.go
corverroos: your change flips the logic which isn't correct.

Issue #2151, opened 2023-05-03
Error while combining 1000 distributed keys


tdahar:
# 🐞 Bug Report

### Description

 A clear and concise description of the problem...

While using the combine feature of the tool, there is an error prompted and the keys are never combined. This was done for 1000 distributed keys in 4 nodes.

### Has this worked before in a previous version?

 Yes, the previous version in which this bug was not present was: ....

## 🔬 Minimal Reproduction

Place your distributed keys in a folder following the structure as [here](https://docs.obol.tech/docs/int/quickstart/quickstart-combine): 

## 🔥 Error

<pre><code>

ethereum@Ubuntu-2004-focal-64-minimal-hwe:/home/ethereum/charon_playground/charon_keys/cluster7$ docker run --rm -v "$(pwd):/opt/charon" obolnetwork/charon:v0.15.0 combine --cluster-dir /opt/charon/validators-to-be-combined
14:58:28.425 INFO            Recombining key shares                   {"input_dir": "/opt/charon/validators-to-be-combined", "output_dir": "/opt/charon/validator_keys"}
15:09:06.252 INFO            Recombining key share                    {"validator_number": 346}
Error: generated and lockfile public key for validator DO NOT match
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 20.04.4 LTS
  </code>
</pre>

**What version of Charon are you running? (Which release)**

<pre>
  <code>
obolnetwork/charon:v0.15.0
  </code>
</pre>

**Anything else relevant (validator index / public key)?**

tdahar:
I have also tried the same operation with 100 keys and 10 keys, the tool takes less time to throw the error, but the output is the same.

PR #2178, opened 2023-05-11
cmd: improve error logs

corverroos:
Improve all command fatal error logs:
 - Remove duplicates
 - Use app/log to print errors, so structured fields are included.
 - Only show usage on flag parsing errors.  
 
 Also improve `combine` mismatch error logs.

category: misc
ticket: #2151

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -120,11 +120,6 @@ type TestConfig struct {
 // to the life cycle manager which handles starting and graceful shutdown.
 func Run(ctx context.Context, conf Config) (err error) {
 	ctx = log.WithTopic(ctx, "app-start")
-	defer func() {
-		if err != nil {
-			log.Error(ctx, "Fatal run error", err)
-		}
-	}()
 
 	_, _ = maxprocs.Set()
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -58,10 +58,27 @@ func newRootCmd(cmds ...*cobra.Command) *cobra.Command {
 	}
 
 	root.AddCommand(cmds...)
+	root.SilenceErrors = true // Disable default error printing.
 
 	titledHelp(root)
+	silenceUsage(root)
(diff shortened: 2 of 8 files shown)


app/app.go
```diff
@@ -120,11 +120,6 @@ type TestConfig struct {
 // to the life cycle manager which handles starting and graceful shutdown.
 func Run(ctx context.Context, conf Config) (err error) {
 	ctx = log.WithTopic(ctx, "app-start")
-	defer func() {
```
corverroos, line 123: not needed anymore

cmd/cmd.go
```diff
@@ -137,8 +154,6 @@ func titledHelp(cmd *cobra.Command) {
 
 // printFlags INFO logs all the given flags in alphabetical order.
 func printFlags(ctx context.Context, flags *pflag.FlagSet) {
-	ctx = log.WithTopic(ctx, "cmd")
```
corverroos, line 140: not needed anymore

cmd/combine/combine.go
```diff
@@ -113,7 +113,8 @@ func Combine(ctx context.Context, inputDir, outputDir string, force bool, opts .
 		}
 
 		if valPk != genPubkey {
-			return errors.New("generated and lockfile public key for validator DO NOT match", z.Int("validator_index", idx))
+			return errors.New("unexpected resulting combined validator public key",
+				z.Int("validator_index", idx), z.Hex("actual", genPubkey[:]), z.Hex("expected", valPk[:]))
```
gsora, line 117: I would still keep expected and actual pubkeys, helps with debugging.

gsora (approved):

cmd/combine/combine.go
corverroos: indeed, I added them here, they weren't there previously.

PR #2182, opened 2023-05-11
cluster/state: legacy lock json and transform

corverroos:
Add legacy lock serialisation and transform logic. 

category: feature
ticket: #1886

### DIFF
--- a/cluster/state/cluster.go
+++ b/cluster/state/cluster.go
@@ -4,5 +4,24 @@ package state
 
 // Cluster represents the state of a cluster after applying a sequence of mutations.
 type Cluster struct {
-	// TODO(corver): Implement
+	Name         string
+	Threshold    int
+	DKGAlgorithm string
+	ForkVersion  []byte
+	Operators    []Operator
+	Validators   []Validator
+}
+
+// Operator represents the operator of a node in the cluster.
+type Operator struct {
+	Address string
+	ENR     string
+}
+
+// Validator represents a validator in the cluster.
+type Validator struct {
+	PubKey              []byte
+	PubShares           [][]byte
+	FeeRecipientAddress string
+	WithdrawalAddress   string
 }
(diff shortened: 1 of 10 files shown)


cluster/state/cluster.go
```diff
@@ -4,5 +4,24 @@ package state
 
 // Cluster represents the state of a cluster after applying a sequence of mutations.
 type Cluster struct {
-	// TODO(corver): Implement
+	Name         string
+	Threshold    int
+	DKGAlgorithm string
+	ForkVersion  []byte
```
xenowits, line 10: nit: why not use `Network string`?

corverroos: I'm using the `cluster.Definition` types

cluster/state/helpers.go
```diff
@@ -66,3 +87,40 @@ func verifyK1SignedMutation(signed SignedMutation) error {
 
 	return nil
 }
+
+// ethHex represents a byte slices that is json formatted as 0x prefixed hex.
```
dB2510, line 91:
```suggestion
// ethHex represents a byte slice that is json formatted as 0x prefixed hex.
```

xenowits (approved):

cluster/state/mutationlegacylock.go
```diff
@@ -85,7 +89,36 @@ func transformLegacyLock(_ Cluster, signed SignedMutation) (Cluster, error) {
 		return Cluster{}, errors.Wrap(err, "verify legacy lock")
 	}
 
-	// TODO(corver): Implement legacy lock transform.
+	lock := signed.Mutation.Data.(lockWrapper)
```
dB2510, line 92: nit: check comma ok please, same as line 64-66 in the same file

dB2510 (approved):

cluster/state/mutationlegacylock.go
corverroos: verify function already checks that type is correct, I'll add a comment

PR #2190, opened 2023-05-12
app: refactor to use cluster state

corverroos:
- Replace usage of `cluster.Lock` in app wiring with `state.Cluster`. 
- This means charon can run with a legacy lock file or new mutable cluster state file.
- Align `state.Cluster` API with that of `cluster.Lock`.
- Remove `Hash` in `Mutation` field, as this isn't useful for humans to know.
- Add top level `Cluster.Hash`
- Use lockhash for legacy lock cluster hash.

category: refactor
ticket: #1886


gsora (approved):

app/app.go
```diff
@@ -173,28 +174,28 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+	tcpNode, err := wireP2P(ctx, life, conf, cState, p2pKey, lockHashHex)
 	if err != nil {
 		return err
 	}
 
-	nodeIdx, err := lock.NodeIdx(tcpNode.ID())
+	nodeIdx, err := cState.NodeIdx(tcpNode.ID())
 	if err != nil {
-		return errors.Wrap(err, "private key not matching lock file")
+		return errors.Wrap(err, "private key not matching cState file")
```
xenowits, line 185:
```suggestion
		return errors.Wrap(err, "private key not matching cluster state file")
```

cluster/state/cluster.go
```diff
@@ -12,6 +25,81 @@ type Cluster struct {
...
+	}
+
+	return resp
+}
+
+// FeeRecipientAddresses is a convenience function to return all fee-recipient address from the validators slice.
+func (c Cluster) FeeRecipientAddresses() []string {
+	var resp []string
+	for _, vaddrs := range c.Validators {
+		resp = append(resp, vaddrs.FeeRecipientAddress)
```
xenowits, line 76:
```suggestion
	for _, val := range c.Validators {
		resp = append(resp, val.FeeRecipientAddress)
```

cluster/state/cluster.go
```diff
@@ -12,6 +25,81 @@ type Cluster struct {
...
+
+	return resp, nil
+}
+
+// WithdrawalAddresses is a convenience function to return all withdrawal address from the validators slice.
+func (c Cluster) WithdrawalAddresses() []string {
+	var resp []string
+	for _, vaddrs := range c.Validators {
+		resp = append(resp, vaddrs.WithdrawalAddress)
+	}
```
xenowits, line 67:
```suggestion
	for _, val := range c.Validators {
		resp = append(resp, val.WithdrawalAddress)
	}
```

cluster/state/cluster.go
```diff
@@ -12,6 +25,81 @@ type Cluster struct {
...
+
+	return resp
+}
+
+// FeeRecipientAddresses is a convenience function to return all fee-recipient address from the validators slice.
```
xenowits, line 72:
```suggestion
// FeeRecipientAddresses is a convenience function that return fee-recipient addresses for all the validators in the cluster state.
```

dB2510 (approved):

PR #2191, opened 2023-05-14
core/validatorapi: get all validators

dB2510:
Adds support to get all validators from validatorapi if no validator index or pubkey is provided. This issue came up while testing with prysm VC which needs to query beacon node to get the information of total validators active on beacon chain.
Refer: https://github.com/prysmaticlabs/prysm/blob/5c00fcb84ffef6e85d3d105d7288281ae44b44b2/validator/client/wait_for_activation.go#L139

category: misc
ticket: none

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -1152,10 +1152,6 @@ func getValidatorIDs(query url.Values) []string {
 
 // getValidatorsByID returns the validators with ids being either pubkeys or validator indexes.
 func getValidatorsByID(ctx context.Context, p eth2client.ValidatorsProvider, stateID string, ids ...string) ([]v1Validator, error) {
-	if len(ids) == 0 {
-		return nil, errors.New("no validator ids provided")
-	}
-
 	flatten := func(kvs map[eth2p0.ValidatorIndex]*eth2v1.Validator) []v1Validator {
 		var vals []v1Validator
 		for _, v := range kvs {
@@ -1165,7 +1161,7 @@ func getValidatorsByID(ctx context.Context, p eth2client.ValidatorsProvider, sta
 		return vals
 	}
 
-	if strings.HasPrefix(ids[0], "0x") {
+	if len(ids) > 0 && strings.HasPrefix(ids[0], "0x") {
 		var pubkeys []eth2p0.BLSPubKey
 		for _, id := range ids {
 			coreBytes, err := core.PubKey(id).Bytes()
--- a/core/validatorapi/validatorapi.go
+++ b/core/validatorapi/validatorapi.go
@@ -1000,11 +1000,10 @@ func (c Component) convertValidators(vals map[eth2p0.ValidatorIndex]*eth2v1.Vali
 		pubshare, ok := c.getPubShareFunc(val.Validator.PublicKey)
 		if !ok && !ignoreNotFound {
 			return nil, errors.New("pubshare not found")
-		}
-
-		if ok {
+		} else if ok {
 			val.Validator.PublicKey = pubshare
 		}
+
 		resp[vIdx] = val
 	}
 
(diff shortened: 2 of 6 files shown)


core/validatorapi/router.go
```diff
@@ -1152,10 +1152,6 @@ func getValidatorIDs(query url.Values) []string {
 
 // getValidatorsByID returns the validators with ids being either pubkeys or validator indexes.
 func getValidatorsByID(ctx context.Context, p eth2client.ValidatorsProvider, stateID string, ids ...string) ([]v1Validator, error) {
```
corverroos, line 1154: This will fail in `validatorapi` component in `convertValidators` with "pubshare not found"

corverroos (changes requested):

core/validatorapi/router.go
dB2510: Thanks for the catch 🙏

So we have 2 options I suppose, either only convert the validators that are there in the cluster with the rest of the validators OR just return ALL validators if no indices/pubkeys are specified with no pubkey<>pubshare conversion.

I would prefer the latter one. What do you think?

corverroos: not converting dv public keys will result in the VC not being able to finds it's own validator in the list, resulting in it thinking it is not activated yet. So we have to convert DV public keys to pubshare ALWAYS. 

Suggest somehow detecting that convert should ignore validators if not found in the list. So still erroring by default but allowing only conerting "some" in the case of "querying all validators"

core/validatorapi/validatorapi.go
```diff
@@ -958,7 +958,7 @@ func (c Component) Validators(ctx context.Context, stateID string, validatorIndi
 		return nil, err
 	}
 
-	return c.convertValidators(vals)
+	return c.convertValidators(vals, validatorIndices == nil)
```
corverroos, line 961: never use nil checks for empty slice checks, always use `len(x) == 0`

core/validatorapi/validatorapi.go
```diff
@@ -989,19 +989,22 @@ func (Component) NodeVersion(context.Context) (string, error) {
 	return fmt.Sprintf("obolnetwork/charon/%s-%s/%s-%s", version.Version, commitSHA, runtime.GOARCH, runtime.GOOS), nil
 }
 
-// convertValidators returns the validator map with all root public keys replaced by public shares.
-func (c Component) convertValidators(vals map[eth2p0.ValidatorIndex]*eth2v1.Validator) (map[eth2p0.ValidatorIndex]*eth2v1.Validator, error) {
+// convertValidators returns the validator map with root public keys replaced by public shares for all validators that are part of the cluster.
+func (c Component) convertValidators(vals map[eth2p0.ValidatorIndex]*eth2v1.Validator, allVals bool) (map[eth2p0.ValidatorIndex]*eth2v1.Validator, error) {
```
corverroos, line 993: this function should know about its use-cases, rather, variables should be names according to their meaning in the function. So rather `ignoreNotFound bool`

testutil/beaconmock/options.go
```diff
@@ -149,6 +149,12 @@ func WithValidatorSet(set ValidatorSet) Option {
 				}
 			}
 
+			if len(resp) == 0 {
```
corverroos, line 152: move this to the top, so first check if "querying for all", in that case return all, else pick from set and log errs.

core/validatorapi/validatorapi.go
```diff
@@ -989,19 +989,22 @@ func (Component) NodeVersion(context.Context) (string, error) {
...
 
-		var ok bool
-		val.Validator.PublicKey, ok = c.getPubShareFunc(val.Validator.PublicKey)
-		if !ok {
+		pubshare, ok := c.getPubShareFunc(val.Validator.PublicKey)
+		if !ok && !ignoreNotFound {
 			return nil, errors.New("pubshare not found")
 		}
+
+		if ok {
```
corverroos, line 1005: nit: else if (since it is part of the same logic flow)

corverroos (approved):

PR #2192, opened 2023-05-15
tbls: support deterministic key gen for testing

corverroos:
Adds support for deterministic tbls key generation for use in golden file tests.

category: test
ticket: none

### DIFF
--- a/cluster/test_cluster.go
+++ b/cluster/test_cluster.go
@@ -39,7 +39,7 @@ func NewForT(t *testing.T, dv, k, n, seed int, opts ...func(*Definition)) (Lock,
 	}
 
 	for i := 0; i < dv; i++ {
-		rootSecret, err := tbls.GenerateSecretKey()
+		rootSecret, err := tbls.GenerateInsecureKey(t, random)
 		require.NoError(t, err)
 
 		rootPublic, err := tbls.SecretToPublicKey(rootSecret)
--- a/tbls/herumi.go
+++ b/tbls/herumi.go
@@ -4,10 +4,13 @@ package tbls
 
 import (
 	"fmt"
+	"io"
 	"strconv"
 	"sync"
+	"testing"
 
 	"github.com/herumi/bls-eth-go-binary/bls"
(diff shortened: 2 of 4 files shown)


cluster/test_cluster.go
```diff
@@ -39,7 +39,7 @@ func NewForT(t *testing.T, dv, k, n, seed int, opts ...func(*Definition)) (Lock,
 	}
 
 	for i := 0; i < dv; i++ {
-		rootSecret, err := tbls.GenerateSecretKey()
+		rootSecret, err := tbls.GenerateInsecureKey(t, random)
```
corverroos, line 42: the clusters generated by this function should be deterministic.

tbls/herumi.go
```diff
@@ -36,6 +39,27 @@ func init() {
...
+func (Herumi) GenerateInsecureKey(t *testing.T, random io.Reader) (PrivateKey, error) {
+	t.Helper()
+	for i := 0; i < 100; i++ {
+		b := make([]byte, 32)
+		_, err := random.Read(b)
+		require.NoError(t, err)
+
+		p := new(bls.SecretKey)
+		err = p.Deserialize(b)
+		if err != nil {
```
corverroos, line 53: not all random values are valid, seems around 50% are valid.

gsora: AFAIK `bls.SetRandFunc` can be used to tell the Herumi library to use a specific source of entropy.

Might help with this issue?

tbls/tbls.go
```diff
@@ -25,6 +29,10 @@ type Implementation interface {
 	// GenerateSecretKey generates a secret key and returns its compressed serialized representation.
 	GenerateSecretKey() (PrivateKey, error)
 
+	// GenerateInsecureKey generates a insecure deterministic secret key using the provided random number generator
```
xenowits, line 32: nit:
```suggestion
	// GenerateInsecureKey generates an insecure deterministic secret key using the provided random number generator
```

xenowits (approved):

PR #2199, opened 2023-05-15
cluster/state: implement composite add validators

corverroos:
Implements composite mutation to add validators which wraps `gen_validators` (renamed from `add_validators`) and `node_approvals`.

category: feature
ticket: #1886


cluster/state/mutationaddvalidator.go
```diff
@@ -33,6 +41,85 @@ func (v addValidators) MarshalJSON() ([]byte, error) {
...
+	}, nil
+}
+
+type addValidators struct {
+	GenValidators SignedMutation `ssz:"Composite" json:"gen_validators"`
+	NodeApprovals SignedMutation `ssz:"Composite" json:"node_approvals"`
+}
+
+func (a addValidators) MarshalJSON() ([]byte, error) {
+	b, err := json.Marshal(struct {
```
dB2510, line 109: nit: this struct can be `addValidatorsJSON`

cluster/state/cluster.go
```diff
@@ -136,11 +136,16 @@ func (v Validator) toSSZ() validatorSSZ {
 		pubshares = append(pubshares, sszPubkey{Pubkey: share})
 	}
 
+	// Convert to bytes ignoring errors, assume that verify has been called.
```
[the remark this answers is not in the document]
corverroos: because genssz doesn't support errors at this point, so we cannot return an error here at this point.

cluster/state/mutationaddvalidator.go
corverroos: cool

PR #2206, opened 2023-05-18
eth2util/keystore: finite amount of goroutines in read/write

gsora:
Avoids resource exhaustion due to high amount of validators to be read/written concurrently.

category: bug
ticket: none

### DIFF
--- a/eth2util/keystore/keystore.go
+++ b/eth2util/keystore/keystore.go
@@ -6,6 +6,7 @@
 package keystore
 
 import (
+	"context"
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


gsora:
@corverroos this PR doesn't add a test case for e.g. 1000 validators DKG because it would slow down the entire test suite.

Shall I add a integration test instead?

eth2util/keystore/keystore.go
```diff
@@ -119,51 +139,67 @@ func loadFiles(dir string, sortKeyfiles func([]string) ([]string, error)) ([]tbl
...
 	}
 
-	if err := eg.Wait(); err != nil {
-		return nil, errors.Wrap(err, "write keys")
+	for result := range join() {
+		if result.Err != nil {
+			return nil, errors.Wrap(err, "write keys")
+		}
+
+		// PSA: this is a concurrent array write, and it works because we're not
```
corverroos, line 194: PSA can be removed, this isn't concurrent anymore.

eth2util/keystore/keystore.go
```diff
@@ -55,48 +61,62 @@ func StoreKeys(secrets []tbls.PrivateKey, dir string) error {
 }
 
 func storeKeysInternal(secrets []tbls.PrivateKey, dir string, filenameFmt string, opts ...keystorev4.Option) error {
```
corverroos, line 63: can we add a test to store/load 100 keys?

corverroos (approved):

eth2util/keystore/keystore.go
gsora: Will add one in a separate PR as discussed offline.

PR #2218, opened 2023-05-22
core/consensus: add 2 of 3 test

corverroos:
Add a qbft component test for different cluster sizes and count of offline nodes, including 2 of 3. This shows that a 2of3 cluster works, even though it isnt' BFT.

category: test
ticket: none


gsora (approved):
A microscopic nit, lftm otherwise

Issue #2172, opened 2023-05-10
Pre-generate validator registrations


corverroos:
# 🎯 Problem to be solved
The current builder-api validator registration endpoint isn't compatible with DV middleware, see https://github.com/ObolNetwork/charon/issues/1643 and this [design doc](https://docs.google.com/document/d/1J9Kv2LNZtJk_4rG88UGs5ngtL7d14suSsnlgZ5KPcKA/edit#heading=h.abi7h3595rh9). 
 Changing the spec and getting all VC implementations to implement it is the proper long term solution. 

This is short term workaround is to pre-generate validator registration messages during the DKG, to store them in the lock file, and to broadcast them once an epoch when `--builder-api=true`.

# 🛠️ Proposed solution
- [x] Approved design doc: [link](https://docs.google.com/document/d/1lCYiEv7JmVYymp4CBrEvExj3WNmSqWVvZwl9dwATCmA/edit#)
- [x] Core team consensus on the proposed solution

See design doc.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

boulder225:
https://grafana.monitoring.gcp.obol.tech/d/TcqKfPx4z/engineering-okrs?orgId=1&from=now-90d&to=now

boulder225:
update docs pls @gsora @dB2510

PR #2219, opened 2023-05-22
core/bcast: add pre-generate registrations

dB2510:
Adds support to broadcast pre-generate validator registrations through recaster component.

category: feature
ticket: #2172


app/app.go
```diff
@@ -556,11 +559,37 @@ func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, t
...
-func wireRecaster(sched core.Scheduler, sigAgg core.SigAgg, broadcaster core.Broadcaster) {
+func wireRecaster(ctx context.Context, sched core.Scheduler, sigAgg core.SigAgg, broadcaster core.Broadcaster, validators []state.Validator) error {
 	recaster := bcast.NewRecaster()
+
+	for _, val := range validators {
+		// Check if the current cluster state supports pre-generate validator registrations.
+		if len(val.BuilderRegistration.Signature) == 0 ||
+			len(val.BuilderRegistration.Message.FeeRecipient) == 0 ||
+			len(val.BuilderRegistration.Message.PubKey) == 0 {
+			break
```
corverroos, line 570: what about `continue` rather?

core/bcast/recast.go
```diff
@@ -69,21 +72,62 @@ func (r *Recaster) Store(_ context.Context, duty core.Duty,
...
+) error {
+	if duty.Type != core.DutyBuilderRegistration {
+		return nil
+	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	tuple, ok := r.pregenTuples[pubkey]
+	if ok && tuple.duty.Slot >= duty.Slot {
```
xenowits, line 87: i think this should be
```suggestion
	if ok || tuple.duty.Slot >= duty.Slot {
```
instead ?

app/app.go
```diff
@@ -556,11 +559,37 @@ func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, t
...
+		if err != nil {
+			return errors.Wrap(err, "core pubkey from bytes")
+		}
+
+		signedData, err := core.NewVersionedSignedValidatorRegistration(builderRegistrationToETH2(val.BuilderRegistration))
+		if err != nil {
+			return errors.Wrap(err, "new versioned signed validator registration")
+		}
+
+		if err = recaster.StorePregen(ctx, core.NewBuilderRegistrationDuty(-1), pubkey, signedData); err != nil {
```
corverroos, line 583: lets put this in slot 0 and change the proper registration endpoint to use slot 1. Also hardcode network genesis timestamp in `eth2util` package and use genesis time for registration timestamp.

Can do this in a subsequent PR actually

core/bcast/recast.go
```diff
@@ -97,6 +141,12 @@ func (r *Recaster) SlotTicked(ctx context.Context, slot core.Slot) error {
 				log.Error(ctx, "Rebroadcast duty error (will retry next epoch)", err)
 			}
 		}
+
+		if tuple.pregenerate {
+			pregenerateRegistrationGauge.WithLabelValues(pubkey.String()).Set(1)
+		} else {
+			pregenerateRegistrationGauge.WithLabelValues(pubkey.String()).Set(0)
+		}
```
xenowits, line 149: nit:
```suggestion
		pregenerateRegistrationGauge.WithLabelValues(pubkey.String()).Set(0)
		if tuple.pregenerate {
			pregenerateRegistrationGauge.WithLabelValues(pubkey.String()).Set(1)
                 }
```

core/bcast/recast.go
```diff
@@ -69,21 +72,62 @@ func (r *Recaster) Store(_ context.Context, duty core.Duty,
...
+	// Copy locked things before doing IO.
 	var (
 		clonedTuples = make(map[core.PubKey]recastTuple)
 		clonedSubs   []func(context.Context, core.Duty, core.PubKey, core.SignedData) error
 	)
 
 	r.mu.Lock()
 	clonedSubs = append(clonedSubs, r.subs...)
+
+	// Populate pre-generated registrations first which can be overridden by VC submitted ones.
```
corverroos, line 123: not a fan of all this custom logic for pre-generated registrations. Really do not think recast should even know about this use-case. What about just tracking the latest duty slot we are recasting for each pubkey and then you can query for `slot=0` in grafana to display "pre-generated" registrations.

corverroos (changes requested):

core/bcast/recast.go
dB2510: @corverroos do we need to capture broadcast delay for builder registration duty? we calculate broadcast delay from genesis start time:
```
slotStart := genesis.Add(slotDuration * time.Duration(slot))
delay := time.Since(slotStart)
```

If we don't care about the broadcast delay of this duty then I think we are good. Also IMO broadcast delay for this duty isn't that critical.

app/app.go
```diff
@@ -556,11 +559,37 @@ func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, t
...
+		if err != nil {
+			return errors.Wrap(err, "core pubkey from bytes")
+		}
+
+		signedData, err := core.NewVersionedSignedValidatorRegistration(builderRegistrationToETH2(val.BuilderRegistration))
+		if err != nil {
+			return errors.Wrap(err, "new versioned signed validator registration")
+		}
+
+		if err = recaster.Store(ctx, core.NewBuilderRegistrationDuty(0), pubkey, signedData); err != nil {
```
corverroos, line 583: can we do `slotFromTimestamp` instead of hardcoding 0?

core/bcast/metrics.go
```diff
@@ -26,6 +26,13 @@ var broadcastDelay = promauto.NewHistogramVec(prometheus.HistogramOpts{
 	Buckets:   []float64{.05, .1, .25, .5, 1, 2.5, 5, 10, 20, 30, 60},
 }, []string{"duty"})
 
+var registrationGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
corverroos, line 29: Cardinatlity of "slot" is very high, it will break prometheus. 

Suggest:
`core_bcast_recast_registration_slot`

"The slot of the latest validator registration successfully broadcast by pubkey"

core/bcast/recast.go
corverroos: indeed, delay not critical

app/app.go
```diff
@@ -944,3 +978,38 @@ func hex7(input []byte) string {
...
+				Timestamp:    reg.Message.Timestamp,
+				Pubkey:       eth2p0.BLSPubKey(reg.Message.PubKey),
+			},
+			Signature: eth2p0.BLSSignature(reg.Signature),
+		},
+	}
+}
+
+// slotFromTimestamp returns slot from the provided timestamp.
+func slotFromTimestamp(ctx context.Context, eth2Cl eth2wrap.Client, timestamp time.Time) (int64, error) {
```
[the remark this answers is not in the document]
dB2510: Added 👍

app/app.go
dB2510: Okay so I have added timestamp in eth2util package for each network in this PR. This was needed for simnet tests.

core/bcast/recast.go
```diff
@@ -19,15 +21,17 @@ type recastTuple struct {
...
 }
 
 // Recaster rebroadcasts core.DutyBuilderRegistration aggregate signatures every epoch.
 type Recaster struct {
-	mu     sync.Mutex
-	tuples map[core.PubKey]recastTuple
-	subs   []func(context.Context, core.Duty, core.PubKey, core.SignedData) error
+	mu           sync.Mutex
+	tuples       map[core.PubKey]recastTuple
+	uniqueTuples map[core.PubKey]map[eth2p0.Root]core.SignedData
```
corverroos, line 33: no need for this, can remove, the `tuple.duty.Slot >= duty.Slot` already ensures we store unique tuples.

corverroos (approved):

corverroos (approved):

Issue #2130, opened 2023-04-25
Incorrect attestation inclusion check


dB2510:
# 🐞 Bug Report

### Description

Seems like there is a bug in the way we calculated attestation inclusion in recent refactor of inclusion checker. Some of the attestation that are marked as never included on-chain were actually included in a block. Here are the logs from core team cluster node `outrageous-taxi`:
![image](https://user-images.githubusercontent.com/37813203/234282360-0d1481c1-6c38-4396-9971-8a987098b372.png)

For example, check for first log slot number `5492676`, here's our validator https://goerli.beaconcha.in/validator/0xafd55095b1390d6f075aaac8673bde501d9ec0459cc32b96ba311b4f36d47cf3b2dd846be3833260c92703c26f8c6423#attestations. It says our attestation was included in `5492677`.

## 🔬 Minimal Reproduction

Run charon with tag `96af658`.

PR #2223, opened 2023-05-23
core/tracker: handle duplicate attestations per block

corverroos:
Fix inclusion tracker to handle duplicate attestations per block. That means attestations with identical attestation data, but  different aggregation bits. 

category: bug
ticket: #2130

### DIFF
--- a/core/tracker/inclusion_internal_test.go
+++ b/core/tracker/inclusion_internal_test.go
@@ -14,7 +14,71 @@ import (
 	"github.com/obolnetwork/charon/app/eth2wrap"
 	"github.com/obolnetwork/charon/core"
 	"github.com/obolnetwork/charon/testutil"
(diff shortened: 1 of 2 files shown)


dB2510 (approved):

core/tracker/inclusion_internal_test.go
```diff
@@ -14,8 +14,63 @@ import (
 	"github.com/obolnetwork/charon/app/eth2wrap"
 	"github.com/obolnetwork/charon/core"
 	"github.com/obolnetwork/charon/testutil"
+	"github.com/obolnetwork/charon/testutil/beaconmock"
 )
 
+func TestDuplicateAttDate(t *testing.T) {
```
xenowits, line 20:
```suggestion
func TestDuplicateAttData(t *testing.T) {
```

xenowits (approved):

PR #2230, opened 2023-05-24
p2p: use direct connection for ping

dB2510:
The Ping service was reporting wrong latencies. This is because the ping service creates a long lived stream to the peer using the relay connection. It uses relay connection to report metrics even if direct connection is now available. This is resolved by using new stream/connection for each ping.

category: bug
ticket: none

### DIFF
--- a/p2p/ping.go
+++ b/p2p/ping.go
@@ -86,29 +86,27 @@ func pingPeerOnce(ctx context.Context, svc *ping.PingService, p peer.ID,
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	isDirectConn := isDirectConnAvailable(svc.Host.Network().ConnsToPeer(p))
-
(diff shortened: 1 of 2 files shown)


p2p/ping.go
```diff
@@ -86,24 +86,42 @@ func pingPeerOnce(ctx context.Context, svc *ping.PingService, p peer.ID,
...
 
-	for result := range svc.Ping(ctx, p) {
-		if IsRelayError(result.Error) || errors.Is(result.Error, context.Canceled) {
-			// Just exit if relay error or context cancelled.
+	for {
+		select {
+		case <-ctx.Done():
 			return
-		}
+		case result := <-svc.Ping(ctx, p):
```
corverroos, line 93: add comment: `only ping once to always use "best" connection`

corverroos (approved):

Issue #2097, opened 2023-04-13
Set app/version at build time for tagged releases


corverroos:
# 🎯 Problem to be solved
See the release v2 design doc. We need to set the `app.version` to the value of the git tag (and docker image tag) for official tagged releases.

# 🛠️ Proposed solution
- [x] Approved design doc: [link](https://docs.google.com/document/d/1A_TtUB54rhxz_Nxw3HHLfR6bVvJpUNe2S8AaIlP0rL0/edit#)
- [x] Core team consensus on the proposed solution

We cannot just tag an existing image anymore when publishing a tagged release. Instead we need to build a new image and set the app/version dynamically at build time using:

```
go build -ldflags "-X github.com/obolnetwork/charon/app/version.Version=${GIT_TAG} ..."
```
So we should:
 - Refactor charon Dockerfile to support optional app version overrides using env vars
 - Refactor build-push-deploy action to set the env var accordingly when triggered by a tag event

PR #2232, opened 2023-05-24
github: overwrite version with tagged releases

corverroos:
Sets the app/version.Version variable to the git tag when building tagged docker images. 

category: misc
ticket: #2097

### DIFF
--- a/.github/workflows/build-push-deploy.yml
+++ b/.github/workflows/build-push-deploy.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: docker/setup-buildx-action@v1
     - name: Define docker image meta data tags
       id: meta
-      uses: docker/metadata-action@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/build-push-deploy.yml
```diff
@@ -44,12 +44,16 @@ jobs:
         username: obolnetwork
         password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+    - name: Maybe overwrite app/version.Version with git tag
```
xenowits, line 47: why `Maybe`?

corverroos: this means we do not do this for normal main branch commits. In that case, we just use the default `app/version.Version` value in the code.

dB2510 (approved):

PR #2234, opened 2023-05-24
cluster: correct ssz hash calculation

dB2510:
Corrects SSZ hash calculation for deposit data and builder registrations for v1.7.0. This bug was identified in v1.6.0 cluster lock version.

category: bug
ticket: none


cluster/ssz.go
```diff
@@ -635,19 +637,56 @@ func hashDepositData(d DepositData, hh ssz.HashWalker) error {
 	return putBytesN(hh, d.Signature, sszLenBLSSig)
```
corverroos, line 637: add comment that there is a bug in this code

corverroos (approved):

Issue #2204, opened 2023-05-17
Add node_signatures to lock file


corverroos:
# 🎯 Problem to be solved

A valid cluster lock can be generated by anyone from a valid cluster definition. This is because the operators or nodes do not sign the resulting generated validators, the `signature_aggregate` is only signed by the validators themselves. This has been addressed in the new cluster state, but should be addressed in the legacy lock as well.

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Steps:
- Add a `node_signatures` field to the cluster lock that contains all charon_enr_private_key k1 signatures of the lock hash. 
- Ensure that lock verification checks this 
- Generate and exchange these signatures during the DKG

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

boulder225:
@OisinKyne asks "can we call it `enr_signatures` or `operator_enr_signature` or something maybe? `node` is a new concept not mentioned in the rest of our code to my knowledge" @gsora FYI

PR #2240, opened 2023-05-25
dkg: harden k1 node signatures exchange

gsora:
Dedup signatures, check that the received index is admissible, check that the signature verifies with the expected pubkey.

category: refactor
ticket: #2204

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -31,6 +31,7 @@ import (
 	"github.com/obolnetwork/charon/dkg/sync"
 	"github.com/obolnetwork/charon/eth2util"
 	"github.com/obolnetwork/charon/eth2util/deposit"
+	"github.com/obolnetwork/charon/eth2util/enr"
 	"github.com/obolnetwork/charon/eth2util/keymanager"
 	"github.com/obolnetwork/charon/eth2util/registration"
 	"github.com/obolnetwork/charon/p2p"
@@ -256,8 +257,18 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	var operatorPubkeys []*k1.PublicKey
+	for _, operator := range lock.Operators {
+		parsedEnr, err := enr.Parse(operator.ENR)
+		if err != nil {
+			return errors.Wrap(err, "operator enr parse")
+		}
+
+		operatorPubkeys = append(operatorPubkeys, parsedEnr.PubKey)
+	}
+
 	// Sign, exchange K1 signatures over Lock Hash
-	lock.NodeSignatures, err = nodeSigCaster.exchange(ctx, lock.LockHash, key, nodeIdx)
+	lock.NodeSignatures, err = nodeSigCaster.exchange(ctx, lock.LockHash, key, operatorPubkeys, nodeIdx)
 	if err != nil {
 		return errors.Wrap(err, "k1 lock hash signature exchange")
 	}
--- a/dkg/nodesigs.go
+++ b/dkg/nodesigs.go
@@ -4,6 +4,7 @@ package dkg
 
 import (
 	"context"
+	"encoding/hex"
 
 	k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -37,7 +38,7 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
 	ret := nodeSigBcast{
 		// bcast returns len(peers)-1 messages to each peer, so that senders don't get their own message
 		// hence wait for operatorAmt-1 messages
-		otherSigs:   make(chan *dkgpb.MsgNodeSig, operatorAmt-1),
+		otherSigs:   make(chan *dkgpb.MsgNodeSig),
 		bcastFunc:   bcastComp.Broadcast,
 		operatorAmt: operatorAmt,
 	}
@@ -50,13 +51,17 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
 }
 
 // broadcastCallback is the default bcast.Callback for nodeSigBcast.
-func (n *nodeSigBcast) broadcastCallback(_ context.Context, _ peer.ID, _ string, msg proto.Message) error {
+func (n *nodeSigBcast) broadcastCallback(ctx context.Context, _ peer.ID, _ string, msg proto.Message) error {
 	response, ok := msg.(*dkgpb.MsgNodeSig)
 	if !ok {
 		return errors.New("invalid node sig type")
 	}
 
-	n.otherSigs <- response
+	select {
+	case <-ctx.Done():
+		return errors.Wrap(ctx.Err(), "nodeSigs exchange")
+	case n.otherSigs <- response:
+	}
 
 	return nil
 }
@@ -66,6 +71,7 @@ func (n *nodeSigBcast) exchange(
 	ctx context.Context,
 	lockHash []byte,
 	key *k1.PrivateKey,
+	peerKeys []*k1.PublicKey,
 	nodeIdx cluster.NodeIdx,
 ) ([][]byte, error) {
 	sig, err := k1util.Sign(key, lockHash)
@@ -84,21 +90,46 @@ func (n *nodeSigBcast) exchange(
 		return nil, errors.Wrap(err, "k1 lock hash signature broadcast")
 	}
 
+	dedup := make(map[string]*dkgpb.MsgNodeSig)
+
 	ret := make([][]byte, n.operatorAmt)
 	ret[nodeIdx.PeerIdx] = sig
 
-	// see newLockHashK1Bcast comment
-	for idx := 0; idx < n.operatorAmt-1; idx++ {
+	// see newNodeSigBcast comment
+	for len(dedup) != n.operatorAmt-1 {
 		select {
 		case <-ctx.Done():
-			return nil, errors.New("context canceled")
+			return nil, errors.Wrap(ctx.Err(), "nodeSigs exchange")
 		case otherSig := <-n.otherSigs:
+			dedup[hex.EncodeToString(otherSig.Signature)] = otherSig
 
 			// TODO: remove if appears too verbose down the line
 			log.Debug(ctx, "Received node signature", z.Uint("sender", uint(otherSig.GetPeerIndex())))
-
-			ret[otherSig.GetPeerIndex()] = otherSig.GetSignature()
 		}
+	}
+
+	for _, elem := range dedup {
+		eidx := int(elem.GetPeerIndex())
+		if eidx == nodeIdx.PeerIdx {
+			// ignore, someone tried to send us a signature with our index
+			continue
+		}
+
+		if eidx < 0 || eidx > n.operatorAmt {
+			// ignore, element index is malformed
+			continue
+		}
+
+		sig := elem.GetSignature()
+
+		verified, err := k1util.Verify(peerKeys[eidx], lockHash, sig[:len(sig)-1])
+		if err != nil {
+			return nil, errors.Wrap(err, "dedup signature failure")
+		} else if !verified {
+			return nil, errors.New("signature verification failed on peer lock hash")
+		}
+
+		ret[eidx] = sig
 	}
 
 	return ret, nil
--- a/dkg/nodesigs_internal_test.go
+++ b/dkg/nodesigs_internal_test.go
@@ -1,0 +1,171 @@
+// Copyright © 2022-2023 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package dkg
+
+import (
+	"bytes"
+	"context"
+	"testing"
+	"time"
+
+	k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/obolnetwork/charon/cluster"
+	"github.com/obolnetwork/charon/dkg/bcast"
+	"github.com/obolnetwork/charon/testutil"
+)
+
+func TestSigsExchange(t *testing.T) {
+	n := 32
+
+	var (
+		ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+
+		secrets  []*k1.PrivateKey
+		pubkeys  []*k1.PublicKey
+		tcpNodes []host.Host
+		peers    []peer.ID
+		nsigs    []nodeSigBcast
+		results  [][][]byte
+	)
+
+	defer cancel()
+
+	// Create secretes and libp2p nodes
+	for i := 0; i < n; i++ {
+		secret, err := k1.GeneratePrivateKey()
+		require.NoError(t, err)
+		secrets = append(secrets, secret)
+		pubkeys = append(pubkeys, secret.PubKey())
+
+		tcpNode := testutil.CreateHostWithIdentity(t, testutil.AvailableAddr(t), secret)
+		tcpNodes = append(tcpNodes, tcpNode)
+
+		peers = append(peers, tcpNode.ID())
+	}
+
+	// Connect peers
+	for i := 0; i < n; i++ {
+		for j := 0; j < n; j++ {
+			tcpNodes[i].Peerstore().AddAddrs(tcpNodes[j].ID(), tcpNodes[j].Addrs(), peerstore.PermanentAddrTTL)
+		}
+	}
+
+	for i := 0; i < n; i++ {
+		i := i
+		component := bcast.New(tcpNodes[i], peers, secrets[i])
+		nsigs = append(nsigs, newNodeSigBcast(n, component))
+	}
+
+	results = make([][][]byte, n)
+
+	var eg errgroup.Group
+	for i := 0; i < n; i++ {
+		i := i
+		eg.Go(func() error {
+			res, err := nsigs[i].exchange(
+				ctx,
+				bytes.Repeat([]byte{42}, 32),
+				secrets[i],
+				pubkeys,
+				cluster.NodeIdx{PeerIdx: i},
+			)
+			if err != nil {
+				return err
+			}
+
+			results[i] = res
+
+			return nil
+		})
+	}
+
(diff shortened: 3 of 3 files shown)


dkg/dkg.go
```diff
@@ -256,8 +257,18 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+		if err != nil {
+			return errors.Wrap(err, "operator enr parse")
+		}
+
+		operatorPubkeys = append(operatorPubkeys, parsedEnr.PubKey)
+	}
+
 	// Sign, exchange K1 signatures over Lock Hash
-	lock.NodeSignatures, err = nodeSigCaster.exchange(ctx, lock.LockHash, key, nodeIdx)
+	lock.NodeSignatures, err = nodeSigCaster.exchange(ctx, lock.LockHash, key, operatorPubkeys, nodeIdx)
```
corverroos, line 271: can just pass in `peers p2p.Peer` which has a pubkey method/field

dkg/nodesigs.go
```diff
@@ -50,13 +51,17 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
 // broadcastCallback is the default bcast.Callback for nodeSigBcast.
-func (n *nodeSigBcast) broadcastCallback(_ context.Context, _ peer.ID, _ string, msg proto.Message) error {
+func (n *nodeSigBcast) broadcastCallback(ctx context.Context, _ peer.ID, _ string, msg proto.Message) error {
 	response, ok := msg.(*dkgpb.MsgNodeSig)
 	if !ok {
 		return errors.New("invalid node sig type")
 	}
 
-	n.otherSigs <- response
+	select {
```
corverroos, line 60: lets do verification here, as soon as possible. Since your logic below can be broken if byzantine peers spam invalid data. Also as suggested, store in the map here after verification.

dkg/nodesigs.go
```diff
@@ -84,21 +90,46 @@ func (n *nodeSigBcast) exchange(
...
 
+	dedup := make(map[string]*dkgpb.MsgNodeSig)
+
 	ret := make([][]byte, n.operatorAmt)
 	ret[nodeIdx.PeerIdx] = sig
 
-	// see newLockHashK1Bcast comment
-	for idx := 0; idx < n.operatorAmt-1; idx++ {
+	// see newNodeSigBcast comment
+	for len(dedup) != n.operatorAmt-1 {
```
corverroos, line 99: you are basically using the first `len-1` non-validated msgs, so this is BFT.

corverroos (changes requested):

dkg/nodesigs.go
```diff
@@ -26,20 +28,30 @@ func nodeSigMsgIDs() []string {
 
 // nodeSigBcast handles broadcasting of K1 signatures over the lock hash via the bcast protocol.
 type nodeSigBcast struct {
-	otherSigs   chan *dkgpb.MsgNodeSig
+	otherSigs   sync.Map
```
corverroos, line 31: nit: Add a comment that the key value types are

dkg/nodesigs.go
```diff
@@ -26,20 +28,30 @@ func nodeSigMsgIDs() []string {
...
+	operatorAmt int,
+	peers []p2p.Peer,
+	nodeIdx cluster.NodeIdx,
+	bcastComp *bcast.Component,
+	lockHashFn func() []byte,
+) *nodeSigBcast {
+	ret := &nodeSigBcast{
+		otherSigs:   sync.Map{},
 		bcastFunc:   bcastComp.Broadcast,
 		operatorAmt: operatorAmt,
```
corverroos, line 51: can remove `operatorAmt` and replace with `len(peers)`

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
 	response, ok := msg.(*dkgpb.MsgNodeSig)
 	if !ok {
 		return errors.New("invalid node sig type")
 	}
 
-	n.otherSigs <- response
+	sig := response.GetSignature()
+	msgPeerIdx := int(response.GetPeerIndex())
+
+	if (msgPeerIdx == n.nodeIdx.PeerIdx) || (msgPeerIdx < 0 || msgPeerIdx > n.operatorAmt) {
```
corverroos, line 74: off-by-one `msgPeerIdx >= n.operatorAmt`

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+		return errors.New("wrong peer index", z.Str("peer", peer.String()))
```
corverroos, line 75: nit: we just use `invalid` in most cases. Nit, you can also remove the "peer" since that caller of this function knows what peer it is, so no need to tell it. It actually already logs the peer. So only return fields in errors that are unknown to the caller.

Also, just FYI, we have `p2p.PeerName` when logging peer names

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	}
+
+	peerPubk, err := n.peers[msgPeerIdx].PublicKey()
+	if err != nil {
+		return errors.Wrap(err, "can't get peer public key", z.Str("peer", peer.String()))
```
corverroos, line 80: nit: keep wrapping error messages super concise. It isn't always required to say "this failed while doing X", since it is clear from the context that this is an error that stuff failed while doing X. So you can just do: `errors.Wrap(err, "get peer pubkey")`

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	}
+
+	// wait until lock hash becomes available
+	for {
```
corverroos, line 84: this will spin, using 100% CPU. As suggested, rather use a channel and storing the resulting lockhash in `nodeSigBcast` behind a mutex:

```
func (b nodeSigBcast) getLockHash(ctx context.Context) lockHash, error {
  b.mu.Lock()
  defer b.mu.Lock()
  
  if b.lockHash {
    return b.lockHash
  } 
  
  select {
     case <-ctx.Done: return ctx.Err()
     case b.LockHash <- b.lockHashCh: return b.lockHash
  }
}
```

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+		}
+
+		if done {
+			break
+		}
+	}
+
+	verified, err := k1util.Verify(peerPubk, n.lockHashFn(), sig[:len(sig)-1])
+	if err != nil {
+		return errors.Wrap(err, "dedup signature failure")
```
corverroos, line 100: `errors.Wrap(err, "verify signature")

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	} else if !verified {
+		return errors.New("signature verification failed on peer lock hash")
```
corverroos, line 102: `errors.New("invalid node signature")`

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
 }
 
 // broadcastCallback is the default bcast.Callback for nodeSigBcast.
-func (n *nodeSigBcast) broadcastCallback(_ context.Context, _ peer.ID, _ string, msg proto.Message) error {
+func (n *nodeSigBcast) broadcastCallback(ctx context.Context, peer peer.ID, _ string, msg proto.Message) error {
 	response, ok := msg.(*dkgpb.MsgNodeSig)
```
corverroos, line 66: nit: `nodeSig, ok := msg.(*dkgpb.MsgNodeSig)`

dkg/nodesigs.go
```diff
@@ -50,32 +62,64 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
 	key *k1.PrivateKey,
-	nodeIdx cluster.NodeIdx,
 ) ([][]byte, error) {
-	sig, err := k1util.Sign(key, lockHash)
+	sig, err := k1util.Sign(key, n.lockHashFn())
 	if err != nil {
 		return nil, errors.Wrap(err, "k1 lock hash signature")
 	}
 
 	bcastData := &dkgpb.MsgNodeSig{
```
corverroos, line 120: `localSig` or `nodeSig`

dkg/nodesigs.go
```diff
@@ -85,21 +129,53 @@ func (n *nodeSigBcast) exchange(
...
-	for idx := 0; idx < n.operatorAmt-1; idx++ {
 		select {
 		case <-ctx.Done():
-			return nil, errors.New("context canceled")
-		case otherSig := <-n.otherSigs:
+			return nil, ctx.Err()
+		default:
+			// bcast returns len(peers)-1 messages to each peer, so that senders don't get their own message
+			// hence wait for operatorAmt-1 messages
+			done = mapLen(&n.otherSigs) == n.operatorAmt-1
```
corverroos, line 143: why not store `localSig` in the `n.otherSigs` then it follows the same flow as other sigs and no comments are required?

dkg/nodesigs.go
```diff
@@ -85,21 +129,53 @@ func (n *nodeSigBcast) exchange(
 	}
 
 	ret := make([][]byte, n.operatorAmt)
-	ret[nodeIdx.PeerIdx] = sig
+	ret[n.nodeIdx.PeerIdx] = sig
+
+	for {
```
corverroos, line 134: this will spin using 100% CPU, suggest using a ticker with period 100ms to trigger checking the map.

dkg/nodesigs.go
corverroos: also, what about refactoring this as a fixed size slice, so basically peerIdx as keys? Then we do not need to convert from map the a slice, since it is already correct. We also dedup using peerIdx keys. Only thing is we would need a special function to check is all sigs populated
```
func (b nodeSigBcast) allsigs() bool {
 b.mu.Lock()
 defer unlock
 
 for sig := range b.sigs {
     if sig == nil { return false }
 } 
 
 return true
}
```

dkg/nodesigs.go
```diff
@@ -49,33 +62,94 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	}
+}
+
+// allSigs returns true if all the node signatures have been received.
+func (n *nodeSigBcast) allSigs() bool {
+	n.osLock.Lock()
+	defer n.osLock.Unlock()
+
+	for _, sig := range n.otherSigs {
+		if sig == nil {
```
corverroos, line 89: always do `len(sig) == 0` comparison when check for empty slices, not `== nil`.

dkg/nodesigs.go
```diff
@@ -49,33 +62,94 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	verified, err := k1util.Verify(peerPubk, lockHash, sig[:len(sig)-1])
+	if err != nil {
+		return errors.Wrap(err, "verify signature")
+	} else if !verified {
+		return errors.New("invalid node signature")
+	}
+
+	n.osLock.Lock()
+	n.otherSigs[msgPeerIdx] = sig
+	n.osLock.Unlock()
```
corverroos, line 130: always extract dedicated function when reading/writing to mutexed data:
```
func () setSet(sig, idx) {
  lock
  defer unlock
  n.otherSigs[idx] = sig
}
```

dkg/nodesigs.go
```diff
@@ -84,22 +158,25 @@ func (n *nodeSigBcast) exchange(
 		return nil, errors.Wrap(err, "k1 lock hash signature broadcast")
 	}
 
-	ret := make([][]byte, n.operatorAmt)
-	ret[nodeIdx.PeerIdx] = sig
+	n.otherSigs[n.nodeIdx.PeerIdx] = localSig
```
corverroos, line 161: you are not using mutex here... use suggested `setSig` function.

dkg/nodesigs.go
```diff
@@ -26,20 +28,31 @@ func nodeSigMsgIDs() []string {
 
 // nodeSigBcast handles broadcasting of K1 signatures over the lock hash via the bcast protocol.
 type nodeSigBcast struct {
-	otherSigs   chan *dkgpb.MsgNodeSig
-	bcastFunc   bcast.BroadcastFunc
-	operatorAmt int
+	otherSigs [][]byte
```
corverroos, line 31: nit: this includes local sig now, so name incorrect, suggest: `sigs`

dkg/nodesigs.go
```diff
@@ -84,22 +158,25 @@ func (n *nodeSigBcast) exchange(
...
 		select {
 		case <-ctx.Done():
-			return nil, errors.New("context canceled")
-		case otherSig := <-n.otherSigs:
-
-			// TODO: remove if appears too verbose down the line
-			log.Debug(ctx, "Received node signature", z.Uint("sender", uint(otherSig.GetPeerIndex())))
+			return nil, ctx.Err()
+		case <-tick.C:
+			done = n.allSigs()
```
corverroos, line 172: can inline:
```
if n.allsigs() {
  tick.Stop()
  break
}
```

dkg/nodesigs.go
```diff
@@ -49,33 +62,94 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	lockHash []byte,
 ) ([][]byte, error) {
-	sig, err := k1util.Sign(key, lockHash)
+	localSig, err := k1util.Sign(key, lockHash)
 	if err != nil {
 		return nil, errors.Wrap(err, "k1 lock hash signature")
 	}
 
+	go func() {
+		n.lockHashCh <- lockHash
```
corverroos, line 147: ok, yeah, like that fact that you are passing lockhash here. Wonder if the channel is worth it still? Other option would be simple `n.setLockHash()` that uses the mutex and then a `n.awaitLockHash(ctx)` that just uses a for loop and a ticker to check if the value has been set. Not a fan of using channels AND mutexes, I would prefer to either or.

dkg/nodesigs.go
gsora: yeah that's a leftover from a refactor, originally it was a slice of pointers then moved to array of bytes

dkg/nodesigs.go
gsora: written like this the break will only get out of the `select`, not the `for` as well - i could use `goto` but i'd rather leave it like this

dkg/nodesigs.go
gsora: i see your point, and in one of my original designs `exchange()` set `n.lockHash` directly so that the broadcast callbacks could use it, and another one where the caller had to set the lock hash manually

IMO this solution is fine, yeah we're using a lock + a channel but all in all the latter is only used once: it's doing its thing just once

dkg/nodesigs.go
gsora: See my other comment locking: concurrent writing only happens in one place, it's okay leaving manual locking/unlocking in place.

If it was used in more than one place, then I would've moved to a separate method.

dkg/nodesigs.go
```diff
@@ -26,20 +28,31 @@ func nodeSigMsgIDs() []string {
 
 // nodeSigBcast handles broadcasting of K1 signatures over the lock hash via the bcast protocol.
 type nodeSigBcast struct {
-	otherSigs   chan *dkgpb.MsgNodeSig
-	bcastFunc   bcast.BroadcastFunc
-	operatorAmt int
+	sigs   [][]byte
+	osLock sync.Mutex
```
corverroos, line 32: nit: rename to `sigsLock`

dkg/nodesigs.go
```diff
@@ -49,33 +62,94 @@ func newNodeSigBcast(operatorAmt int, bcastComp *bcast.Component) nodeSigBcast {
...
+	verified, err := k1util.Verify(peerPubk, lockHash, sig[:len(sig)-1])
+	if err != nil {
+		return errors.Wrap(err, "verify signature")
+	} else if !verified {
+		return errors.New("invalid node signature")
+	}
+
+	n.osLock.Lock()
+	n.sigs[msgPeerIdx] = sig
+	n.osLock.Unlock()
```
corverroos, line 130: strongly suggest to stick the pattern of dedicated functions for mutexed data reads/writes: `func (b) setSig`

dkg/nodesigs.go
corverroos: ok cool

dkg/nodesigs.go
corverroos: ok cool

dkg/nodesigs.go
```diff
@@ -84,22 +158,25 @@ func (n *nodeSigBcast) exchange(
 		return nil, errors.Wrap(err, "k1 lock hash signature broadcast")
 	}
 
-	ret := make([][]byte, n.operatorAmt)
-	ret[nodeIdx.PeerIdx] = sig
+	n.sigs[n.nodeIdx.PeerIdx] = localSig
```
corverroos, line 161: The micro-optimisation to not lock sigs when writing the localsig doesn't feel good. Sure it might be technically correct. But it requires thinking. It requires understanding and remembering that setting the other sigs cannot clash with this sig. This makes the code hard to read and increases cognitive load. Taking the performance hit of always locking and always using `setSig` mitigates that at insignificant cost.

dkg/nodesigs.go
```diff
@@ -84,22 +170,27 @@ func (n *nodeSigBcast) exchange(
 		return nil, errors.Wrap(err, "k1 lock hash signature broadcast")
 	}
 
-	ret := make([][]byte, n.operatorAmt)
-	ret[nodeIdx.PeerIdx] = sig
+	n.setSig(localSig, n.nodeIdx.PeerIdx)
+
+	tick := time.NewTicker(100 * time.Millisecond)
```
corverroos, line 175: nit: can defer `tick.Stop()` here and then:

```
	case <-tick.C:
		sigs, ok = n.allSigs() // Note `ok` makes a bit more sense than `done` as a result from `allSigs`
		if ok {
			return sigs, nil
		}
```

corverroos (approved):

PR #2241, opened 2023-05-26
testutil/integration: add test for pre-registrations

dB2510:
Adds a simnet test for pre-generate builder registrations.

category: test
ticket: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -107,6 +107,8 @@ type TestConfig struct {
 	SimnetBMockOpts []beaconmock.Option
 	// BroadcastCallback is called when a duty is completed and sent to the broadcast component.
 	BroadcastCallback func(context.Context, core.Duty, core.PubKey, core.SignedData) error
+	// RecastCallback is called when a duty is completed and sent to the recaster component.
+	RecastCallback func(context.Context, core.Duty, core.PubKey, core.SignedData) error
 	// BuilderRegistration provides a channel for tests to trigger builder registration by the validator mock.
 	BuilderRegistration <-chan *eth2api.VersionedValidatorRegistration
 	// PrioritiseCallback is called with priority protocol results.
@@ -465,7 +467,7 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 		return err
 	}
 
-	if err = wireRecaster(ctx, eth2Cl, sched, sigAgg, broadcaster, cState.Validators, conf.BuilderAPI); err != nil {
+	if err = wireRecaster(ctx, eth2Cl, sched, sigAgg, broadcaster, cState.Validators, conf.BuilderAPI, conf.TestConfig.RecastCallback); err != nil {
 		return errors.Wrap(err, "wire recaster")
 	}
 
@@ -561,12 +563,19 @@ func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, t
 
 // wireRecaster wires the rebroadcaster component to scheduler, sigAgg and broadcaster.
 // This is not done in core.Wire since recaster isn't really part of the official core workflow (yet).
-func wireRecaster(ctx context.Context, eth2Cl eth2wrap.Client, sched core.Scheduler, sigAgg core.SigAgg, broadcaster core.Broadcaster, validators []state.Validator, builderAPI bool) error {
+func wireRecaster(ctx context.Context, eth2Cl eth2wrap.Client, sched core.Scheduler, sigAgg core.SigAgg,
+	broadcaster core.Broadcaster, validators []state.Validator, builderAPI bool,
+	callback func(context.Context, core.Duty, core.PubKey, core.SignedData) error,
+) error {
 	recaster := bcast.NewRecaster()
 
 	sched.SubscribeSlots(recaster.SlotTicked)
 	sigAgg.Subscribe(recaster.Store)
 	recaster.Subscribe(broadcaster.Broadcast)
+
+	if callback != nil {
+		recaster.Subscribe(callback)
+	}
 
 	if !builderAPI {
 		return nil
--- a/testutil/integration/simnet_test.go
+++ b/testutil/integration/simnet_test.go
@@ -37,54 +37,67 @@ import (
 	"github.com/obolnetwork/charon/testutil/beaconmock"
 )
 
+// vcType defines a validator client type.
+type vcType string
+
+// TODO(dhruv): Add support of more VC types to simnet tests.
+const (
+	vcMock vcType = "mock"
+	vcTeku vcType = "teku"
+)
+
 //go:generate go test . -integration -v -run=TestSimnetDuties
 
 func TestSimnetDuties(t *testing.T) {
 	skipIfDisabled(t)
 
 	tests := []struct {
-		name                string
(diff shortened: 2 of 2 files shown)


testutil/integration/simnet_test.go
```diff
@@ -37,87 +37,109 @@ import (
...
-		scheduledType       core.DutyType
-		duties              []core.DutyType
-		builderAPI          bool
-		builderRegistration bool
-		exit                bool
-		teku                bool
+		name                    string
+		scheduledType           core.DutyType
+		duties                  []core.DutyType
+		vcTypes                 []vcType
```
corverroos, line 58: why slice, do we test with multiple types?

app/app.go
```diff
@@ -107,6 +107,8 @@ type TestConfig struct {
 	SimnetBMockOpts []beaconmock.Option
 	// BroadcastCallback is called when a duty is completed and sent to the broadcast component.
 	BroadcastCallback func(context.Context, core.Duty, core.PubKey, core.SignedData) error
+	// RecastCallback is called when a duty is sent to the broadcast component from recaster.
```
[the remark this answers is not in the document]
corverroos: #lessismore

testutil/integration/simnet_test.go
dB2510: not as of now, but i'd like to test multiple types in simnet test. 
An argument against testing multiple types in simnet is we already do some multiple types testing in compose smoke tests so these tests would be more or less same other than not involving docker compose files. What do you think?

app/app.go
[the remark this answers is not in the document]
dB2510: Done!

testutil/integration/simnet_test.go
corverroos: not sure this adds a lot of value here, I'd rather keep it simple for now, single VC. We can do matrix E2E testing.

app/app.go
```diff
@@ -561,13 +561,20 @@ func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, t
...
+	callback func(context.Context, core.Duty, core.PubKey, core.SignedData) error,
+) error {
 	recaster := bcast.NewRecaster()
 
 	sched.SubscribeSlots(recaster.SlotTicked)
 	sigAgg.Subscribe(recaster.Store)
 	recaster.Subscribe(broadcaster.Broadcast)
 
+	if callback != nil {
+		recaster.Subscribe(callback)
```
corverroos, line 575: I was more thinking that you can use the existing wired BroadcastCallback being called by braodcaster. Why do you need a callback from recast itself if it already called broadcast and we can use that callback?

dB2510: How can we use existing wired BroadcastCallback? It is called by sigagg component as broadcaster component has no subscribers:
https://github.com/ObolNetwork/charon/blob/23797a010dc3a98f67f953696df1176f5f186793/app/app.go#L494-L496

That's why I utilized the same callback function as a subscriber for the recaster component. Since the pre-generated registrations are already aggregated, they won't be routed through the sigagg component.

testutil/integration/simnet_test.go
dB2510: cool, I'll revert this back to single VC.

testutil/integration/simnet_test.go
```diff
@@ -155,25 +164,30 @@ func TestSimnetDuties(t *testing.T) {
 				args.BMockOpts = append(args.BMockOpts, beaconmock.WithDeterministicSyncCommDuties(2, 2))
 			}
 
+			if test.pregenerateRegistration {
```
corverroos, line 167: so `test.pregenerateRegistration` basically disables vmock, rather do that directly. Suggest configuring specific VC type per test and in this case, no VC type at all.

testutil/integration/simnet_test.go
```diff
@@ -492,7 +506,7 @@ func startTeku(t *testing.T, args simnetArgs, node int) simnetArgs {
 				args.ErrChan <- errors.New("beaconmock wasn't available after 10s")
 				return
 			default:
-				_, err := http.Get("http://" + args.VAPIAddrs[node])
+				_, err := http.Get("http://" + args.VAPIAddrs[node] + "/up")
```
corverroos, line 509: why add this?

app/app.go
corverroos: ah ok, that is fine then

testutil/integration/simnet_test.go
dB2510: Without this, when it calls bmock to check if it up or not we get the following WARN logs from bmock:
Screenshot 2023-05-29 at 12 21 23 PM

bmock support `/up` to check if bmock is up or not:
https://github.com/ObolNetwork/charon/blob/494f33fdf89f52618bb212c92e351a847f1ca6de/testutil/beaconmock/server.go#L62-L64

testutil/integration/simnet_test.go
dB2510: Done!

testutil/integration/simnet_test.go
```diff
@@ -49,75 +49,86 @@ func TestSimnetDuties(t *testing.T) {
...
 			name:          "voluntary exit with teku",
 			scheduledType: 0,
 			duties:        []core.DutyType{core.DutyExit},
 			exit:          true,
-			teku:          true,
+			vcType:        "teku",
+		},
+		{
+			name:          "pre-generate registrations",
+			scheduledType: 0,
```
corverroos, line 129: can omit zero values

testutil/integration/simnet_test.go
```diff
@@ -162,18 +177,19 @@ func TestSimnetDuties(t *testing.T) {
...
+	P2PKeys                  []*k1.PrivateKey
+	SimnetKeys               []tbls.PrivateKey
+	BMockOpts                []beaconmock.Option
+	Lock                     cluster.Lock
+	ErrChan                  chan error
+	BuilderAPI               bool
+	BuilderRegistration      bool
+	SyntheticProposals       bool
+	VoluntaryExit            bool
+	PregenerateRegistrations bool
```
corverroos, line 192: not used anymore?

testutil/integration/simnet_test.go
```diff
@@ -188,18 +204,14 @@ func newSimnetArgs(t *testing.T) simnetArgs {
...
+	var vapiAddrs []string
 	for i := 0; i < n; i++ {
-		vmocks = append(vmocks, true)
 		vapiAddrs = append(vapiAddrs, testutil.AvailableAddr(t).String())
 	}
 
 	return simnetArgs{
 		N:          n,
-		VMocks:     vmocks,
+		VMocks:     make([]bool, n),
```
corverroos, line 214: can remove this `VMocks` field if constant

testutil/integration/simnet_test.go
```diff
@@ -49,75 +49,86 @@ func TestSimnetDuties(t *testing.T) {
 		builderAPI          bool
 		builderRegistration bool
 		exit                bool
-		teku                bool
+		vcType              string
```
corverroos, line 52: suggest using a int enum with 0 as unknown

corverroos (approved):

PR #2249, opened 2023-05-29
core/consensus: propose returns instance errors

corverroos:
Consensus has two "start" methods, `Participate` and `Propose` either of which can run the instance (depending on timer type). This causes problems with tracker since it doens't know which method to track and the fact that `Participate` is called independent of fetcher, means it breaks tracker step ordering logic.

Refactor consensus to always return the instance error from `Propose`, regardless of whether it was started from `Participate` or `Propose`. `Participate` doesn't need to be instrumented, and the order of events in tracker remain consistent.

category: bug
ticket: none

### DIFF
--- a/core/consensus/component.go
+++ b/core/consensus/component.go
@@ -159,6 +159,7 @@ func New(tcpNode host.Host, sender *p2p.Sender, peers []p2p.Peer, p2pKey *k1.Pri
 	c.mutable.recvBuffers = make(map[core.Duty]chan msg)
 	c.mutable.inputValues = make(map[core.Duty]chan proto.Message)
 	c.mutable.inputHashes = make(map[core.Duty]chan [32]byte)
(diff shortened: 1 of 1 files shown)


core/consensus/component.go
```diff
@@ -299,7 +304,7 @@ func (c *Component) Participate(ctx context.Context, duty core.Duty) error {
 		return nil // Not an eager start timer, wait for Propose to start.
 	}
 
-	if _, _, running := c.getInputChannels(duty); running {
+	if _, _, _, running := c.getInstanceChans(duty); running {
 		return nil // Instance already running.
```
dB2510, line 308: why no error return here by reading from errCh? like above

dB2510 (approved):

xenowits (approved):

PR #2257, opened 2023-05-31
dkg: sync privkeylock before exiting

gsora:
Block `dkg` exit until `privkeylock.Run()` enclosing goroutine exits, ensuring that the private key lock file is always deleted.

`Run()` is supposed to be executed in a goroutine, and since we can't control the scheduler it might be scheduled *after* the `main` one exits, leading to the `ctx.Done()` code path to sometimes never be executed.

category: bug
ticket: #2258

### DIFF
--- a/app/privkeylock/privkeylock.go
+++ b/app/privkeylock/privkeylock.go
@@ -67,23 +67,23 @@ type Service struct {
 }
 
 // Run runs the service, updating the lock file every second and deleting it on context cancellation.
-func (h *Service) Run(ctx context.Context) error {
-	tick := time.NewTicker(h.updatePeriod)
+func (s *Service) Run(ctx context.Context) error {
+	tick := time.NewTicker(s.updatePeriod)
 	defer tick.Stop()
 
-	defer h.wg.Done()
+	defer s.wg.Done()
 
 	for {
 		select {
 		case <-ctx.Done():
-			if err := os.Remove(h.path); err != nil {
+			if err := os.Remove(s.path); err != nil {
 				return errors.Wrap(err, "deleting private key lock file failed")
 			}
 
 			return nil
 		case <-tick.C:
 			// Overwrite lockfile with new metadata
-			if err := writeFile(h.path, h.command, time.Now()); err != nil {
+			if err := writeFile(s.path, s.command, time.Now()); err != nil {
 				return err
 			}
 		}
@@ -91,8 +91,8 @@ func (h *Service) Run(ctx context.Context) error {
 }
 
 // Done waits until Service has finished deleting the private key lock file.
-func (h *Service) Done() {
-	h.wg.Wait()
+func (s *Service) Done() {
+	s.wg.Wait()
 }
 
 // metadata is the metadata stored in the lock file.
--- a/app/privkeylock/privkeylock.go
+++ b/app/privkeylock/privkeylock.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/obolnetwork/charon/app/errors"
@@ -21,20 +22,20 @@ var (
 )
 
 // New returns new private key locking service. It errors if a recently-updated private key lock file exists.
-func New(path, command string) (Service, error) {
+func New(path, command string) (*Service, error) {
 	content, err := os.ReadFile(path)
 	if errors.Is(err, os.ErrNotExist) { //nolint:revive // Empty block is fine.
 		// No file, we will create it in run
 	} else if err != nil {
-		return Service{}, errors.Wrap(err, "cannot read private key lock file", z.Str("path", path))
+		return nil, errors.Wrap(err, "cannot read private key lock file", z.Str("path", path))
 	} else {
 		var meta metadata
 		if err := json.Unmarshal(content, &meta); err != nil {
-			return Service{}, errors.Wrap(err, "cannot decode private key lock file content", z.Str("path", path))
+			return nil, errors.Wrap(err, "cannot decode private key lock file content", z.Str("path", path))
 		}
 
 		if time.Since(meta.Timestamp) <= staleDuration {
-			return Service{}, errors.New(
+			return nil, errors.New(
 				"existing private key lock file found, another charon instance may be running on your machine",
 				z.Str("path", path),
 				z.Str("command", meta.Command),
@@ -43,14 +44,18 @@ func New(path, command string) (Service, error) {
 	}
 
 	if err := writeFile(path, command, time.Now()); err != nil {
-		return Service{}, err
+		return nil, err
 	}
 
-	return Service{
+	s := &Service{
 		command:      command,
 		path:         path,
 		updatePeriod: updatePeriod,
-	}, nil
+	}
+
+	s.wg.Add(1)
+
+	return s, nil
 }
 
 // Service is a private key locking service.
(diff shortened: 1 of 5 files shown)


dkg/dkg.go
```diff
@@ -313,6 +313,10 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 	log.Info(ctx, "Successfully completed DKG ceremony 🎉")
 
+	// explicitly cancel the context and wait until the privkey lock is deleted
+	cancel()
```
dB2510, line 317: what if we call ctrl+C before getting to this line? This `cancel()` and `lockSvc.Done()` will not be called in that case. But defer cancel() at the start of this function will be called.

gsora: even better, we can just defer `lockSvc.Done()` once since it'll be executed right before the exit, but after the `cancel()` is called at the end of the file

app/privkeylock/privkeylock.go
```diff
@@ -82,6 +90,11 @@ func (h Service) Run(ctx context.Context) error {
 	}
 }
 
+// Done waits until Service has finished deleting the private key lock file.
+func (h *Service) Done() {
```
xenowits, line 94: nit: any reason why the receiver is named `h` instead of `s`?

gsora: no! thanks for pointing that out

dkg/dkg.go
```diff
@@ -77,15 +77,23 @@ func (c Config) HasTestConfig() bool {
...
 	ctx = log.WithTopic(ctx, "dkg")
 
 	lockSvc, err := privkeylock.New(p2p.KeyPath(conf.DataDir)+".lock", "charon dkg")
 	if err != nil {
+		// cancel manually here because we'll defer cancel() and lockSvc.Done() later
+		cancel()
 		return err
 	}
 
+	// make sure to always wait for lockSvc to be done.
```
xenowits, line 90:
```suggestion
	// Make sure to always wait for lockSvc to be done.
```

dkg/dkg.go
```diff
@@ -77,15 +77,23 @@ func (c Config) HasTestConfig() bool {
...
+	defer func() {
+		// explicitly cancel the context and wait until the privkey lock is deleted
```
xenowits, line 92:
```suggestion
		// Explicitly cancel the context and wait until the privkey lock is deleted.
```

xenowits:
@gsora hear me out on this:
* Instead of using a `sync.WaitGroup`, let's just add a `quit channel`:
```go
// Service is a private key locking service.
type Service struct {
         ...
	quit            chan struct{}   // New
}
```
* In the `Stop()` method, we simply close the `quit chan`:
```go
func (s *Service) Done() {
	close(s.quit)
}
```
* We add another case in the select block:
```go
case <-ctx.Done():
	cleanup = true
case <-h.quit:
	cleanup = true
```
* Break out of the loop if `cleanup == true`
* Finally, do the delete file stuff

It essentially does the same thing as the waitgroup but the `quit chan` pattern is very common in the charon codebase (see scheduler, aggsigdb etc). The DKG process also doesn't have to wait for privkeylock deletion to complete, since it will eventually happen.

Wdyt?

gsora:
> The DKG process also doesn't have to wait for privkeylock deletion to complete, since it will eventually happen.

That is not correct: since the deletion happens in a different goroutine you are not sure if it'll be scheduled.

I just implemented your solution as a test, and while it does work more often than just relying on `defer cancel()`, it sometimes doesn't work.

The idea here is having the `main` goroutine **waiting until** the child one has actually deleted the file.

dkg/dkg.go
```diff
@@ -77,16 +77,30 @@ func (c Config) HasTestConfig() bool {
...
+	defer func() {
+		// Explicitly cancel the context and wait until the privkey lock is deleted.
+		cancel()
+		<-lockSvcDone
+	}()
+
 	go func(ctx context.Context) {
+		defer func() {
+			lockSvcDone <- struct{}{}
+		}()
```
dB2510, line 102: can just do:
```suggestion
		defer close(lockSvcDone)
```

gsora: It's pretty much the same thing, `struct{}` has zero size in memory.

dkg/dkg.go
```diff
@@ -77,16 +77,30 @@ func (c Config) HasTestConfig() bool {
 //nolint:maintidx // Refactor into smaller steps.
 func Run(ctx context.Context, conf Config) (err error) {
 	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
 
 	ctx = log.WithTopic(ctx, "dkg")
 
 	lockSvc, err := privkeylock.New(p2p.KeyPath(conf.DataDir)+".lock", "charon dkg")
 	if err != nil {
+		// cancel manually here because we'll defer cancel() and lockSvc.Done() later
```
xenowits, line 85:
```suggestion
		// Cancel manually here because we'll defer cancel() and lockSvc.Done() later.
```

dkg/dkg.go
```diff
@@ -77,16 +77,30 @@ func (c Config) HasTestConfig() bool {
...
+		cancel()
 		return err
 	}
 
+	// Make sure to always wait for lockSvc to be done.
+	lockSvcDone := make(chan struct{})
+
```
xenowits, line 92: nit: can remove newline

xenowits (approved):
LGTM! I think this is a more simpler approach 👍

dkg/dkg.go
```diff
@@ -77,16 +77,30 @@ func (c Config) HasTestConfig() bool {
...
+	defer func() {
+		// Explicitly cancel the context and wait until the privkey lock is deleted.
+		cancel()
+		<-lockSvcDone
+	}()
+
 	go func(ctx context.Context) {
+		defer func() {
```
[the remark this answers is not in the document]
xenowits: yeah, true, defer makes sense here!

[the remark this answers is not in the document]
gsora: admittedly there's no error return to be handled here, but having defer there is close to zero cost, and will help not having bugs in the future if we'll change what happens there, and add early returns

dkg/dkg.go
```diff
@@ -81,16 +81,23 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+
+		// Start it async
+		go func() {
+			if err := lockSvc.Run(); err != nil {
+				log.Error(ctx, "Error locking private key file", err)
+			}
+		}()
+
+		// Stop it on exit.
+		defer lockSvc.Close()
```
dB2510, line 99: nit: move this line before go func, after we initialise lockSvc

dkg/dkg.go
```diff
@@ -81,16 +81,23 @@ func Run(ctx context.Context, conf Config) (err error) {
...
 
-	lockSvc, err := privkeylock.New(p2p.KeyPath(conf.DataDir)+".lock", "charon dkg")
-	if err != nil {
-		return err
-	}
-
-	go func(ctx context.Context) {
-		if err := lockSvc.Run(ctx); err != nil {
-			log.Error(ctx, "Error locking private key file", err)
+	{
```
dB2510, line 84: @corverroos why do we need a separate block for this?

dkg/dkg.go
```diff
@@ -81,16 +81,23 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+		// Setup private key locking.
+		lockSvc, err := privkeylock.New(p2p.KeyPath(conf.DataDir)+".lock", "charon dkg")
+		if err != nil {
+			return err
 		}
-	}(ctx)
+
+		// Start it async
```
dB2510, line 91:
```suggestion
		// Start private key lock service async.
```

dB2510 (approved):

Issue #2259, opened 2023-05-31
Stream scope ping errors 


dB2510:
# 🐞 Bug Report

### Description

We are seeing ping errors in our core team cluster as `stream scope not attached to a protocol`. Because of these errors we are seeing ping connections and disconnections frequently.

## 🔥 Error

<pre><code>
2023-05-31 16:14:25.683 | INFO ping       Peer connected                           {"peer": "outrageous-taxi", "rtt": "409.06674ms"}
2023-05-31 16:13:58.606 | WARN ping       Peer ping failing                        {"peer": "outrageous-taxi", "error": "stream scope not attached to a protocol"}
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 22.04 LTS
  </code>
</pre>

**What version of Charon are you running? (Which release/git commit)**

<pre>
  <code>
0229636
  </code>
</pre>

**Anything else relevant (validator index / public key)?**
Cluster name: `Obol Core Team 3 of 4`

dB2510:
Closed by #2260

PR #2260, opened 2023-05-31
p2p: close existing ping streams

dB2510:
Close existing streams after each ping. This solves the errors from ping service `stream scope not attached to a protocol` which was caused by creating new streams for each ping and NOT closing them.

Note: We need to test this on our core team cluster.

category: bug
ticket: #2259

### DIFF
--- a/p2p/ping.go
+++ b/p2p/ping.go
@@ -80,26 +80,24 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID, callback fu
 }
 
 // pingPeerOnce starts a long lived ping connection with the peer and returns on first error.
-func pingPeerOnce(parentCtx context.Context, svc *ping.PingService, p peer.ID,
+func pingPeerOnce(ctx context.Context, svc *ping.PingService, p peer.ID,
 	logFunc func(context.Context, ping.Result), callback func(peer.ID, host.Host),
 ) {
-	ctx, cancel := context.WithCancel(parentCtx)
-	defer cancel()
+	ticker := time.NewTicker(time.Second * 30)
+	defer ticker.Stop()
 
-	tick := time.NewTicker(time.Minute * 10)
-	defer tick.Stop()
-
-	pingChan := svc.Ping(ctx, p)
+	pingChan, closeStream := getPingChan(ctx, svc, p)
+	defer closeStream()
 	for {
 		select {
-		case <-parentCtx.Done():
+		case <-ctx.Done():
 			return
-		case <-tick.C:
+		case <-ticker.C:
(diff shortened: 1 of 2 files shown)


p2p/ping.go
```diff
@@ -83,14 +83,22 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID, callback fu
...
+	ticker := time.NewTicker(time.Second * 30)
+	defer ticker.Stop()
 
+	pingChan, closeStream := getPingChan(ctx, svc, p)
+	defer closeStream()
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case result := <-svc.Ping(ctx, p): // Only ping once to always use "best" connection.
```
dB2510, line 93: This creates new streams for each leaving existing streams open.

p2p/ping.go
```diff
@@ -83,14 +83,22 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID, callback fu
 func pingPeerOnce(ctx context.Context, svc *ping.PingService, p peer.ID,
 	logFunc func(context.Context, ping.Result), callback func(peer.ID, host.Host),
 ) {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	ticker := time.NewTicker(time.Second * 30)
+	defer ticker.Stop()
 
+	pingChan, closeStream := getPingChan(ctx, svc, p)
+	defer closeStream()
```
xenowits, line 90: nit: can try to use the `onStartup` pattern here
```go
        onStartup := make(chan struct{}, 1)
        onStartup <- struct{}{}
```

p2p/ping.go
```diff
@@ -83,14 +83,22 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID, callback fu
...
 		select {
 		case <-ctx.Done():
 			return
-		case result := <-svc.Ping(ctx, p): // Only ping once to always use "best" connection.
+		case <-ticker.C:
+			// Signal ping service to close the existing stream to avoid having orphaned streams.
+			closeStream()
+
+			// Create new stream to use the "best" connection.
+			pingChan, closeStream = getPingChan(ctx, svc, p)
```
corverroos, line 100: why not do the simplest solution, create a new stream for each ping?

p2p/ping.go
corverroos: why then not just close the existing streams. Simplest sounds like new stream for each ping, making sure it is closed.

dB2510: Sounds good! Pushed the changes.

xenowits (approved):
Looks good! Let's merge and test it, we can monitor if those logs are gone

Issue #2239, opened 2023-05-25
Nightly tests are failing


corverroos:
# 🐞 Bug Report

### Description

The nightly tests are failing every time.

### Has this worked before in a previous version?

I think so, not sure actually.

## 🔬 Minimal Reproduction

https://github.com/ObolNetwork/charon/actions/runs/5079707417

gsora:
After merging #2253, nightly tests are now passing: https://github.com/ObolNetwork/charon/actions/runs/5121234674

Closing this issue as completed.

PR #2267, opened 2023-06-01
testutil/integration: 700 validators in nightly

gsora:
Nightly tests are sometimes not able to be executed in 1h30m, so we're dropping them to 700.

category: test
ticket: #2239


xenowits (approved):

corverroos:
Cool, are we still aiming for 20min test duration?

gsora:
Yes, but we have to do so by trial and error :-/

Issue #2277, opened 2023-06-05
Connections with relay temporarily lost


boulder225:
Noticed these errors on our core team cluster:
```
WARN ping  Peer ping failing                        
{"peer": "wondrous-computer", "error": "stream-1627777: transient: cannot reserve stream: resource limit exceeded"}
```

We lost p2p connection after these error logs. This happened around 16:19 UTC yesterday (4th june).
We even lost connection with relay.
We connected again this morning without any restart around 7:20 am. I am still not connected with jittery-pond.
We need to debug this asap.

First impression is to check whether libp2p only counts active streams or ALL streams (open+closed).

Also noticing these error logs:
`concurrent active dial through the same relay failed with a protocol error`

boulder225:
Summary by @dB2510  :

> This issue does not exist in v0.15.0.
> We found this issue when we upgraded our core team cluster to v0.16.0-rc1 last friday to monitor over weekend.
> This issue popped up on Sunday midnight or Monday morning. We were getting frequent resource limit exceeded logs. Here are more logs for reference: https://grafana.monitoring.gcp.obol.tech/explore?orgId=1&left=%7B%22datasource%22:%22ALmVcoFVk%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster_name%3D%5C%22Obol%20Core%20Team%203%20of%204%5C%22,%20cluster_hash%3D%5C%22736c122%5C%22,%20cluster_peer%3D%5C%22wondrous-computer%5C%22%7D%20%7C%3D%20%60resource%20limit%20exceeded%60%20%7C%20logfmt%20%7C%20line_format%20%60%7B%7B.pretty%7D%7D%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22ALmVcoFVk%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D
> 
> We added a metric to catch the cause of this issue but no luck till now. Here's the metric: p2p_peer_streams

boulder225:
History of charon-perf- by @sugh01 

> Interesting to note that we do not see this error in charon-perf, rel016test01 and issue98-test02 clusters, all running 16.0-rc5,  but during that timeframe, ObolCore3/4 had an interesting mix of direct connection and relay.  I think we should reproduce that scenario on canary

[...]

PR #2278, opened 2023-06-05
p2p: add stream counter metric

gsora:
Adds a Prometheus metric to track the opened streams by protocol.

category: bug
ticket: #2277

### DIFF
--- a/p2p/metrics.go
+++ b/p2p/metrics.go
@@ -56,6 +56,12 @@ var (
 		Help:      "Current number of libp2p connections by peer and type ('direct' or 'relay'). Note that peers may have multiple connections.",
 	}, []string{"peer", "type"})
 
+	peerStreamGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "p2p",
+		Name:      "peer_streams",
+		Help:      "Current number of libp2p streams by peer and direction ('inbound' or 'outbound' or 'unknown').",
+	}, []string{"peer", "direction"})
+
 	peerConnCounter = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "p2p",
 		Name:      "peer_connection_total",
--- a/p2p/metrics.go
+++ b/p2p/metrics.go
@@ -56,6 +56,7 @@ var (
 		Help:      "Current number of libp2p connections by peer and type ('direct' or 'relay'). Note that peers may have multiple connections.",
 	}, []string{"peer", "type"})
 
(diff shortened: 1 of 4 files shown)


p2p/ping.go
```diff
@@ -79,50 +79,53 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID, callback fu
...
 ) {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
 	// newPingChan creates a new stream and returns the ping result channel to listen for results
 	// and a close function to close the stream.
 	// Note: The only way to close a stream opened by ping service is to cancel the context.
-	newPingChan := func() (<-chan ping.Result, func()) {
-		pingCtx, pingCancel := context.WithCancel(ctx)
+	newPingChan := func(parentCtx context.Context) (<-chan ping.Result, func()) {
```
xenowits, line 89: nit: can simply name `ctx`
```suggestion
	newPingChan := func(ctx context.Context) (<-chan ping.Result, func()) {
```

xenowits (approved):
LGTM!

p2p/ping.go
```diff
@@ -79,50 +79,53 @@ func pingPeer(ctx context.Context, svc *ping.PingService, p peer.ID, callback fu
 	}
 }
 
-// pingPeerOnce starts a long lived ping connection with the peer and returns on first error.
+// pingPeerOnce starts a long-lived ping connection with the peer and returns on first error.
 func pingPeerOnce(ctx context.Context, svc *ping.PingService, p peer.ID,
```
corverroos, line 83: all these closures and callbacks and stuff is very confusing, lets try and simplify into a simple loop. Take a step back, zoom out, think what the problem is, then solve it from scratch, we are building on top of previous stuff, we are digging the hole deeper and deeper

p2p/metrics.go
```diff
@@ -56,6 +56,12 @@ var (
 		Help:      "Current number of libp2p connections by peer and type ('direct' or 'relay'). Note that peers may have multiple connections.",
 	}, []string{"peer", "type"})
 
+	peerStreamGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "p2p",
+		Name:      "peer_streams",
+		Help:      "Current number of libp2p streams by peer and direction ('inbound' or 'outbound' or 'unknown').",
```
corverroos, line 62: help test needs to be updated

corverroos (changes requested):

dB2510 (approved):

Issue #2203, opened 2023-05-17
Switch to SSZ eth2 encoding in protobufs


corverroos:
# 🎯 Problem to be solved

We currently encode eth2 types to json when sending them over the wire as protobufs. JSON is more than 2x the bandwidth of SSZ. We should decrease our bandwidth requirements to improve the performance of large proposal wire payloads. 

Smaller payloads result in less latency in low-throughput clusters, which result in quicker and more robust consensus which result in less missed duties and more rewards.

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

This is a multi step process, affecting `core. UnsignedDataSetToProto` and `core.UnsignedDataSetFromProto`:
1: add support for unmarshalling SSZ encoded values in the proto (in addition to JSON)
2: release v0.16
3: switch to marshalling SSZ encoded protos
4: release v0.17
5: remove support for unmarshalling and marshalling JSON encoded protos.

> ⚠️⚠️⚠️
> This affects backwards compatibility:
> V0.17 is only compatible with v0.16

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2286, opened 2023-06-06
core: enable ssz encoding in protos

corverroos:
Enables SSZ encoding in protos by default. This decreases wire bandwitdh requirements. 

> Note this breaks compatibility with v0.15 and earlier. So v0.17 will only be compatible with v0.16.

category: feature 
ticket: #2203

### DIFF
--- a/core/ssz_test.go
+++ b/core/ssz_test.go
@@ -92,13 +92,13 @@ func TestMarshalUnsignedProto(t *testing.T) {
 
 	seed := time.Now().Unix()
 
-	for _, enabledSSZ := range []bool{true, false} {
+	for _, disableSSZ := range []bool{true, false} {
 		for i, test := range tests {
 			f := testutil.NewEth2Fuzzer(t, seed+int64(i)) // Use the same seed for ssz vs json for each type.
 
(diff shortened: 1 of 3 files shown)


core/ssz_test.go
```diff
@@ -202,7 +202,7 @@ func TestMarshalParSignedProto(t *testing.T) {
 
 				b, err := proto.Marshal(pb)
 				require.NoError(t, err)
-				if enabledSSZ {
+				if disabledSSZ {
 					sszSizes[fmt.Sprintf("%T", signedPtr)] = len(b)
 				} else {
 					jsonSizes[fmt.Sprintf("%T", signedPtr)] = len(b)
```
dB2510, line 208: Shouldn't this be?:
```suggestion
				if disabledSSZ {
					jsonSizes[fmt.Sprintf("%T", signedPtr)] = len(b)
				} else {
					sszSizes[fmt.Sprintf("%T", signedPtr)] = len(b)
```

dB2510 (approved):

core/ssz_test.go
corverroos: cheers thanks fixed

PR #2289, opened 2023-06-08
app/promauto: implement reset gauge

corverroos:
Implement a ResetGauge to simplify "resetting of previously set labels".

category: misc
ticket: none


gsora (approved):

gsora:
@corverroos CI is red because the metrics docs hasn't been updated, approved regardless because it looks fine otherwise

PR #2290, opened 2023-06-08
app: disable resource manager

dB2510:
Disable resource manager in charon to avoid `resource limit exceeded` error from libp2p.

category: misc
ticket: #2277

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -21,6 +21,7 @@ import (
 	k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/network"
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -286,7 +287,10 @@ func wireP2P(ctx context.Context, life *lifecycle.Manager, conf Config,
 	}
 
 	// Start libp2p TCP node.
-	opts := []libp2p.Option{p2p.WithBandwidthReporter(peerIDs)}
+	opts := []libp2p.Option{
+		p2p.WithBandwidthReporter(peerIDs),
+		libp2p.ResourceManager(&network.NullResourceManager{}),
```
corverroos, line 292: nit: `new(network.NullResourceManager)`

corverroos (approved):

app/app.go
corverroos, line 292:
```suggestion
		libp2p.ResourceManager(new(network.NullResourceManager)),
```

Issue #2200, opened 2023-05-16
Add inclusion results to tracker for accurate success/failure metrics


corverroos:
# 🎯 Problem to be solved

Tracker currently measures broadcasts as duty success. But this excludes the inclusion result which is the actual objective duty result. This is generating conflicting data, duty succeed when they didn't, or fail when then did (other node's broadcasted).

 For accurate data, we need to incorporate inclusion results in tracker calculations.
 
# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

 - Delay tracker deadliners such that inclusion results are included.
 - Push inclusion results to tracker
 - Add inclusion results as last step in tracker.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2299, opened 2023-06-12
core/tracker: track inclusion events

corverroos:
Extends tracker duty failure reasons to include "chain inclusion" results. Duties will subsequently only be marked as success if included onchain, and accordinlgy marked as failed if not included onchain.

Note that chain inclusion is only checked if the duty was broadcasted, so this doesn't actually "make duty failures objective", since if the local node failed to broadcast, it will not check inclusion and will mark the duty as failed regardless of actual results.

category: feature
ticket: #2200

### DIFF
--- a/core/tracker/inclusion.go
+++ b/core/tracker/inclusion.go
@@ -18,12 +18,12 @@ import (
 )
 
 const (
-	// inclCheckLag is the number of slots to lag before checking inclusion.
+	// InclCheckLag is the number of slots to lag before checking inclusion.
 	// Half an epoch is good compromise between finality and responsiveness.
-	inclCheckLag = 16
-	// inclTrimLag is the number of slots after which we delete cached submissions.
-	// This matches scheduler trimEpochOffset.
-	inclTrimLag = 32 * 3
+	InclCheckLag = 16
+	// InclMissedLag is the number of slots after which we assume the duty was not included and we
+	// delete cached submissions.
+	InclMissedLag = 32
 )
 
 // submission represents a duty submitted to the beacon node/chain.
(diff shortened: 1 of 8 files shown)


core/tracker/inclusion.go
```diff
@@ -18,12 +18,12 @@ import (
 )
 
 const (
-	// inclCheckLag is the number of slots to lag before checking inclusion.
+	// InclCheckLag is the number of slots to lag before checking inclusion.
 	// Half an epoch is good compromise between finality and responsiveness.
-	inclCheckLag = 16
-	// inclTrimLag is the number of slots after which we delete cached submissions.
-	// This matches scheduler trimEpochOffset.
-	inclTrimLag = 32 * 3
```
corverroos, line 26: decrease from 3 to 1 epoch since that is max at this point.

core/tracker/tracker.go
```diff
@@ -29,6 +29,7 @@ const (
 	sigAgg                // Partial signed data aggregated; emitted from sigagg
 	aggSigDB              // Aggregated signed data stored in aggsigdb
 	bcast                 // Aggregated data submitted to beacon node
+	chainInclusion        // Aggregated data submitted to beacon node
```
dB2510, line 32:
```suggestion
	chainInclusion        // Aggregated data included in canonical chain
```

core/tracker/inclusion.go
dB2510: just wondering, what if we are running charon in a network with `slots_per_epoch` != 32, for example, gnosis? Shouldn't we read this kind of information from beacon node?

If we are taking slot as smallest unit in any network then its fine to hardcode it to 32.

dB2510 (approved):
LGTM!
Let's test this.

docs/reasons.md
```diff
@@ -17,6 +17,10 @@ maintain system performance.
   - *Summary*: failed to broadcast duty to beacon node
   - *Details*: Reason `bcast` indicates that beacon node returned an error while submitting aggregated duty signature to beacon node.
 
+### Failure Reason: `chain_inclusion`
+  - *Summary*: duty not included on-chain
+  - *Details*: Reason `chain_inclusion` indicates that even though charon broadcasted the duty successfully, it wasn`t included in the beacon chain. This is expected for up to 20% of attestations. It may however indicate problematic charon broadcast delays or beacon node network problems.
```
xenowits, line 22: nit: do we have data that only 80% attestations end up on-chain?

xenowits (approved):

docs/reasons.md
corverroos: Although that might not be accurate, here is better data. Seems like 5% is more realistic for goerli at least. Well see how mainnet does after releasing v0.16.

image

Issue #1887, opened 2023-03-15
Implement alpha add validator local command


corverroos:
# 🎯 Problem to be solved
Once the new cluster state datastructure has been implemented, #1846, including the `add_validator` operation. We need to add support for generating additional keys and generating the cluster state file to a cli command. We can start with a similar command to `charon create cluster` as this is easiest to implement.

# 🛠️ Proposed solution

See design [doc](https://docs.google.com/document/d/1fpC4BCPeqOL_zKlDWYetuSL87ucVWE4COroIIo1c-lY/edit#).

Create a new command: `charon alpha add-validators-local`

- It takes a legacy lock file as input
- It takes withdrawal addresses and fee_recipients as input
- It generates keys
- It generates a new cluster state file with the `legacy_lock` and `add_validators` mutations

xenowits:
Let's solve this in the following phases:
1. Add `charon alpha add-validators-solo` command to `cmd` package. It does not contain the actual run logic.
2. Add `add validators` logic to the subcommand. This involves using the `state` package. Ex: `state.NewGenValidators`, `state.NewNodeApprovalsComposite` etc.
  * This may involve some number of PRs.
3. Save the new cluster state file to disk

PR #2306, opened 2023-06-13
cmd: add-validators command

xenowits:
Adds logic to `charon alpha add-validators-solo` command using mutable cluster APIs.

category: feature 
ticket: #1887

### DIFF
--- a/cmd/addvalidators.go
+++ b/cmd/addvalidators.go
@@ -75,7 +75,7 @@ func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err erro
 	}
 
 	if err = validateConf(conf, len(cState.Operators)); err != nil {
-		return err
+		return errors.Wrap(err, "validate config")
 	}
 
 	p2pKeys, err := getP2PKeys(conf)
@@ -320,11 +320,11 @@ func genNewVals(numOps, threshold int, forkVersion []byte, conf addValidatorsCon
 
 // getP2PKeys returns a list of p2p private keys either by loading from disk or from test config.
 func getP2PKeys(conf addValidatorsConfig) ([]*k1.PrivateKey, error) {
-	var p2pKeys []*k1.PrivateKey
 	if len(conf.TestConfig.P2PKeys) > 0 {
-		p2pKeys = conf.TestConfig.P2PKeys
+		return conf.TestConfig.P2PKeys, nil
 	}
 
+	var p2pKeys []*k1.PrivateKey
 	for _, enrKeyFile := range conf.EnrPrivKeyfiles {
 		p2pKey, err := k1util.Load(enrKeyFile)
 		if err != nil {


cmd/addvalidators_internal_test.go
```diff
@@ -87,3 +112,22 @@ func TestValidateConfigAddValidators(t *testing.T) {
 		})
 	}
 }
+
+// TODO(xenowits): Add more extensive tests, this is just a very simple unit test.
```
xenowits, line 116: Will add more tests in the subsequent PR

corverroos:
From ChatGPT:

Here are some suggestions based on the provided code:

**cmd/addvalidators.go**
1. Line 12: Import declarations should be grouped in the following order:
    1. Standard library packages.
    2. Related third-party packages.
    3. Own organization's or application's packages.
2. Line 26, Line 30: Provide comments for the `TestConfig` struct and its fields. This could help understand the code better.
3. Line 60: It's usually a good idea to pass context.Context as the first argument to all functions that could potentially take a long time, or need to be cancellable or time out.
4. Line 72: In Go, it's idiomatic to return the zero value when returning an error. It would be better to `return nil, err` here for consistency and to avoid confusion.
5. Line 105: Here and in many other places, it's preferable to include the value that caused the error in the error message. This will help you debug issues much quicker.
6. Line 124: Instead of repeating the loop of appending the single address to the list for each validator, consider using the `repeat` function from the `strings` package.
7. Line 143: Don't ignore the error returned by the `tbls.GenerateSecretKey()` function. Always handle errors immediately, unless you're intentionally ignoring them (and if so, you should probably leave a comment explaining why).
8. Line 223: The function `runAddValidatorsSolo` is quite large and complex, and performs many different tasks. Consider breaking it up into smaller functions, each of which performs one clear task.
9. Line 247, Line 272: It's recommended to encapsulate the error handling within the called function itself, instead of returning errors and handling them in the caller.
10. Line 285: Try to minimize the use of global state or functions that alter global state. Here, you're loading keys from files directly within this function. Instead, consider passing these keys into the function, so the function becomes easier to test.
11. Line 343: Avoid hard-coding version strings. It would be a good idea to define it as a constant at the start of the file.
12. Line 383, Line 385: These if-else statements can be a bit confusing. Consider refactoring them to make the code clearer. 

Remember, code maintainability and readability are very important in long-term software projects. It's worth taking the time to refactor the code to make it easier to read and maintain.

cmd/addvalidators.go
```diff
@@ -36,20 +54,208 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
+	cmd.Flags().StringSliceVar(&config.EnrPrivKeyfiles, "private-key-files", nil, "Comma separated list of paths to charon enr private key files. This should be in the same order as the operators, ie, first private key file should correspond to the first operator and so on.")
 	cmd.Flags().StringSliceVar(&config.FeeRecipientAddrs, "fee-recipient-addresses", nil, "Comma separated list of Ethereum addresses of the fee recipient for each new validator. Either provide a single fee recipient address or fee recipient addresses for each validator.")
 	cmd.Flags().StringSliceVar(&config.WithdrawalAddrs, "withdrawal-addresses", nil, "Comma separated list of Ethereum addresses to receive the returned stake and accrued rewards for each new validator. Either provide a single withdrawal address or withdrawal addresses for each validator.")
 }
 
-func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err error) {
-	// TODO(xenowits): Implement this in next PR, see issue https://github.com/ObolNetwork/charon/issues/1887.
-	return validateConf(conf)
+func runAddValidatorsSolo(conf addValidatorsConfig) (err error) {
```
corverroos, line 66: best retain ctx

cmd/addvalidators.go
```diff
@@ -36,20 +54,208 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
+	if err != nil {
+		return err
+	}
+
+	if err = validateConf(conf, cState.Operators); err != nil {
+		return err
+	}
+
+	// If a single address is provided, use the same address for all the validators.
+	if len(conf.FeeRecipientAddrs) == 1 {
```
corverroos, line 78: extract this a separate function

cmd/addvalidators.go
```diff
@@ -36,20 +54,208 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
+		for i := 0; i < conf.NumVals; i++ {
+			feeRecipients = append(feeRecipients, conf.FeeRecipientAddrs[0])
+			withdrawalAddrs = append(withdrawalAddrs, conf.WithdrawalAddrs[0])
+		}
+
+		conf.FeeRecipientAddrs = feeRecipients
+		conf.WithdrawalAddrs = withdrawalAddrs
+	}
+
+	// Generate new validators
```
corverroos, line 89: extract this a separate function

cmd/addvalidators.go
```diff
@@ -36,20 +54,208 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
+	for _, enrKeyFile := range conf.EnrPrivKeyfiles {
+		enrKey, err := k1util.Load(enrKeyFile)
+		if err != nil {
+			return errors.Wrap(err, "load enr private key")
+		}
+
+		enrKeys = append(enrKeys, enrKey)
+	}
+
+	if conf.TestConfig.Lock != nil {
```
corverroos, line 168: doesn't load keys from disk if you provide them via test config

cmd/addvalidators.go
```diff
@@ -69,6 +275,38 @@ func validateConf(conf addValidatorsConfig) error {
...
+	}
+
+	// Ensure ENR private keys are ordered by peer index.
+	// TODO(xenowits): Add unit test for this.
+	for i, op := range ops {
+		var enrKey *k1.PrivateKey
+		if conf.TestConfig.Lock != nil {
+			enrKey = conf.TestConfig.P2PKeys[i]
+		} else {
+			key, err := k1util.Load(conf.EnrPrivKeyfiles[i])
```
corverroos, line 293: you are loading keys from disk twice, that is not good

corverroos (changes requested):

cmd/addvalidators.go
xenowits: Right, refactored the load keys functionality to a separate function

corverroos (approved):

PR #2325, opened 2023-06-16
eth2util/keystore: handle single keystore case

dB2510:
Handle single keystore case in `orderByKeystoreNum` function. Prior to this, it was returning nil as indices slice for single keystore file. This is because comparator function of sort is not called for a single element slice and this comparator function is responsible for populating indices slice.

category: bug
ticket: none

### DIFF
--- a/eth2util/keystore/keystore.go
+++ b/eth2util/keystore/keystore.go
@@ -395,7 +395,7 @@ func randomHex32() (string, error) {
 	return hex.EncodeToString(b), nil
 }
 
-// validateKeystoreFilename validates keystore filename and returns index of the keystore file.
+// validateKeystoreFilename checks if the given keystore filename is valid and returns index of the keystore file.
 func validateKeystoreFilename(file string) (int, error) {
 	prefix := filepath.Dir(file)
 	extractor := regexp.MustCompile(`keystore-(?:insecure-)?([0-9]+).json`)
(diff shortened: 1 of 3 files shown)


eth2util/keystore/keystore.go
```diff
@@ -411,3 +394,24 @@ func randomHex32() (string, error) {
 
 	return hex.EncodeToString(b), nil
 }
+
+// validateKeystoreFilename validates keystore filename and returns index of the keystore file.
+func validateKeystoreFilename(file string) (int, error) {
```
gsora, line 399: What does `validate` means in this context?

gsora, line 399: Can you write some tests for this function as well please?

dB2510: not sure what you mean by the context, IMO if I read `validateKeystoreFilename` it is clear that it checks if the given keystore filename is valid or not.

Can modify godoc to:
// validateKeystoreFilename checks if the given keystore filename is valid. and returns index of the keystore file.

corverroos: validate shouldn't return data, suggest:`getKeystoreIndex`

gsora: > not sure what you mean by the context

I meant to say to explain in the comment doc when `file`  is valid, and when it isn't.

PR #2326, opened 2023-06-16
eth2util/keystore: refactor Load to return KeyFile type

corverroos:
Refactors `keystore.Load` to return a `KeyFile` type with explicit filename and file index (and private key) to fix issues with implicit index wrangling.

This also fixes the bug if only a single keystore is loaded.

category: bug
ticket: none


eth2util/keystore/load.go
```diff
@@ -0,0 +1,134 @@
+package keystore
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/forkjoin"
```
xenowits, line 7: padding not aligned

eth2util/keystore/load.go
```diff
@@ -0,0 +1,134 @@
...
+	"path"
+	"path/filepath"
+	"regexp"
+	"strconv"
+)
+
+type KeyFiles []KeyFile
+
+// Keys returns the private keys in random order.
+func (s KeyFiles) Keys() []tbls.PrivateKey {
```
xenowits, line 20: nit:
```suggestion
func (k KeyFiles) Keys() []tbls.PrivateKey {
```

eth2util/keystore/load.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, ks := range s {
+		resp = append(resp, ks.PrivateKey)
+	}
+
+	return resp
+}
+
+// SequencedKeys returns the private keys in strict sequential file index order from 0 to N.
+// If the indexes are unknown or not sequential or there are duplicates, an error is returned.
+func (s KeyFiles) SequencedKeys() ([]tbls.PrivateKey, error) {
```
dB2510, line 31: what if there is a case where we have KeyFiles as: `keystore-1.json`, `keystore-5.json `and `keystore-17.json`. Length of keyfiles is 3 but indices are 1, 5 and 17. I think this function will panic in this case with out of bounds message.

eth2util/keystore/load.go
corverroos: fixed

eth2util/keystore/load.go
corverroos: cool

eth2util/keystore/load.go
```diff
@@ -0,0 +1,134 @@
...
+	defer cancel()
+
+	return joinResults.Flatten()
+}
+
+var extractor = regexp.MustCompile(`keystore-(?:insecure-)?([0-9]+).json`)
+
+// extractFileIndex extracts the index from a keystore file name or returns -1
+// if an index cannot be extracted.
+func extractFileIndex(filename string) int {
```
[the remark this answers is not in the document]
corverroos: nope, that is handled higher up, this extracts the index whether it is present or not. Some keyfiles do not have an index.

eth2util/keystore/load.go
corverroos: I added a test, it does not

xenowits (approved):

eth2util/keystore/load.go
```diff
@@ -0,0 +1,142 @@
...
+	"regexp"
+	"strconv"
+
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/forkjoin"
+	"github.com/obolnetwork/charon/app/z"
+	"github.com/obolnetwork/charon/tbls"
+)
+
+type KeyFiles []KeyFile
```
gsora, line 20: Missing godoc

dB2510 (approved):

eth2util/keystore/load.go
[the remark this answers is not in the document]
dB2510: sure, but it would be better to just return for those cases explicitly as there can be more than one reason behind `-1`. And it is also aligned with fail fast pattern.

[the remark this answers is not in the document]
gsora: The regexp never matches `keystore--1.json` or `keystore-insecure--1.json` so there's no way of having that value pop up in the pattern matching process.

Considering the godoc is pretty precise in describing the process, it's well known when -1 is returned. No big deal IMO.

gsora (approved):
Small nitpicks, but overall this approach seems sound. I'm okay with replacing the old one!

eth2util/keystore/load.go
corverroos: added

PR #2330, opened 2023-06-17
core/validatorapi: swallow non-dv registrations

corverroos:
Swallow incorrect non-dv validator registrations. This is a known limitation. Most VCs do not support valid dv-middleware validator registrations (only teku does at this point). But with pre-generated registrations, we do enable builder-api on VCs to enable mev-boost, so that results in these invalid registration being submitted which results in a lot of `unknown public key` errors in both charon as well as errors the VC logs. These errors are just noise, resulting in operators worrying unnecessarily. 

The risk is that we do not identify actual bugs, but that is why we do the debug log (with log filtering).

category: misc
ticket: none

### DIFF
--- a/core/validatorapi/validatorapi.go
+++ b/core/validatorapi/validatorapi.go
@@ -129,6 +129,7 @@ func NewComponent(eth2Cl eth2wrap.Client, allPubSharesByKey map[core.PubKey]map[
 		shareIdx:           shareIdx,
 		feeRecipientFunc:   feeRecipientFunc,
 		builderEnabled:     builderEnabled,
+		swallowRegFilter:   log.Filter(),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


dB2510 (approved):

core/validatorapi/validatorapi.go
```diff
@@ -495,22 +497,29 @@ func (c Component) SubmitBlindedBeaconBlock(ctx context.Context, block *eth2api.
...
+
+	pubkey, err := core.PubKeyFromBytes(eth2Pubkey[:])
 	if err != nil {
 		return err
 	}
 
-	// Note this is the group pubkey
-	eth2Pubkey, err := registration.PubKey()
+	if _, ok := c.getPubShareFunc(eth2Pubkey); !ok {
+		log.Debug(ctx, "Swallowing non-dv registration, "+
```
xenowits, line 512:
```suggestion
		log.Debug(ctx, "Swallowing partial registration, "+
```

PR #2332, opened 2023-06-19
cmd: write cluster manifest to disk

xenowits:
Writes cluster updated by `charon alpha add-validators-solo` command to disk. The new file is named `cluster-manifest.pb`.

category: feature 
ticket: #1887

### DIFF
--- a/cmd/addvalidators.go
+++ b/cmd/addvalidators.go
@@ -89,11 +89,8 @@ func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err erro
 		return err
 	}
 
-	// Write backup of the current cluster manifest to disk
-	err = writeClusterBackup(conf.ClusterBackupDir, cluster)
-	if err != nil {
-		return err
-	}
+	// Will be used to create cluster backup
+	oldCluster := proto.Clone(cluster).(*manifestpb.Cluster)
 
 	if err = validateConf(conf, len(cluster.Operators)); err != nil {
 		return errors.Wrap(err, "validate config")
@@ -157,6 +154,12 @@ func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err erro
 	cluster, err = manifest.Transform(cluster, addVals)
 	if err != nil {
 		return errors.Wrap(err, "transform cluster manifest")
+	}
+
+	// Save cluster backup to disk
+	err = writeClusterBackup(conf.ClusterBackupDir, oldCluster)
+	if err != nil {
+		return err
 	}
 
 	// Save cluster manifest to disk
--- a/cmd/addvalidators.go
+++ b/cmd/addvalidators.go
@@ -6,7 +6,11 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
+	"path"
+	"strings"
+	"time"
 
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
@@ -29,13 +33,14 @@ import (
 
 // addValidatorsConfig is config for the `add-validators` command.
 type addValidatorsConfig struct {
-	NumVals           int
-	WithdrawalAddrs   []string
-	FeeRecipientAddrs []string
+	NumVals           int      // No of validators to add
+	WithdrawalAddrs   []string // Withdrawal address of each validator
+	FeeRecipientAddrs []string // Fee recipient address of each validator
 
-	Lockfile            string
-	ClusterManifestFile string
-	EnrPrivKeyfiles     []string
+	Lockfile            string   // Path to the legacy cluster lock file
+	ClusterManifestFile string   // Path to the cluster manifest file
+	ClusterBackupDir    string   // Directory containing cluster manifest backup files
+	EnrPrivKeyfiles     []string // Paths to node enr private keys
 
 	TestConfig TestConfig
 }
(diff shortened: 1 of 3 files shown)


cmd/addvalidators.go
```diff
@@ -180,8 +189,8 @@ func builderRegistration(secret tbls.PrivateKey, pubkey tbls.PublicKey, feeRecip
 	}, nil
 }
 
-// loadClusterState returns the cluster manifest from the given file path.
-func loadClusterState(conf addValidatorsConfig) (*manifestpb.Cluster, error) {
+// loadClusterManifest returns the cluster manifest from the given file path.
+func loadClusterManifest(conf addValidatorsConfig) (*manifestpb.Cluster, error) {
```
dB2510, line 193: Add TODO to load existing `cluster-manifest.pb` file.

cmd/addvalidators.go
```diff
@@ -27,12 +29,14 @@ import (
...
 // addValidatorsConfig is config for the `add-validators` command.
 type addValidatorsConfig struct {
-	Lockfile          string
-	EnrPrivKeyfiles   []string
 	NumVals           int
 	WithdrawalAddrs   []string
 	FeeRecipientAddrs []string
 
+	Lockfile            string
+	ClusterManifestFile string
```
dB2510, line 37: seems like we will first read the `cluster-manifest.pb` file and then we will override the same after `add-validators` command. I'd suggest to write the new cluster manifest file in a different location rather than overriding the existing one on disk. For ex: having a `--data-dir` flag to write the new cluster manifest file.

cmd/addvalidators.go
xenowits: This TODO is already written in the `manifest.Load` function

cmd/addvalidators.go
xenowits: Cool, added a new flag `--cluster-backup-dir` to store the cluster manifest backups. The backups are in the format `cluster-manifest-backup-1687250009.pb`

xenowits:
Backup files preview:
Screenshot 2023-06-20 at 14 12 40

cmd/addvalidators.go
```diff
@@ -65,15 +74,23 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
-	// Read lock file to load mutable cluster manifest.
-	cluster, err := loadClusterState(conf)
+	// Read lock file to load mutable cluster manifest
+	cluster, err := loadClusterManifest(conf)
+	if err != nil {
+		return err
+	}
+
+	// Write backup of the current cluster manifest to disk
+	err = writeClusterBackup(conf.ClusterBackupDir, cluster)
```
dB2510, line 93: can we move this to end of this function? like write backup file when validators are added and we have a new cluster manifest. 
This is way we can avoid the situation where we have written the backup file first but the command failed after that. In that case, our existing cluster manifest will be same as the backup one and it can be confusing to the operators.

xenowits: Yeah, good point. This also ensures backups are created only when it's time to override the `cluster-manifest.pb` since we can encounter errors before getting to the final point of storing to disk.

cmd/addvalidators.go
```diff
@@ -137,12 +151,22 @@ func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err erro
...
 	// Finally, perform a cluster transformation.
-	_, err = manifest.Transform(cluster, addVals)
+	cluster, err = manifest.Transform(cluster, addVals)
 	if err != nil {
 		return errors.Wrap(err, "transform cluster manifest")
 	}
 
-	// TODO(xenowits): Write new cluster manifest to disk, see issue https://github.com/ObolNetwork/charon/issues/1887.
+	// Save cluster backup to disk
+	err = writeClusterBackup(conf.ClusterBackupDir, oldCluster)
```
corverroos, line 160: this doesn't make sense if you loaded from legacy lock file, it only makes sense if loaded from manifest file.

cmd/addvalidators.go
```diff
@@ -65,19 +74,24 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
 	cmd.Flags().StringSliceVar(&config.FeeRecipientAddrs, "fee-recipient-addresses", nil, "Comma separated list of Ethereum addresses of the fee recipient for each new validator. Either provide a single fee recipient address or fee recipient addresses for each validator.")
 	cmd.Flags().StringSliceVar(&config.WithdrawalAddrs, "withdrawal-addresses", nil, "Comma separated list of Ethereum addresses to receive the returned stake and accrued rewards for each new validator. Either provide a single withdrawal address or withdrawal addresses for each validator.")
+	cmd.Flags().StringVar(&config.Lockfile, "lock-file", ".charon/cluster-lock.json", "The path to the legacy cluster lock file defining distributed validator cluster.")
+	cmd.Flags().StringVar(&config.ClusterManifestFile, "manifest-file", ".charon/cluster-manifest.pb", "The path to the cluster manifest file.")
+	cmd.Flags().StringVar(&config.ClusterBackupDir, "cluster-backup-dir", ".charon/cluster-manifest-backups", "The directory containing cluster manifest backup files.")
```
corverroos, line 81: `"manifest-backup-dir", ".charon/cluster-manifest-backups", "The directory to backup input cluster manifest file before over writing."`

cmd/addvalidators.go
```diff
@@ -206,6 +230,49 @@ func loadClusterState(conf addValidatorsConfig) (*manifestpb.Cluster, error) {
...
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return errors.Wrap(err, "error retrieving backup dir info", z.Str("backup-dir", backupDir))
+	} else if err != nil && errors.Is(err, fs.ErrNotExist) {
+		err := os.Mkdir(backupDir, 0o755)
+		if err != nil {
+			return errors.Wrap(err, "create backup dir", z.Str("backup-dir", backupDir))
+		}
+	} else if err == nil && !info.IsDir() {
+		return errors.New("backup dir already exists and is a file", z.Str("backup-dir", backupDir))
+	}
```
corverroos, line 262:
```
if errors.Is(err, fs.ErrNotExist){
  err := os.Mkdir(backupDir, 0o755)
  if err != nil {
    return errors.Wrap(err, "create backup dir", z.Str("backup-dir", backupDir))
  }
} else if err != nil {
  return errors.Wrap(err, "error retrieving backup dir info", z.Str("backup-dir", backupDir))
} else if info.IsDir() {
  return errors.New("backup dir already exists and is a file", z.Str("backup-dir", backupDir))
}

File: cmd/addvalidators.go (hunk 4 above)
Comment (xenowits) [reply cmd/addvalidators.go]: Good point. But, removing the cluster-backup stuff from this PR, will add this in the subsequent PR for ticket: https://github.com/ObolNetwork/charon/issues/2345

File: cmd/addvalidators.go [hunk 7]
```diff
@@ -65,15 +69,16 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
-	cmd.Flags().StringSliceVar(&config.EnrPrivKeyfiles, "private-key-files", nil, "Comma separated list of paths to charon enr private key files. This should be in the same order as the operators, ie, first private key file should correspond to the first operator and so on.")
 	cmd.Flags().StringSliceVar(&config.FeeRecipientAddrs, "fee-recipient-addresses", nil, "Comma separated list of Ethereum addresses of the fee recipient for each new validator. Either provide a single fee recipient address or fee recipient addresses for each validator.")
 	cmd.Flags().StringSliceVar(&config.WithdrawalAddrs, "withdrawal-addresses", nil, "Comma separated list of Ethereum addresses to receive the returned stake and accrued rewards for each new validator. Either provide a single withdrawal address or withdrawal addresses for each validator.")
+	cmd.Flags().StringVar(&config.Lockfile, "lock-file", ".charon/cluster-lock.json", "The path to the legacy cluster lock file defining distributed validator cluster.")
+	cmd.Flags().StringVar(&config.ClusterManifestFile, "manifest-file", ".charon/cluster-manifest.pb", "The path to the cluster manifest file.")
```
corverroos, line 75: add a comment that this takes presedence over lock file if found. Also add a comment to lock file that manifest file takes presendence

corverroos (approved):

cmd/addvalidators.go
xenowits: Screenshot 2023-06-22 at 11 38 43

xenowits:
`charon alpha add-validators-solo` command help:

Screenshot 2023-06-22 at 11 39 53

Issue #2347, opened 2023-06-22
Fix panic in testutil/validatormock/synccomm.go


dB2510:
# 🐞 Bug Report

### Description

A panic has been observed by compose tests in github actions: https://github.com/ObolNetwork/charon/actions/runs/5335228338/attempts/1
Might be due to a race condition.

### Has this worked before in a previous version?

Yes

## 🔥 Error

<pre><code>
panic: close of closed channel
[35mnode2_1       |[0m 
[35mnode2_1       |[0m goroutine 478 [running]:
[35mnode2_1       |[0m github.com/obolnetwork/charon/testutil/validatormock.(*SyncCommMember).setDuties(0x1d76ea0?, 0xc002d69c80?, {0x2a468b0?, 0xc001f4bd00?, 0x1d76ea0?})
[35mnode2_1       |[0m 	/app/charon/testutil/validatormock/synccomm.go:153 +0x105
[35mnode2_1       |[0m github.com/obolnetwork/charon/testutil/validatormock.(*SyncCommMember).PrepareEpoch(0xc001fc6b40, {0x1d76ea0, 0xc002d69c80})
[35mnode2_1       |[0m 	/app/charon/testutil/validatormock/synccomm.go:182 +0x9f
[35mnode2_1       |[0m github.com/obolnetwork/charon/app.wireValidatorMock.func1.2({0x1d76ea0?, 0xc002d69c80?}, {{0x1d92f00, 0xc001f4bd00}, 0xc001b0e600, 0xc001d94e70, 0xc001fc6b40})
[35mnode2_1       |[0m 	/app/charon/app/vmock.go:47 +0x45
[35mnode2_1       |[0m github.com/obolnetwork/charon/app.newVMockWrapper.func1.1()
[35mnode2_1       |[0m 	/app/charon/app/vmock.go:182 +0xac
[35mnode2_1       |[0m created by github.com/obolnetwork/charon/app.newVMockWrapper.func1
[35mnode2_1       |[0m 	/app/charon/app/vmock.go:178 +0x792

</code></pre>

**Anything else relevant (validator index / public key)?**
Here's complete log of the failed test: [run_version_matrix_with_dkg.log](https://github.com/ObolNetwork/charon/files/11829595/run_version_matrix_with_dkg.log)

PR #2349, opened 2023-06-22
testutil/validatormock: error instead of panic

corverroos:
Refactors validator mock logic to error when state already set (instead of panic). Also use explicit slot/poch to store stateful objects, rather then implicit "last slot" since out of order slot scheduling should cause problems. 

category: test
ticket: #2347


dB2510 (approved):
LGTM

Issue #2334, opened 2023-06-19
load cluster manifest and lock file


xenowits:
# 🎯 Problem to be solved

Currently, we have two files representing charon's config:
* `cluster-lock.json`: Legacy lock file (old)
* `cluster-manifest.pb`: Mutable cluster manifest (new)

We need a way to figure out what happens if both of these files are provided.

# Proposed solution

Refactor the `manifest.Load` function to take 2 parameters, `manifestFile` & `legacyLockFile`:
```go
func Load(manifestFile, legacyLockFile string, lockCallback func(cluster.Lock) error) (*manifestpb.Cluster, error) {}
```
The following is the intended flow:
* Try loading the `manifest` file first.
  * If successful, use the manifest and return
* If manifest file is not found, try loading the `legacyLockFile`
  * If successful, use the lock and return
* If none of the files are found, return an error

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

# 👐 Additional acceptance criteria
None

# ❌ Out of Scope
None

gsora:
@OisinKyne how would you handle the situation in which a user wants to load a cluster created in v0.16 in v0.17, assuming we only support either lock or manifest files? 

This is what we're trying to solve: supporting old cluster lock files on charon versions that support cluster manifest.

An alternative might be writing a migration tool/command, but that opens a can of worms I don't think we should close.

I believe we should support cluster lock files at some point, but we can live with this bit of tech debt right now.

corverroos:
Another safety feature could be:
 
If both legacy lock and manifest files are found, ensure that the cluster hashes match, ie, that the manifest contains that legacy lock.  Else error with mismatching manifest and legacy lock found

xenowits:
The functionality to load from both files would help us in slowly phasing out clusters with legacy lock files. Since we won't be building a migration tool to bump "legacy clusters" to "manifest clusters", this feature would help in easy migration:
* load legacy lock
* run any alpha command, like `charon alpha add-validators-solo`
* a new `cluster-manifest.pb` file is created
* this new cluster manifest can now be used to run charon

xenowits:
> Another safety feature could be:
> 
> If both legacy lock and manifest files are found, ensure that the cluster hashes match, ie, that the manifest contains that legacy lock. Else error with mismatching manifest and legacy lock found

[...]

Issue #2345, opened 2023-06-21
Create backup manifests


xenowits:
# 🎯 Problem to be solved

When we run `charon alpha add-validators-solo` command, it changes (mutates) the loaded cluster and writes the mutated cluster to a `cluster-manifest.pb` file. This means the `cluster-manifest.pb` is overriden each time a new mutation is performed.

This maybe problematic if we want to revert to an older state in case there's a bug in our code.

# 🛠️ Proposed solution

The solution is to create backups of manifest files before overriding.

* Write a backup file before modifying `cluster-manifest.pb`.
* Each backup file should be named with a lexicographic datetime (fancy way of saying reverse order so it sorts well). Timestamps in unix time aren't user friendly despite being sortable. cluster-manifest.pb.20230721183000"

* Note that backup file should only be created when the cluster is loaded from a cluster manifest file, not a legacy lock file.

^ We shouldn't be dragging tech debt along here, think we need to spec more carefully. 

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

OisinKyne:
Don't think this needs to be in a different folder. don't think it needs a flag. 

I think we just need to make a copy, can be like `cluster-manifest.pb.20230721183000`, and we just spit a new one of these out every time the `charon alpha add-validators` command runs, and if the program hits a panic, it has the ability to unwind the situation and leave a working `charon-manifest.pb` no matter what.

PR #2358, opened 2023-06-26
cluster/manifest: load cluster from disk

xenowits:
Loads cluster either from `cluster manifest` or `legacy lock` file. If both files are provided, `cluster manifest` is read first.

category: feature 
ticket: #2334

### DIFF
--- a/cluster/manifest/load.go
+++ b/cluster/manifest/load.go
@@ -13,47 +13,32 @@ import (
 	manifestpb "github.com/obolnetwork/charon/cluster/manifestpb/v1"
 )
 
-// LoadMetadata represents the result of loading a cluster from cluster manifest or legacy lock file.
-type LoadMetadata struct {
-	Filename     string // Name of the file from which the cluster was loaded
-	IsLegacyLock bool   // True if cluster was loaded from legacy lock file
-}
-
-// Load loads a cluster from disk. It supports reading from both cluster manifest and legacy lock files.
+// Load loads a cluster from disk and returns true if cluster was loaded from a legacy lock file.
+// It supports reading from both cluster manifest and legacy lock files.
 // If both files are provided, it first reads the manifest file before reading the legacy lock file.
 // TODO(xenowits): Remove loading from legacy lock when we fully adopt mutable cluster manifests.
-func Load(manifestFile, legacyLockFile string, lockCallback func(cluster.Lock) error) (*manifestpb.Cluster, LoadMetadata, error) {
+func Load(manifestFile, legacyLockFile string, lockCallback func(cluster.Lock) error) (*manifestpb.Cluster, bool, error) {
 	b, err := os.ReadFile(manifestFile)
 	if err == nil {
 		manifest := new(manifestpb.Cluster)
 		if err := proto.Unmarshal(b, manifest); err != nil {
-			return nil, LoadMetadata{}, errors.Wrap(err, "unmarshal cluster manifest")
+			return nil, false, errors.Wrap(err, "unmarshal cluster manifest")
 		}
 
-		return manifest, LoadMetadata{Filename: manifestFile, IsLegacyLock: false}, nil
+		return manifest, false, nil
 	}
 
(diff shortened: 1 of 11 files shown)


cluster/manifest/load.go
```diff
@@ -13,21 +13,47 @@ import (
...
+		return manifest, LoadMetadata{Filename: manifestFile, IsLegacyLock: false}, nil
+	}
+
+	b, err = os.ReadFile(legacyLockFile)
 	if err != nil {
-		return nil, errors.Wrap(err, "read file")
+		return nil, LoadMetadata{}, errors.Wrap(err, "read legacy lock file")
 	}
 
 	rawDAG := new(manifestpb.SignedMutationList)
```
gsora, line 41: Why not call `loadLegacyLock` straight away?

```go
	m, err := loadLegacyLock(b, lockCallback)
	if err != nil {
		return nil, LoadMetadata{}, errors.Wrap(err, "load legacy lock")
	}

	return m, LoadMetadata{Filename: legacyLockFile, IsLegacyLock: true}, nil
```

cmd/addvalidators.go
```diff
@@ -147,10 +147,14 @@ func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err erro
 		return errors.Wrap(err, "transform cluster manifest")
 	}
 
-	// TODO(xenowits): Write cluster backup, see https://github.com/ObolNetwork/charon/issues/2345.
+	// Save cluster backup to disk
+	//nolint:revive,staticcheck
+	if loadMetadata.IsLegacyLock {
+		// TODO(xenowits): Write cluster backup, see https://github.com/ObolNetwork/charon/issues/2345.
```
gsora, line 153: #2345 says otherwise?

> Note that backup file should only be created when the cluster is loaded from a cluster manifest file, not a legacy lock file.

cluster/manifest/load.go
```diff
@@ -13,21 +13,47 @@ import (
...
+// LoadMetadata represents the result of loading a cluster from cluster manifest or legacy lock file.
+type LoadMetadata struct {
+	Filename     string // Name of the file from which the cluster was loaded
+	IsLegacyLock bool   // True if cluster was loaded from legacy lock file
+}
+
+// Load loads a cluster from disk. It supports reading from both cluster manifest and legacy lock files.
+// If both files are provided, it first reads the manifest file before reading the legacy lock file.
+// TODO(xenowits): Remove loading from legacy lock when we fully adopt mutable cluster manifests.
+func Load(manifestFile, legacyLockFile string, lockCallback func(cluster.Lock) error) (*manifestpb.Cluster, LoadMetadata, error) {
```
gsora, line 25: Instead of returning a `LoadMetadata`, would a simple `bool` work?

Caller know at what path manifest and lock file live, so just telling them "hey I loaded a legacy lock file" should be enough IMO.

xenowits: Yeah, that would be simpler, lm update it

cluster/manifest/load.go
```diff
@@ -13,21 +13,32 @@ import (
...
-// Load loads a cluster manifest from disk. It supports both legacy lock files and raw DAG files.
-// TODO(xenowits): Refactor, load either from manifest or lock file, see https://github.com/ObolNetwork/charon/issues/2334.
-func Load(file string, lockCallback func(cluster.Lock) error) (*manifestpb.Cluster, error) {
-	b, err := os.ReadFile(file)
-	if err != nil {
-		return nil, errors.Wrap(err, "read file")
+// Load loads a cluster from disk and returns true if cluster was loaded from a legacy lock file.
+// It supports reading from both cluster manifest and legacy lock files.
+// If both files are provided, it first reads the manifest file before reading the legacy lock file.
+// TODO(xenowits): Remove loading from legacy lock when we fully adopt mutable cluster manifests.
```
corverroos, line 19: can remove this todo as this will be required for a long time

corverroos (approved):

gsora (approved):
LFTM

PR #2360, opened 2023-06-26
core/bcast: don't register inactive validators

gsora:
Make sure to check the list of active validators before broadcasting the associated validator registration.

category: bug
ticket: none


core/bcast/recast.go
```diff
@@ -101,8 +109,25 @@ func (r *Recaster) SlotTicked(ctx context.Context, slot core.Slot) error {
...
 		ctx := log.WithCtx(ctx, z.Any("duty", tuple.duty))
+
+		ethPk, err := pubkey.ToETH2()
+		if err != nil {
+			log.Error(ctx, "Can't convert pubkey to eth2 format", err)
+			continue
+		}
+
+		if _, found := activeVals[ethPk]; !found {
+			log.Debug(ctx, "Ignoring non-active validator", z.Str("pubkey", pubkey.String()))
```
corverroos, line 127: inactive validators is expected, so this is going to be very verbose expeseically for large clusters, can remove I think.

corverroos (approved):

core/bcast/recast.go
gsora: yea was about to ask you for an opinion on that debug message, will remove!

app/app.go
```diff
@@ -571,7 +571,23 @@ func wireRecaster(ctx context.Context, eth2Cl eth2wrap.Client, sched core.Schedu
...
 ) error {
-	recaster := bcast.NewRecaster()
+	recaster, err := bcast.NewRecaster(func(ctx context.Context) (map[eth2p0.BLSPubKey]struct{}, error) {
+		valList, err := eth2Cl.ActiveValidators(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		ret := make(map[eth2p0.BLSPubKey]struct{})
+
```
xenowits, line 581: nit: remove newline

xenowits (approved):

app/app.go
```diff
@@ -571,7 +571,23 @@ func wireRecaster(ctx context.Context, eth2Cl eth2wrap.Client, sched core.Schedu
...
 	callback func(context.Context, core.Duty, core.PubKey, core.SignedData) error,
 ) error {
-	recaster := bcast.NewRecaster()
+	recaster, err := bcast.NewRecaster(func(ctx context.Context) (map[eth2p0.BLSPubKey]struct{}, error) {
+		valList, err := eth2Cl.ActiveValidators(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		ret := make(map[eth2p0.BLSPubKey]struct{})
```
xenowits, line 580: nit: can simply have a map[pubkey]bool
```suggestion
		ret := make(map[eth2p0.BLSPubKey]bool)
```

gsora: using `struct{}` as a dummy value for maps is a well-known practice, i don't think it needs changing

PR #2365, opened 2023-06-27
cmd/combine: always load the same set of files

gsora:
`manifest.Load()` gives precedence to manifest files instead of locks, so that even in case a charon directory contains both, manifest is always chosen.

Make sure that combine checks the second return value of that method and stop the process if a lock file is loaded instead of a manifest.

TLDR: combined directories must all load manifest files OR lock files.

category: refactor
ticket: none

### DIFF
--- a/cmd/combine/combine.go
+++ b/cmd/combine/combine.go
@@ -196,11 +196,12 @@ type options struct {
 }
 
 // loadManifest loads a cluster manifest from one of the charon directories contained in dir.
-// It checks that all the directories containing a validator_keys subdirectory contain the same cluster_lock.json file.
+// It checks that all the directories containing a validator_keys subdirectory contain the same manifest file, or lock file.
+// loadManifest gives precedence to the manifest file.
+// loadManifest will fail if some of the directories contain a different set of manifest and lock file.
+// For example, if 3 out of 4 directories contain both manifest and lock file, and the fourth only contains lock, loadManifest will return error.
 // It returns the v1.Cluster read from the manifest, and a list of directories that possibly contains keys.
 func loadManifest(ctx context.Context, dir string, noverify bool) (*manifestpb.Cluster, []string, error) {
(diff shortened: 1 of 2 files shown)


cmd/combine/combine.go
```diff
@@ -220,13 +222,19 @@ func loadManifest(ctx context.Context, dir string, noverify bool) (*manifestpb.C
...
-		cl, _, err := manifest.Load(manifestFile, lockFile, func(lock cluster.Lock) error {
+		cl, legacyLockLoaded, err := manifest.Load(manifestFile, lockFile, func(lock cluster.Lock) error {
 			return verifyLock(ctx, lock, noverify)
 		})
 		if err != nil {
 			return nil, nil, errors.Wrap(err, "manifest load error", z.Str("name", sd.Name()))
 		}
 
+		if (didReadLegacyLock != legacyLockLoaded) && idx != 0 {
+			return nil, nil, errors.New("loaded legacy lock, but not all other directories being combined contain one")
```
xenowits, line 233:
```suggestion
			return nil, nil, errors.New("missing legacy lock in some directories")
```

gsora: Hmmm while I agree the error I proposed isn't worded the best, it highlights the fact that finding a cluster lock and no manifest is unexpected... @corverroos you're the error message magician, do you have any suggestion?

cmd/combine/combine.go
```diff
@@ -209,6 +211,7 @@ func loadManifest(ctx context.Context, dir string, noverify bool) (*manifestpb.C
 	var (
 		possibleValKeysDir []string
 		lastCluster        *manifestpb.Cluster
+		lastMutationHash   []byte
```
corverroos, line 214: I think you can remove this variable and just use `lastCluster.LatestMutationHash`

corverroos (approved):

Issue #2367, opened 2023-06-27
Workflows using wrong image tag


LukeHackett12:
The image tag being deployed for promrated and the dev relay is 1 character too long

PR #2368, opened 2023-06-27
workflows: add length to shorten command

LukeHackett12:
Fix image sha for deployments

category: bug
ticket: #2367


aly-obol (approved):
LGTM

PR #2369, opened 2023-06-27
cluster/manifest: refactor cluster hashes

corverroos:
Refactor `manifestpb.Cluster` to expose `InitialMutationHash` and `LatestMutationHash` explicitly.

`LatestMutationHash` is required for:
- Identify manifest "iteration", allowing to comparison of different manifest iterations.
- Use of parent hash when adding subsequent mutations. 

Also regenerate all protobufs with latest protogengo version.

category: refactor
ticket: none


cmd/addvalidators.go
```diff
@@ -108,7 +108,7 @@ func runAddValidatorsSolo(_ context.Context, conf addValidatorsConfig) (err erro
 	}
 
 	// Perform a `gen_validators/v0.0.1` mutation using the newly created validators.
-	genVals, err := manifest.NewGenValidators(cluster.Hash, vals)
```
corverroos, line 111: Note this was a bug.

@xenowits can you please add a unit test that calls add_validators 2 times, so adds validators to the same manifest twice.

xenowits (approved):
LGTM!

gsora (approved):

PR #2370, opened 2023-06-27
cmd: fix bug in `add-validators-solo` command

xenowits:
Fixes bug in `add-validators-solo` command identified [here](https://github.com/ObolNetwork/charon/pull/2369/files#r1243873252).

Also adds a unit test which runs the command twice and ensures that the final manifest has two extra validators as expected.

category: bug 
ticket: none

### DIFF
--- a/cmd/addvalidators_internal_test.go
+++ b/cmd/addvalidators_internal_test.go
@@ -119,39 +119,85 @@ func TestRunAddValidators(t *testing.T) {
 		n        = 3
 		valCount = 1
 	)
-	lock, p2pKeys, _ := cluster.NewForT(t, valCount, n, n, 0)
 
-	tmp := t.TempDir()
-	manifestFile := path.Join(tmp, "cluster-manifest.pb")
-	conf := addValidatorsConfig{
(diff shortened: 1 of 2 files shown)


cmd/addvalidators_internal_test.go
```diff
@@ -118,33 +118,79 @@ func TestRunAddValidators(t *testing.T) {
...
+		b, err := os.ReadFile(manifestFile)
+		require.NoError(t, err)
+
+		var msg manifestpb.Cluster
+		require.NoError(t, proto.Unmarshal(b, &msg))
+		require.Equal(t, len(msg.Validators), 2) // valCount+1
+		require.Equal(t, msg.Validators[1].FeeRecipientAddress, feeRecipientAddr)
+	})
+
+	t.Run("add validators twice", func(t *testing.T) {
```
xenowits, line 152: Note that this test can be easily generalized to test that `n` consecutive runs actually add `n` new validators to the cluster manifest. I have just considered `n == 2` for simplicity

corverroos (approved):

PR #2374, opened 2023-06-28
app/health: implement health checks

corverroos:
PoC of Health Checks V2. See `checks.go`

category: feature 
ticket: none

### DIFF
--- a/app/health/checker.go
+++ b/app/health/checker.go
@@ -19,6 +19,7 @@ const (
 	maxScrapes   = 10
 )
 
+// NewChecker returns a new health checker.
 func NewChecker(metadata Metadata, gatherer prometheus.Gatherer) *Checker {
 	return &Checker{
 		metadata:     metadata,
@@ -30,6 +31,7 @@ func NewChecker(metadata Metadata, gatherer prometheus.Gatherer) *Checker {
 	}
 }
 
+// Checker is a health checker.
 type Checker struct {
 	metadata     Metadata
 	checks       []check
@@ -40,6 +42,7 @@ type Checker struct {
 	logFilter    z.Field
 }
 
+// Run runs the health checker until the context is canceled.
 func (c *Checker) Run(ctx context.Context) {
 	ticker := time.NewTicker(c.scrapePeriod)
 	defer ticker.Stop()
@@ -59,6 +62,7 @@ func (c *Checker) Run(ctx context.Context) {
 	}
 }
 
+// Instrument runs all health checks and updates the check gauge.
 func (c *Checker) instrument(ctx context.Context) {
 	for _, check := range c.checks {
 		failing, err := check.Func(c.query, c.metadata)
@@ -76,6 +80,7 @@ func (c *Checker) instrument(ctx context.Context) {
 	}
 }
 
+// scrape scrapes metrics from the gatherer.
 func (c *Checker) scrape() error {
 	metrics, err := c.gatherer.Gather()
 	if err != nil {
--- a/app/health/checker.go
+++ b/app/health/checker.go
@@ -1,0 +1,118 @@
+// Copyright © 2022-2023 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package health
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	pb "github.com/prometheus/client_model/go"
+
+	"github.com/obolnetwork/charon/app/errors"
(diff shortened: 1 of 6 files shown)


app/health/checks.go
```diff
@@ -0,0 +1,100 @@
...
+		Severity:    severityCritical,
+		Func: func(s store, m metadata) bool {
+			counts := s.Get("p2p_peer_connection_total", countNonZeroLabels)
+			return gaugeMin(counts) < float64(m.QuorumPeers)
+		},
+	},
+	{
+		Name:        "pending_validators",
+		Description: "Pending validators detected. Activate them to start validating.",
+		Type:        typeNegative,
```
xenowits, line 82: would we still call this negative if there are no pending validators, ie, `gaugeMax(counts) == 0` ?

xenowits (approved):

app/health/select.go
```diff
@@ -0,0 +1,137 @@
...
+			val = metric.Counter.GetValue()
+		case pb.MetricType_GAUGE:
+			val = metric.Gauge.GetValue()
+		default:
+			panic("invalid metric type for simple value labelSelector")
+		}
+
+		if val > max.Counter.GetValue() {
+			max = metric
+		}
```
dB2510, line 32: how do you know that `max` metric will always be a counter metric? i think it depends on `metric` variable of the for loop which gets assigned to max in this if block

dB2510: maybe it is better to have two variables: `maxValue` and `maxMetric`. `maxValue` compares the values of the metrics and `maxMetric` stores the max metric based on the value.

app/health/select.go
```diff
@@ -0,0 +1,137 @@
...
+
+func countNonZeroLabels(metricsFam *pb.MetricFamily) *pb.Metric {
+	gauge := &pb.Metric{
+		Gauge:       new(pb.Gauge),
+		TimestampMs: metricsFam.Metric[0].TimestampMs,
+	}
+
+	for _, metric := range metricsFam.Metric {
+		if metric.Gauge.GetValue() != 0 {
+			*guage.Gauge.Value++
```
dB2510, line 46:
```suggestion
			*gauge.Gauge.Value++
```

app/health/select.go
```diff
@@ -0,0 +1,137 @@
...
+			panic("histogram sum not implemented yet")
+		default:
+			panic("invalid metric type for simple value labelSelector")
+		}
+	}
+
+	return sum
+}
+
+func l(name, val string) *pb.LabelPair {
```
dB2510, line 79: can we have a proper name of this function?

dB2510 (commented):
the approach looks good to me, looking forward to how it will interact with other components.

app/health/reducers.go
```diff
@@ -0,0 +1,48 @@
+// Copyright © 2022-2023 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package health
+
+import pb "github.com/prometheus/client_model/go"
```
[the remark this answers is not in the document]
corverroos: it is protobufs, so pb is common pattern

app/health/select.go
corverroos: updated

app/health/select.go
corverroos: fixed

app/health/select.go
corverroos: no 😄

gsora (approved):

app/health/checker.go
```diff
@@ -0,0 +1,126 @@
...
+	checks       []check
+	metrics      [][]*pb.MetricFamily
+	gatherer     prometheus.Gatherer
+	scrapePeriod time.Duration
+	maxScrapes   int
+	logFilter    z.Field
+}
+
+// Run runs the health checker until the context is canceled.
+func (c *Checker) Run(ctx context.Context) {
```
dB2510, line 49: i assume this will be a lifecycle hook in app package later.

app/health/checker.go
```diff
@@ -0,0 +1,126 @@
...
+				log.Warn(ctx, "Failed to scrape metrics", err)
+				continue
+			}
+
+			c.instrument(ctx)
+		}
+	}
+}
+
+// Instrument runs all health checks and updates the check gauge.
```
dB2510, line 68:
```suggestion
// instrument runs all health checks and updates the check gauge.
```

xenowits (approved):

dB2510 (approved):

Issue #2341, opened 2023-06-20
Remove default values for critical flags


corverroos:
# 🎯 Problem to be solved
Users do not understand how default values of flags work, see https://discord.com/channels/849256203614945310/1084021363406274560/1120691357766398003.

In `charon create cluster`, the number of validators to add is a critical decision the user should make. Defaulting to 1 doesn't mean that the user actually wants 1 validator, especially if they then provide 4 addresses.

# 🛠️ Proposed solution
Remove the `--num-validators` default value and error if 0, requesting the user explicitly provide the value.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2380, opened 2023-06-30
cmd: simplify `create cluster` logic

gsora:
This PR refactors `create cluster` logic to make it slightly simpler.

New behaviors:
 - if `--split-existing-keys` is enabled, `--num-validators` cannot be set
 - `--num-validators` must be present unless `--split-existing-keys` is enabled, otherwise `create cluster` exits with error
 - split key generation and disk read
 - don't fill in with random keys when splitting: during `--split-existing-keys` before this PR, if the user specified a `--num-validators` greater than the amount of keys contained in `--split-keys-dir` Charon would fill in the blanks with random keys
 - if definition file validator amount and `--num-validators` don't match, the creation process stops

category: refactor
ticket: #2341

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -107,7 +107,7 @@ func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.StringSliceVar(&config.WithdrawalAddrs, "withdrawal-addresses", nil, "Comma separated list of Ethereum addresses to receive the returned stake and accrued rewards for each validator. Either provide a single withdrawal address or withdrawal addresses for each validator.")
 	flags.StringVar(&config.Network, "network", defaultNetwork, "Ethereum network to create validators for. Options: mainnet, goerli, gnosis, sepolia.")
 	flags.BoolVar(&config.Clean, "clean", false, "Delete the cluster directory before generating it.")
-	flags.IntVar(&config.NumDVs, "num-validators", 1, "The number of distributed validators needed in the cluster.")
+	flags.IntVar(&config.NumDVs, "num-validators", 0, "The number of distributed validators needed in the cluster.")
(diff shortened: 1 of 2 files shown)


cmd/createcluster.go
```diff
@@ -139,6 +139,11 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return errors.New("number of --keymanager-addresses do not match --keymanager-auth-tokens. Please fix configuration flags")
 	}
 
+	if conf.SplitKeys && conf.NumDVs == 0 {
+		// set conf.NumDVs to 1 if the user didn't specify num-validators
+		conf.NumDVs = 1
+	}
```
corverroos, line 145: I do not understand why we do this?

corverroos (changes requested):

OisinKyne:
> Artificially set it to 1 if --split-existing-keys is enabled and user didn't provide --num-validators, so that the code flow can continue.

What happens if I point `--split-existing-keys` and `--split-existing-keys-dir` at a folder with a number of keys in it? Will it produce only 1, or will it know to produce as many validators as there are keystores?

gsora:
I’m reworking the whole PR, I’ll ask for review later. The approach I used in this PR didn’t seem sound.

gsora:
@OisinKyne please review the behavior changes now

cmd/createcluster.go
```diff
@@ -139,6 +139,31 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return errors.New("number of --keymanager-addresses do not match --keymanager-auth-tokens. Please fix configuration flags")
 	}
 
+	var secrets []tbls.PrivateKey
+
+	switch conf.SplitKeys {
+	case true:
```
corverroos, line 145: what about a good old `if else` rather?

corverroos (approved):

cmd/createcluster.go
gsora: the effect is the same, i can replace it if you want

gsora: done

PR #2404, opened 2023-07-05
app/health: add remaining check tests

corverroos:
Add tests for all remaining health checks.

category: test
ticket: none


xenowits (approved):

app/health/checks.go
```diff
@@ -111,21 +111,13 @@ var checks = []check{
 		Description: "Proposal failures detected. See <link to troubleshoot proposal failures>.",
 		Severity:    severityWarning,
 		Func: func(q query, m Metadata) (bool, error) {
-			fullIncrease, err := q("core_tracker_failed_duties_total",
-				selectLabel(l("duty", "proposal")),
-				counterIncrease)
+			increase, err := q("core_tracker_failed_duties_total",
+				sumLabels(l("duty", "proposal")), increase)
```
dB2510, line 115: i think this should be:
```suggestion
				sumLabels(l("duty", "proposer"), l("duty", "builder_proposer")), increase)
```

dB2510 (approved):

app/health/checks.go
corverroos: This is will not work, label selectors uses AND not OR. I changed to `.*propser` to make it more explicit what is going on.

Issue #2382, opened 2023-06-30
Create keystore files for new validators


xenowits:
# 🎯 Problem to be solved

The `charon alpha add-validators-alpha` command doesn't create keystore files (`keystore-i.json` & `keystore-i.txt`) for newly created validators.

# 🛠️ Proposed solution

Create keystore files for each new validator and store them in each node's individual directory:
* `node0/validator_keys`
* `node1/validator_keys`
* ....
* `nodeN/validator_keys`

Keystore files should be named starting with `keystore-<numVals>.json`. For example, if a cluster already has 5 validator keys in its `validator_keys/` directory and we try adding 2 more, the new keystore files should be:
* `keystore-5.json`
* `keystore-5.txt`
* `keystore-6.json`
* `keystore-6.txt`

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2406, opened 2023-07-06
cmd: write keystores

xenowits:
Write keystores for new validators to disk.

category: feature 
ticket: #2382

### DIFF
--- a/cmd/addvalidators.go
+++ b/cmd/addvalidators.go
@@ -101,6 +101,8 @@ func runAddValidatorsSolo(ctx context.Context, conf addValidatorsConfig) (err er
 	if err != nil {
 		return err
 	}
+	nextKeystoreIdx := len(cluster.Validators)
+
 	log.Info(ctx, "Cluster manifest loaded",
 		z.Str("cluster_name", cluster.Name),
 		z.Str("cluster_hash", hex7(cluster.InitialMutationHash)),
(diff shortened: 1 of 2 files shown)


xenowits:
Output when `add-validators-solo` adds two new validators to a cluster with originally a single validator:
Screenshot 2023-07-06 at 13 09 54

Output when `add-validators-solo` adds two new validators again to the above cluster(total 5 validators):
Screenshot 2023-07-06 at 13 10 07

cmd/addvalidators.go
```diff
@@ -167,6 +170,12 @@ func runAddValidatorsSolo(ctx context.Context, conf addValidatorsConfig) (err er
 		return errors.Wrap(err, "transform cluster manifest")
 	}
 
+	nextKeystoreIdx := len(cluster.Validators) - conf.NumVals
```
dB2510, line 173: Add comment on this

dB2510 (approved):

corverroos (approved):

cmd/addvalidators.go
xenowits: No, since the `nextKeystoreIdx` value is calculated after the cluster is transformed, so the cluster contains both old & new validators. Hence, needed to subtract the new validator count.

Although, since this doesn't seem to be intuitive, I'll just move this before `manifest.Transform()` is called.

PR #2413, opened 2023-07-10
cmd: add unsafe command

dB2510:
Adds 'unsafe' command to provide support for testing features from the CLI.

- The 'unsafe' command introduces additional flags in its subcommands, specifically designed for testing purposes.
- The 'charon unsafe' command is hidden from the 'charon --help' output.
- Regular usage of commands will not include any test-only flags, whereas these flags will be available in the 'unsafe' command for testing purposes.

category: feature
ticket: none


cmd/run.go
```diff
@@ -69,13 +73,7 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
 	cmd.Flags().StringVar(&config.MonitoringAddr, "monitoring-address", "127.0.0.1:3620", "Listening address (ip and port) for the monitoring API (prometheus, pprof).")
 	cmd.Flags().StringVar(&config.JaegerAddr, "jaeger-address", "", "Listening address for jaeger tracing.")
 	cmd.Flags().StringVar(&config.JaegerService, "jaeger-service", "charon", "Service name used for jaeger tracing.")
-	cmd.Flags().BoolVar(&config.SimnetBMock, "simnet-beacon-mock", false, "Enables an internal mock beacon node for running a simnet.")
-	cmd.Flags().BoolVar(&config.SimnetVMock, "simnet-validator-mock", false, "Enables an internal mock validator client when running a simnet. Requires simnet-beacon-mock.")
```
corverroos, line 73: lets keep these existing simnet stuff here for now, I still use it, compose also uses it. Can migrate this later once we are sure about this new unsafe approach

dB2510: aligned, reverted it!

corverroos (approved):

Issue #2335, opened 2023-06-19
cluster hashes should match in manifest and lock


xenowits:
# 🎯 Problem to be solved

Users can provide both `cluster-manifest.pb` and `cluster-lock.json` files to charon commands such as `charon run`. In such cases, for safety, we can verify that the cluster hashes in both these files match. 

# 🛠️ Proposed solution

If both legacy lock and manifest files are found, load both files. Ensure that the cluster hashes match, ie, that the manifest contains that legacy lock. Else error with mismatching manifest and legacy lock found.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

# 👐 Additional acceptance criteria

None

# ❌ Out of Scope

None

OisinKyne:
I don't think we should have definition and lock files as part of the manifest flow, what is the benefit? what does the user journey look like? How does the launchpad flow change to accomodate this?

corverroos:
@OisinKyne see our comments in #2334. Adding support for both legacy lock files and new manifest files is required to support both old and new cluster configuration with the same code base.

PR #2414, opened 2023-07-10
manifest: cluster hash mismatch error

xenowits:
Return error if clusters loaded from legacy lock and manifest files don't have the same cluster hash.

category: feature 
ticket: #2335

### DIFF
--- a/cluster/manifest/load.go
+++ b/cluster/manifest/load.go
@@ -28,7 +28,7 @@ func Load(manifestFile, legacyLockFile string, lockCallback func(cluster.Lock) e
 	switch {
 	case err1 == nil && err2 == nil:
 		// Both files loaded successfully, check if cluster hashes match
-		if !bytes.Equal(manifestCluster.InitialMutationHash, legacyCluster.LatestMutationHash) {
+		if !bytes.Equal(manifestCluster.InitialMutationHash, legacyCluster.InitialMutationHash) {
 			return nil, errors.New("manifest and legacy cluster hashes don't match",
 				z.Str("manifest_hash", hex.EncodeToString(manifestCluster.InitialMutationHash)),
 				z.Str("legacy_hash", hex.EncodeToString(legacyCluster.InitialMutationHash)))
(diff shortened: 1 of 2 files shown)


cluster/manifest/load.go
```diff
@@ -3,47 +3,75 @@
...
-		manifest := new(manifestpb.Cluster)
-		if err := proto.Unmarshal(b, manifest); err != nil {
-			return nil, errors.Wrap(err, "unmarshal cluster manifest")
+	manifestCluster, err1 := loadClusterManifest(manifestFile)
+	legacyCluster, err2 := loadLegacyLock(legacyLockFile, lockCallback)
+
+	switch {
+	case err1 == nil && err2 == nil:
+		// Both files loaded successfully, check if cluster hashes match
+		if !bytes.Equal(manifestCluster.InitialMutationHash, legacyCluster.LatestMutationHash) {
```
corverroos, line 31: rather use `InitialMutationHash` for both (like you do in the error)

cluster/manifest/load.go
```diff
@@ -3,47 +3,75 @@
...
+	b, err := os.ReadFile(manifestFile)
 	if err != nil {
-		return nil, errors.Wrap(err, "read legacy lock file")
+		return nil, errors.Wrap(err, "read manifest file")
 	}
 
-	m, err := loadLegacyLock(b, lockCallback)
-	if err != nil {
-		return nil, errors.Wrap(err, "load legacy lock")
+	manifest := new(manifestpb.Cluster)
```
corverroos, line 57: mmm, the `Cluster` proto isn't stored on disk, a `SignedMutationList` is the file on disk... seems like we lack proper testing of this load logic.

corverroos (changes requested):

cluster/manifest/load.go
xenowits: @corverroos this is fixed in https://github.com/ObolNetwork/charon/pull/2473

xenowits: Closing this in favour of https://github.com/ObolNetwork/charon/pull/2485

Issue #2375, opened 2023-06-29
Fuzz charon with malicious charon node


dB2510:
# 🎯 Problem to be solved
Currently we have simnet beacon mock fuzz which acts as a beacon node fuzzer by returning random responses for all the beacon node API calls.
As part of our fuzz testing efforts for charon, we need to implement a malicious charon node which should act as a charon fuzzer in a cluster. The malicious charon node should send random responses to its peers in a cluster over defined libp2p protocols.

# 🛠️ Proposed solution

- Add `charon_peer_fuzz` feature flag to enable charon peer fuzzing over p2p network in a cluster via CLI.
- Wrap existing p2p default writers in [`sendRecvOpts.writersByProtocol`](https://github.com/ObolNetwork/charon/blob/305a2e5353423bb492af09c2f4e2f735d82163e1/p2p/sender.go#L162) to write random messages when called by send functions in `p2p/fuzz.go` as something like:

```
type fuzzWriter struct {
	w pbio.Writer
}

func (fw *fuzzWriter) WriteMsg(msg proto.Message) error {
	cloneMsg := proto.Clone(msg)
	fuzz.New().Fuzz(cloneMsg)

	return fw.w.WriteMsg(cloneMsg)
}
```

- Add an `SendRecvOption` function as:
```
func WithFuzzWriter() func(*sendRecvOpts) {
	return func(opts *sendRecvOpts) {
		for pID := range opts.writersByProtocol {
			opts.writersByProtocol[pID] = func(s network.Stream) pbio.Writer {
				return &fuzzWriter{w: pbio.NewDelimitedWriter(s)}
			}
		}
	}
}
```

- Use `WithFuzzWriter` when `charon_peer_fuzz` feature flag is enabled.

# 🧪 Tests
- [ ] Tested by new automated compose tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2426, opened 2023-07-13
p2p: add peer to peer fuzzing support

dB2510:
Adds support for peer to peer fuzzing by adding a fuzz reader writer which intercepts all the libp2p messages and reads/writes fuzzed data.

category: feature
ticket: #2375

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -224,7 +224,7 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 	// Enable p2p fuzzing if --p2p-fuzz is set.
 	if conf.P2PFuzz {
-		p2p.EnableP2PFuzz()
+		p2p.SetFuzzerDefaultsUnsafe()
 	}
 
 	sender := new(p2p.Sender)
--- a/app/app.go
+++ b/app/app.go
@@ -88,7 +88,6 @@ type Config struct {
 	SyntheticBlockProposals bool
 	BuilderAPI              bool
 	SimnetBMockFuzz         bool
-	P2PFuzz                 bool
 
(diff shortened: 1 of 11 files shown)


p2p/sender.go
```diff
@@ -165,6 +172,29 @@ func WithDelimitedProtocol(pID protocol.ID) func(*sendRecvOpts) {
 	}
 }
 
+// WithFuzzReaderWriter returns an option that sets a fuzz reader writer to all the protocols if p2p fuzz is enabled.
+//
+// If p2p fuzz is enabled, this option sets a fuzz reader writer for each protocol in the provided sendRecvOpts.
+// The fuzz reader writer is responsible for creating a customized reader and writer for each network stream
+// associated with a specific protocol. The reader and writer implement the pbio.Reader and pbio.Writer interfaces,
```
xenowits, line 179:
```suggestion
// associated with a specific protocol. The reader and writer implement the pbio.Reader and pbio.Writer interfaces respectively
```

xenowits (approved):

app/app.go
```diff
@@ -88,7 +88,7 @@ type Config struct {
 	SyntheticBlockProposals bool
 	BuilderAPI              bool
 	SimnetBMockFuzz         bool
-	CharonP2PFuzz           bool
+	P2PFuzz                 bool
```
corverroos, line 91: shouldn't this be in the TestConfig?

corverroos (approved):

app/app.go
dB2510: yeah, moved this field to TestConfig

Issue #2400, opened 2023-07-05
Update vmock to submit selections at the start of epoch for current and next epoch


dB2510:
# 🎯 Problem to be solved
Currently we submit beacon committee selections and sync committee selections at the start of a slot but we want to test our assumptions around submitting beacon committee selections and sync committee selections at the start of an epoch for current and next epoch. This is related to the update that we are proposing in beacon API specs: https://github.com/ethereum/beacon-APIs/pull/334

# 🛠️ Proposed solution

Update validatormock to submit beacon committee selections and sync committee selections at the start of an epoch for current and next epoch [here](https://github.com/ObolNetwork/charon/blob/268107c53f0dda23376fe381b6065188b35a63d0/app/vmock.go#L39) and [here](https://github.com/ObolNetwork/charon/blob/268107c53f0dda23376fe381b6065188b35a63d0/app/vmock.go#L73) similar what we have when we prepare sync committee message at the start of an epoch [here](https://github.com/ObolNetwork/charon/blob/268107c53f0dda23376fe381b6065188b35a63d0/app/vmock.go#L44).
Test this change in core team cluster.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

dB2510:
closed by #2545

PR #2427, opened 2023-07-13
vmock: query attester duties for 2 epochs

xenowits:
Update validatormock to query attestations at the start of the current epoch for the current and next epochs .

This enables testing of selections endpoints by simulating behaviour of real validator clients which query attester duties for two epochs, current and next. They do this to calculate all subnets to subscribe for attester and aggregator duties.

category: feature 
ticket: #2400


corverroos (changes requested):
we need to rethink this approach, this feels very messy and very prone to bugs. This approach just throws everything into a single bag and hopes for the best.

xenowits:
Closing this in favour of a simpler approach - https://github.com/ObolNetwork/charon/pull/2431

PR #2430, opened 2023-07-14
cluster: align hash timestamp to unix

dB2510:
Aligns timestamp format of validator registration in hash calculation to unix format. This is done to avoid any confusion between timestamp in `cluster-lock.json` and the format used to calculate lock hash. Also it aligns the timestamp format with that of `go-eth2-client` types.

category: bug
ticket: #2436


cluster/ssz.go
```diff
@@ -695,7 +697,7 @@ func hashRegistration(r Registration, hh ssz.HashWalker) error {
 	hh.PutUint64(uint64(r.GasLimit))
 
 	// Field (2) 'Timestamp' Bytes48
-	hh.PutBytes([]byte(r.Timestamp.String()))
+	hh.PutBytes([]byte(r.Timestamp.Format(time.RFC3339)))
```
corverroos, line 700: please use eth2 standard of ssz marshalling, also please ensure this hash equals canonical ssz hashing of validator registration

corverroos (changes requested):

cluster/ssz.go
dB2510: Done 🫡

corverroos (approved):

PR #2431, opened 2023-07-14
vmock: query attester duties for 2 epochs

xenowits:
Update validatormock to query attestations at the start of the current epoch for the current and next epochs .

This enables testing of selections endpoints by simulating behaviour of real validator clients which query attester duties for two epochs, current and next. They do this to calculate all subnets to subscribe for attester and aggregator duties.

This is an alternate approach to this PR: https://github.com/ObolNetwork/charon/pull/2427.

category: feature 
ticket: #2400


testutil/validatormock/attest.go
```diff
@@ -26,6 +26,68 @@ type (
...
+	if err != nil {
+		return err
+	}
+
+	slotStart := uint64(a.epoch) * slotsPerEpoch
+	slotEnd := uint64(a.epoch+2) * slotsPerEpoch
+	// Prepare attesters for current and next epochs and store them in a map for each slot.
+	for slot := slotStart; slot < slotEnd; slot++ {
+		attester := NewSlotAttester(a.eth2Cl, eth2p0.Slot(slot), a.signFunc, a.pubkeys)
+		err := attester.Prepare(ctx)
```
corverroos, line 64: note this is going to block for a long time, especially for next epoch slots.

corverroos (changes requested):

testutil/validatormock/attest.go
corverroos: or rather, it must be called async, since it may block for along time.

xenowits:
Closing this for a different solution that involves refactoring the vmock quite a bit

PR #2432, opened 2023-07-14
app: add mutex to `onStartup`

xenowits:
Add mutex to `onStartup` variable in `wireValidatorMock()`. This fixes data race identified in this build - https://github.com/ObolNetwork/charon/actions/runs/5551674526/jobs/10138073971?pr=2431#logs. 

The data race arises from two goroutines trying to modify the value of `onStartup`:
1. First goroutine which handles slot X tries to read the value of `onStartup`
2. Second goroutine which handles slot Y tries to update the value to `false`

category: bug 
ticket: none

### DIFF
--- a/app/vmock.go
+++ b/app/vmock.go
@@ -43,7 +43,7 @@ func wireValidatorMock(conf Config, pubshares []eth2p0.BLSPubKey, sched core.Sch
 
 		return startingUp
 	}
-	setOnStartup := func() {
+	clearOnStartup := func() {
 		mu.Lock()
 		defer mu.Unlock()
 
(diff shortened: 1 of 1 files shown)


app/vmock.go
```diff
@@ -33,7 +33,10 @@ func wireValidatorMock(conf Config, pubshares []eth2p0.BLSPubKey, sched core.Sch
 		return err
 	}
 
-	onStartup := true
+	var (
+		mu        sync.Mutex
```
corverroos, line 37: please wrap accessing shared data in a function. You have a bug, you need to lock when writing AND reading.

corverroos (changes requested):

app/vmock.go
```diff
@@ -67,7 +83,7 @@ func wireValidatorMock(conf Config, pubshares []eth2p0.BLSPubKey, sched core.Sch
 			})
 		}
 
-		onStartup = false
+		setOnStartup()
```
corverroos, line 86: nit, rename to `clearOnStartup` or `startedUp`. `setOnStartup` sounds like you are setting "onStartup==true"

corverroos (approved):

Issue #2433, opened 2023-07-14
`parsigex` can't unmarshal blinded blocks


gsora:
![image](https://github.com/ObolNetwork/charon/assets/145909/932f3e91-071d-4d7b-a2a0-f277e0b93d6b)

# 🐞 Bug Report

### Description
Node `tense-gas` in the 3o4 cluster, as well as others albeit with smaller frequency, cannot unmarshal data from `parsigex` seemingly randomly.

Git hash for `tense-gas`: `ffe4966`.

PR #2434, opened 2023-07-14
parsigex: unmarshal data that is about to be broadcasted

gsora:
Also log full hex encoded byte arrays that fail to get unmarshaled in their original type.

Needed to troubleshoot https://github.com/ObolNetwork/charon/issues/2433.

category: misc
ticket: #2433

### DIFF
--- a/core/parsigex/parsigex.go
+++ b/core/parsigex/parsigex.go
@@ -105,11 +105,8 @@ func (m *ParSigEx) Broadcast(ctx context.Context, duty core.Duty, set core.ParSi
 	}
 
 	// TODO: remove this from prod once we know what's going on
-	for _, data := range pb.Set {
-		_, err = core.ParSignedDataFromProto(duty.Type, data)
-		if err != nil {
-			log.Warn(ctx, "ParSignedDataSet which was just marshaled can't be unmarshaled back", err)
-		}
+	if _, err = core.ParSignedDataSetFromProto(duty.Type, pb); err != nil {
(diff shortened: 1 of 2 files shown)


core/proto.go
```diff
@@ -251,13 +251,13 @@ func unmarshal(data []byte, v any) error {
 			return nil
 		} else if !bytes.HasPrefix(bytes.TrimSpace(data), []byte("{")) {
 			// No json prefix, so no point attempting json unmarshalling.
-			return errors.Wrap(err, "unmarshal ssz")
+			return errors.Wrap(err, "unmarshal ssz", z.Hex("data", data))
```
corverroos, line 254: add TODO to remove this before v0.17 release

core/parsigex/parsigex.go
```diff
@@ -104,6 +104,14 @@ func (m *ParSigEx) Broadcast(ctx context.Context, duty core.Duty, set core.ParSi
 		return err
 	}
 
+	// TODO: remove this from prod once we know what's going on
+	for _, data := range pb.Set {
+		_, err = core.ParSignedDataFromProto(duty.Type, data)
```
corverroos, line 109: what about doing this inside `ParSignedDataSetToProto`?

Also, rather please use `ParSignedDataSetFromProto`

gsora: > what about doing this inside ParSignedDataSetToProto?

I need to pass duty type to this function, I would rather use `core.ParSignedDataSetFromProto` as suggested.

corverroos (approved):

Issue #2438, opened 2023-07-15
new memory leak alert on v.0.17-dev


boulder225:
# 🐞 Bug Report

### Description
We have received a memory leak alert via PagerDuty for the most recent two versions of the main. At the moment, we have temporarily snoozed the alert. This memory leak may affect the system's performance and needs to be fixed urgently.

![image.png](https://images.zenhubusercontent.com/648c138feb5dc8a72ad3911a/5954b128-cbb1-4fe9-8ae6-5e9d2d8b33b9) 
![image.png](https://images.zenhubusercontent.com/648c138feb5dc8a72ad3911a/c12b1601-e8c1-4d3b-a4fc-9861537d9c11)
![image.png](https://images.zenhubusercontent.com/648c138feb5dc8a72ad3911a/b192e00e-43f3-402e-8ec0-2e03c59b938a)

### 🔬 Minimal Reproduction
1. Monitor the memory usage of the two most recent versions of the main.
2. Notice the abnormal increase in memory usage over time, indicating a potential memory leak.

### 🔥 Error
Memory usage continues to increase in an abnormal pattern when the most recent two versions of the main are running, causing the memory leak alert in PagerDuty.

### 🌍 Your Environment

**What version of Charon are you running? (Which release)**
- `v.0.17-dev`

**Anything else relevant (validator index / public key)?**
- 

### Additional Context
- This issue was first identified via a PagerDuty alert.
- Current workaround is to snooze the alert, but this is not a long-term solution.

### Resolution
- Immediate investigation and resolution are required to prevent any significant impact on the system performance.

corverroos:
Memory issue was due to bug in prometheus metric refactor that included duty slot as a label which has infinite cardinality.

PR #2440, opened 2023-07-17
app/health: zero discarded metric

gsora:
Zero out the last metric of c.metrics before reslicing it, marking it as ready to be garbage collected.

category: bug
ticket: #2438


corverroos (changes requested):
I'm pretty sure this isn't required

PR #2445, opened 2023-07-18
cmd: fix `create cluster` regression

### LINKED ISSUE #2444: Missing flags error on create cluster command
# 🎯 Problem to be solved
create cluster command throws an error about missing --nodes, --num-validators flags when --definition-file is passed.

Image

# 🛠️ Proposed solution

gsora:
After the `create cluster` refactor a couple PR's back, definition file was essentially ignored.

This PR restores that functionality back.

category: bug
ticket: #2444

Closes #2444

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -142,6 +142,7 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 			return err
 		}
 
+		// Needed if --split-existing-keys is called without a definition file.
 		conf.NumDVs = len(secrets)
 	}
 
@@ -152,13 +153,6 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		if err != nil {
 			return err
 		}
-
-		conf.FeeRecipientAddrs = def.FeeRecipientAddresses()
-		conf.WithdrawalAddrs = def.WithdrawalAddresses()
-		conf.Name = def.Name
-		conf.NumNodes = len(def.Operators)
-		conf.NumDVs = def.NumValidators
-		conf.Threshold = def.Threshold
 	} else { // Create new definition from cluster config
 		def, err = newDefFromConfig(ctx, conf)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/createcluster.go
```diff
@@ -152,22 +147,38 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		if err != nil {
 			return err
 		}
+
+		defWAddrs := def.WithdrawalAddresses()
+		defFAddrs := def.FeeRecipientAddresses()
+
+		if len(defFAddrs) != 0 {
```
corverroos, line 154: I'd stick to "definition overrides all other configuration flags". So It doesn't state "non-zero values in the definition overrides any provided in other flags."

corverroos (changes requested):

cmd/createcluster.go
```diff
@@ -152,22 +147,30 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		if err != nil {
 			return err
 		}
+
+		conf.FeeRecipientAddrs = def.FeeRecipientAddresses()
+		conf.WithdrawalAddrs = def.WithdrawalAddresses()
+
+		if strings.TrimSpace(def.Name) != "" {
+			conf.Name = def.Name
+		}
```
corverroos, line 156: why this conditional?

cmd/createcluster.go
```diff
@@ -152,22 +147,30 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		if err != nil {
 			return err
 		}
+
+		conf.FeeRecipientAddrs = def.FeeRecipientAddresses()
```
corverroos, line 151: it feels like definition is the source of truth from this point, why is it required to update the config? The else statement creates a definition from the config, feels like we should just use definition from this point forward

cmd/createcluster.go
gsora: removing since as you suggested, definition is source of truth

corverroos (approved):

Issue #2439, opened 2023-07-17
goroutine leak in `charon-perf-1`


dB2510:
# 🐞 Bug Report

### Description

We have observed a goroutine leak in `charon-perf-1` cluster for peers `faithful-sports` and `cute-cycle`.

## 🔬 Minimal Reproduction

Run a 3 of 4 cluster with commit hash = `79116f9` and with 2 nodes down.

## 🔥 Error

Here's pprof profile of goroutine of `faithful-sports`:
[charon-perf-1-faithful-sports-goroutine.gz](https://github.com/ObolNetwork/charon/files/12067201/charon-perf-1-faithful-sports-goroutine.gz)

Check the above profile with the following command:
```
go tool pprof -http=: charon-perf-1-faithful-sports-goroutine.gz
```

**What version of Charon are you running? (Which release)**

```
79116f9
```

corverroos:
Seems to be due to gorougtines getting stuck in consensus "waiting for running consensus to exit", this is probably because it already exitted and the Propose method is being called again.

```
goroutine profile: total 758
454 @ 0x454156 0x41f11d 0x41ec18 0x10ea5f7 0x10ea272 0x1034f43 0x102fd54 0x1023e78 0x103652a 0x4886e1
#	0x10ea5f6	github.com/obolnetwork/charon/core/consensus.(*Component).propose+0x296	/app/charon/core/consensus/component.go:318
#	0x10ea271	github.com/obolnetwork/charon/core/consensus.(*Component).Propose+0x71	/app/charon/core/consensus/component.go:271
#	0x1034f42	github.com/obolnetwork/charon/core.WithTracing.func1.3+0x102		/app/charon/core/tracing.go:72
#	0x102fd53	github.com/obolnetwork/charon/core.WithTracking.func1.2+0x53		/app/charon/core/tracking.go:24
#	0x1023e77	github.com/obolnetwork/charon/core.WithAsyncRetry.func1.3.1+0x37	/app/charon/core/retry.go:32
#	0x1036529	github.com/obolnetwork/charon/app/retry.(*Retryer[...]).DoAsync+0x8c9	/app/charon/app/retry/retry.go:118
```

dB2510:
Here's the complete analysis and fix of the bug: https://docs.google.com/document/d/16Y2TSx1IsAMJMxVwNQKCIMb0MIuuSWMWDvParJZ1I6g/edit

dB2510:
closed by #2452

PR #2448, opened 2023-07-18
core/consensus: error on subsequent calls

corverroos:
Error when either `Participate` or `Propose` is called multiple times for the same duty since this results in goroutines hanging forever. 

Also removes debug logs.

category: refactor
ticket: #2439

### DIFF
--- a/core/consensus/component.go
+++ b/core/consensus/component.go
@@ -129,22 +129,52 @@ func newDefinition(nodes int, subs func() []subscriber, roundTimer roundTimer,
 // newInstanceIO returns a new instanceIO.
 func newInstanceIO() instanceIO {
 	return instanceIO{
-		recvBuffer:  make(chan msg, recvBuffer),
-		hashCh:      make(chan [32]byte, 1),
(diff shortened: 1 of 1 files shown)


core/consensus/component.go
```diff
@@ -129,22 +129,52 @@ func newDefinition(nodes int, subs func() []subscriber, roundTimer roundTimer,
...
+	proposed     chan struct{}      // Closed when Propose was called for this instance.
+	recvBuffer   chan msg           // Outer receive buffers.
+	hashCh       chan [32]byte      // Async input hash channel.
+	valueCh      chan proto.Message // Async input value channel.
+	errCh        chan error         // Async output error channel.
+	decidedAtCh  chan time.Time     // Async output decided timestamp channel.
+}
+
+// MarkParticipated marks the instance as participated, it returns
+// and error if the instance was already marked as participated.
```
xenowits, line 155:
```suggestion
// MarkParticipated marks the instance as participated.
// It returns an error if the instance was already marked as participated.
```

xenowits (approved):

PR #2452, opened 2023-07-19
core/consensus: add ShouldRun method

dB2510:
Adds a `ShouldRun` method to `instanceIO` to check if the instance was actually "running" rather than returning boolean from `getInstanceIO`.

category: bug
ticket: #2439

### DIFF
--- a/core/consensus/component.go
+++ b/core/consensus/component.go
@@ -131,6 +131,7 @@ func newInstanceIO() instanceIO {
 	return instanceIO{
 		participated: make(chan struct{}),
 		proposed:     make(chan struct{}),
+		running:      make(chan struct{}),
 		recvBuffer:   make(chan msg, recvBuffer),
 		hashCh:       make(chan [32]byte, 1),
 		valueCh:      make(chan proto.Message, 1),
@@ -144,6 +145,7 @@ func newInstanceIO() instanceIO {
 type instanceIO struct {
 	participated chan struct{}      // Closed when Participate was called for this instance.
 	proposed     chan struct{}      // Closed when Propose was called for this instance.
+	running      chan struct{}      // Closed when runInstance was already called.
 	recvBuffer   chan msg           // Outer receive buffers.
 	hashCh       chan [32]byte      // Async input hash channel.
 	valueCh      chan proto.Message // Async input value channel.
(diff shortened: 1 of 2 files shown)


corverroos (approved):

corverroos:
A test that calls the handle function with some data before Propose/Participate would be great

core/consensus/component.go
```diff
@@ -177,6 +179,19 @@ func (io *instanceIO) MarkProposed() error {
 	return nil
 }
 
+// ShouldRun marks the instance as running.
```
xenowits, line 182:
```suggestion
// ShouldRun returns true and marks the instance as running if the instance was not already running.
```

core/consensus/component.go
```diff
@@ -177,6 +179,19 @@ func (io *instanceIO) MarkProposed() error {
...
+// It returns false if the instance was already marked as running.
```
xenowits, line 183:
```suggestion
// ShouldRun checks the current status of the instance.
// If the instance is not already running, it returns true and marks the instance as running.
// If the instance is already running, it returns false.
```

xenowits (approved):

core/consensus/component_internal_test.go
```diff
@@ -454,6 +454,73 @@ func TestComponentHandle(t *testing.T) {
...
+
+	t.Run("ShouldRun after handle", func(t *testing.T) {
+		var c Component
+		c.deadliner = testDeadliner{}
+		c.mutable.instances = make(map[core.Duty]instanceIO)
+
+		msg := &pbv1.ConsensusMsg{
+			Msg: randomMsg(t),
+		}
+		p2pKey := testutil.GenerateInsecureK1Key(t, 0)
```
xenowits, line 472: nit: add newline

core/consensus/component_internal_test.go
```diff
@@ -454,6 +454,73 @@ func TestComponentHandle(t *testing.T) {
...
+		c.pubkeys = make(map[int64]*k1.PublicKey)
+		c.pubkeys[0] = p2pKey.PubKey()
+
+		duty := core.Duty{Slot: 42, Type: 1}
+		msg = signConsensusMsg(t, msg, p2pKey, duty)
+
+		// It should create new instance of instanceIO for the give duty.
```
xenowits, line 479:
```suggestion
		// It should create new instance of instanceIO for the given duty.
```

core/consensus/component_internal_test.go
```diff
@@ -454,6 +454,73 @@ func TestComponentHandle(t *testing.T) {
...
+		_, _, err := c.handle(ctx, "peerID", msg)
+		require.NoError(t, err)
+
+		pubkey := testutil.RandomCorePubKey(t)
+
+		// It should mark instance as running by calling inst.ShouldRun().
```
xenowits, line 513:
```suggestion
		// Propose should internally mark instance as running by calling inst.ShouldRun().
```

xenowits (approved):

PR #2454, opened 2023-07-20
*: bump golangci-lint-action

xenowits:
Bump `golangci-lint-action` to `v3` and `golangci-lint` version to `v1.52.2`.

category: misc
ticket: none

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -77,7 +77,7 @@ linters-settings:
       - name: comment-spacings
         disabled: true
       - name: use-any
-        disabled: true
+        disabled: false
       # Some configured revive rules
       - name: unhandled-error
         arguments:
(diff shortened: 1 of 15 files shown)


.golangci.yml
```diff
@@ -74,6 +74,10 @@ linters-settings:
         disabled: true
       - name: cognitive-complexity
         disabled: true
+      - name: comment-spacings
+        disabled: true
+      - name: use-any
```
xenowits, line 79: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#use-any

.golangci.yml
```diff
@@ -74,6 +74,10 @@ linters-settings:
         disabled: true
       - name: cognitive-complexity
         disabled: true
+      - name: comment-spacings
```
xenowits, line 77: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#comment-spacings

corverroos: seems like you need a few pragmas: `go:generate and nolint`?

.golangci.yml
corverroos: sounds like a good rule, why can't we enable this?

xenowits: enabled this in latest changes 👍

.golangci.yml
xenowits: yeah, these pragmas are needed, but this linter fails in code blocks like https://github.com/ObolNetwork/charon/blob/main/app/errors/errors.go#L32-L47. I think this is a bug, so i chose to disable it
Screenshot 2023-07-24 at 13 50 30

corverroos: ok cool, add a comment that it doesn't support the latest go spec comments when disabling it.

corverroos (approved):

PR #2462, opened 2023-07-23
eth2util: clone go-eth2-client data version

corverroos:
Clones the go-eth2-client `github.com/attestantio/go-eth2-client/spec.DataVersion` type aligning with pre-v0.18 values. 

This is required since we serialise these version numbers over the wire so they MUST remain constant to ensure backwards compatibility between charon versions. go-eth2-client changed the `DataVersion` numbers in the recent v0.18.0 upgrade which we had to roll-back. 

This change allows the following:
- Can upgrade to go-eth2-client to v0.18.0
- Can start a migration to using "data version" strings instead of numbers.

The migration will look something like:
- Introduce a new "string" based data version format, versus the old number based format.
- v0.17 should support receiving both old and new message formats (incoming)
- v0.17 should send old messages (outgoing) to remain compatible with v0.16.
- v0.18 can then start sending new format (outgoing), it still supports receiving both old and new (incoming).
- v0.19 can remove support for old format message, only receiving and sending the new format

category: refactor
ticket: #2433


core/ssz.go
```diff
@@ -42,7 +43,7 @@ func (b VersionedSignedBeaconBlock) MarshalSSZ() ([]byte, error) {
 
 // MarshalSSZTo ssz marshals the VersionedSignedBeaconBlock object to a target array.
 func (b VersionedSignedBeaconBlock) MarshalSSZTo(buf []byte) ([]byte, error) {
-	return marshalSSZVersionedTo(buf, b.Version, b.sszValFromVersion)
+	return marshalSSZVersionedTo(buf, eth2util.DataVersion(b.Version), b.sszValFromVersion)
```
corverroos, line 46: note that casting isn't enough, we need to ensure the numbers we are working with are aligned with pre-v0.18.0

xenowits (approved):

core/ssz.go
corverroos: this has been refactored to convert instead of cast.

eth2util/types.go
```diff
@@ -6,12 +6,147 @@ import (
...
+	return errors.New("unknown data version")
+}
+
+// ToUint64 returns the integer value used by go-eth2-client pre-v0.18.
+func (v DataVersion) ToUint64() uint64 {
+	return uint64(legacyDataVersionValues[v])
+}
+
+// ToETH2 returns a eth2spec.DataVersion equivalent to the DataVersion.
+func (v DataVersion) ToETH2() eth2spec.DataVersion {
```
dB2510, line 94: This method will return different values once we update go-eth2-client to v0.18, right?

corverroos: yup

PR #2468, opened 2023-07-25
cluster/manifest: refactor to 65 byte k1 sigs

corverroos:
Refactor manifest K1 sigs to 65 bytes to align with Ethereum standard.

Also improve cluster lock signature verification.

category: refactor
ticket: none


cluster/lock.go
```diff
@@ -182,46 +182,68 @@ func (l Lock) VerifySignatures() error {
...
-			}
+	return l.verifyNodeSignatures()
+}
 
-			sig := l.NodeSignatures[idx]
-			verified, err := k1util.Verify(record.PubKey, l.LockHash, sig[:len(sig)-1])
-			if err != nil {
-				return errors.Wrap(err, "operator node signature check")
-			}
+// verifyNodeSignatures returns true an error if the node signatures field is not correctly
```
dB2510, line 191:
```suggestion
// verifyNodeSignatures returns an error if the node signatures field is not correctly
```

cluster/lock.go
```diff
@@ -182,46 +182,68 @@ func (l Lock) VerifySignatures() error {
...
+				z.Int("peer_index", idx),
+			)
 		}
 	}
 
 	return nil
 }
 
-// verifyBuilderRegistrations returns an error if populated builder registrations from json are invalid.
+// verifyBuilderRegistrations returns an error if the populated builder registrations are invalid.
```
dB2510, line 227:
```suggestion
// verifyBuilderRegistrations returns an error if the builder registrations are not correctly 
// populated or otherwise invalid.
```

dB2510 (approved):

xenowits (approved):

dkg/dkg.go
```diff
@@ -106,6 +106,11 @@ func Run(ctx context.Context, conf Config) (err error) {
 		return err
 	}
 
+	// This DKG only supports a few specific config versions.
+	if def.Version != "v1.6.0" && def.Version != "v1.7.0" {
+		return errors.New("only v1.6.0 and v1.7.0 cluster definition version supported")
+	}
```
dB2510, line 112: why do we need to enforce this?

corverroos: because it fails for earlier versions

dkg/dkg_test.go
```diff
@@ -313,11 +326,16 @@ func verifyDKGResults(t *testing.T, def cluster.Definition, dir string) {
 		require.NoError(t, lock.VerifySignatures())
 		locks = append(locks, lock)
 
-		for _, val := range lock.Validators {
+		for j, val := range lock.Validators {
 			// Assert Deposit Data
 			require.EqualValues(t, val.PubKey, val.DepositData.PubKey)
 			require.EqualValues(t, 32_000_000_000, val.DepositData.Amount)
 
+			if !cluster.SupportPregenRegistrations(lock.Version) {
```
dB2510, line 334: can we have similar verification for node signatures as well?

corverroos: yes I added that

PR #2475, opened 2023-07-26
log/loki: limit line size

gsora:
This PR limits the maximum size of each single log line to 4096 bytes in production workloads (i.e. for consumers of `loki.New()`).

This prevents pushing log lines too long for our instance of Loki, avoiding potential DoS.

As with `batchMax`, `maxLogLineLen` is not configurable except in tests.

category: misc
ticket: none


app/log/loki/client.go
```diff
@@ -101,10 +104,15 @@ func (c *Client) Run() {
 	for {
 		select {
 		case line := <-c.input:
+			if len(line) > c.maxLogLineLen && c.maxLogLineLen != 0 {
+				continue // Silently drop log line longer than maxLogLineLen if set
```
corverroos, line 108: not sure about silent, what about using a log filter?

gsora: I'm doing what `batch` does pretty much, we discussed this briefly on Slack and you suggested to just drop it.

What behavior would you like to see instead?

corverroos: well, a log it using a log filter

but sure, this is only loki client itself, which is not our primary logging framework, it is best effort, so yeah, maybe silent drop is fine.

PR #2496, opened 2023-08-01
dkg: exchange sigTypes concurrently

### LINKED ISSUE #2491: Possible race condition in DKG exchanger
# 🐞 Bug Report

### Description

The DKG uses the `exchanger` component for:
```
exhanges partial signatures of lockhash/deposit-data among dkg participants and returns all the partial
BLS signatures of the cluster according to public key of each vallidator.
```

It used to exchange the following signatures:
- `signAndAggDepositData`
- `signAndAggValidatorRegistrations`
- `signAndAggLockHash`

Note that timing in distributed systems can be unintuitive and can follow unexpected sequences.

The code requires all three above "types" to be executed in sequence on each peer. We drop unexpected duty types silently, which could happen in large geo-distributed clusters where Q nodes are close to each other, they might already be done with all three while one need is far away and might receive quorum messages out of order. Silently dropping that and getting stuck.
```
		case peerSet := <-e.sigChan:
			if sigType != peerSet.sigType {
				// Do nothing if duty doesn't match
				continue
			}
			sets[peerSet.pubkey] = peerSet.psigs
		}
```

### Proposed Solution

Refactor exchanger to something like DutyDB with a query API allowing out of order queries and responses.

Oh, and this is another good example of improper error handling. We should have error edabove instead of silently dropping it.

gsora:
Before this commit, calling `exchange()` would listen for new incoming messages and discard everything that didn't carry the expected `sigType`.

In a DKG one node might fall behind the others (for example, while DKG'ing 3000 validator keys) and send `sigTypes` that the other peers have already received threshold for.

This PR makes sure a given instance of `exchanger` will hold onto all messages whose `sigTypes` are specified during its initialization phase.

Calling `exchange()` will return as soon as there are enough messages for the specified `sigType`, but different `sigType`s messages will still be stored and will be returned when requested by means of another `exchange()` call, with a different `sigType` argument.

category: bug
ticket: #2491

Closes #2491.

### DIFF
--- a/dkg/exchanger.go
+++ b/dkg/exchanger.go
@@ -11,7 +11,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 
 	"github.com/obolnetwork/charon/app/errors"
-	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/app/z"
 	"github.com/obolnetwork/charon/core"
 	"github.com/obolnetwork/charon/core/parsigdb"
@@ -75,16 +74,8 @@ func (stb *dataByPubkey) get(sigType sigType) (map[core.PubKey][]core.ParSignedD
 	return ret, ok
 }
 
-// sigData includes the fields obtained from sigdb when threshold is reached.
-type sigData struct {
-	sigType sigType
-	pubkey  core.PubKey
-	psigs   []core.ParSignedData
-}
-
 // exchanger is responsible for exchanging partial signatures between peers on libp2p.
 type exchanger struct {
-	sigChan  chan sigData
 	sigex    *parsigex.ParSigEx
 	sigdb    *parsigdb.MemDB
 	numVals  int
@@ -92,7 +83,7 @@ type exchanger struct {
 	sigData  dataByPubkey
 }
 
-func newExchanger(ctx context.Context, tcpNode host.Host, peerIdx int, peers []peer.ID, vals int, sigTypes []sigType) *exchanger {
+func newExchanger(tcpNode host.Host, peerIdx int, peers []peer.ID, vals int, sigTypes []sigType) *exchanger {
 	// Partial signature roots not known yet, so skip verification in parsigex, rather verify before we aggregate.
 	noopVerifier := func(ctx context.Context, duty core.Duty, key core.PubKey, data core.ParSignedData) error {
 		return nil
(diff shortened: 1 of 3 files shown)


dkg/exchanger.go
```diff
@@ -29,6 +34,46 @@ const (
...
+
+	stb.store[sigType][pubKey] = data
+}
+
+// get gets all the core.ParSignedData for a given core.PubKey.
+func (stb *dataByPubkey) get(sigType sigType) map[core.PubKey][]core.ParSignedData {
+	stb.lock.Lock()
+	defer stb.lock.Unlock()
+
+	if len(stb.store[sigType]) == 0 {
```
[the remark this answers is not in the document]
corverroos: optional nit: bool checks are slightly more readable than nil checks

dkg/exchanger.go
```diff
@@ -29,6 +34,46 @@ const (
...
+type sigTypeStore map[sigType]map[core.PubKey][]core.ParSignedData
+
+// dataByPubkey maps a sigType to its map of public key to slice of core.ParSignedData..
+type dataByPubkey struct {
+	store sigTypeStore
+	lock  sync.Mutex
+}
+
+// set sets data for the given sigType and core.PubKey.
+func (stb *dataByPubkey) set(pubKey core.PubKey, sigType sigType, data []core.ParSignedData) {
```
corverroos, line 47: can refactor this to a single `sigdatas` parameter

dkg/exchanger.go
```diff
@@ -29,6 +34,46 @@ const (
 	sigValidatorRegistration sigType = 103
 )
 
+// sigTypeStore is a shorthand for a map of sigType to map of core.PubKey to slice of core.ParSignedData.
+type sigTypeStore map[sigType]map[core.PubKey][]core.ParSignedData
+
+// dataByPubkey maps a sigType to its map of public key to slice of core.ParSignedData..
+type dataByPubkey struct {
```
corverroos, line 41: personal style: I would not introduce more types. I would just put the lock  and a `map[sigType]sigData` in `exchanger` itself. 

func (e exchanger) set(sigdata) {...}
func (e exchanger) get(sigtype) (sigdata, bool) {...}

But that is just me #lessismore

corverroos (approved):

corverroos:
LGTM

PR #2504, opened 2023-08-02
testutil/compose: integrate p2p fuzz to compose

dB2510:
Integrates p2p fuzz to compose.

category: test
ticket: #2375

### DIFF
--- a/testutil/compose/fuzz/fuzz_test.go
+++ b/testutil/compose/fuzz/fuzz_test.go
@@ -17,18 +17,16 @@ import (
 	"github.com/obolnetwork/charon/testutil/compose"
 )
 
-//go:generate go test . -run=TestBeaconFuzz -integration -v
-
 var (
-	fuzz        = flag.Bool("fuzz", false, "Enable docker based integration test")
+	fuzzer      = flag.Bool("fuzzer", false, "Enables docker based fuzzer tests")
 	sudoPerms   = flag.Bool("sudo-perms", false, "Enables changing all compose artefacts file permissions using sudo.")
 	logDir      = flag.String("log-dir", "", "Specifies the directory to store test docker-compose logs. Empty defaults to stdout.")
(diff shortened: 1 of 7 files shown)


testutil/compose/fuzz/fuzz_test.go
```diff
@@ -58,3 +58,35 @@ func TestBeaconFuzz(t *testing.T) {
...
+		t.Skip("Skipping p2p fuzz integration test")
+	}
+
+	dir, err := os.MkdirTemp("", "")
+	require.NoError(t, err)
+
+	conf := compose.NewDefaultConfig()
+	conf.DisableMonitoringPorts = true
+	conf.BuildLocal = true
+	conf.P2PFuzz = true
```
corverroos, line 73: seems like all nodes are fuzzing so all duties will fail.

testutil/compose/run.go
```diff
@@ -41,10 +41,16 @@ func Run(ctx context.Context, dir string, conf Config) (TmplData, error) {
 		nodes = append(nodes, n)
 	}
 
+	charonCmd := cmdRun
+	if conf.P2PFuzz {
+		nodes[0].EnvVars = append(nodes[0].EnvVars, kv{"p2p-fuzz", fmt.Sprintf(`"%v"`, conf.P2PFuzz)})
+		charonCmd = cmdUnsafeRun
+	}
```
dB2510, line 48: @corverroos only node0 has `p2p-fuzz` enabled.

testutil/compose/fuzz/fuzz_test.go
dB2510: this config field is introduced to enable p2p fuzz in "one" of the nodes in compose cluster.

testutil/compose/run.go
corverroos: ah cool

corverroos (approved):

corverroos (approved):

PR #2510, opened 2023-08-03
dkg: wait for all validator key sigs

gsora:
Wait until all peer's validator signatures over exchanger-handled data.

This should fix nightly test.

category: bug
ticket: none

### DIFF
--- a/dkg/exchanger.go
+++ b/dkg/exchanger.go
@@ -25,9 +25,9 @@ import (
 type sigType int
 
 const (
-	// dutyLock is responsible for lock hash signed partial signatures exchange and aggregation.
+	// sigLock is responsible for lock hash signed partial signatures exchange and aggregation.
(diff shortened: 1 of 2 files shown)


dkg/exchanger.go
```diff
@@ -65,6 +66,10 @@ func (stb *dataByPubkey) get(sigType sigType) (map[core.PubKey][]core.ParSignedD
 		return nil, ok
 	}
 
+	if len(data) != stb.numVals {
```
corverroos, line 69: I thought the data is only set once and then it includes all the validators? I don't understand why this check is required or why this works?

corverroos, line 69: feels like we need a test for this.

corverroos (approved):

corverroos (approved):

PR #2514, opened 2023-08-04
dkg: lockstep success message

gsora:
Make every peer participating in the DKG wait for the "slow" one, and synchronize the `Successfully completed DKG ceremony 🎉` message output.

category: misc
ticket: none

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -345,12 +345,6 @@ func Run(ctx context.Context, conf Config) (err error) {
 	}
 	log.Debug(ctx, "Saved deposit data file to disk")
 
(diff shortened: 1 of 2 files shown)


dkg/dkg.go
```diff
@@ -355,8 +351,17 @@ func Run(ctx context.Context, conf Config) (err error) {
 	log.Debug(ctx, "Graceful shutdown delay", z.Int("seconds", int(conf.ShutdownDelay.Seconds())))
```
corverroos, line 351: I think this shutdown delay needs to be after the call to `stopsync`

corverroos (approved):

corverroos (approved):

Issue #2435, opened 2023-07-14
Soft shutdown not working


corverroos:
# 🐞 Bug Report

### Description

When stopping the charon docker container it doesn't always do a proper clean soft shutdown. Docker needs to send a sig kill command after 10s for the container to exit. 

This is probably related to long running vapi proxy event requests that are not force closed on context cancel.

### Proposed Solution

Refactor the proxy in vapi to cancel proxied requests when the application context is closed.

PR #2516, opened 2023-08-06
core/validatorapi: close proxy requests on shutdown

corverroos:
Fixes soft shutdown by cancelling long running proxied requests.

This is because long lived connections are not interrupted by `http.Server.Shutdown`:
```
// Shutdown gracefully shuts down the server without interrupting any
// active connections. Shutdown works by first closing all open
// listeners, then closing all idle connections, and then waiting
// indefinitely for connections to return to idle and then shut down...
```

category: bug
ticket: #2435

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -80,7 +80,7 @@ type Handler interface {
 // NewRouter returns a new validator http server router. The http router
 // translates http requests related to the distributed validator to the Handler.
 // All other requests are reverse-proxied to the beacon-node address.
-func NewRouter(h Handler, eth2Cl eth2wrap.Client) (*mux.Router, error) {
+func NewRouter(ctx context.Context, h Handler, eth2Cl eth2wrap.Client) (*mux.Router, error) {
 	// Register subset of distributed validator related endpoints.
 	endpoints := []struct {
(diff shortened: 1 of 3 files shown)


core/validatorapi/router.go
```diff
@@ -908,30 +913,13 @@ func proxyHandler(eth2Cl eth2wrap.Client) http.HandlerFunc {
...
-		proxy.ServeHTTP(proxyResponseWriter{w.(writeFlusher)}, r)
+		proxy.ServeHTTP(proxyResponseWriter{w.(writeFlusher)}, r.Clone(ctx))
 	}
 }
 
 // getBeaconNodeAddress returns an active beacon node proxy target address.
-func getBeaconNodeAddress(ctx context.Context, eth2Cl eth2wrap.Client) (*url.URL, error) {
+func getBeaconNodeAddress(eth2Cl addressProvider) (*url.URL, error) {
 	addr := eth2Cl.Address()
-	if addr == "none" {
```
corverroos, line 918: this logic was only applicable to go-eth2-client's "multi" package, which we don't use anymore

dB2510 (approved):

gsora (approved):

Issue #2423, opened 2023-07-12
Implement Launchpad URL Return for `--publish` Command


boulder225:
## 🎯 Problem to be solved
The `--publish` command, used at the end of the Distributed Key Generation (DKG) ceremony, currently does not provide a user-friendly way to access the new cluster dashboard. Users have to manually locate and navigate to the dashboard, which may be inconvenient and challenging for those unfamiliar with the system.

## 🛠️ Proposed solution
To improve user experience, the `--publish` command should be enhanced to return a launchpad URL directly pointing to the new cluster dashboard. This means that upon completing the DKG ceremony, users will receive an immediate link to their cluster dashboard, simplifying the process and improving usability. This solution will require modifications to the `--publish` command's functionality, specifically in the section dealing with the DKG ceremony's conclusion. The implementation should ensure the generated URL is accurate and functional, leading the user to the correct cluster dashboard.

corverroos:
Cool, but given a `--publish-address` (default value is `https://api.obol.tech`) what should the url be we print in the terminal?

boulder225:
@Battenfield

boulder225:
`{base_url_endpoint}/clusters/details/?lockHash={lock_hash}` @f1lander

PR #2518, opened 2023-08-07
app/obolapi: create cluster launchpad URL

gsora:
Add the `LaunchpadURLForLock` method, which generates a Launchpad cluster dashboard URL for a given cluster lock.

Reference this method in `dkg` and `create cluster` commands so that at the end of those processes, if the user has specified `--publish`, the Launchpad dashboard URL will be printed on the console.

category: feature
ticket: #2423 

Closes https://github.com/ObolNetwork/charon/issues/2423.


gsora:
Solving merge conflicts rn.

gsora:
Done, ready for review.

app/obolapi/api.go
```diff
@@ -15,46 +16,68 @@ import (
...
-	addr, err := url.JoinPath(c.baseURL, "lock")
-	if err != nil {
-		return errors.Wrap(err, "invalid address")
-	}
-
-	url, err := url.Parse(addr)
-	if err != nil {
-		return errors.Wrap(err, "invalid endpoint")
-	}
+	addr := *c.baseURL
```
corverroos, line 49: I would keep baseURL a string since this pointer trick is easy to miss when adding more functions which can result in bugs.

gsora:
Screenshot 2023-08-07 at 11 35 41

For reference, this is how it appears to users when a DKG with `--publish` completes.

/cc @Battenfield

app/obolapi/api.go
```diff
@@ -15,46 +16,68 @@ import (
...
+// LaunchpadURLForLock returns the Launchpad cluster dashboard page for a given lock, on the given
+// Obol API client.
+func (c Client) LaunchpadURLForLock(lock cluster.Lock) string {
+	lURL := *c.baseURL
+
+	lURL.Path = launchpadReturnPath
+
+	qs := url.Values{}
+
+	qs.Set(lockQueryStr, hex.EncodeToString(lock.LockHash))
```
corverroos, line 74: the hash might be 0x prefixed... not sure

cmd/createcluster.go
```diff
@@ -944,16 +957,19 @@ func randomHex64() (string, error) {
 }
 
 // writeLockToAPI posts the lock file to obol-api.
```
corverroos, line 959: update godoc

corverroos (approved):

app/obolapi/api.go
gsora: Yeah I tagged the launchpad in the release thread to sort this kind of stuff out, I'm not going to merge until we have everything sorted.

app/obolapi/api.go
gsora: What do you think about this? Validate url in `New()`, and provide a `url()` method that returns it as `*url.URL`, panic-ing if parsing fails: since validation happens in `New()`, if it fails later it's a very weird situation:

```go
// New returns a new Client.
func New(urlStr string) (Client, error) {
	_, err := url.ParseRequestURI(urlStr) // check that urlStr is valid
	if err != nil {
		return Client{}, errors.Wrap(err, "could not parse Obol API URL")
	}

	return Client{
		baseURL: urlStr,
	}, nil
}

// Client is the REST client for obol-api requests.
type Client struct {
	baseURL string // Base obol-api URL
}

func (c Client) url() *url.URL {
	baseURL, err := url.ParseRequestURI(c.baseURL)
	if err != nil {
		panic(errors.Wrap(err, "could not parse Obol API URL, this should never happen"))
	}

	return baseURL
}
```

corverroos: sure, could also just return the error from `url()`, don't think the effort to omit errors is super valuable.

PR #2532, opened 2023-08-10
app/version: bump to v0.17-rc2

gsora:
Bumps version to v0.17-rc2

category: misc
ticket: none

### DIFF
--- a/app/version/version.go
+++ b/app/version/version.go
@@ -15,7 +15,7 @@ import (
 )
 
 // version a string since it is overwritten at build-time with the git tag for official releases.
-var version = "v0.17-rc"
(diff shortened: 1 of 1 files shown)


app/version/version.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // version a string since it is overwritten at build-time with the git tag for official releases.
-var version = "v0.17-rc"
+var version = "v0.17-rc2"
```
xenowits, line 18: For release candidates, we don't bump the `version` string. The `rc2` candidate is actually `git tagged`

corverroos (changes requested):

Issue #2536, opened 2023-08-11
Ensure the desired folder exists, before writing the combined keys


boulder225:
## 🎯 Problem to be solved
When the `combine` command is executed, it writes new combined keys to disk. However, if the target folder does not exist (which is typically the default scenario), the command results in an error.

## 🛠️ Proposed solution
Before writing the combined keys, the system should ensure that the desired folder exists. If it does not, the folder should be created to allow the `combine` command to execute without issues.

## Incidence
The exact number of occurrences is unknown since we don't have logs specifically for this command. However, no users have raised complaints. This might be because the combine command is seldom used or possibly not by most of our user base.

## Impact
Essentially, for the default case, the combine command is broken. This could deter new or existing users who want to utilize this functionality.

PR #2537, opened 2023-08-14
cmd/combine: ensure output dir exists

corverroos:
Ensures the output folder exists before writing combined keys to it.

category: bug
ticket: #2536

### DIFF
--- a/eth2util/keystore/keystore.go
+++ b/eth2util/keystore/keystore.go
@@ -52,11 +52,17 @@ func StoreKeysInsecure(secrets []tbls.PrivateKey, dir string, _ confirmInsecure)
 
 // StoreKeys stores the secrets in dir/keystore-%d.json EIP 2335 Keystore files
 // with new random passwords stored in dir/Keystore-%d.txt.
+//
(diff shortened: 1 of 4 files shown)


eth2util/keystore/keystore.go
```diff
@@ -52,11 +52,17 @@ func StoreKeysInsecure(secrets []tbls.PrivateKey, dir string, _ confirmInsecure)
 
 // StoreKeys stores the secrets in dir/keystore-%d.json EIP 2335 Keystore files
 // with new random passwords stored in dir/Keystore-%d.txt.
+//
+// Note it doesn't ensure the folder dir exists.
```
dB2510, line 56: It does, right? when it calls `checkDir` in `storeKeysInternal` ?

dB2510 (approved):

eth2util/keystore/keystore.go
corverroos: "ensure" doesn't mean "check", it means "create". It doesn't create the folder if it doesn't exist.

dB2510: cool

Issue #2546, opened 2023-08-15
Race in AggSigDB unit test


corverroos:
# 🐞 Bug Report

### Description

AggSigDB test `TestDutyExpiration` has a race condition which causes builds and PRs to fail, [see](https://github.com/ObolNetwork/charon/actions/runs/5857191067/job/15878503330?pr=2543). The test has been disabled.

Fix the race and enable the test.

Aleksao998:
Hey, nice to meet you all! 👋
I've been going through your repo and really like the work you've put into it. 

I've identified the root cause of the race condition in the TestDutyExpiration test. And open PR with a fix:
https://github.com/ObolNetwork/charon/pull/2561. Looking forward to your feedback 🚀

PR #2561, opened 2023-08-23
core/aggsigdb: fix race condition in TestDutyExpiration

Aleksao998:
This PR addresses a detected race condition in TestDutyExpiration tests. The issue arose from simultaneous read/write operations on the db.keysByDuty map. Instead of relying on timeouts, which proved unreliable, I switched to using a sync.WaitGroup to ensure effective coordination. 
Along with the race condition fix, this PR also enhances the test suite's performance by introducing t.Parallel() at the beginning of tests to allow for parallel execution of tests within the package.

category: bug
ticket: https://github.com/ObolNetwork/charon/issues/2546

### DIFF
--- a/core/aggsigdb/memory_internal_test.go
+++ b/core/aggsigdb/memory_internal_test.go
@@ -14,8 +14,6 @@ import (
 )
 
 func TestDutyExpiration(t *testing.T) {
-	t.Parallel()
-
 	var wg sync.WaitGroup
 	ctx, cancel := context.WithCancel(context.Background())
 
@@ -56,8 +54,6 @@ func TestDutyExpiration(t *testing.T) {
 }
 
 func TestCancelledQuery(t *testing.T) {
-	t.Parallel()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/core/aggsigdb/memory_internal_test.go
+++ b/core/aggsigdb/memory_internal_test.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


core/aggsigdb/memory_internal_test.go
```diff
@@ -15,14 +14,20 @@ import (
 )
 
 func TestDutyExpiration(t *testing.T) {
-	t.Skip("Disabled due to race condition, see https://github.com/ObolNetwork/charon/issues/2546")
+	t.Parallel()
```
xenowits, line 17: can remove this as this doesn't bring significant performance improvements

core/aggsigdb/memory_internal_test.go
```diff
@@ -15,14 +14,20 @@ import (
...
 	deadliner := newTestDeadliner()
 	db := NewMemDB(deadliner)
 
-	go db.Run(ctx)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		db.Run(ctx)
+	}()
```
xenowits, line 30:
```suggestion
	go func() {
		db.Run(ctx)
		wg.Done()
	}()
```
would this work?

core/aggsigdb/memory_internal_test.go
Aleksao998: Sure, no problem!

Personally, I like to make unit tests parallel not only for the speed but also because they can catch concurrency issues. However, I agree that these tests are fairly simple, and removing the parallelization is reasonable 😃

Aleksao998: Removed ca5a65632bacf4f1b8ea78de322c8fb1ef210c0a

core/aggsigdb/memory_internal_test.go
Aleksao998: Using `defer wg.Done()` is a cleaner solution as it ensures that `wg.Done()` will always be called right before the goroutine exits, irrespective of how the function returns or if unexpected panics occur. Placing `wg.Done()` directly at the end depends on the function's flow to reach that point, there's a risk it might not get executed. Using `defer` offers more robustness and clarity in the code's intention.

However, if you believe the direct approach is more appropriate for this example, I respect that viewpoint and it should work in this specific case.

xenowits: I understand & agree with your reasoning, let's do `defer wg.Done()`

xenowits (approved):
Thanks for fixing this!

Issue #2597, opened 2023-09-27
Implement deneb support


xenowits:
# 🎯 Problem to be solved

Implement support for deneb.

# 🛠️ Proposed solution
- [x] Approved design doc: https://docs.google.com/document/d/1WcJ4P-_JO937biy4nzwrYyWrYapldiTlAyOQU0qoCo0/edit#heading=h.yhsadc1regoc
- [x] Core team consensus on the proposed solution

Implement solution 2

# 🧪 Tests

We need to test deneb support on devnets: https://github.com/ethpandaops/dencun-testnet

Also, see https://dencun-devnet-8.ethpandaops.io/

boulder225:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2597?planning-poker) with Zenhub @gsora @dB2510 @xenowits

xenowits:
Closing this as changes to core interfaces is merged now

PR #2635, opened 2023-10-10
core: update interfaces

xenowits:
Update `SignedData` interface to include:
* `Signatures() []Signature` - Replaces `Signature()`
* `SetSignatures([]Signature) (SignedData, error)` - Replaces `SetSignature()`
* `MessageRoots() ([][32]byte, error)` - Replaces `MessageRoot()`
* `HashRoot() ([32]byte, error)`

Also, update `Eth2SignedData` interface to include:
* `DomainNames() []signing.DomainName` - Replaces `DomainName()`

These changes to the core interfaces are done in order to prepare for the upcoming deneb ethereum upgrade.

category: feature 
ticket: #2597


core/fetcher/fetcher.go
```diff
@@ -230,7 +230,7 @@ func (f *Fetcher) fetchProposerData(ctx context.Context, slot int64, defSet core
 			return nil, err
 		}
 
-		randao := randaoData.Signature().ToETH2()
+		randao := randaoData.Signatures()[0].ToETH2()
```
gsora, line 233: Thougth: can we ever unmarshal a `SignedData` whose signatures slice is empty or nil?

core/parsigdb/memory.go
```diff
@@ -204,11 +204,11 @@ func getThresholdMatching(typ core.DutyType, sigs []core.ParSignedData, threshol
 		return sigs, len(sigs) == threshold, nil
 	}
 
-	sigsByMsgRoot := make(map[[32]byte][]core.ParSignedData)
+	sigsByMsgRoot := make(map[[32]byte][]core.ParSignedData) // map[Root][]ParSignedData
```
gsora, line 207: See, this is what I meant with my Slack message:

> i would like to avoid having random [32]byte all over the place

I feel we should introduce a specific type like `SignedDataFingerprint` or something, so it's apparent what those bytes's purpose is.

core/sigagg/sigagg.go
```diff
@@ -86,39 +86,84 @@ func (a *Aggregator) aggregate(ctx context.Context, pubkey core.PubKey, parSigs
...
-	// Get all partial signatures.
-	blsSigs := make(map[int]tbls.Signature)
+	// The following tables shows the layout of the data to be aggregated.
+	// Note that the signature aggregation will be performed for each individual column.
+	// +=============+============+============+============+============+
+	// | Data/Charon |   Data0    |   Data1    |   Data2    |   Data3    |
+	// +=============+============+============+============+============+
+	// | Charon 0    | ParSig[00] | ParSig[01] | ParSig[02] | ParSig[03] |
+	// 	+-------------+------------+------------+------------+------------+
+	// | Charon 1    | ParSig[10] | ParSig[01] | ParSig[12] | ParSig[13] |
```
gsora, line 96:
```suggestion
	// | Charon 1    | ParSig[10] | ParSig[11] | ParSig[12] | ParSig[13] |
```

core/sigagg/sigagg.go
```diff
@@ -86,39 +86,84 @@ func (a *Aggregator) aggregate(ctx context.Context, pubkey core.PubKey, parSigs
...
+		parsigsPerPeer[parSig.ShareIdx] = sigs
+	}
+
+	if len(parsigsPerPeer) < a.threshold {
+		return nil, errors.New("number of partial signatures less than threshold", z.Int("threshold", a.threshold), z.Int("got", len(parsigsPerPeer)))
 	}
 
-	if len(blsSigs) < a.threshold {
-		return nil, errors.New("number of partial signatures less than threshold", z.Int("threshold", a.threshold), z.Int("got", len(blsSigs)))
+	// Check if each peer has the same number of signatures.
```
gsora, line 124: I feel you could move this check in the previous for loop:

1. if `len(parsigsPerPeer) == 0`, then continue
2. if `len(parsigsPerPeer) > 0`, check that `len(parsigsPerPeer[len(parsigsPerPeer)-1]) == len(sigs)`

core/sigagg/sigagg.go
```diff
@@ -86,39 +86,84 @@ func (a *Aggregator) aggregate(ctx context.Context, pubkey core.PubKey, parSigs
...
+	}
+	if sigLen == 0 {
+		return nil, errors.New("empty signatures")
+	}
+	if len(sigLens) > 1 {
+		return nil, errors.New("number of signatures for each peer doesn't match")
 	}
 
 	// Aggregate signatures
 	_, span := tracer.Start(ctx, "tbls.Aggregate")
```
gsora, line 139:
```suggestion
	_, span := tracer.Start(ctx, "tbls.ThresholdAggregate")
```

core/eth2signeddata.go
```diff
@@ -27,25 +29,34 @@ var (
...
+	sigs := data.Signatures()
+	domainNames := data.DomainNames()
+	msgRoots, err := data.MessageRoots()
 	if err != nil {
 		return err
 	}
 
-	return signing.Verify(ctx, eth2Cl, data.DomainName(), epoch, sigRoot, data.Signature().ToETH2(), pubkey)
+	for i, sig := range sigs {
+		err = signing.Verify(ctx, eth2Cl, domainNames[i], epoch, msgRoots[i], sig.ToETH2(), pubkey)
```
gsora, line 47: If `domainNames` or `msgRoots` don't contain the `i`-th element, this code will panic.

Please add some bounds checks before falling into this for loop.

core/signeddata.go
```diff
@@ -178,19 +183,54 @@ type VersionedSignedBeaconBlock struct {
...
 }
 
-func (b VersionedSignedBeaconBlock) MessageRoot() ([32]byte, error) {
+func (b VersionedSignedBeaconBlock) HashRoot() ([32]byte, error) {
+	roots, err := b.MessageRoots()
+	if err != nil {
+		return [32]byte{}, err
+	}
+
+	// TODO(xenowits/deneb): Return hash tree root of all roots when len(roots) > 1
```
gsora, line 192: I believe you could implement this already?

core/signeddata.go
```diff
@@ -178,19 +183,54 @@ type VersionedSignedBeaconBlock struct {
...
-		return b.Capella.Message.HashTreeRoot()
+		root, err := b.Capella.Message.HashTreeRoot()
+		if err != nil {
+			return nil, errors.Wrap(err, "hash signed block message", z.Str("version", eth2spec.DataVersionCapella.String()))
+		}
+
+		return [][32]byte{root}, nil
 	case eth2spec.DataVersionDeneb:
-		return b.Deneb.Message.HashTreeRoot()
+		root, err := b.Deneb.Message.HashTreeRoot()
```
gsora, line 228: This code path will return block root and sidecars root in the Deneb support PR, right?

core/types.go
```diff
@@ -385,12 +385,14 @@ func (s UnsignedDataSet) Clone() (UnsignedDataSet, error) {
...
-	// MessageRoot returns the unsigned data message root.
-	MessageRoot() ([32]byte, error)
+	// Signatures returns the signed duty data's signatures.
+	Signatures() []Signature
+	// SetSignatures returns a copy of signed duty data with the signatures replaced.
+	SetSignatures([]Signature) (SignedData, error)
+	// MessageRoots returns the unsigned data message roots.
+	MessageRoots() ([][32]byte, error)
+	// HashRoot returns the final root of the unsigned data.
+	HashRoot() ([32]byte, error)
```
gsora, line 395: Please rename this method as specified by the outcome of the Slack thread, as well as explaining what this field does a little more.

gsora (commented):
Generally LFTM, left some comments that need addressing before giving the green light.

core/proto.go
```diff
@@ -144,10 +144,15 @@ func ParSignedDataToProto(data ParSignedData) (*pbv1.ParSignedData, error) {
...
+	for _, sig := range data.Signatures() {
+		sigs = append(sigs, sig)
+	}
+
 	return &pbv1.ParSignedData{
-		Data:      d,
-		Signature: data.Signature(),
-		ShareIdx:  int32(data.ShareIdx),
+		Data:       d,
+		Signatures: sigs,
```
dB2510, line 154: We should also set `Signature` field, right? as the first signature in sigs to be backwards compatible with previous charon versions.

core/sigagg/sigagg.go
dB2510, line 96:
```suggestion
	// | Charon 1    | ParSig[10] | ParSig[11] | ParSig[12] | ParSig[13] |
```

core/sigagg/sigagg.go
```diff
@@ -86,39 +86,84 @@ func (a *Aggregator) aggregate(ctx context.Context, pubkey core.PubKey, parSigs
...
+	// Check if each peer has the same number of signatures.
+	sigLens := make(map[int]struct{})
+	var sigLen int
+	for _, row := range parsigsPerPeer {
+		sigLen = len(row)
+		sigLens[sigLen] = struct{}{}
+	}
+	if sigLen == 0 {
+		return nil, errors.New("empty signatures")
+	}
```
dB2510, line 133: this check can be moved to above for loop where `parsigsPerPeer` is set.
#failfast

core/fetcher/fetcher.go
xenowits: Yes, such `SignedData` can be unmarshalled. In general, marshalling/unmarshalling is not related to the signatures.

Take for example the `Attestation` object:
```go
func (a Attestation) Signatures() []Signature {
	return []Signature{SigFromETH2(a.Attestation.Signature)}
}
```
```go
func (a *Attestation) UnmarshalJSON(b []byte) error {
	return a.Attestation.UnmarshalJSON(b)
}
```
Unmarshalling can fail, but not because of nil signatures as these two are not directly co-related

core/parsigdb/memory.go
xenowits: These [32]byte are generally roots, we can replace them with `eth2p0.Root` or introduce a `core.Root` type of our own

core/sigagg/sigagg.go
xenowits: Yeah, wrote a slightly simpler version in the previous for loop:
```go
...
		if sigLen == 0 {
			sigLen = len(sigs)
		} else if sigLen > 0 && sigLen != len(sigs) {
			// Check if each peer has the same number of signatures.
			return nil, errors.New("number of signatures for each peer doesn't match", z.Int("peer_x", sigLen), z.Int("peer_y", len(sigs)))
		}
...
```

core/types.go
xenowits: Removed the `HashRoot()` interface method in favour of the function `HashMessageRoots(data)`, see updated code

core/signeddata.go
xenowits: Yeah, absolutely. Something like this:
```go
...
	case eth2spec.DataVersionDeneb:
		var msgRoots [][32]byte
		blockRoot, err := v.Deneb.SignedBlock.Message.HashTreeRoot()
		if err != nil {
			return nil, errors.Wrap(err, "hash signed block message")
		}

		msgRoots = append(msgRoots, blockRoot)

		for _, blob := range v.Deneb.SignedBlobSidecars {
			blobRoot, err := blob.Message.HashTreeRoot()
			if err != nil {
				return nil, errors.Wrap(err, "hash blob sidecar")
			}

			msgRoots = append(msgRoots, blobRoot)
		}

		return msgRoots, nil
...
```

core/proto.go
xenowits: yeah, to ensure v0.18.0 is compatible with v0.17.0. Although, this can be removed when we release v0.19.0

core/sigagg/sigagg.go
xenowits: yeah done, please check update code

core/eth2signeddata.go
xenowits: yes correct, added checks before this for loop is executed:
```go
	if len(domainNames) != len(msgRoots) {
		return errors.New("mismatching lengths", z.Int("domain_names", len(domainNames)), z.Int("message_roots", len(msgRoots)))
	}
	if len(domainNames) != len(sigs) {
		return errors.New("mismatching lengths", z.Int("domain_names", len(domainNames)), z.Int("signatures", len(sigs)))
	}
```

core/fetcher/fetcher.go
gsora: Then maybe we should add bound checks?

gsora (approved):

gsora (approved):
LFTM

core/fetcher/fetcher.go
xenowits: I don't think we need to add bound checks, as the unmarshalling is done by the go-eth2-client itself and it does bound checks

dB2510 (approved):

PR #2645, opened 2023-10-18
testutil/validatormock: fix simnet tests

dB2510:
Fixes panics in simnet tests by introducing a delay in validatormock of 2 slots. The panics were caused by the race condition between scheduling duties of two slots. Note: In simnet tests we have 1s slot duration and 1 slot per epoch. This is happening when scheduler sends slot tick event in between the slot. 

For example, in above scenario on startup first peer got slot tick of 1st slot at 900th Millisecond but it took 200ms to schedule duties for that slot which results in slot tick of 2nd slot. And since we set attester duties once at the start of each epoch, this results in panic as it sets duties twice just because first two slots were out of sync.

category: fixbuild
ticket: none

### DIFF
--- a/testutil/validatormock/component.go
+++ b/testutil/validatormock/component.go
@@ -20,6 +20,10 @@ import (
 // epochWindow is the window size of the lookahead.
 const epochWindow = 2
 
+// delayStartSlots is the number slots to wait before starting validatormock.
+// This is needed to avoid inconsistencies between peers on startup in simnet tests .
+const delayStartSlots = 2
+
 // scheduleTuple is a tuple of a duty and the time it should be performed.
 type scheduleTuple struct {
 	duty      core.Duty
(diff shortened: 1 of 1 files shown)


xenowits (approved):

testutil/validatormock/component.go
```diff
@@ -20,6 +20,10 @@ import (
 // epochWindow is the window size of the lookahead.
 const epochWindow = 2
 
+// delayStartSlots is the number slots to wait before starting validatormock.
+// This is needed to avoid inconsistencies between peers on startup in simnet tests .
+const delayStartSlots = 2
```
gsora, line 25: nit: combine the two `const` declarations in a single block

gsora (approved):

testutil/validatormock/component.go
dB2510: done

PR #2649, opened 2023-10-18
vmock: change start time of sync selections to start of slot

dB2510:
Changes start time of sync selections in vmock to start of slot as proposed.

category: refactor
ticket: none


xenowits (approved):

gsora (approved):
I wonder if we should leave comments explaining the choice of start times... but that's a task for another day.

PR #2651, opened 2023-10-19
testutil/validatormock: remove pseudo code

dB2510:
Removes sync contribution pseudo code as it is not needed now. It was written before we even implemented sync committee duties in vmock to understand the workflow.

category: refactor
ticket: none


xenowits (approved):
LGTM! 

Yeah, this is dead code, we should have removed it immediately after code for sync committee was written.

Issue #2658, opened 2023-10-26
upper bound checks for integer conversion


xenowits:
# 🎯 Problem to be solved

Fix codeql errors about unsafe integer conversion:
```
Incorrect conversion of an unsigned 64-bit integer from  to a lower bit size type int64 without an upper bound check.
```
See https://github.com/ObolNetwork/charon/pull/2650/checks?check_run_id=18076838602

# 🛠️ Proposed solution

Convert int64s to uint64s according to the ethereum specification doc.

PR #2668, opened 2023-10-31
*: convert to uint64

xenowits:
Convert `Duty.Slot` from int64 to uint64.

This is recommended according to ethereum consensus specs: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#custom-types

category: feature 
ticket: #2658


core/scheduler/scheduler.go
```diff
@@ -319,11 +319,11 @@ func (s *Scheduler) resolveAttDuties(ctx context.Context, slot core.Slot, vals v
 			continue
 		}
 
-		duty := core.NewAttesterDuty(int64(attDuty.Slot))
+		duty := core.NewAttesterDuty(uint64(attDuty.Slot))
```
gsora, line 322: Can `attDuty.Slot` be moved to uint64 as well?

core/tracing.go
```diff
@@ -29,7 +30,7 @@ func StartDutyTrace(ctx context.Context, duty Duty, spanName string, opts ...tra
 	ctx, outerSpan = tracer.Start(tracer.RootedCtx(ctx, traceID), fmt.Sprintf("core/duty.%s", strings.Title(duty.Type.String())))
 	ctx, innerSpan = tracer.Start(ctx, spanName, opts...)
 
-	outerSpan.SetAttributes(attribute.Int64("slot", duty.Slot))
+	outerSpan.SetAttributes(attribute.Int64("slot", safeInt64(duty.Slot)))
```
gsora, line 33: Same here, i wonder if we can have Slot as uint64 and eventually do the safeint64 checks when creating the struct.

dkg/exchanger.go
```diff
@@ -131,7 +131,7 @@ func newExchanger(tcpNode host.Host, peerIdx int, peers []peer.ID, vals int, sig
 // signatures of the group according to public key of each DV.
 func (e *exchanger) exchange(ctx context.Context, sigType sigType, set core.ParSignedDataSet) (map[core.PubKey][]core.ParSignedData, error) {
 	// Start the process by storing current peer's ParSignedDataSet
-	duty := core.NewSignatureDuty(int64(sigType))
+	duty := core.NewSignatureDuty(uint64(sigType))
```
gsora, line 134: Would change sigType to uint64 instead.

xenowits: I don't that's required, what'd be the reason to change type for `sigType` which is currently an `int` to `uint64`?

slots, committee indices etc should be uint64 because the spec mandates that.

core/tracing.go
xenowits: attribute has only 3 methods: `attribute.Int()`, `attribute.Int64()` & `attribute.String()`

q: why would we need to perform safeint checks if we define slots are uint64, won't the compiler already catch this?

core/scheduler/scheduler.go
xenowits: `attDuty.Slot` is of `eth2p0.Slot` type. Since, `NewAttesterDuty(slot uint64)`, we need to typecast the eth2p0.Slot to uint64 type

core/consensus/component.go
```diff
@@ -743,7 +743,7 @@ func fmtStepPeers(step roundStep) string {
 
 // leader return the deterministic leader index.
 func leader(duty core.Duty, round int64, nodes int) int64 {
```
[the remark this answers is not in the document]
xenowits: it can, but uint64 is not really required

gsora (approved):

PR #2673, opened 2023-11-03
cmd/createcluster: fix pre-gen builder registrations

dB2510:
This PR fixes a bug in the `create cluster` command that was generating incorrect builder registrations in cluster-lock.json. The issue stemmed from the usage of withdrawal addresses instead of fee recipient addresses, which resulted in signature verification failures when the command was used for the 'charon run' operation.

Here's the error log caused by this bug:
image

category: bug
ticket: #2674

### DIFF
--- a/cmd/createcluster_internal_test.go
+++ b/cmd/createcluster_internal_test.go
@@ -189,6 +189,24 @@ func TestCreateCluster(t *testing.T) {
 				return data
 			},
 		},
+		{
+			Name: "test with fee recipient and withdrawal addresses",
+			Config: clusterConfig{
+				Name:      "test_cluster",
+				NumNodes:  3,
+				Threshold: 4,
+				NumDVs:    5,
+				Network:   "goerli",
+			},
+			Prep: func(t *testing.T, config clusterConfig) clusterConfig {
+				t.Helper()
+
+				config.FeeRecipientAddrs = []string{testutil.RandomChecksummedETHAddress(t, 1)}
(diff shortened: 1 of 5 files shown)


cmd/createcluster_internal_test.go
```diff
@@ -189,6 +189,24 @@ func TestCreateCluster(t *testing.T) {
 				return data
 			},
 		},
+		{
+			Name: "test with fee recipient and withdrawal addresses",
```
xenowits, line 193: let's add tests for:
* different fee & withdrawal addresses (done)
* same fee & withdrawal addresses
* no fee & withdrawal address

xenowits (approved):

OisinKyne:
Small nit/feedback @dB2510 

>This PR fixes a critical bug in the create cluster command 

I think we should develop a shared framework for how we name bug severities. This would have made mev registrations fail for a specific subset of scenarios, and local block production may have worked. Either way, I would not put 'could miss block' in the critical category here, and would save that for loss of funds not missed opportunities.

dB2510:
Hey @OisinKyne 
Yeah agreed, this is not a critical bug which can cause loss of funds. This bug was not letting charon to start as it was failing initial cluster-lock verification which is why I mentioned it as a critical bug. It can be mitigated with the `--no-verify` flag and then everything will work as expected even the block proposals.
Updated the PR description 👍

cmd/createcluster_internal_test.go
dB2510: These tests are there in this file as part of `TestMultipleAddresses`.

PR #2675, opened 2023-11-06
core/bcast: fix builder registrations delays

dB2510:
This PR fixes delay reports by bcast for builder registration duty. As reported multiple times:

```
12:32:23.134 INFO bcast      Successfully submitted validator registration to beacon node {"delay": "25416h32m0.13494757s", "pubkey": "858_9bf", "duty": "0/builder_registration"}
```

category: bug
ticket: #2681

### DIFF
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -110,6 +110,14 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, set core.Sig
 		return err
 
 	case core.DutyBuilderRegistration:
+		slot, err := firstSlotInCurrentEpoch(ctx, b.eth2Cl)
+		if err != nil {
+			return errors.Wrap(err, "calculate first slot in epoch")
+		}
+
+		// Use first slot in current epoch for accurate delay calculations while submitting builder registrations.
+		// This is because builder registrations are submitted in first slot of every epoch.
+		duty.Slot = slot
 		registrations, err := setToRegistrations(set)
 		if err != nil {
 			return err
@@ -298,4 +306,28 @@ func newDelayFunc(ctx context.Context, eth2Cl eth2wrap.Client) (func(slot uint64
 		slotStart := genesis.Add(slotDuration * time.Duration(slot))
 		return time.Since(slotStart)
 	}, nil
+}
+
+// firstSlotInCurrentEpoch calculates first slot number of the current ongoing epoch.
+func firstSlotInCurrentEpoch(ctx context.Context, eth2Cl eth2wrap.Client) (int64, error) {
+	genesis, err := eth2Cl.Genesis(ctx)
+	if err != nil {
+		return 0, errors.Wrap(err, "fetch genesis")
+	}
+
+	slotDuration, err := eth2Cl.SlotDuration(ctx)
+	if err != nil {
+		return 0, errors.Wrap(err, "fetch slot duration")
+	}
+
+	slotsPerEpoch, err := eth2Cl.SlotsPerEpoch(ctx)
+	if err != nil {
+		return 0, errors.Wrap(err, "fetch slots per epoch")
+	}
+
+	chainAge := time.Since(genesis.Data.GenesisTime)
+	currentSlot := chainAge / slotDuration
+	currentEpoch := uint64(currentSlot) / slotsPerEpoch
+
+	return int64(currentEpoch * slotsPerEpoch), nil
 }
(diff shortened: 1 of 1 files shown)


core/bcast/bcast.go
```diff
@@ -110,6 +110,14 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, set core.Sig
 		return err
 
 	case core.DutyBuilderRegistration:
+		slot, err := firstSlotInCurrentEpoch(ctx, b.eth2Cl)
+		if err != nil {
+			return errors.Wrap(err, "calculate first slot in epoch")
```
gsora, line 115: If this function call fails we'll not submit a builder registration.

What if the upstream beacon node briefly wasn't able to fulfill this request, but later it'll be able to submit a validator registration?

Bottom line is: should we allow validator registration to fail if the resulting metric *might* be off?

dB2510: @gsora yeah good thought, but here's the catch:
We need these three value to calculate the first slot in epoch: `genesis_time`, `slots_per_epoch` and `slot_duration`. `go-eth2-client` caches (since these are constants) these values when we create an instance of `eth2http` and it returns cached values every time we call the associated methods. So I think we should be fine calling these methods here.

Refer:
- https://github.com/attestantio/go-eth2-client/blob/122f335c2f6a20558846872dc4b47cf61ad615d6/http/service.go#L69
- https://github.com/attestantio/go-eth2-client/blob/122f335c2f6a20558846872dc4b47cf61ad615d6/http/service.go#L145

Other approach could be store these constants in `Broadcaster` struct and populate those values in constructor.

xenowits: I think this won't return an error almost all of the time since as you mentioned, these are cached constants, ofc unless there's an error with eth2http configuration itself. So, i think it's safe to catch and return an error.

core/bcast/bcast.go
```diff
@@ -299,3 +307,27 @@ func newDelayFunc(ctx context.Context, eth2Cl eth2wrap.Client) (func(slot uint64
...
 }
+
+// firstSlotInCurrentEpoch calculates first slot number of the current ongoing epoch.
+func firstSlotInCurrentEpoch(ctx context.Context, eth2Cl eth2wrap.Client) (uint64, error) {
+	genesis, err := eth2Cl.Genesis(ctx)
+	if err != nil {
+		return 0, errors.Wrap(err, "fetch genesis")
+	}
+
+	slotDuration, err := eth2Cl.SlotDuration(ctx)
```
xenowits, line 318: SlotDuration is deprecated, you should use `Spec` instead

core/bcast/bcast.go
```diff
@@ -110,6 +110,14 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, set core.Sig
...
 
 	case core.DutyBuilderRegistration:
+		slot, err := firstSlotInCurrentEpoch(ctx, b.eth2Cl)
+		if err != nil {
+			return errors.Wrap(err, "calculate first slot in epoch")
+		}
+
+		// Use first slot in current epoch for accurate delay calculations while submitting builder registrations.
+		// This is because builder registrations are submitted in first slot of every epoch.
+		duty.Slot = slot
```
xenowits, line 120: I'm wondering about an alternate approach:

We can probably do the slot from timestamp conversion at the validatorapi component: https://github.com/ObolNetwork/charon/blob/main/core/validatorapi/validatorapi.go#L522

This will create the accurate slot and initialize the duty with the correct slot (NewBuilderRegistrationDuty) starting at validatorapi itself

core/bcast/bcast.go
```diff
@@ -299,3 +307,27 @@ func newDelayFunc(ctx context.Context, eth2Cl eth2wrap.Client) (func(slot uint64
...
+	if err != nil {
+		return 0, errors.Wrap(err, "fetch genesis")
+	}
+
+	slotDuration, err := eth2Cl.SlotDuration(ctx)
+	if err != nil {
+		return 0, errors.Wrap(err, "fetch slot duration")
+	}
+
+	slotsPerEpoch, err := eth2Cl.SlotsPerEpoch(ctx)
```
xenowits, line 323: This is deprecated too, use `Spec` instead

core/bcast/bcast.go
dB2510: yeah thought about this too, we need to do this at two places, first is here in validatorapi when VC submits registration and second is where we submit pre-gen registrations. Currently we use slot 0 for pregens and slot 1 for VC submissions which helps us in differentiating both types of registrations. 
Wanted to keep the differentiating part which is why I did this just before we submit in bcast and where we log/report the delays.

gsora (approved):

core/bcast/bcast.go
dB2510: Ignoring this for now

core/bcast/bcast.go
dB2510: Ignoring this for now as it will be part of next go-eth2-client version.

Issue #2659, opened 2023-10-26
update to new deneb changes


xenowits:
# 🎯 Problem to be solved

There have been changes to the deneb spec recently(see [PR](https://github.com/ethereum/consensus-specs/pull/3531)).

We need to update our code to comply with the updated deneb specs.

# 🛠️ Proposed solution

Note that in the updated spec, we don't require signatures on blob sidecars. Hence, the change to our [core interfaces](https://github.com/ObolNetwork/charon/pull/2635) doesn't make sense anymore as we now have a single signature per signed object.

* Revert PR#2635
* In all `eth2spec.DataVersionDeneb` cases, use only the `SignedBeaconBlock` signature.
  * There may be changes on the go-eth2-client side

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2682, opened 2023-11-07
core: revert core interfaces

xenowits:
Revert changes to core interfaces.

This is done since the deneb upgrade no longer requires multiple signatures per signed object. Reverting to the original implementation makes the interfaces simple and code much easier to reason about.

category: refactor
ticket: #2659


dB2510 (approved):
why the hell sonarlint failing 😂

gsora (approved):
lfg

also we should have a synchronous session to review sonarcloud's rules

Issue #2695, opened 2023-11-09
fix synthetic proposal when block not found


xenowits:
# 🎯 Problem to be solved

Fix synthetic proposal errors when beacon block for a slot is not found
```
Permanent failure calling fetcher/fetch: fetch builder proposer data: beacon api signed_beacon_block: GET failed with status 404: {"code":404,"message":"NOT_FOUND: beacon block at slot 6917368","stacktraces":[]} {"label": "signed_beacon_block", "duty": "6917369/builder_proposer"}
```

# 🛠️ Proposed solution

If a not found error is received (404), continue looking for blocks in earlier slots until found instead of erroring.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

dB2510:
Fixed by #2702

PR #2702, opened 2023-11-10
app/eth2wrap: fix error comparison for synthetic blocks

dB2510:
There was a small issue when comparing eth2api errors when searching for signed beacon block. The error returned should be part of eth2wrap multi implementation which was not able to decompose eth2 errors properly. This PR fixes decomposition of eth2 errors and adds a function to check if the relevant field is present as part of the error chain.

category: bug
ticket: #2695

### DIFF
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -330,7 +330,7 @@ func incError(endpoint string) {
 // wrapError returns the error as a wrapped structured error.
 func wrapError(ctx context.Context, err error, label string, fields ...z.Field) error {
 	// Decompose go-eth2-client http errors
-	if apiErr := new(eth2api.Error); errors.As(err, apiErr) {
+	if apiErr := new(eth2api.Error); errors.As(err, &apiErr) {
 		err = errors.New("nok http response",
 			z.Int("status_code", apiErr.StatusCode),
 			z.Str("endpoint", apiErr.Endpoint),
--- a/app/eth2wrap/synthproposer.go
+++ b/app/eth2wrap/synthproposer.go
@@ -20,9 +20,11 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/capella"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	shuffle "github.com/protolambda/eth2-shuffle"
+	"go.uber.org/zap"
 
 	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -330,7 +330,7 @@ func incError(endpoint string) {
 // wrapError returns the error as a wrapped structured error.
 func wrapError(ctx context.Context, err error, label string, fields ...z.Field) error {
 	// Decompose go-eth2-client http errors
-	if apiErr := new(eth2api.Error); errors.As(err, apiErr) {
+	if apiErr := new(eth2api.Error); errors.As(err, &apiErr) {
```
dB2510, line 333: Seems like this was not working at all.

app/eth2wrap/synthproposer.go
```diff
@@ -149,10 +151,8 @@ func (h *synthWrapper) syntheticProposal(ctx context.Context, slot eth2p0.Slot,
 		}
 		signed, err := h.Client.SignedBeaconBlock(ctx, opts)
 		if err != nil {
-			if apiErr := new(eth2api.Error); errors.As(err, apiErr) { // Continue if block is not found in the given slot.
-				if apiErr.StatusCode == http.StatusNotFound {
-					continue
-				}
+			if fieldExists(err, zap.Int("status_code", http.StatusNotFound)) {
```
dB2510, line 154: This was needed as the returned error will be a wrapped error from eth2wrap.Multi struct

app/eth2wrap/synthproposer.go
```diff
@@ -204,6 +204,34 @@ func (h *synthWrapper) syntheticProposal(ctx context.Context, slot eth2p0.Slot,
 	return proposal, nil
 }
 
+// fieldExists checks if the given field exists as part of the given error.
+func fieldExists(err error, field zap.Field) bool {
```
xenowits, line 208: Please add tests for this

xenowits (approved):

app/eth2wrap/synthproposer.go
dB2510: done

app/eth2wrap/eth2wrap.go
gsora: There must be a linter somewhere that checks for this...

app/eth2wrap/synthproposer.go
gsora, line 208: I wonder if instead of declaring this function here, we should create a `z.Fields()` function that returns `[]z.Field`.

The casting logic seems to be a 1:1 copy off `z.Err`.

gsora (approved):
Approving and opening an issue to later address https://github.com/ObolNetwork/charon/pull/2702/files#r1389222292

Issue #2671, opened 2023-11-02
On-disk logging facility


gsora:
# 🎯 Problem to be solved

Charon has been designed from the beginning with a no-disk-write approach: aside for the DKG and `create cluster`, Charon never   writes anything on disk.

The container-exclusive deployment strategy leaves the burden of log retention to the user: since on container restart logs are lost, the user needs to devise a plan to persist them somewhere in order to audit and eventually find and report bugs to us.

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

I propose we consider the implementation of a on-disk log retention strategy, where Charon writes logs on the already existing facilities (Loki, stdout) as well as files -- plural here because log files need to be [periodically rotated](https://en.wikipedia.org/wiki/Log_rotation).

Zap provides multiple ways of configuring custom logging outputs, and they document a way of automating log rotation and compression from within the codebase itself: https://github.com/uber-go/zap/blob/master/FAQ.md#usage.

corverroos:
A simple solution would be to add optional logging driver config to cdvn repo. Either manually to each compose service or globally as docker default logging driver. A good option is [journald](https://docs.docker.com/config/containers/logging/journald/) which has log rotation built in.

PR #2713, opened 2023-11-14
app/log: on-disk log writing

gsora:
Write debug logs on disk if `--log-output-path` is specified.

Log files are rotated automatically once the file size reaches 50MB.

Rotated log files will be kept for 7 days - max 10 rotated files will be kept, whatever comes first.

category: feature
ticket: #2671


app/log/config.go
```diff
@@ -145,6 +159,25 @@ func InitLogger(config Config) error {
...
 	}
 
+	var registerError error
+	registerZapSink.Do(func() {
+		registerError = zap.RegisterSink("lumberjack", func(u *url.URL) (zap.Sink, error) {
+			return lumberjackSink{
+				Logger: &lumberjack.Logger{
+					Filename: u.Path,
+
+					MaxBackups: 10,
```
dB2510, line 169: nit: add a `const` for MaxBackups field and a godoc for that const

dB2510 (approved):

Issue #2690, opened 2023-11-08
Add `--testnet-file` flag to support custom test networks.


dB2510:
# 🎯 Problem to be solved
Currently we only support the cluster creation for public networks like mainnet, goerli, sepolia, holesky, etc. In order to test charon in a custom testnet with dynamic chain configuration, we need to add support for cluster creation for custom testnets.

# 🛠️ Proposed solution

Add a `--testnet-file` flag to `charon create cluster` and `charon create dkg` to support custom networks. The expected file should look like the following:
```json
{
  "chain_id": 5,
  "name": "goerli",
  "genesis_fork_version": "0x00001020",
  "genesis_timestamp": 1616508000
}
```

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

-->

PR #2715, opened 2023-11-17
cmd: add custom testnet flags

dB2510:
Adds custom testnet flags to `charon create cluster` and `charon run` commands.

category: feature
ticket: #2690

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -146,10 +146,7 @@ func Run(ctx context.Context, conf Config) (err error) {
 		life.RegisterStop(lifecycle.StopPrivkeyLock, lifecycle.HookFuncMin(lockSvc.Close))
 	}
 
-	if conf.TestnetConfig.Name != "" &&
-		conf.TestnetConfig.GenesisForkVersionHex != "" &&
-		conf.TestnetConfig.GenesisTimestamp != 0 &&
-		conf.TestnetConfig.ChainID != 0 {
+	if conf.TestnetConfig.IsNonZero() {
 		eth2util.AddTestNetwork(conf.TestnetConfig)
 	}
 
--- a/app/app.go
+++ b/app/app.go
@@ -150,7 +150,7 @@ func Run(ctx context.Context, conf Config) (err error) {
 		conf.TestnetConfig.GenesisForkVersionHex != "" &&
 		conf.TestnetConfig.GenesisTimestamp != 0 &&
 		conf.TestnetConfig.ChainID != 0 {
-		eth2util.AddNetwork(conf.TestnetConfig)
+		eth2util.AddTestNetwork(conf.TestnetConfig)
 	}
 
 	if err := wireTracing(life, conf); err != nil {
--- a/app/app.go
+++ b/app/app.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"strings"
 	"sync"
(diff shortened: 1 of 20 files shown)


eth2util/network.go
```diff
@@ -59,12 +61,22 @@ var (
 	}
 )
 
-var supportedNetworks = []Network{
-	Mainnet, Goerli, Gnosis, Sepolia, Holesky,
+var (
+	addNetworkOnce    sync.Once
```
xenowits, line 65: nit: can remove this if `supportedNetworks` is written only once on startup, then only read

wdyt?

dB2510: agreed, removed sync once and updated the godoc.

cmd/createcluster.go
```diff
@@ -1030,3 +1050,31 @@ func builderRegistrationFromETH2(reg core.VersionedSignedValidatorRegistration)
 		Signature: reg.Signature(),
 	}, nil
 }
+
+// validateNetworkConfig returns an error if the network configuration is invalid in the given cluster configuration.
+func validateNetworkConfig(conf clusterConfig) error {
+	if conf.Network != "" {
+		if conf.Network == eth2util.Prater {
```
gsora, line 1057: Why are you checking this here and in the outer/caller function?

eth2util/network.go
```diff
@@ -63,8 +64,14 @@ var supportedNetworks = []Network{
 	Mainnet, Goerli, Gnosis, Sepolia, Holesky,
 }
 
+// AddTestNetwork adds given network to list of supported networks.
+// Note: This function should only be called once.
```
gsora, line 68: Why?

xenowits (approved):

eth2util/network.go
dB2510: Because in this function we are writing to `supportedNetworks` slice. This can introduce race condition if called twice at the same time by different goroutines as it is not protected by and mutex/once.
Other option can be to have `sync.Once` so that this _will_ be called once.
Open to suggestions to improve godoc for more clarity.

The usage of AddTestNetwork is to add the testnet config as the first step for any process (create cluster OR charon run) so that other functions in this file will behave as expected.

gsora: My first suggestion is to fix the go doc: this function can be called more than once, but it can’t be called *concurrently*. 

One thing that we could do is “seal” the slice: add a bool flag that is true when it’s sealed. If true, don’t allow appends. 

Another solution (arguably simpler to reason about) is to add locking around slice access. 

Honestly even just documenting the racy nature of this function would suffice, but adding a lock and remove the ambiguity is the most complete solution IMO.

dB2510: yeah, I initially went with sync.Once but then thought that just documenting would suffice. I'll go with fixing the godoc.

dB2510: yeah that is why I went with `sync.Once` to protect this function. Now i am thinking about an integration test which can call `app.Run(..)` in different goroutines 🤔 

Cool, I am going with adding a lock 👍

gsora: > Now i am thinking about an integration test which can call app.Run(..) in different goroutines 🤔

Now this is interesting 👀 

The race detector would go wild!

dB2510: Done, I added lock 👍

app/app.go
```diff
@@ -145,6 +146,13 @@ func Run(ctx context.Context, conf Config) (err error) {
 		life.RegisterStop(lifecycle.StopPrivkeyLock, lifecycle.HookFuncMin(lockSvc.Close))
 	}
 
+	if conf.TestnetConfig.Name != "" &&
```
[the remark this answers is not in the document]
dB2510: Added NonZero method 👍

Issue #2736, opened 2023-12-01
Return dependent root from validatorapi


dB2510:
# 🎯 Problem to be solved
We need to return a [proper](https://github.com/ObolNetwork/charon/blob/cb4c265a6c3d85da14f701c6da3f74a96e86a614/core/validatorapi/router.go#L469) `dependent_root` field as response to the duties endpoints. This is causing lodestar to flood WARN logs every slot:

```
Potential current epoch attester duties reorg slot=408896, dutyEpoch=12778, priorDependentRoot=0xea63000000000000000000000000000000000000000000000000000000000000
```

# 🛠️ Proposed solution
As part of recent go-eth2-client upgrade, it has started returning the `dependent_root` field as part of metadata which was absent in earlier versions. We need to add support of this field while returning response through validatorapi for VCs.
https://github.com/attestantio/go-eth2-client/blob/677e48992f5edabadc5acbaf11b8db654ac47c07/http/attesterduties.go#L84

# 🧪 Tests
- [x] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [x] Manually tested on local compose simnet

boulder225:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2736?planning-poker) with Zenhub @pinebit

PR #2739, opened 2023-12-06
validatorapi: handling duties response metadata

pinebit:
As part of recent go-eth2-client upgrade, `ProposerDuties` and `AttesterDuties` calls started to return Metadata fields: 
```
  "dependent_root": "0xc01a8003a974cea31fd9e91c7d2cec8120ea3cc71edcbb836b6fbede6c289a69",
  "execution_optimistic": false,
```

This change updates the corresponding handlers of validator API to properly parse and propagate these new fields. 

category: feature
ticket: #2736

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -429,9 +429,19 @@ func proposerDuties(p eth2client.ProposerDutiesProvider) handlerFunc {
 			data = []*eth2v1.ProposerDuty{}
 		}
 
+		executionOptimistic, err := getExecutionOptimisticFromMetadata(eth2Resp.Metadata)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "failed to decode ProposerDuties response metadata")
+		}
+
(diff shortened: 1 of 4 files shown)


gsora (approved):

core/validatorapi/router.go
```diff
@@ -429,9 +428,19 @@ func proposerDuties(p eth2client.ProposerDutiesProvider) handlerFunc {
 			data = []*eth2v1.ProposerDuty{}
 		}
 
+		executionOptimistic, err := getExecutionOptimisticFromMetadata(eth2Resp.Metadata)
```
dB2510, line 431: go-eth2-client doesn't support execution optimistic field yet.
Refer: https://github.com/attestantio/go-eth2-client/blob/3087274ef2ac950392c3e362e0d0572aa1c6dada/http/json.go#L12

So I'd suggest to keep the TODO to add execution optimistic field later when they add support for it.

dB2510 (approved):
Congrats on your first contribution 🎉 

Let's test this change in our core team cluster with lodestar.

PR #2762, opened 2023-12-21
app/featureset: split `qbft_timers_ab_test` feature flag

dB2510:
This PR splits `qbft_timers_ab_test` feature flag into two feature flags to enable consensus participate and eager double linear round timer separately. 

category: misc
ticket: none

### DIFF
--- a/core/consensus/component.go
+++ b/core/consensus/component.go
@@ -18,6 +18,7 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/featureset"
 	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/app/z"
 	"github.com/obolnetwork/charon/core"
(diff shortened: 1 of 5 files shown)


core/consensus/roundtimer_internal_test.go
```diff
@@ -113,10 +113,10 @@ func TestDoubleEagerLinearRoundTimer(t *testing.T) {
 }
 
 func TestGetTimerFunc(t *testing.T) {
```
gsora, line 115: Do we test if timerFunc ever returns something that isn't `timerEagerDoubleLinear`?

core/consensus/component.go
```diff
@@ -382,8 +383,8 @@ func (c *Component) Participate(ctx context.Context, duty core.Duty) error {
 		return nil // No eager consensus for potential no-op aggregation duties.
```
gsora, line 383: I believe this method's godoc needs to change, because we don't check if the timer is Eager anymore.

core/consensus/component.go
```diff
@@ -382,8 +383,8 @@ func (c *Component) Participate(ctx context.Context, duty core.Duty) error {
...
 	}
 
-	if !c.timerFunc(duty).Type().Eager() {
-		return nil // Not an eager start timer, wait for Propose to start.
+	if featureset.Enabled(featureset.ConsensusParticipate) {
+		return nil // Wait for Propose to start.
```
gsora, line 387: Shouldn't this check be flipped?

i.e. if `ConsensusParticipate` is disabled, don't proceed.

dB2510: Fixed

gsora (approved):

PR #2783, opened 2024-01-12
core: addressed some panic instances

pinebit:
Removed some unnecessary panic() calls and added recover() where appropriate.

category: refactor
ticket: none

### DIFF
--- a/core/proto.go
+++ b/core/proto.go
@@ -5,6 +5,7 @@ package core
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	ssz "github.com/ferranbt/fastssz"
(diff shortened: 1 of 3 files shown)


core/proto.go
```diff
@@ -44,10 +45,15 @@ func DutyFromProto(duty *pbv1.Duty) Duty {
 }
 
 // ParSignedDataFromProto returns the data from a protobuf.
-func ParSignedDataFromProto(typ DutyType, data *pbv1.ParSignedData) (ParSignedData, error) {
-	// TODO(corver): This can panic due to json unmarshalling unexpected data.
-	//  For now, it is a good way to catch compatibility issues. But we should
-	//  recover panics and return an error before launching mainnet.
+func ParSignedDataFromProto(typ DutyType, data *pbv1.ParSignedData) (odata ParSignedData, oerr error) {
```
xenowits, line 48: where is `odata` returned?

xenowits (approved):
LGTM!

core/proto.go
pinebit: nowhere, but I think I cannot omit one while having names on the others..

core/tracing.go
```diff
@@ -30,7 +30,8 @@ func StartDutyTrace(ctx context.Context, duty Duty, spanName string, opts ...tra
 	ctx, outerSpan = tracer.Start(tracer.RootedCtx(ctx, traceID), fmt.Sprintf("core/duty.%s", strings.Title(duty.Type.String())))
 	ctx, innerSpan = tracer.Start(ctx, spanName, opts...)
 
-	outerSpan.SetAttributes(attribute.Int64("slot", safeInt64(duty.Slot)))
+	slotStr := strconv.FormatUint(duty.Slot, 10)
+	outerSpan.SetAttributes(attribute.String("slot", slotStr))
```
gsora, line 34: @xenowits @dB2510 this line changes the "slot" tracing attribute from int64 to string.

I don't think it'll lead to any worrisome behavior, but since you guys have the most experience on the codebase, wdyt?

gsora (approved):

core/tracing.go
xenowits: I don't think it should cause any ripple behaviors

PR #2797, opened 2024-01-17
dkg: strengthen logic for steps update

pinebit:
Improved logic for steps update in DKG process.

category: misc
ticket: none

### DIFF
--- a/dkg/dkg.go
+++ b/dkg/dkg.go
@@ -478,8 +478,9 @@ func startSyncProtocol(ctx context.Context, tcpNode host.Host, key *k1.PrivateKe
 			break
 		}
 
-		// Sleep for 100ms to let clients connect with each other.
-		time.Sleep(time.Millisecond * 100)
+		// Sleep for 250ms to let clients connect with each other.
+		// Must be at least two times greater than the sync messages period specified in client.go NewClient().
+		time.Sleep(time.Millisecond * 250)
 	}
 
 	// Disable reconnecting clients to other peer's server once all clients are connected.
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -38,7 +38,7 @@ func NewClient(tcpNode host.Host, peer peer.ID, hashSig []byte, version version.
 		done:      make(chan struct{}),
 		reconnect: true,
 		version:   version,
-		period:    250 * time.Millisecond,
+		period:    100 * time.Millisecond, // must be at least two times lower than the sync timeout (dkg.go, startSyncProtocol)
 	}
 
 	for _, opt := range opts {
--- a/dkg/sync/client.go
+++ b/dkg/sync/client.go
@@ -53,7 +53,7 @@ func NewClient(tcpNode host.Host, peer peer.ID, hashSig []byte, version version.
 // supports reestablishing on relay circuit recycling, and supports soft shutdown.
 type Client struct {
 	// Mutable state
-	mu        sync.Mutex
+	mu        sync.RWMutex
 	connected bool
 	reconnect bool
 	step      int
@@ -113,16 +113,16 @@ func (c *Client) SetStep(step int) {
 
 // getStep returns the current step.
 func (c *Client) getStep() int {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	return c.step
 }
 
 // IsConnected returns if client is connected to the server or not.
 func (c *Client) IsConnected() bool {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	return c.connected
 }
@@ -255,8 +255,8 @@ func (c *Client) DisableReconnect() {
 
 // shouldReconnect returns true if clients should re-attempt connecting to peers.
 func (c *Client) shouldReconnect() bool {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	return c.reconnect
 }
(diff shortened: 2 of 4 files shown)


dkg/dkg.go
```diff
@@ -478,8 +478,9 @@ func startSyncProtocol(ctx context.Context, tcpNode host.Host, key *k1.PrivateKe
 			break
 		}
 
-		// Sleep for 100ms to let clients connect with each other.
-		time.Sleep(time.Millisecond * 100)
+		// Sleep for 250ms to let clients connect with each other.
+		// Must be at least two times greater than the sync messages period specified in client.go NewClient().
```
xenowits, line 482: can you please point me to the code where this is specified?

pinebit: https://github.com/ObolNetwork/charon/blob/d2748d5c96dde7c790ae6a18b6e1ef8ecb8e37d7/dkg/sync/client.go#L41

dkg/sync/client.go
```diff
@@ -38,7 +38,7 @@ func NewClient(tcpNode host.Host, peer peer.ID, hashSig []byte, version version.
 		done:      make(chan struct{}),
 		reconnect: true,
 		version:   version,
-		period:    250 * time.Millisecond,
+		period:    100 * time.Millisecond, // must be at least two times lower than the sync timeout (dkg.go, startSyncProtocol)
```
xenowits, line 41: nit: Start comments with uppercase
```suggestion
		period:    100 * time.Millisecond, // Must be at least two times lower than the sync timeout (dkg.go, startSyncProtocol)
```

dkg/sync/client.go
```diff
@@ -53,7 +53,7 @@ func NewClient(tcpNode host.Host, peer peer.ID, hashSig []byte, version version.
 // supports reestablishing on relay circuit recycling, and supports soft shutdown.
 type Client struct {
 	// Mutable state
-	mu        sync.Mutex
+	mu        sync.RWMutex
```
xenowits, line 56: why do we need a read-write mutex here?

pinebit: Because in many places we need just read. Using `RWMutex` optimizes performance (less locks and waits) and is widely adopted in golang applications. However, I found that Charon often uses normal Mutex, and it is unclear why. I would wish to eventually replace everything to RWMutex where appropriate.

dkg/sync/server.go
```diff
@@ -170,26 +170,40 @@ func (s *Server) setShutdown(pID peer.ID) {
...
+	currentPeerStep, hasCurrentPeerStep := s.steps[pID]
+	if hasCurrentPeerStep {
+		if step < currentPeerStep {
+			return errors.New("peer reported step is behind the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
+		} else if step > currentPeerStep+1 {
+			return errors.New("peer reported step is ahead the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
+		}
+	} else if step < 0 || step > 1 {
+		return errors.New("peer reported abnormal initial step, expected 0 or 1", z.Int("peer_step", step))
+	}
```
xenowits, line 188: Instead of branching out, i'd suggest we "return early". That is the guideline that we follow across the charon codebase. 

Suggest this instead:
```suggestion
	if hasCurrentPeerStep && step < currentPeerStep {
		return errors.New("peer reported step is behind the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
	}
	
	if hasCurrentPeerStep && step > currentPeerStep+1 {
		return errors.New("peer reported step is ahead the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
	}

	if step < 0 || step > 1 {
		return errors.New("peer reported abnormal initial step, expected 0 or 1", z.Int("peer_step", step))
	}
```

pinebit: Thank you for the hint, accepted!

dkg/sync/server.go
```diff
@@ -170,26 +170,42 @@ func (s *Server) setShutdown(pID peer.ID) {
...
+	
+	if hasCurrentPeerStep && step > currentPeerStep+1 {
+		return errors.New("peer reported step is ahead the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
+	}
+
+	if step < 0 || step > 1 {
+		return errors.New("peer reported abnormal initial step, expected 0 or 1", z.Int("peer_step", step))
+	}
+
```
xenowits, line 191: Instead of branching out, I suggest "return early". That's the general guideline we follow across the codebase.

So, suggest this instead:
```suggestion
	if hasCurrentPeerStep && step < currentPeerStep {
		return errors.New("peer reported step is behind the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
	}

	if hasCurrentPeerStep && step > currentPeerStep+1 {
		return errors.New("peer reported step is ahead the last known step", z.Int("peer_step", step), z.Int("last_step", currentPeerStep))
	}

	if step < 0 || step > 1 {
		return errors.New("peer reported abnormal initial step, expected 0 or 1", z.Int("peer_step", step))
	}
```

dkg/sync/client.go
xenowits: @pinebit Absolutely agree that `RWMutex`es are more optimized. But wondering if we necessarily require this. In charon, we have tried to refrain from doing "premature optimizations" wherever possible, hence you would see more mutexes used than rw-mutexes.

By what magnitude do you believe using `RWMutex` can optimize the performance here?

Else, we can discuss this async and create a ticket to refactor mutexes in the codebase to use `RWMutex` in a separate PR

xenowits (approved):

dkg/sync/client.go
pinebit: This is more about training muscle memory and develop good habits to use the appropriate API. Not necessary to optimize the performance in this particular case. Also, this improves readability - you show an intention. For example if I see RLock(), I can quickly realize this function should not modify the state and then I will think twice when introducing a change.

xenowits: > For example if I see RLock(), I can quickly realize this function should not modify the state and then I will think twice when introducing a change.
Yes precisely, this removes `congnitive load`!

PR #2798, opened 2024-01-17
dkg/bcast: exchange full messages

gsora:
Instead of passing around hashes, share the whole message with the target for them to validate.

category: refactor
ticket: none


dkg/bcast/impl_test.go
```diff
@@ -70,10 +72,20 @@ func TestBCast(t *testing.T) {
 			return nil
 		}
 
+		checkMessage := func(ctx context.Context, peerID peer.ID, msgAny *anypb.Any) error {
```
pinebit, line 75: Looks like `ctx` and `peerID` are not used and can be muted `_`.

dkg/bcast/server.go
```diff
@@ -48,32 +49,41 @@ type dedupKey struct {
...
+	fn, found := s.msgIDFuncs[msgID]
 
 	return fn, found
 }
 
-func (s *server) registerCallback(msgID string, cb Callback) {
-	s.callbacksMutex.Lock()
-	defer s.callbacksMutex.Unlock()
+func (s *server) registerMessageIDFuncs(msgID string, cb Callback, cm CheckMessage) {
+	s.msgIDFuncsMutex.Lock()
```
pinebit, line 80: I am wondering why don't we use `RWLock`s?

dkg/bcast/server.go
```diff
@@ -92,18 +102,32 @@ func (s *server) dedupHash(pID peer.ID, msgID string, hash []byte) error {
...
 	}
 
+	fn, found := s.getMessageIDFunc(req.Id)
+	if !found {
+		return nil, false, errors.New("unknown message id", z.Str("message_id", req.Id))
+	}
+
+	reqMessageHash, err := s.hashFunc(req.Message)
+	if err != nil {
+		return nil, false, errors.Wrap(err, "hash any")
```
pinebit, line 118: Weird unclear error message.

dkg/dkgpb/v1/bcast.proto
```diff
@@ -8,7 +8,7 @@ option go_package = "github.com/obolnetwork/charon/dkg/dkgpb/v1";
 
 message BCastSigRequest {
   string id   = 1;
-  bytes  hash = 2;
```
pinebit, line 11: Can't we just mark it as `[deprecated = true]` to avoid reusing the field in a wrong way in the future?
Or, we can make it a reserved field: https://protobuf.dev/programming-guides/proto3/#fieldreserved

dkg/frostp2p.go
```diff
@@ -70,7 +76,45 @@ func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, bcastComp
...
 		round2CastsRecv: round2CastsRecv,
+	}, nil
+}
+
+// newCheckMsg returns a bcast.CheckMessage function for round 1 and 2.
+func newCheckMsg(messageID string) (bcast.CheckMessage, error) {
+	found := false
+	for _, mID := range frostMessageIDs() {
+		if mID == messageID {
+			found = true
```
pinebit, line 87: `break`?

dkg/frostp2p.go
```diff
@@ -70,7 +76,45 @@ func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, bcastComp
 		round1CastsRecv: round1CastsRecv,
 		round1P2PRecv:   round1P2PRecv,
 		round2CastsRecv: round2CastsRecv,
+	}, nil
+}
+
+// newCheckMsg returns a bcast.CheckMessage function for round 1 and 2.
+func newCheckMsg(messageID string) (bcast.CheckMessage, error) {
```
dB2510, line 83: so here are we checking if the message is of the correct format to validate?

dkg/frostp2p.go
```diff
@@ -70,7 +76,45 @@ func newFrostP2P(tcpNode host.Host, peers map[peer.ID]cluster.NodeIdx, bcastComp
...
+				r := pb.FrostRound2Casts{}
+				return &r
+			default:
+				return nil // Note: impossible due to the above check, but needed because compilers.
+			}
+		}
+
+		target := targetFn(messageID)
+
+		err := msgAny.UnmarshalTo(target)
```
dB2510, line 111: this is still an incomplete solution as a malicious peer can still broadcast a malformed message in the correct `pb.FrostRount1Casts`. They can still be able to get the sign on this data from other peers.

dkg/bcast/server.go
gsora: Considering this mutex is only used at initialization phase, and that is essentially always only read, I don't see a point to use RWMutex here.

Generally though, I agree we should check whether we can use RWMutex widely.

dkg/bcast/server.go
gsora: That's the standard way we do errors: https://github.com/ObolNetwork/charon/blob/main/docs/goguidelines.md#error-handling

The wrapped error will eventually have enough context, the user-facing error will be much clearer and contain more context.

dkg/dkgpb/v1/bcast.proto
gsora: Good call! I'll reserve 2.

dkg/frostp2p.go
gsora: Correct.

dkg/frostp2p.go
gsora: They don't control the input anymore though, i.e. it's not just a slice of bytes.

What do you suggest we check?

dkg/bcast/server.go
```diff
@@ -92,18 +102,32 @@ func (s *server) dedupHash(pID peer.ID, msgID string, hash []byte) error {
...
+	}
+
+	if err := fn.checkMessage(ctx, pID, req.Message); err != nil {
+		return nil, false, errors.Wrap(err, "signature request message check")
+	}
```
dB2510, line 123: nit: move this before hashing the message to sign

dB2510 (approved):
LGTM

dkg/frostp2p.go
dB2510: not sure if we can check each field in `pb.FrostRound1Cast`

pinebit (approved):

dkg/bcast/server.go
pinebit: My concern was about the wording "hash any" that does not make a big sense to me..

dkg/bcast/server.go
gsora: Done

PR #2801, opened 2024-01-18
core: removed unused allowlist and denylist flags

pinebit:
Removed unused p2p's allowlist and denylist flags to avoid user confusion and false assumptions.

category: bug
ticket: none


OisinKyne:
Not directly related to this PR but because I saw it in the diff I'll flag here. We really really need at the bare minimum need a second relay default in charon. I think we should include one for an 0.19.1 fix

dB2510 (approved):

xenowits (approved):

PR #2806, opened 2024-01-19
app: eth2wrap best selector refactoring

pinebit:
A quick refactor of eth2wrap to avoid panic() call, to remove extra loops and to remove ambiguity around address<->index translation.

category: refactor
ticket: none

### DIFF
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -272,6 +272,7 @@ func provide[O any](ctx context.Context, clients []Client,
 			if bestSelector != nil {
 				bestSelector.Increment(res.Input.Address())
 			}
+
 			return res.Output, nil
 		} else {
 			nokResp = res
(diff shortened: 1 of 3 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -114,17 +114,17 @@ type multi struct {
...
-	m.selector.Increment(i)
-}
-
 func (multi) Name() string {
 	return "eth2wrap.multi"
 }
 
 func (m multi) Address() string {
-	return m.clients[m.selector.Best()].Address()
+	address, ok := m.selector.BestAddress()
```
gsora, line 122: In what case could `BestAddress` return `ok == false`?

gsora (approved):

app/eth2wrap/eth2wrap.go
pinebit: In theory this should never happen, because even after `period` elapsed, we always add at least one entry to the map. So, this is a safeguard in case somebody changes the logic in `Increment` and not look into this function.

xenowits (approved):

PR #2813, opened 2024-01-23
*: more warnings on HTTP URLs

gsora:
Warn if relay or definition file URLs are HTTP.

category: refactor
ticket: none

### DIFF
--- a/dkg/disk.go
+++ b/dkg/disk.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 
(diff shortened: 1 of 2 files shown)


xenowits (approved):

dkg/disk.go
```diff
@@ -34,8 +35,14 @@ func loadDefinition(ctx context.Context, conf Config) (cluster.Definition, error
 
 	// Fetch definition from URI or disk
 
+	parsedURL, err := url.ParseRequestURI(conf.DefFile)
+
 	var def cluster.Definition
-	if validURI(conf.DefFile) {
+	if err == nil && parsedURL.Host != "" {
+		if strings.HasPrefix(parsedURL.Scheme, "https") {
```
pinebit, line 42: Reversed the condition?

pinebit (changes requested):
Wrong condition

dkg/disk.go
xenowits: So, we have in the validURI function:
```go
...
return err == nil && (u.Scheme == "http" || u.Scheme == "https") && u.Host != ""
```

The changes in this PR just uses this condition except the `u.Scheme` condition as that is required for adding a warn log logic

gsora: I believe @pinebit is referring to the HasPrefix check

xenowits: okay, great

apologies, i misunderstood

pinebit (approved):

PR #2814, opened 2024-01-23
gomod: update Herumi BLS to v1.33.0

gsora:
In order to support the latest version of the Herumi BLS library Charon links against, we had to update our Docker container to build and run on Debian Bookworm.

category: fixbuild
ticket: none

Closes #2814.


pinebit (approved):

dB2510 (approved):
Just curious what is the change in image size because of this change?

gsora:
It's about 86mb bigger.

PR #2818, opened 2024-01-24
github/workflows: fuzz tests on nightly

gsora:
Move fuzz tests on nightly tests rather than running them in each PR.

This should reduce our runtime greatly.

Also remove unused Discord dependency.

category: misc
ticket: none


pinebit (approved):

xenowits (approved):

dB2510 (approved):
Are we notified on slack like we were on discord?

gsora:
I'm not sure, but considering we migrated away from discord and that you get emails when builds fail, we should remove it regardless.

PR #2830, opened 2024-01-26
p2p: fixed flakey TestWithSendTimeout

pinebit:
Fixed flakey `TestWithSendTimeout`.

category: bug
ticket: none

### DIFF
--- a/p2p/sender_test.go
+++ b/p2p/sender_test.go
@@ -44,15 +44,18 @@ func TestWithSendTimeout(t *testing.T) {
 	client.Peerstore().AddAddrs(server.ID(), server.Addrs(), time.Hour)
 
 	protocolID := protocol.ID("testprotocol")
+	sendTimeout := time.Millisecond
 	p2p.RegisterHandler("test", server, protocolID, func() proto.Message { return new(pbv1.Duty) },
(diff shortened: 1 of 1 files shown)


p2p/sender_test.go
```diff
@@ -44,15 +44,18 @@ func TestWithSendTimeout(t *testing.T) {
 	client.Peerstore().AddAddrs(server.ID(), server.Addrs(), time.Hour)
 
 	protocolID := protocol.ID("testprotocol")
+	sendTimeout := time.Millisecond
 	p2p.RegisterHandler("test", server, protocolID, func() proto.Message { return new(pbv1.Duty) },
 		func(ctx context.Context, peerID peer.ID, req proto.Message) (proto.Message, bool, error) {
+			// The delay must be much greater than the send timeout to trigger the deadline error.
```
xenowits, line 50: What's the send timeout?

xenowits (approved):

p2p/sender_test.go
pinebit: See `github.com/libp2p/go-yamux/v4@v4.0.1/stream.go` for details.
It creates a timer in "future" and then `select` takes either message or a timer signal. If the test runs fast (the reason for being flakey) `select` chooses the message and skips the timer event in the first pass. Then it closes the stream. That's how I understood the problem..

Issue #2728, opened 2023-11-28
create new cluster lock version v1.8


xenowits:
# 🎯 Problem to be solved

We want to add the partial deposits feature in charon.

Partial deposits is the feature where operators can create clusters with arbitrary validator deposit amounts (not necessarily 32ETH). 

For example, an entity would like to create a validator with 3 partial deposit data files: 20ETH, 10ETH & 2 ETH. We don’t support this functionality now and only create full deposit data of 32ETH per validator.

# 🛠️ Proposed solution
- [ ✅ ] Approved design doc: https://docs.google.com/document/d/1S2HrktE8IXvY6hika8pWPbzMiACbRobEONL1l7El4c4/edit#heading=h.alakie8lxcfb
- [ ✅ ] Core team consensus on the proposed solution

* Add a new version v1.8 for cluster locks.
* Add a new field `DepositAmounts` to cluster definition (`cluster/definition.go`):
```json
{
...
   "threshold":3,
   "deposit_amounts":[		<-- NEW
"200000000000",	// 20 ETH
"100000000000",	// 10 ETH
"20000000000"	// 02 ETH
],
   "validators":[
      {
         "fee_recipient_address":"0x52f...74d",
         "withdrawal_address":"0x818...2c4",		
      },
      {
         "fee_recipient_address":"0xeb9...a25",
         "withdrawal_address":"0x5fb...e7c",
      }
   ],
   "dkg_algorithm":"default"
...
}
```
* See the constraints for the `DepositAmounts` field [here](https://docs.google.com/document/d/1S2HrktE8IXvY6hika8pWPbzMiACbRobEONL1l7El4c4/edit#heading=h.mmo04ceqztco).

# 🧪 Tests
- [x] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #2839, opened 2024-01-30
cluster: partial deposits output

pinebit:
This PR enables writing of multiple deposit files in according with partial amount values.
Also, this enables v1.8.0 to be the default one.

category: feature
ticket: #2728


cmd/addvalidators.go
```diff
@@ -244,7 +244,7 @@ func writeDepositDatas(ctx context.Context, clusterDir string, numOps int, secre
 
 	var depositDatas []eth2p0.DepositData
 	for i, val := range vals {
-		depositMsg, err := deposit.NewMessage(eth2p0.BLSPubKey(val.PublicKey), val.WithdrawalAddress, deposit.MaxValidatorAmount)
+		depositMsg, err := deposit.NewMessage(eth2p0.BLSPubKey(val.PublicKey), val.WithdrawalAddress, deposit.MaxDepositAmount)
```
gsora, line 247: Does this mean that when manually adding validators we can't specify partial deposits?

cmd/createcluster.go
```diff
@@ -558,27 +574,39 @@ func generateKeys(numDVs int) ([]tbls.PrivateKey, error) {
...
+	// The loop across partial amounts
+	for i := range depositDatas {
+		// Serialize the deposit data into bytes
+		bytes, err := deposit.MarshalDepositData(depositDatas[i], network)
 		if err != nil {
-			return errors.Wrap(err, "write deposit data")
+			return err
+		}
+
+		filename := "deposit-data.json"
```
gsora, line 598: I wonder if it's worth differentiating file names when there are multiple deposit amounts vs a single one.

I think it's fine keeping it the same, i.e. when there's a single deposit amount, create a file called `deposit-data-0-32eth.json`.

@dB2510 @xenowits thoughts?

cmd/createcluster.go
```diff
@@ -672,17 +687,33 @@ func getValidators(
 			return nil, errors.Wrap(err, "builder registration to cluster object")
 		}
 
+		var partialDepositData []cluster.DepositData
+
+		// The loop over partial amounts to collect PartialDepositData
+		for i := range depositDatas {
+			for j, dd := range depositDatas[i] {
+				if [48]byte(dd.PublicKey) != dv {
```
gsora, line 695: Should we error here if we never find a pubkey == dv?

dkg/dkg.go
```diff
@@ -109,8 +109,8 @@ func Run(ctx context.Context, conf Config) (err error) {
 	}
 
 	// This DKG only supports a few specific config versions.
-	if def.Version != "v1.6.0" && def.Version != "v1.7.0" {
-		return errors.New("only v1.6.0 and v1.7.0 cluster definition version supported")
+	if def.Version != "v1.6.0" && def.Version != "v1.7.0" && def.Version != "v1.8.0" {
```
gsora, line 112: Wonder if we should only support v1.8.0?

cmd/createcluster.go
```diff
@@ -156,6 +156,11 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		conf.NumDVs = len(secrets)
 	}
 
+	if len(conf.DepositAmounts) == 0 {
+		// If partial deposit amounts were not specified, default to single amount of 32ETH.
+		conf.DepositAmounts = []int{int(deposit.MaxDepositAmount / deposit.OneEthInGwei)}
```
dB2510, line 161: can just be?
```suggestion
		conf.DepositAmounts = []int{32}
```

cmd/createcluster.go
```diff
@@ -164,6 +169,10 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 			return err
 		}
 
+		if len(def.DepositAmounts) == 0 {
+			def.DepositAmounts = deposit.EthsToGweis(conf.DepositAmounts)
```
dB2510, line 173: we can't modify a definition object that is loaded from a file as it will invalidate the definition hash in the def object

dB2510 (changes requested):

cmd/createcluster.go
pinebit: Yes, but I wanted to be explicit here.

cmd/addvalidators.go
pinebit: The spec did not mention this. Shall we support this command as well?

dkg/dkg.go
pinebit: This breaks tests like `TestDKG/frost_v16` that aimed for older versions.

cmd/createcluster.go
pinebit: I understand this would change the hash, but I see in other places we do this, e.g. https://github.com/ObolNetwork/charon/blob/3e12d09b357b1679788a517ba29836f05eff670b/cmd/createcluster.go#L215
Anyway for extra safety I reworked this.

cmd/createcluster.go
pinebit: Fixed

cmd/addvalidators.go
pinebit: If we should support this and update manifest schema, then I will leave it for another PR, otherwise this one will be too big.

pinebit: @xenowits confirmed this has to be done later for other commands as well under a different ticket

cluster/version.go
```diff
@@ -5,11 +5,11 @@ package cluster
 import "testing"
 
 const (
-	currentVersion = v1_7
+	currentVersion = v1_8
```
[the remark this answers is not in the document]
pinebit: Done

cmd/createcluster.go
```diff
@@ -230,7 +230,13 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return err
 	}
 
-	depositDatas, err := createDepositDatas(def.WithdrawalAddresses(), network, secrets)
+	depositAmounts := deposit.EthsToGweis(conf.DepositAmounts)
```
dB2510, line 233: `conf.DepositAmounts` will be nil if `conf.DefFile` is set. I suggest to set the `conf.DepositAmounts` from def object if `conf.DefFile` is provided considering def is of v1.8.0 version.

cmd/createcluster.go
```diff
@@ -672,17 +684,39 @@ func getValidators(
 			return nil, errors.Wrap(err, "builder registration to cluster object")
 		}
 
+		var partialDepositData []cluster.DepositData
+
+		// The loop over partial amounts to collect PartialDepositData
+		for i := range depositDatas {
```
dB2510, line 690: what if `depositDatas` looks like `map[pubkey][]eth2p0.DepositData` where pubkey is the dv pubkey? Would this code segment looked more clean?

cmd/createcluster.go
```diff
@@ -374,48 +378,56 @@ func detectNodeDirs(clusterDir string, nodeAmount int) error {
 	return nil
 }
 
-// signDepositDatas returns Distributed Validator pubkeys and deposit data signatures corresponding to each pubkey.
-func signDepositDatas(secrets []tbls.PrivateKey, withdrawalAddresses []string, network string) ([]eth2p0.DepositData, error) {
+// signDepositDatas returns array of []DepositData for each partial amount.
```
xenowits, line 381:
```suggestion
// signDepositDatas returns a list of DepositData for each partial amount.
```

cmd/createcluster.go
```diff
@@ -374,48 +378,56 @@ func detectNodeDirs(clusterDir string, nodeAmount int) error {
...
+func signDepositDatas(secrets []tbls.PrivateKey, withdrawalAddresses []string, network string, depositAmounts []eth2p0.Gwei) ([][]eth2p0.DepositData, error) {
 	if len(secrets) != len(withdrawalAddresses) {
 		return nil, errors.New("insufficient withdrawal addresses")
 	}
+	if len(depositAmounts) == 0 {
+		return nil, errors.New("deposit amounts must not be empty")
```
xenowits, line 387: nit
```suggestion
		return nil, errors.New("empty deposit amounts")
```

cmd/createcluster.go
```diff
@@ -558,27 +570,40 @@ func generateKeys(numDVs int) ([]tbls.PrivateKey, error) {
 }
 
 // createDepositDatas creates a slice of deposit datas using the provided parameters and returns it.
-func createDepositDatas(withdrawalAddresses []string, network string, secrets []tbls.PrivateKey) ([]eth2p0.DepositData, error) {
+func createDepositDatas(withdrawalAddresses []string, network string, secrets []tbls.PrivateKey, depositAmounts []eth2p0.Gwei) ([][]eth2p0.DepositData, error) {
 	if len(secrets) != len(withdrawalAddresses) {
 		return nil, errors.New("insufficient withdrawal addresses")
 	}
+	if len(depositAmounts) == 0 {
+		return nil, errors.New("deposit amounts must not be empty")
```
xenowits, line 578: nit (same as above):
```suggestion
		return nil, errors.New("empty deposit amounts")
```

xenowits (approved):
LGTM!

cmd/createcluster.go
pinebit: Reworked.

cmd/createcluster.go
pinebit: Reworked.

dkg/dkg.go
```diff
@@ -109,8 +109,8 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+		return errors.New("only v1.6.0, v1.7.0 and v1.8.0 cluster definition version supported")
```
OisinKyne, line 113:
```suggestion
		return errors.New("only v1.6.0, v1.7.0 and v1.8.0 cluster definition versions supported")
```

cmd/testdata/TestCreateCluster_four_partial_deposits_files.golden
```diff
@@ -5,18 +5,30 @@
  "node3",
  "node0/charon-enr-private-key",
  "node0/cluster-lock.json",
- "node0/deposit-data.json",
+ "node0/deposit-data-0-8eth.json",
+ "node0/deposit-data-1-8eth.json",
+ "node0/deposit-data-2-8eth.json",
+ "node0/deposit-data-3-8eth.json",
```
OisinKyne, line 11: Is there a reason to create duplicates files here? Shouldn't the be all the same? We get people to specify the amounts in `deposit_amounts` mostly as a safety to make sure they don't create too few deposits, but in my opinion we should only do one deposit signing process and output one deposit file for each unique amount, as they are deterministic and identical copies. Otherwise I fear people will have to sweat more about activating exactly the correct files when it doesn't matter in practice.

cmd/createcluster.go
```diff
@@ -852,7 +885,7 @@ func writeOutput(out io.Writer, splitKeys bool, clusterDir string, numNodes int,
 	_, _ = sb.WriteString(fmt.Sprintf("├─ node[0-%d]/\t\t\tDirectory for each node\n", numNodes-1))
 	_, _ = sb.WriteString("│  ├─ charon-enr-private-key\tCharon networking private key for node authentication\n")
 	_, _ = sb.WriteString("│  ├─ cluster-lock.json\t\tCluster lock defines the cluster lock file which is signed by all nodes\n")
-	_, _ = sb.WriteString("│  ├─ deposit-data.json\t\tDeposit data file is used to activate a Distributed Validator on DV Launchpad\n")
+	_, _ = sb.WriteString("│  ├─ deposit-data-*.json\tDeposit data files are used to activate a Distributed Validator on DV Launchpad\n")
```
OisinKyne, line 888:
```suggestion
	_, _ = sb.WriteString("│  ├─ deposit-data-*.json\tDeposit data files are used to activate a Distributed Validator on the DV Launchpad\n")
```

dB2510 (approved):
LGTM!

cluster/definition.go
```diff
@@ -55,7 +56,7 @@ func WithLegacyVAddrs(feeRecipientAddress, withdrawalAddress string) func(*Defin
 // NewDefinition returns a new definition populated with the latest version, timestamp and UUID.
 // The hashes are also populated accordingly. Note that the hashes need to be recalculated when any field is modified.
 func NewDefinition(name string, numVals int, threshold int, feeRecipientAddresses []string, withdrawalAddresses []string,
-	forkVersionHex string, creator Creator, operators []Operator, random io.Reader, opts ...func(*Definition),
+	forkVersionHex string, creator Creator, operators []Operator, depositAmounts []int, random io.Reader, opts ...func(*Definition),
```
OisinKyne, line 59: Does the fact that we're accepting `int`s here mean no one can make a deposit for e.g 3.5 ether via the CLI? (Probably okay for a CLI restriction once we specify it in the flag description/help text, we can just tell people to craft a definition with the SDK if needs be? (though our createCluster Payload will have to be in Gwei not eth if that is the case))

OisinKyne (commented):
I think after taking input and validation and what not, when we move to the crafting/signing phase, we deduplicate identical deposit amounts and only produce one deposit for each amount, negating the need for an index in the filename, and reducing the amount of signatures to be shared during DKG and making things look simpler. :)

cluster/definition.go
pinebit: > Does the fact that we're accepting ints here mean no one can make a deposit for e.g 3.5 ether via the CLI?
Correct. To avoid floating points weirdness as we need to be very precise here. We can describe in the documentation the users can leverage SDK when Gweis are the input.

cmd/testdata/TestCreateCluster_four_partial_deposits_files.golden
pinebit: Implemented as suggested.

Issue #2840, opened 2024-01-30
Implement POST get validators endpoint


xenowits:
# 🎯 Problem to be solved

A [new](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/postStateValidators) endpoint has been recently added to the `beacon APIs` to query validators from the beacon chain
```
POST /eth/v1/beacon/states/{state_id}/validators
```

Note that we already have an endpoint for the same purpose of querying validators, but it's a `GET` that takes in validator pubkeys/indices in a query parameter
```
GET /eth/v1/beacon/states/{state_id}/validators
```

# 🛠️ Proposed solution

Implement the new POST endpoint in `core/validatorapi`.

This will probably require us to read the bytes in the request body (`router.go`):
```
...
func getValidators(p eth2client.ValidatorsProvider) handlerFunc {
	return func(ctx context.Context, params map[string]string, query url.Values, _ contentType, _ []byte) (any, http.Header, error) {
		stateID := params["state_id"]

		resp, err := getValidatorsByID(ctx, p, stateID, getValidatorIDs(query)...)
...
```

So, one approach might be to:
1. Read query parameters for validator pubkeys/indices
2. If no query parameters provided, read request body
3. If query parameters provided, don't read request body

# 🧪 Tests
- [x] Tested by new automated unit/integration/smoke tests
- [x] Manually tested on core team/canary/test clusters
- [x] Manually tested on local compose simnet

dB2510:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2840?planning-poker) with Zenhub @dB2510 @gsora @pinebit @xenowits

pinebit:
Tested

PR #2849, opened 2024-02-01
validatorapi: supporting getValidators with post

pinebit:
Implemented POST get validators endpoint.

category: feature
ticket: #2840

### DIFF
--- a/core/validatorapi/router_internal_test.go
+++ b/core/validatorapi/router_internal_test.go
@@ -320,6 +320,71 @@ func TestRawRouter(t *testing.T) {
 		testRawRouter(t, handler, callback)
 	})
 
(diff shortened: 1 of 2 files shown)


core/validatorapi/router_internal_test.go
```diff
@@ -320,6 +320,71 @@ func TestRawRouter(t *testing.T) {
 		testRawRouter(t, handler, callback)
 	})
 
+	t.Run("get validators with post", func(t *testing.T) {
```
dB2510, line 323: can you add one more test where neither query params are provided nor request body is provided? In this case it should respond with the data of all the validators.

dB2510 (approved):

xenowits (approved):

core/validatorapi/router_internal_test.go
pinebit: My test was designed to not return any results if indexes are empty, but now fixed that to match the expected behavior and added another test.

Issue #2851, opened 2024-02-05
create cluster doesn't error when run in same directory


xenowits:
# 🐞 Bug Report

### Description

The `create cluster` command has a bug that results in no error being logged if the same command is run twice in the same directory. Interestingly, this happens only when the cluster definition file is loaded via network or disk.

### Has this worked before in a previous version?

No

## 🔬 Minimal Reproduction

Run the following command in the same directory:
```
docker run -u $(id -u):$(id -g) --rm -v "$(pwd)/:/opt/charon" obolnetwork/charon:v0.19.1-rc1 create cluster --definition-file="https://obol-api-dev.gcp.obol.tech/dv/0xdf8b633657c35f21170b30aecf0033e4b3369f446778c64ac0cb76e36b8d380c" --network=holesky --cluster-dir="cluster" --publish --publish-address="https://obol-api-dev.gcp.obol.tech/"
```

## 🔥 Error

<pre><code>
10:05:20.674 ERRO cmd        Fatal error: write keystore: open cluster/node0/validator_keys/keystore-1.json: permission denied {"filename": "cluster/node0/validator_keys/keystore-1.json"}
        eth2util/keystore/keystore.go:93 .func1
        app/forkjoin/forkjoin.go:195 .func2

</code></pre>

Expected error
```
existing node directory found, please delete it before running this command
```

## 🌍  Your Environment

MacOS, M1

**What version of Charon are you running? (Which release)**

`v0.19.1-rc1`

**Anything else relevant (validator index / public key)?**

PR #2852, opened 2024-02-05
cmd: fix create cluster bug

xenowits:
Fix bug in `create cluster` command.

--

From @xenowits's explanation on slack:

> When create cluster command is run twice in the same directory, it doesn’t error out with “Fatal error: existing node directory found, please delete it before running this command” while it should

> Explanation: When we provide a definition file to the create cluster command, we don’t set the "conf.NumNodes" field when verifying the config

```go
//.....
func validateCreateConfig(ctx context.Context, conf clusterConfig) error {
    if conf.NumNodes == 0 && conf.DefFile == "" { // if there's a definition file, infer this value from it later
       return errors.New("missing --nodes flag")
    }

    // Check for valid network configuration.
    if err := validateNetworkConfig(conf); err != nil {
       return errors.Wrap(err, "get network config")
    }

    // BUG: conf.NumNodes is zero here!!
    if err := detectNodeDirs(conf.ClusterDir, conf.NumNodes); err != nil {
       return err
    }
```

category: bug 
ticket: #2851

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -137,6 +137,16 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		conf.Network = eth2util.Goerli.Name
 	}
 
+	var def cluster.Definition
+	if conf.DefFile != "" { // Load definition from DefFile
+		def, err = loadDefinition(ctx, conf.DefFile)
+		if err != nil {
+			return err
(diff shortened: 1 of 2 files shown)


gsora:
Could you drop a couple lines on how this PR fixes the specified bug?

gsora (approved):

PR #2866, opened 2024-02-07
core/validatorapi: handle HTTP methods

gsora:
Instead of handling all HTTP methods for all validator API endpoints, only support a subset of them for each.

Selected the appropriate HTTP method by matching them from the Beacon Node API spec: https://ethereum.github.io/beacon-APIs/#/.

category: bug
ticket: none

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -86,142 +86,172 @@ func NewRouter(ctx context.Context, h Handler, eth2Cl eth2wrap.Client) (*mux.Rou
 		Name    string
 		Path    string
 		Handler handlerFunc
+		Methods []string
 	}{
 		{
 			Name:    "attester_duties",
 			Path:    "/eth/v1/validator/duties/attester/{epoch}",
 			Handler: attesterDuties(h),
+			Methods: []string{http.MethodPost},
 		},
 		{
 			Name:    "proposer_duties",
 			Path:    "/eth/v1/validator/duties/proposer/{epoch}",
 			Handler: proposerDuties(h),
+			Methods: []string{http.MethodGet},
 		},
 		{
 			Name:    "sync_committee_duties",
 			Path:    "/eth/v1/validator/duties/sync/{epoch}",
 			Handler: syncCommitteeDuties(h),
+			Methods: []string{http.MethodPost},
 		},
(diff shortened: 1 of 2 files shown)


core/validatorapi/router.go
```diff
@@ -86,142 +86,172 @@ func NewRouter(ctx context.Context, h Handler, eth2Cl eth2wrap.Client) (*mux.Rou
...
 		},
 	}
 
 	r := mux.NewRouter()
 	for _, e := range endpoints {
-		r.Handle(e.Path, wrap(e.Name, e.Handler))
+		handler := r.Handle(e.Path, wrap(e.Name, e.Handler))
+		if len(e.Methods) != 0 {
+			handler.Methods(e.Methods...)
+		}
```
xenowits, line 254:
```suggestion
		handler.Methods(e.Methods...)
```
i wonder if this could work?

pinebit (approved):
LGTM

core/validatorapi/router_internal_test.go
```diff
@@ -576,6 +576,36 @@ func TestRouter(t *testing.T) {
 		"dependent_root":       dependentRoot,
 	}
 
+	t.Run("wrong http method", func(t *testing.T) {
```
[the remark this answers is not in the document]
gsora: Correct, we must always specify the list of supported methods for each VAPI endpoint.

core/validatorapi/router.go
```diff
@@ -86,142 +86,172 @@ func NewRouter(ctx context.Context, h Handler, eth2Cl eth2wrap.Client) (*mux.Rou
 		Name    string
 		Path    string
 		Handler handlerFunc
+		Methods []string
```
dB2510, line 89: why a slice rather than a string? In this case, we will need to write a single handler which will handle all the http statuses. But if specify just one method per endpoint then we can separate handlers for each method which might be more clean than the current approach. WDYT?

gsora: Generally when designing APIs like this I found that one might want to handle the same endpoint in different HTTP methods.

dB2510 (approved):

pinebit (approved):

PR #2868, opened 2024-02-08
core/validatorapi: return 404 for propose block v3

dB2510:
Returns 404 with message `endpoint not supported` for produce block v3 endpoint to signal VC it is not supported by charon yet. Otherwise charon will proxy this request to beacon node.

category: feature
ticket: #2749

Issue #2749, opened 2023-12-12
Add support for produce block v3 endpoint


dB2510:
# 🎯 Problem to be solved
We need to add support for fallback mechanism to local beacon node in case of blinded block failures. This can be done by adding [produceBlockV3](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/ValidatorRequiredApi/produceBlockV3) endpoint to charon node. This endpoint can return full block as well as blinded block which provides fallback to local block in case if builder fails to respond on time.

# 🛠️ Proposed solution
One possible solution can be to combine `DutyProposer` and `DutyBuilderProposer` duties. The reason being our current consensus workflow does consensus round by duty type which separates consensus rounds of both of these duties. Our consensus workflow doesn't care what is the underlying type of signed data. 

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

pinebit:
This work is not relevant anymore, new ticket: #2936

PR #2870, opened 2024-02-08
core/validatorapi: return 404 for propose block v3

gsora:
Cherrypick of #2868.

--

Returns 404 with message `endpoint not supported` for produce block v3 endpoint to signal VC it is not supported by charon yet. Otherwise charon will proxy this request to beacon node.

category: feature
ticket: #2749


dB2510 (approved):

gsora:
Closing in favor of another pr.

PR #2878, opened 2024-02-13
dkg/sync: increase by max 2 steps

gsora:
Allow step increase by a maximum amount of 2 steps.

Since the sync protocol runs in the context of a distributed system, peer messages might be received in a different order than expected.

This is especially true when the nodes are e.g. writing data to disk.

This change still keeps the monotonic property of the function: step count will never decrease.

category: bug
ticket: none

### DIFF
--- a/dkg/dkg_test.go
+++ b/dkg/dkg_test.go
@@ -11,6 +11,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


pinebit (approved):

gsora:
@dB2510 @xenowits please review this PR when you have time 🙏

PR #2899, opened 2024-02-22
cmd: remove todos

xenowits:
Removes TODOs in the `cmd` package.

category: misc
ticket: none

### DIFF
--- a/cluster/manifest/helpers.go
+++ b/cluster/manifest/helpers.go
@@ -131,8 +131,6 @@ func SignK1(m *manifestpb.Mutation, secret *k1.PrivateKey) (*manifestpb.SignedMu
 }
 
 // verifyK1SignedMutation verifies that the signed mutation is signed by a k1 key.
-//
-// TODO(corver): Figure out no-verify case.
 func verifyK1SignedMutation(signed *manifestpb.SignedMutation) error {
 	pubkey, err := k1.ParsePubKey(signed.Signer)
 	if err != nil {
--- a/cmd/addvalidators_internal_test.go
+++ b/cmd/addvalidators_internal_test.go
@@ -104,7 +104,6 @@ func TestValidateConfigAddValidators(t *testing.T) {
 	}
 }
 
-// TODO(xenowits): Add more extensive tests, this is just a very simple unit test.
 func TestRunAddValidators(t *testing.T) {
 	const (
 		n        = 3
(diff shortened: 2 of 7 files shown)


cluster/manifest/helpers.go
```diff
@@ -131,8 +131,6 @@ func SignK1(m *manifestpb.Mutation, secret *k1.PrivateKey) (*manifestpb.SignedMu
 }
 
 // verifyK1SignedMutation verifies that the signed mutation is signed by a k1 key.
-//
-// TODO(corver): Figure out no-verify case.
```
xenowits, line 135: We verify k1 signatures, so we don't need a "no-verify" case

cmd/addvalidators_internal_test.go
```diff
@@ -104,7 +104,6 @@ func TestValidateConfigAddValidators(t *testing.T) {
 	}
 }
 
-// TODO(xenowits): Add more extensive tests, this is just a very simple unit test.
```
xenowits, line 107: The test is already comprehensive enough to test the "add validators" feature

In addition to that, the add validators feature is not in production, it is under the `charon alpha` command

cmd/createdkg.go
```diff
@@ -77,7 +77,6 @@ func mustMarkFlagRequired(cmd *cobra.Command, flag string) {
 
 func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
 	// Map prater to goerli to ensure backwards compatibility with older cluster definitions.
-	// TODO(xenowits): Remove the mapping later.
```
xenowits, line 80: We can remove code for goerli entirely when goerli gets deprecated in a couple of weeks

https://blog.ethereum.org/2024/01/10/goerli-dencun-announcement

cmd/relay/relay.go
```diff
@@ -41,7 +41,7 @@ type Config struct {
 	MaxResPerPeer   int
 	MaxConns        int
 	FilterPrivAddrs bool
-	RelayLogLevel   string // TODO(corver): Rename to LibP2PLogLevel.
```
xenowits, line 44: Renamed ✅

cmd/run.go
```diff
@@ -94,7 +94,6 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
 	})
 }
 
-// TODO(dhruv): add more test only flags to this function.
```
xenowits, line 97: We only have one unsafe command for now. We can always add new unsafe commands, no need for a todo for that

pinebit (approved):

PR #2915, opened 2024-02-28
cluster: remove todos

xenowits:
Remove TODOs from the `cluster` package.

category: misc
ticket: none

### DIFF
--- a/cluster/ssz.go
+++ b/cluster/ssz.go
@@ -138,7 +138,6 @@ func hashDefinitionLegacy(d Definition, hh ssz.HashWalker, configOnly bool) erro
 
 	// Field (10) 'timestamp' (optional for backwards compatibility)
 	if configOnly {
-		// TODO(corver): This is a bug, config_hash should use the same check as definitino hash below.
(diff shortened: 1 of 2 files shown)


pinebit (approved):

cluster/ssz.go
```diff
@@ -138,7 +138,6 @@ func hashDefinitionLegacy(d Definition, hh ssz.HashWalker, configOnly bool) erro
 
 	// Field (10) 'timestamp' (optional for backwards compatibility)
 	if configOnly {
-		// TODO(corver): This is a bug, config_hash should use the same check as definitino hash below.
```
xenowits, line 141: This is not a bug. It is more like a devX improvement

cluster/version.go
```diff
@@ -5,7 +5,6 @@ package cluster
 import "testing"
 
 const (
-	// TODO: bump version to 1.8 when ready.
```
xenowits, line 8: No need for this todo. We always bump cluster version when we release a new one

PR #2947, opened 2024-03-06
*: bumped go version

pinebit:
Bumped go version and updated protobuf version.

category: fixbuild
ticket: none


gsora:
do you think we should update go everywhere?

pinebit:
> do you think we should update go everywhere?

Done @gsora

gsora (approved):

PR #2951, opened 2024-03-07
*: partial deposits bugfixes

pinebit:
Several fixes for bugs identified in partial deposits implementation.
Plus, a few improvements to prevent using partial deposits for unsupported lock version.

category: bug
ticket: none

### DIFF
--- a/cluster/definition.go
+++ b/cluster/definition.go
@@ -38,6 +38,13 @@ func WithVersion(version string) func(*Definition) {
 	}
 }
 
+// WithDKGAlgorithm returns an option to set a non-default DKG algorithm in a new definition.
+func WithDKGAlgorithm(algorithm string) func(*Definition) {
+	return func(d *Definition) {
+		d.DKGAlgorithm = algorithm
+	}
+}
+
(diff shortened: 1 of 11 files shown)


cluster/definition.go
```diff
@@ -38,6 +38,13 @@ func WithVersion(version string) func(*Definition) {
 	}
 }
 
+// WithDKGAlgorithm returns an option to set a non-default DKG algorithm in a new definition.
+func WithDKGAlgorithm(algorithm string) func(*Definition) {
```
dB2510, line 42: why is this needed? we only support one dkg algo atm

dB2510 (approved):

cluster/definition.go
pinebit: In createdkg.go it populates opts.. which we need to extend WithVersion.

dB2510: but why do we need to populate the dkg algorithm?

pinebit: I think this is to respect the given configuration. I believe the implementation assumes different algorithms passed from the config. If you think this is odd, then we can have a ticket to remove this field from the config entirely.

Issue #2881, opened 2024-02-13
[CLI] DKG ceremony crash upon existing `validator_keys` folder


crisog:
# 🐞 Bug Report

### Description

This shouldn't have happened. Apparently, all the other nodes in the cluster received their key shares except me. It doesn't seem to be an obvious recovery point, the only ideas that come to head are:
1) Everyone deletes their `validator_keys` folder, which is not practical when you deal with users using dappnode and have no direct access to the OS. (This is assuming DKG can be repeated)
2) A new cluster is created

In my opinion, what should happen is: if `validator_keys` folder exists, then a new directory gets created with a random prefix, i.e. `xyz_validator_keys` and the keys are written there. This way the key shares wouldn't have been lost after a successful DKG ceremony

## 🔬 Minimal Reproduction

```
docker run -u $(id -u):$(id -g) --rm -v "$(pwd)/:/opt/charon" obolnetwork/charon:v0.18.0 dkg --definition-file="https://api.obol.tech/dv/<DV_CLUSTER>" --publish
```

## 🔥 Error

<pre><code>
15:28:54.256 INFO dkg        Connected to peer 2 of 3                 {"peer": "nervous-painter"}
15:48:55.335 INFO dkg        Connected to peer 2 of 3                 {"peer": "strange-jewelry"}
16:28:54.630 INFO dkg        Connected to peer 2 of 3                 {"peer": "nervous-painter"}
16:31:05.277 INFO dkg        Connected to peer 3 of 3                 {"peer": "shy-computer"}
16:31:06.453 INFO dkg        All peers connected, starting DKG ceremony
16:31:11.644 ERRO cmd        Fatal error: mkdir /validator_keys: mkdir .charon/validator_keys: file exists
    dkg/disk.go:128 .writeKeysToDisk
    dkg/dkg.go:315 .Run
    cmd/dkg.go:35 .func1
    cmd/cmd.go:80 .func1
    main.go:19 .main
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
 Docker Image: obolnetwork/charon:v0.18.0
  </code>
</pre>

**What version of Charon are you running? (Which release)**

<pre>
  <code>
v0.18.0
  </code>
</pre>

**Anything else relevant (validator index / public key)?**

gsora:
Hi @crisog!

First and foremost, could you try reproducing this issue with our latest release, v0.19.1?

[...]

PR #2958, opened 2024-03-13
cluster: unique validator_keys dir

pinebit:
In according with the issue description #2881, this is UX improvement - if the target validator_keys folder exists, the implementation checks if the folder is empty. If the existing directory is not empty, or in case of any IO error, the process fails.

category: feature
ticket: #2881

### DIFF
--- a/cluster/helpers.go
+++ b/cluster/helpers.go
@@ -10,6 +10,9 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"os"
+	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


OisinKyne:
I'm not sure this is better. People would struggle to figure out why their validator keys aren't matching with the expected ones because their cluster has imported the default `validator_keys/` one. What we should do is, if the `validator_keys` folder is present, we check the directory is empty, and if so, we continue without erroring. When people clean folders, they often don't think to delete the folder they just delete the files.

pinebit:
> I'm not sure this is better. People would struggle to figure out why their validator keys aren't matching with the expected ones because their cluster has imported the default `validator_keys/` one. What we should do is, if the `validator_keys` folder is present, we check the directory is empty, and if so, we continue without erroring. When people clean folders, they often don't think to delete the folder they just delete the files.

Thank you @OisinKyne, changed as you suggested. Good point.

gsora (approved):

PR #2963, opened 2024-03-14
cmd/combine: fix multiple runs 

### LINKED ISSUE #2538: Confusing error when running combine multiple times
# 🐞 Bug Report

### Description

When running the combine command again (after a previous succesful combine), it errors with a very confusing error. This makes it hard to understand what the problem is and that keys have already been combined.

### Has this worked before in a previous version?

No

## 🔬 Minimal Reproduction

Run combine once:
```
docker build -t obolnetwork/charon:local .

docker run --rm -v "$(pwd):/opt/charon" -it obolnetwork/charon:local create cluster --nodes 3 --network goerli --num-validators 10 --fee-recipient-addresses 0x3C75594181e03E8ECD8468A0037F058a9dAfad79 --withdrawal-addresses 0x3C75594181e03E8ECD8468A0037F058a9dAfad79 --name "Test combine"

docker run -v "$(pwd):/opt/charon" -it obolnetwork/charon:local combine --cluster-dir ./
```
This should succeed. Then run it again which errors:
```
➜  c2 docker run -v "$(pwd):/opt/charon" -it obolnetwork/charon:local combine --cluster-dir ./

06:20:13.102 INFO cmd        Recombining private key shares           {"input_dir": "/opt/charon", "output_dir": "/opt/charon/validator_keys"}
06:20:13.244 ERRO cmd        Fatal error: cannot open manifest file: manifest load error: load dag from disk: couldn't load cluster dag from either manifest or legacy lock file {"name": "validator_keys", "error": "read manifest file: open /opt/charon/validator_keys/cluster-manifest.pb: no such file or directory", "file": "/opt/charon/validator_keys/cluster-manifest.pb"}
	cluster/manifest/load.go:65 .LoadDAG
	cluster/manifest/load.go:26 .LoadCluster
	cmd/combine/combine.go:229 .loadManifest
	cmd/combine/combine.go:61 .Combine
	cmd/combine.go:51 .newCombineFunc
	cmd/combine.go:28 .func1
	cmd/cmd.go:80 .func1
	main.go:19 .main
```
It errors because it cannot read a manifest file from the newly created `validator_keys` folder since it thinks that is a "source input" folder while it is the "output" folder. Suggest checking if the output folder exists and contains keys. Note that we do already do this check, so we can just move this check up to before we attempt to load manifests.  The problem is that `--force` skips this checks when then results in this same error. So we need a better/smarter check for `--force` (or just remove `--force`).

gsora:
Instead of checking if `validator_keys` directory exists at the end of the load manifest loop, do it earlier and exclude potential non-Charon node directories.

Also make so that `--force` deletes the existing keys, if any.

Solves the use case in which the user called `combine` command twice with an input directory that was also used as an output directory.

category: bug
ticket: #2538 

Closes #2538

### DIFF
--- a/cmd/combine/combine.go
+++ b/cmd/combine/combine.go
@@ -146,6 +146,26 @@ func Combine(ctx context.Context, inputDir, outputDir string, force, noverify bo
 		return errors.New("refusing to overwrite existing private key share", z.Str("path", ksPath))
 	}
 
+	if force {
+		matches, err := filepath.Glob(filepath.Join(outputDir, "keystore-*.json"))
+		if err != nil {
+			return errors.Wrap(err, "can't match keystore files")
+		}
+
+		passMatches, err := filepath.Glob(filepath.Join(outputDir, "keystore-*.txt"))
+		if err != nil {
+			return errors.Wrap(err, "can't match keystore password files")
+		}
+
+		matches = append(matches, passMatches...)
+
+		for _, match := range matches {
+			if err := os.RemoveAll(match); err != nil {
+				return errors.Wrap(err, "can't remove existing keystore file")
+			}
+		}
+	}
+
 	if err := o.keyStoreFunc(combinedKeys, outputDir); err != nil {
 		return errors.Wrap(err, "cannot store keystore")
 	}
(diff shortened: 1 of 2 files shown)


pinebit (approved):
LGTM

Issue #2446, opened 2023-07-18
Health check for too many metrics


corverroos:
# 🎯 Problem to be solved
We have had a problem of prometheus metrics using to many labels (high cardinality) which results in memory leak in charon and overloading of the prometheus infra.

We need to monitor and alert when metric labels grows too much. 

# 🛠️ Proposed solution
Two step solution:
- Add a warning health-check that alerts if there are more than `X*validators` metrics in a metrics family. Suggest using X=100 as a start. 
- Add a central monitoring pager duty alert (similar to high memory usage alet) for this check

pinebit:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2446?planning-poker) with Zenhub @gsora

pinebit:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2446?planning-poker) with Zenhub @KaloyanTanev

PR #2967, opened 2024-03-15
app/health: app_health_metrics_high_cardinality

pinebit:
This is to monitor metrics high cardinality, when a given metric gets too many distinct labels which results in high memory consumption in charon.
To this end this PR introduces one new metric: `app_health_metrics_high_cardinality` (gauge) which renders metric names => max labels count, if this metric exceeded the threshold (100 * num_of_validators).
In addition to the above metric which can be used in alerting, this also adds a new "health check" named `metrics_high_cardinality` which is triggered when `app_health_metrics_high_cardinality` reported any offense.

Note: `app_health_metrics_high_cardinality` will not "reset", because the purpose of this feature is to detect and signal an opportunity of memory leak.

category: feature
ticket: #2446

### DIFF
--- a/app/health/checker.go
+++ b/app/health/checker.go
@@ -20,31 +20,34 @@ const (
 	scrapePeriod = 30 * time.Second
 	// maxScrapes is the maximum number of scrapes to keep.
 	maxScrapes = 10
-	// lables cardinality threshold.
+	// labelsCardinalityThreshold is the threshold for single validator;
+	// for N validators, the threshold is N * labelsCardinalityThreshold.
 	labelsCardinalityThreshold = 100
 )
 
 // NewChecker returns a new health checker.
-func NewChecker(metadata Metadata, gatherer prometheus.Gatherer) *Checker {
+func NewChecker(metadata Metadata, gatherer prometheus.Gatherer, numValidators int) *Checker {
 	return &Checker{
-		metadata:     metadata,
-		checks:       checks,
-		gatherer:     gatherer,
-		scrapePeriod: scrapePeriod,
-		maxScrapes:   maxScrapes,
-		logFilter:    log.Filter(),
+		metadata:      metadata,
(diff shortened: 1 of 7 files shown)


pinebit:
Tested with compose (simnet) cluster.

gsora:
Is this metric impacted by the amount of validators in the cluster?

pinebit:
> Is this metric impacted by the amount of validators in the cluster?

There is no direct dependency on the number of validators. If the corresponding labels span for each validator, then it will depend on the number. However, the threshold may vary based on the number... can you suggest a formula?

gsora (approved):

app/health/checker.go
```diff
@@ -91,6 +96,33 @@ func (c *Checker) scrape() error {
...
+			gatherAgain = true
+		}
+	}
+
+	if gatherAgain {
+		metrics, err = c.gatherer.Gather()
+		if err != nil {
+			return errors.Wrap(err, "gather metrics")
+		}
+	}
```
KaloyanTanev, line 124: What is the reason behind this?

pinebit: Line 114: https://github.com/ObolNetwork/charon/pull/2967/files#diff-16436fa4ead5f4fac86aa0910e7dcef9ae20ab0c2fbc25365de419227dbaac2fR114
updates one gauge, we need to re-gather to have new set of metrics including the updated one.

KaloyanTanev: Shouldn't we run then the same for loop for the families again on the re-gathered metrics?

pinebit: This will be endless loop. We make assumption that some metrics may change in between of `Gather()` calls, but this ensures, the updated gauge (line 114) used for health checks is certainly included.

KaloyanTanev (approved):

Issue #2778, opened 2024-01-10
Log the ENR pubkey and full lock hash in `app-start`


OisinKyne:
# 🎯 Problem to be solved

Small ticket, might not be important but logging it so we have something to come back to. 

A user has given me a log output where they fail to connect to relays on boot with the error: 

```json
{"log":"14:23:26.811 INFO cmd        Parsed config                            {\"beacon-node-endpoints\": \"[http://localhost:6002]\", \"builder-api\": \"false\", \"feature-set\": \"stable\", \"feature-set-disable\": \"[]\", \"feature-set-enable\": \"[]\", \"help\": \"false\", \"jaeger-address\": \"\", \"jaeger-service\": \"charon\", \"lock-file\": \".charon/cluster-lock.json\", \"log-color\": \"auto\", \"log-format\": \"console\", \"log-level\": \"info\", \"loki-addresses\": \"[]\", \"loki-service\": \"charon\", \"manifest-file\": \".charon/cluster-manifest.pb\", \"monitoring-address\": \"127.0.0.1:3620\", \"no-verify\": \"false\", \"p2p-allowlist\": \"\", \"p2p-denylist\": \"\", \"p2p-disable-reuseport\": \"false\", \"p2p-external-hostname\": \"\", \"p2p-external-ip\": \"\", \"p2p-relays\": \"[https://charon-relay.dsrvlabs.dev,https//relay-2.prod-relay.721.land,https://relay-1.obol.figment.io,https://0.relay.obol.tech]\", \"p2p-tcp-address\": \"[0.0.0.0:3610]\", \"private-key-file\": \".charon/charon-enr-private-key\", \"private-key-file-lock\": \"false\", \"simnet-beacon-mock\": \"false\", \"simnet-beacon-mock-fuzz\": \"false\", \"simnet-slot-duration\": \"1s\", \"simnet-validator-keys-dir\": \".charon/validator_keys\", \"simnet-validator-mock\": \"false\", \"synthetic-block-proposals\": \"false\", \"validator-api-address\": \"127.0.0.1:3600\"}\n","stream":"stderr","time":"2024-01-05T14:23:26.811586093Z"}

{"log":"14:23:26.811 INFO app-start  Charon starting                          {\"version\": \"v0.18.0\", \"git_commit_hash\": \"67725e4\", \"git_commit_time\": \"2023-11-10T15:18:42Z\"}\n","stream":"stderr","time":"2024-01-05T14:23:26.811751901Z"}
[...]

PR #2972, opened 2024-03-18
app: logging lock hash and enr

pinebit:
Added logging of full lock_hash and node's enr when charon starts.

category: refactor
ticket: #2778

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -203,9 +203,9 @@ func Run(ctx context.Context, conf Config) (err error) {
 	log.Info(ctx, "Lock file loaded",
 		z.Str("peer_name", p2p.PeerName(tcpNode.ID())),
 		z.Int("peer_index", nodeIdx.PeerIdx),
-		z.Str("cluster_hash", lockHashHex),
 		z.Str("cluster_name", cluster.Name),
-		z.Str("lock_hash", hex.EncodeToString(cluster.GetInitialMutationHash())),
+		z.Str("cluster_hash", lockHashHex),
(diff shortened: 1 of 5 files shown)


dB2510 (approved):

app/app.go
```diff
@@ -194,11 +195,18 @@ func Run(ctx context.Context, conf Config) (err error) {
...
 
+	enrRec, err := enr.New(p2pKey)
+	if err != nil {
+		return errors.Wrap(err, "creating enr record from privkey")
+	}
+
 	log.Info(ctx, "Lock file loaded",
 		z.Str("peer_name", p2p.PeerName(tcpNode.ID())),
 		z.Int("peer_index", nodeIdx.PeerIdx),
 		z.Str("cluster_hash", lockHashHex),
```
gsora, line 206: We already log `InitialMutationHash` here albeit in short form (only the first 7 characters), do we need a new `lock_hash` field?

The `labels` mapping at line 213 will keep referencing `lockHashHex`, is this intended behavior?

pinebit: Given the labels mapping, I assumed the existing `cluster_hash` form (short form) might be used for some o11y or alerting. Therefore I preferred to not change this. Hence, the new field which logged just once.

gsora: Makes sense.

Just for consistency sake, since the short form is called `cluster_hash`, should we rename the full field to `cluster_hash_full`?

pinebit: Agreed, done.

gsora (approved):

Issue #2995, opened 2024-03-27
Implement Network Health Check Skeleton for Charon Cluster


boulder225:
## 🎯 Problem to be solved
Develop a skeleton to test the health and connectivity of a Charon cluster before deploying the cluster to the mainnet.

## 🛠️ Proposed solution
- [ ] Implement an HTTP server within the Charon node to expose endpoints for network health checks
- [ ] Add configuration options for the health check server address and request timeout
- [ ] Develop a command-line interface (CLI) command `charon alpha network test` to trigger the health checks
- [ ] Create HTTP handler placeholders for testing peer-to-peer connectivity, beacon node status, and validator client integration
- [ ] Implement a configuration management system for storing and retrieving health check settings and parameters
- [ ] Develop a process to accumulate and evaluate the health check results, providing a comprehensive assessment
- [ ] Implement a reporting module for generating human-readable output with detailed information about the test results, scores, and recommendations
- [ ] Integrate the network test command into the existing Charon CLI structure
- [ ] Develop documentation and user guides for running and interpreting the network health checks

PR #3008, opened 2024-04-01
cmd: charon alpha test skeleton

KaloyanTanev:
Add the skeleton framework for the charon alpha test command. Prepare all the necessities required to start building different test suites. Set the first 3 envisioned categories - peers, beacon node and validator client.

N.B.: Implementation of peers, beacon node and validator client tests are out of the scope of this PR.

category: feature
ticket: #2995


cmd/test.go
```diff
@@ -0,0 +1,290 @@
...
+
+func bindTestFlags(cmd *cobra.Command, config *testConfig) {
+	cmd.Flags().StringVar(&config.OutputFile, "output-file", "", "File path to which output can be written.")
+	cmd.Flags().StringVar(&config.OutputFormat, "output-format", "json", "File format to which output is written. Flag --output-file is required.")
+	cmd.Flags().StringSliceVar(&config.TestCases, "test-cases", nil, "List of comma separated names of tests to be exeucted.")
+	cmd.Flags().DurationVar(&config.Timeout, "timeout", 24*time.Hour, "Execution timeout for all tests.")
+	cmd.Flags().BoolVar(&config.Quiet, "quiet", false, "Do not print test results to stdout.")
+}
+
+type Duration struct {
```
pinebit, line 46: This might be useful for other commands, I suggested to extract this into a separate file and add _test.go

cmd/test.go
```diff
@@ -0,0 +1,290 @@
...
+	default:
+		return errors.New("invalid duration")
+	}
+}
+
+type testVerdict string
+
+const (
+	// boolean tests
+	testVerdictNotOk testVerdict = "NotOK"
```
pinebit, line 85: Isn't this equal to `testVerdictFail`? Especially for boolean tests.

cmd/test.go
```diff
@@ -0,0 +1,290 @@
...
+	categoryScoreA categoryScore = "A"
+	categoryScoreB categoryScore = "B"
+	categoryScoreC categoryScore = "C"
+)
+
+type testResult struct {
+	Verdict     testVerdict `json:"verdict"`
+	Measurement string      `json:"measurement"`
+	Suggestion  string      `json:"suggestion"`
+	Error       string      `json:"error"`
```
pinebit, line 110: Maybe we shall `omitempty` for the `error` field?

cmd/test.go
```diff
@@ -0,0 +1,290 @@
...
+
+		if singleTestRes.Error != "" {
+			testOutput += " - " + singleTestRes.Error
+		}
+		lines = append(lines, testOutput)
+	}
+
+	if len(suggestions) != 0 {
+		lines = append(lines, "")
+		lines = append(lines, "IMPROVEMENT SUGGESTIONS")
```
pinebit, line 229: I would suggest to re-word this to "SUGGESTED IMPROVEMENTS"

cmd/test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	return nil
+}
+
+func calculateScore(results map[string]testResult) categoryScore {
+	// TODO(kalo): calculate score more elaborately (potentially use weights)
+	avg := 0
+	for _, t := range results {
+		if t.Verdict == testVerdictNotOk || t.Verdict == testVerdictBad || t.Verdict == testVerdictFail || t.Verdict == testVerdictTimeout {
```
pinebit, line 251: Why not switch/case construct? It would look more natural IMO

cmd/testbeacon.go
```diff
@@ -0,0 +1,124 @@
...
+	Endpoints []string
+}
+
+func newTestBeaconCmd(runFunc func(context.Context, io.Writer, testBeaconConfig) error) *cobra.Command {
+	var config testBeaconConfig
+
+	cmd := &cobra.Command{
+		Use:   "beacon",
+		Short: "Run multiple tests towards beacon nodes",
+		Long:  `Run multiple tests towards beacon nodes. Verify if the current setup is suitable for mainnet cluster.`,
```
pinebit, line 28: Too vague `Verify if the current setup is suitable for mainnet cluster.`, what if we say "Verify that Charon can interact with Beacon Node."?

cmd/testbeacon.go
```diff
@@ -0,0 +1,124 @@
...
+	queuedTests, err := filterTests(supportedTestCases, cfg.testConfig)
+	if err != nil {
+		return err
+	}
+	sort.Slice(queuedTests, func(i, j int) bool {
+		return queuedTests[i].order < queuedTests[j].order
+	})
+
+	parentCtx := ctx
+	if parentCtx == nil {
```
pinebit, line 66: Parent `ctx` must never be `nil` if we follow best practices. At least for un-exported functions.

cmd/testbeacon.go
```diff
@@ -0,0 +1,124 @@
...
+			break outer
+		case result := <-ch:
+			res.TestsExecuted[qt.name] = result
+		}
+	}
+
+	res.ExecutionTime = Duration{time.Since(startTime)}
+	res.Score = calculateScore(res.TestsExecuted)
+
+	if !cfg.Quiet {
```
pinebit, line 95: if quiet was true and outputfile is empty, user will not see anything?

cmd/testbeacon.go
```diff
@@ -0,0 +1,124 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func runAllBeacon(_ context.Context, queuedTests []testCaseName, allTests map[testCaseName]func(*testBeaconConfig) testResult, cfg testBeaconConfig, ch chan testResult) {
+	for _, t := range queuedTests {
+		ch <- allTests[t](&cfg)
```
pinebit, line 114: I suspect you need to pass context to test functions. Otherwise in case of "timeout" or SIGTERM, this will continue to execute.. Also, I believe the testResult channel shall be closed here in `defer`, this is how you can signal this goroutine is done. And the caller shall wait until this goroutine quits. For the sake of perfectness.

cmd/testbeacon.go
```diff
@@ -0,0 +1,124 @@
...
+	startTime := time.Now()
+	// run all beacon tests, pushing each finished test until all are finished or timeout occurs
+	go runAllBeacon(ctx, queuedTests, testCases, cfg, ch)
+outer:
+	for _, qt := range queuedTests {
+		select {
+		case <-ctx.Done():
+			res.TestsExecuted[qt.name] = testResult{Verdict: testVerdictTimeout}
+			break outer
+		case result := <-ch:
```
pinebit, line 87: Change to `result, ok := <-ch:` and you will be signalled when `runAllBeacon` is done if you close the channel there.

cmd/testpeers.go
```diff
@@ -0,0 +1,124 @@
...
+	ENRs []string
+}
+
+func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) error) *cobra.Command {
+	var config testPeersConfig
+
+	cmd := &cobra.Command{
+		Use:   "peers",
+		Short: "Run multiple tests towards peer nodes",
+		Long:  `Run multiple tests towards peer nodes. Verify if the current setup is suitable for mainnet cluster.`,
```
pinebit, line 28: Same concern as above, too vague long description.

cmd/testpeers.go
```diff
@@ -0,0 +1,124 @@
...
+	ch := make(chan testResult)
+	res := testCategoryResult{
+		TestsExecuted: make(map[string]testResult),
+		CategoryName:  "peers",
+	}
+
+	startTime := time.Now()
+	// run all peers tests, pushing each finished test until all are finished or timeout occurs
+	go runAllPeers(ctx, queuedTests, testCases, cfg, ch)
+outer:
```
pinebit, line 81: I see this pattern "runTestXXX" repeats in every file... Can we design a generic runner with parameters?

cmd/test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	root.AddCommand(cmds...)
+
+	return root
+}
+
+func bindTestFlags(cmd *cobra.Command, config *testConfig) {
+	cmd.Flags().StringVar(&config.OutputFile, "output-file", "", "File path to which output can be written.")
+	cmd.Flags().StringVar(&config.OutputFormat, "output-format", "json", "File format to which output is written. Flag --output-file is required.")
```
pinebit, line 40: What other formats you expect we may want to add?
I would go for the single format - TOML, which is both human and machine readable.

pinebit (commented):
Provided my comments

cmd/test.go
KaloyanTanev: Moved it to `duration.go`. I will push tests for it alongside with the other test files

cmd/test.go
KaloyanTanev: Not really - `testVerdictFail` is used when there is an unexpected error (a function literally returned an error), while here the test was successful, just the outcome of it is negative. Probably I should rename `testVerdictFail` to `testVerdictErr`?

cmd/test.go
KaloyanTanev: Good one, added it

cmd/test.go
KaloyanTanev: Agree, rephrased it

cmd/test.go
KaloyanTanev: Agree, first I had only 3 cases and if/else seemed better and forgot to change it later...

cmd/testbeacon.go
KaloyanTanev: Agree, added `efficiently` as well

cmd/testbeacon.go
KaloyanTanev: This parent context is coming from cobra's `cmd.Context()`, which says:

```
// Context returns underlying command context. If command was executed
// with ExecuteContext or the context was set with SetContext, the
// previously set context will be returned. Otherwise, nil is returned.
```

cmd/testbeacon.go
KaloyanTanev: True, added extra check

cmd/testbeacon.go
KaloyanTanev: Yap, all that sounds good! I have pushed updates on it, let me know what you think

cmd/testbeacon.go
KaloyanTanev: Added

cmd/testpeers.go
KaloyanTanev: Changed here as well

cmd/testpeers.go
KaloyanTanev: I've been thinking about that... for now they are really similar, almost the same, but it is because I don't have any tests and only the templates. I was thinking of extracting more of the common logic to `test.go` after I implement the tests.

cmd/test.go
KaloyanTanev: Yes, TOML is one other on my list. Do you suggest we have single one and get rid of this env var?

cmd/test.go
pinebit: > test was successful

Therefore, `testVerdictOk`, otherwise this is very confusing... IMO

cmd/testbeacon.go
pinebit: I believe we shall always use ExecuteContext, especially for tests, to support graceful shutdown (catching SIGTERM, etc and canceling the root context).

See the skeleton here: https://github.com/pinebit/go-graceful-shutdown

cmd/testpeers.go
pinebit: Alright, maybe eventually this will transform into something different. Up to you, as long as we don't have too much of code duplication.

cmd/test.go
pinebit: This will be much more user-friendly. IMO.
Operators (or DevOps) are very familiar with TOML/YAML via docker-compose experience.

cmd/duration.go
```diff
@@ -0,0 +1,45 @@
+// Copyright © 2022-2024 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package cmd
```
pinebit, line 3: Can this be useful for other Charon components? Then maybe move this to `app` or utils or something that can be shared.

cmd/test.go
KaloyanTanev: By successful I mean it it hasn't thrown any unexpected errors, not that it passed. To give an example:

Send HTTP request isSynced to Beacon node server results in "false" = `testVerdictNotOk`
Send HTTP request isSynced to Beacon node server results in error in `http.Get(url)` function and doesn't even reach the Beacon node server = `testVerdictFail`

cmd/testpeers.go
KaloyanTanev: Yap, I think I will see what would be the best for this function once I start implementing peers/beacon/validator flow.

cmd/test.go
pinebit: I see your point. Considering your example, I would suggest having distinct tests with clear purposes: 1) test Beacon node is alive (we can connect and it responds), 2) The whole DV is healthy (we have `app_health_checks` that includes BN synced state along with other things.
Both tests have clear purposes, and what's important - they have a boolean outcome (OK|FAIL).

cmd/testbeacon.go
KaloyanTanev: Yes, makes sense to me. Looking at it now though, it will require quite a bit of refactoring on the main CLI. Do you think it's okay to leave it like that and open an issue to change it everywhere where suitable (not only test)? Or shall I do this refactoring in this PR?

cmd/test.go
KaloyanTanev: Changed to hardcoded TOML and renamed the flag as `--output-toml` to be a bit more clear when invoked.

cmd/duration.go
pinebit: No problems, leave it here for now.

go.mod
pinebit, line 1: Why go.sum did not update?
Try `go mod tidy`

KaloyanTanev: Why should it have updated? There are only 2 new direct packages that were introduced and they were moved from the indirect ones, not new ones.

pinebit: My bad, I thought you just added go-toml, but it was there. :)

cmd/test.go
KaloyanTanev: Alright! I've removed the `testVerdictNotOk` and `testVerdictTimeout`, should be a bit simpler now. When we have `testVerdictFail` for a failed test we can put the error next to it if applicable, i.e.: `ping                  Fail - not implemented`.

.golangci.yml
```diff
@@ -147,6 +147,7 @@ linters:
     - gochecknoglobals
     - gocognit
     - gocyclo
+    - godot
```
KaloyanTanev, line 150: N.B.: I've disabled [godot](https://golangci-lint.run/usage/linters/#godot), because:

- it doesn't seem to work all the time - sometimes it asks for full stop, other times it doesn't, without any apparent pattern
- some comments really don't require to be a full sentence imo

cmd/duration_internal_test.go
pinebit, line 1: Why did you make this test internal? All functions are exported in `duration.go`

pinebit (approved):

cmd/duration_internal_test.go
KaloyanTanev: Good catch, moved to non internal one!

cmd/test.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func writeResultToWriter(res testCategoryResult, w io.Writer) error {
+	var lines []string
+
+	switch res.CategoryName {
+	case "peers":
+		lines = append(lines, "  ____                                                      ")
+		lines = append(lines, " |  __ \\                                                    ")
+		lines = append(lines, " | |__) |___   ___  _ __  ___                               ")
```
gsora, line 115: Can we have the ASCII art as a multiline const instead?

cmd/test.go
```diff
@@ -0,0 +1,252 @@
...
+	Score         categoryScore         `json:"score"`
+}
+
+func writeResultToFile(res testCategoryResult, path string) error {
+	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o444)
+	if err != nil {
+		return errors.Wrap(err, "create/open file")
+	}
+	defer f.Close()
+	err = toml.NewEncoder(f).Encode(res)
```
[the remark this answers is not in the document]
pinebit: You can easily process TOML with js: https://www.npmjs.com/package/js-toml
There is no difference in UX for TOML.

cmd/testpeers.go
```diff
@@ -0,0 +1,148 @@
...
+		var name string
+		select {
+		case <-timeoutCtx.Done():
+			name = queuedTests[testCounter].name
+			res.TestsExecuted[name] = testResult{Verdict: testVerdictFail}
+
+			break outer
+		case result, ok := <-ch:
+			if !ok {
+				break outer
```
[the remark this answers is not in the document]
pinebit: Agreed, labels are not friendly for readers.

cmd/test.go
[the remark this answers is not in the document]
KaloyanTanev: Did JSON first, but @pinebit suggested TOML 😅

cmd/test.go
KaloyanTanev: I can put them as a function that returns []string? If I use it as a single string with new lines it's a bit harder to attach the score next to it later.

cmd/testbeacon.go
```diff
@@ -0,0 +1,148 @@
...
+				break outer
+			}
+			name = queuedTests[testCounter].name
+			testCounter++
+			res.TestsExecuted[name] = result
+		}
+	}
+
+	res.ExecutionTime = Duration{time.Since(startTime)}
+	res.Score = calculateScore(res.TestsExecuted)
```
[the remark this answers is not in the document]
KaloyanTanev: I've been thinking about that and thought - what use can a user of ours make of the measurements, without our score attached to it? I think almost all of the users wouldn't know if 100ms is good or bad for a specific call? If we don't tell them what is sufficient and what is not, they can't really make up a conclusion on their own?

cmd/testpeers.go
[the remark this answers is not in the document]
KaloyanTanev: Yes, seems better to me. I dislike a bit the label because it is aligned on the most left...

.golangci.yml
KaloyanTanev:
```
const (
	// boolean tests
	testVerdictOk testVerdict = "OK"

	// measurement tests
	testVerdictGood testVerdict = "Good"
	testVerdictAvg  testVerdict = "Avg"
	testVerdictBad  testVerdict = "Bad"

	// failed tests
	testVerdictFail testVerdict = "Fail"
)
```

For this code snippet it forces full stop after `// failed tests`, but not after the others. I also see on some places in the code full stops arbitrary put on top of function names, on top of variables or on the same line (i.e.: `p2p/bootnode.go`).

cmd/test.go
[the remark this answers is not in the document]
gsora: Hm that's an extra package on the UI side but okay LFTM let's go with TOML

cmd/test.go
gsora: Yeah that's better

cmd/testbeacon.go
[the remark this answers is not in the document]
gsora: Thing is, an user might want to upload their measurements to a Launchpad page to e.g. prove to the other peers that their node is up for the task. The launchpad might have a scoring system embedded in it, and can show a pretty UI around it.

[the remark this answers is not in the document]
KaloyanTanev: Ah, okay, got what you mean. But then if we offload this logic to the launchpad, we are coupling the scoring with the launchpad, should those two be related? For me it has more connection to the tests in charon. We can still have the same functionality in the launchpad, it will just consume the already evaluated scores. Then also the non-launchpad users will benefit from the scoring (if I do the whole flow through the CLI, I won't have to go to the launchpad to get scores).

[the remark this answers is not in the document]
pinebit: I generally agree about making it as universal as possible, but maybe let's keep it as simple as possible at this stage of design/development? Just to move forward quicker. I believe this code is not very complex to not change it in the near future. I suspect we will change it anyway as soon as we actually implement a few tests and people start using them.

cmd/testbeacon.go
pinebit: Let's have a separate ticket to ensure we implemented graceful shutdown for charon. I did not check, maybe we have it but not for all commands?
For these tests we must have it 100%, because for continuous tests the user will terminate the app with docker down or ctrl+c/z.

pinebit: This will guarantee the passed context is not `nil`. But until we have this done, leave your check and fallback to Background().

cmd/testbeacon.go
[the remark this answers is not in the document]
gsora: LFTM!

cmd/testbeacon.go
gsora: Agreed with the proposed solution (re-checking for context usage and nilness in CLI).

>Let's have a separate ticket to ensure we implemented graceful shutdown for charon

Charon has a lifecycle manager in `run` but not in CLI commands, which are supposed to run as one-shot instances rather than long-term.

gsora (approved):

PR #3016, opened 2024-04-04
workflows: govulncheck to master

KaloyanTanev:
We have encountered unexpected `govulncheck: invalid finding: if Frame.Function is set, Frame.Package must also be ` error in the pipelines. Seems like an issue on govulncheck's side which is fixed on master and to be released in the upcoming v1.0.5. Bumping the version to master branch in the meantime.

Related: https://github.com/golang/go/issues/66139

category: bug


KaloyanTanev:
Using master seems unstable for now, we will roll with failing govulncheck until they release 1.0.5

PR #3017, opened 2024-04-04
*: update libp2p to version v0.33.2

gsora:
Apart from a nice performance improvement on TCP connections due to defaulting to TLS rather than Noise, this PR also fixes a resource exhaustion issue with an older version of quic-go.

Charon doesn't use quic, but might in the future, so it's good to fix this stuff early.

I had to disable Smart Dialing, introduced in v0.29.0 and enabled for IPv4 and IPv6 connections in v0.31.0 due to timing constraints: we prefer connecting concurrently, as soon as possible.

category: misc
ticket: none


pinebit (approved):

p2p/p2p.go
```diff
@@ -76,6 +77,7 @@ func NewTCPNode(ctx context.Context, cfg Config, key *k1.PrivateKey, connGater C
 			return filterAdvertisedAddrs(externalAddrs, internalAddrs, filterPrivateAddrs)
 		}),
 		libp2p.Transport(tcp.NewTCPTransport, tcpOpts...),
+		libp2p.SwarmOpts(swarm.WithDialRanker(swarm.NoDelayDialRanker)),
```
KaloyanTanev, line 80: Afaict from the [libp2p docs](https://pkg.go.dev/github.com/libp2p/go-libp2p/p2p/net/swarm#DefaultDialRanker), this way we will make attempts at the same time to connect to the private/public libp2p node address and also to the relay address, without prioritising the former? Was it like that before? I suppose practically speaking the attempt for direct connection to the libp2p node will evaluate faster.

gsora: This is essentially what was happening before libp2p 0.32.0, we just restored the original behavior.

Also, to be precise, we're setting the `AddrDelay` provider to be `swarm.NoDelayDialRanker`, which as the docs say:

>NoDelayDialRanker ranks addresses with no delay. This is useful for simultaneous connect requests.

KaloyanTanev: Gotcha, thanks!

Issue #2488, opened 2023-07-31
Improve track failure reason names


corverroos:
# 🎯 Problem to be solved
The tracker failure reasons that we instrument and describe [here](https://github.com/ObolNetwork/charon/blob/main/docs/reasons.md#duty-failure-reasons) use names linked to the components or steps where the duty failed. This results in unintuitive logs and metrics which is hard for users to understand and require lookup in the reference.

We could improve operator UX by renaming these reasons to be more descriptive of the cause where applicable.

# 🛠️ Proposed solution

Few things we can do:
- Prefix "bug" to reasons that are unexpected, also put this in a separate section in the docs.
- If there is only one cause, use that as the reason, not the step.

Some suggestions:
- `bcast` -> `broadcast_bn_error`
- `chain_inclusion` -> `not_included_onchain`
- `consensus` -> `no_consensus`
- `fetcher_aggregator_failed_prepare` -> `failed_aggregator_selection`
- `fetcher_aggregator_few_prepares` -> `insufficient_aggregator_selections`
- `fetcher_aggregator_no_external_prepares` -> `no_aggregator_selections`
- `fetcher_aggregator_no_att_data` -> `missing_aggregator_attestation`
- `fetcher_bn` -> `fetch_bn_error`
- `fetcher_proposer_failed_randao` -> `failed_proposer_randao`
- `par_sig_ex_receive` -> `no_peer_signatures`
- `par_sig_db_insufficient ` -> `insufficient_peer_signatures`
- `validator_api ` -> `no_local_vc_signature`

Bugs:
- `agg_sig_db` -> `bug_aggregation_error`
- `duty_db` -> `bug_duty_db_error`
- `fetcher_error` -> `bug_fetch_error `

gsora:
Closed by https://github.com/ObolNetwork/charon/pull/3033

PR #3033, opened 2024-04-11
core/tracker: improved reasons wording

pinebit:
Improved operator UX by renaming tracker reasons to be more descriptive.

Verified Obol infra alerts and Dashboard V3 - these do not use the affected identifiers explicitly.

category: refactor
ticket: #2488

### DIFF
--- a/docs/reasons.md
+++ b/docs/reasons.md
@@ -8,123 +8,126 @@ prometheus counter when the tracker component detects duty failures.
 By understanding these failure reasons, operators can better monitor, troubleshoot, and
 maintain system performance.
 
+## Errors
 
-### Failure Reason: `agg_sig_db`
-  - *Summary*: bug: failed to store aggregated signature in aggsigdb
-  - *Details*: Reason `agg_sig_db` indicates a bug in the aggregated signature database as it is unexpected.
-
-### Failure Reason: `bcast`
+### Failure Reason: `broadcast_bn_error`
   - *Summary*: failed to broadcast duty to beacon node
-  - *Details*: Reason `bcast` indicates that beacon node returned an error while submitting aggregated duty signature to beacon node.
+  - *Details*: Reason `broadcast_bn_error` indicates that beacon node returned an error while submitting aggregated duty signature to beacon node.
 
-### Failure Reason: `chain_inclusion`
+### Failure Reason: `not_included_onchain`
   - *Summary*: duty not included on-chain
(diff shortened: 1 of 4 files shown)


docs/reasons.md
```diff
@@ -9,122 +9,122 @@ By understanding these failure reasons, operators can better monitor, troublesho
...
+  - *Summary*: couldn`t fetch duty data from the beacon node
+  - *Details*: Reason `fetch_bn_error` indicates a duty failed in the fetcher step when it failed to fetch the required data from the beacon node API. This indicates a problem with the upstream beacon node.
+
+### Failure Reason: `insufficient_aggregator_selections`
   - *Summary*: couldn`t aggregate attestation due to insufficient partial beacon committee selections
-  - *Details*: Reason `fetcher_aggregator_few_prepares` indicates an attestation aggregation duty failed in the fetcher step since it couldn`t fetch the prerequisite aggregated beacon committee selections. This indicates the associated prepare aggregation duty failed due to insufficient partial beacon committee selections submitted by the cluster validator clients.
+  - *Details*: Reason `insufficient_aggregator_selections` indicates an attestation aggregation duty failed in the fetcher step since it couldn`t fetch the prerequisite aggregated beacon committee selections. This indicates the associated prepare aggregation duty failed due to insufficient partial beacon committee selections submitted by the cluster validator clients.
```
gsora, line 58:
```suggestion
  - *Details*: Reason `insufficient_aggregator_selections` indicates an attestation aggregation duty failed in the fetcher step since it couldn't fetch the prerequisite aggregated beacon committee selections. This indicates the associated prepare aggregation duty failed due to insufficient partial beacon committee selections submitted by the cluster validator clients.
```

There might be more instances of back ticks used as apostrophes, it will break Markdown parsing.

pinebit: `reasons.md` is generated from reason.go. I cannot patch it manually as it will break the test.

pinebit: Interesting that the source contains `couldn't`

pinebit: Found it: `result := strings.ReplaceAll(buf.String(), "'", "`")`
https://github.com/ObolNetwork/charon/blob/main/cmd/markdown_internal_test.go#L195

Shall we remove this you think?

pinebit: image

it does not seem break the markdown.

gsora: We're lucky then! Backticks are used for inline code, weird that it doesn't break.

I'm approving then!

pinebit: I believe it looks for the second backtick on the same line. This is not triggered in this case.
Also, note that in our test, we explicitly replace all ' with ` here https://github.com/ObolNetwork/charon/blob/main/cmd/markdown_internal_test.go#L195
The source text (reason.go) contains '

Issue #2997, opened 2024-03-27
Implement Peer-to-Peer Connection Tests


boulder225:
## 🎯 Problem to be solved
Verify the connectivity and health of peer-to-peer connections within the Charon cluster.

## 🛠️ Proposed solution
- [ ] Start libp2p node, connect to relay and subsequently to other libp2p nodes
- [ ] Implement a test to check if our node can ping all other nodes
- [ ] Implement a test to measure packet loss between peers
- [ ] Implement a test to measure round-trip time (RTT) between peers
- [ ] Accumulate and evaluate the peer-to-peer connection test results, contributing to the overall cluster health assessment
- [ ] Include the peer-to-peer connection test results in the report

KaloyanTanev:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2997?planning-poker) with Zenhub @pinebit @gsora

PR #3039, opened 2024-04-12
Draft: cmd: alpha test peers tests

KaloyanTanev:
category: feature
ticket: #2997

### DIFF
--- a/cmd/testpeers.go
+++ b/cmd/testpeers.go
@@ -4,21 +4,42 @@ package cmd
 
 import (
 	"context"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/hex"
 	"io"
-	"math/rand"
+	"math/big"
+	"os"
+	"os/signal"
+	"slices"
+	"strings"
 	"time"
 
+	k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/peerinfo"
+	"github.com/obolnetwork/charon/app/version"
+	"github.com/obolnetwork/charon/app/z"
+	"github.com/obolnetwork/charon/eth2util/enr"
+	"github.com/obolnetwork/charon/p2p"
 )
 
 type testPeersConfig struct {
 	testConfig
 	ENRs      []string
-	P2PRelays []string
+	P2P       p2p.Config
+	DataDir   string
+	KeepAlive time.Duration
 }
+
+const timeoutInterruptedErr = "timeout/interrupted"
 
 func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) error) *cobra.Command {
 	var config testPeersConfig
@@ -38,6 +59,8 @@ func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) e
 
 	bindTestFlags(cmd, &config.testConfig)
 	bindTestPeersFlags(cmd, &config)
+	bindP2PFlags(cmd, &config.P2P)
+	bindDataDirFlag(cmd.Flags(), &config.DataDir)
 
 	return cmd
 }
@@ -45,12 +68,12 @@ func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) e
 func bindTestPeersFlags(cmd *cobra.Command, config *testPeersConfig) {
 	const enrs = "enrs"
 	cmd.Flags().StringSliceVar(&config.ENRs, "enrs", nil, "[REQUIRED] Comma-separated list of each peer ENR address.")
-	cmd.Flags().StringSliceVar(&config.P2PRelays, "p2p-relays", []string{"https://0.relay.obol.tech,https://2.relay.obol.tech"}, "Comma-separated list of each peer P2P relay address.")
+	cmd.Flags().DurationVar(&config.KeepAlive, "keep-alive", 30*time.Minute, "Time to keep TCP node alive after test completion, so connection is open for other peers to test on their end.")
 	mustMarkFlagRequired(cmd, enrs)
 }
 
-func supportedPeerTestCases() map[testCaseName]func(context.Context, *testPeersConfig, string) testResult {
-	return map[testCaseName]func(context.Context, *testPeersConfig, string) testResult{
+func supportedPeerTestCases() map[testCaseName]func(context.Context, *testPeersConfig, host.Host, p2p.Peer) testResult {
+	return map[testCaseName]func(context.Context, *testPeersConfig, host.Host, p2p.Peer) testResult{
 		{name: "ping", order: 1}:        peerPingTest,
 		{name: "pingMeasure", order: 2}: peerPingMeasureTest,
 		{name: "pingLoad", order: 3}:    peerPingLoadTest,
@@ -60,6 +83,142 @@ func supportedPeerTestCases() map[testCaseName]func(context.Context, *testPeersC
 func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersConfig) testResult {
 	return map[testCaseName]func(context.Context, *testPeersConfig) testResult{
 		{name: "natOpen", order: 1}: natOpenTest,
+	}
+}
+
+func startTCPNode(ctx context.Context, cfg testPeersConfig) (host.Host, func(), error) {
+	var peers []p2p.Peer
+	for i, e := range cfg.ENRs {
+		record, err := enr.Parse(e)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "decode enr", z.Str("enr", e))
+		}
+
+		p, err := p2p.NewPeerFromENR(record, i)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		peers = append(peers, p)
+	}
+
+	key, err := p2p.LoadPrivKey(cfg.DataDir)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	r, err := enr.New(key)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	mePeer, err := p2p.NewPeerFromENR(r, len(cfg.ENRs))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	peers = append(peers, mePeer)
+
+	allENRs := cfg.ENRs
+	allENRs = append(allENRs, r.String())
+	slices.Sort(allENRs)
+	allENRsString := strings.Join(allENRs, ",")
+	h := sha256.New()
+	_, err = h.Write([]byte(allENRsString))
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "calculate sha256")
+	}
+	allENRsHash := h.Sum(nil)
+
+	return setupP2P(ctx, key, cfg.P2P, peers, allENRsHash)
+}
+
+func setupP2P(ctx context.Context, key *k1.PrivateKey, cfg p2p.Config, peers []p2p.Peer, enrsHash []byte) (host.Host, func(), error) {
+	var peerIDs []peer.ID
+	for _, p := range peers {
+		peerIDs = append(peerIDs, p.ID)
+	}
+
+	if err := p2p.VerifyP2PKey(peers, key); err != nil {
+		return nil, nil, err
+	}
+
+	relays, err := p2p.NewRelays(ctx, cfg.Relays, hex.EncodeToString(enrsHash))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	connGater, err := p2p.NewConnGater(peerIDs, relays)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	tcpNode, err := p2p.NewTCPNode(ctx, cfg, key, connGater, false)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	p2p.RegisterConnectionLogger(ctx, tcpNode, peerIDs)
+
+	for _, relay := range relays {
+		relay := relay
+		go p2p.NewRelayReserver(tcpNode, relay)(ctx)
+	}
+
+	go p2p.NewRelayRouter(tcpNode, peerIDs, relays)(ctx)
+
+	// Register peerinfo server handler for identification to relays (but do not run peerinfo client).
+	gitHash, _ := version.GitCommit()
+	_ = peerinfo.New(tcpNode, peerIDs, version.Version, enrsHash, gitHash, nil, false)
+
+	return tcpNode, func() {
+		_ = tcpNode.Close()
+	}, nil
+}
+
+func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.Result, error) {
+	select {
+	case <-ctx.Done():
+		return ping.Result{}, errors.New("context done")
+	default:
+		pingSvc := ping.NewPingService(tcpNode)
+		pingCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
+		for {
+			pingChan := pingSvc.Ping(pingCtx, peer.ID)
+			select {
+			case <-pingCtx.Done():
+				return ping.Result{}, errors.New("context done")
+			case result := <-pingChan:
+				return result, nil
+			}
+		}
+	}
+}
+
+func pingPeerContinuously(ctx context.Context, tcpNode host.Host, peer p2p.Peer, resCh chan ping.Result) {
+	for {
+		select {
+		case <-ctx.Done():
+			select {
+			case <-resCh:
+				return
+			default:
+				close(resCh)
+				return
+			}
+		default:
+			r, err := pingPeerOnce(ctx, tcpNode, peer)
+			if err != nil {
+				return
+			}
+			resCh <- r
+			nBig, err := rand.Int(rand.Reader, big.NewInt(100))
+			if err != nil {
+				return
+			}
+			time.Sleep(time.Duration(nBig.Int64()) * time.Millisecond)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/testpeers.go
```diff
@@ -4,22 +4,48 @@ package cmd
...
-	P2PRelays []string
+	ENRs             []string
+	P2P              p2p.Config
+	Log              log.Config
+	DataDir          string
+	KeepAlive        time.Duration
+	LoadTestDuration time.Duration
 }
 
+const timeoutInterruptedErr = "timeout/interrupted"
```
pinebit, line 47: Best practice is to make it of type Error, in this case we can test against it using `errors.Is` for example.

cmd/testpeers.go
```diff
@@ -38,19 +64,30 @@ func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) e
...
 	return cmd
 }
 
 func bindTestPeersFlags(cmd *cobra.Command, config *testPeersConfig) {
 	const enrs = "enrs"
-	cmd.Flags().StringSliceVar(&config.ENRs, "enrs", nil, "[REQUIRED] Comma-separated list of each peer ENR address.")
-	cmd.Flags().StringSliceVar(&config.P2PRelays, "p2p-relays", []string{"https://0.relay.obol.tech,https://2.relay.obol.tech"}, "Comma-separated list of each peer P2P relay address.")
+	cmd.Flags().StringSliceVar(&config.ENRs, enrs, nil, "[REQUIRED] Comma-separated list of each peer ENR address.")
+	cmd.Flags().DurationVar(&config.KeepAlive, "keep-alive", 30*time.Minute, "Time to keep TCP node alive after test completion, so connection is open for other peers to test on their end.")
+	cmd.Flags().DurationVar(&config.LoadTestDuration, "load-test-duration", 30*time.Second, "Time to keep running the load tests in seconds. For each second a new continuous ping instance is spawned.")
```
pinebit, line 78: Previously you implemented great [duration](https://github.com/ObolNetwork/charon/blob/main/cmd/duration.go) structure, why not to leverage it here to let user set arbitrary duration, not just seconds?

cmd/testpeers.go
```diff
@@ -63,7 +100,150 @@ func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersC
...
+
+	log.Info(ctx, "Self p2p name resolved", z.Any("name", mePeer.Name))
+
+	peers = append(peers, mePeer)
+
+	allENRs := cfg.ENRs
+	allENRs = append(allENRs, r.String())
+	slices.Sort(allENRs)
+	allENRsString := strings.Join(allENRs, ",")
+	h := sha256.New()
```
pinebit, line 142: Simplify this to `sha256.Sum256`: https://pkg.go.dev/crypto/sha256#Sum256

cmd/testpeers.go
```diff
@@ -63,7 +100,150 @@ func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersC
...
+	}
+
+	go p2p.NewRelayRouter(tcpNode, peerIDs, relays)(ctx)
+
+	// Register peerinfo server handler for identification to relays (but do not run peerinfo client).
+	gitHash, _ := version.GitCommit()
+	_ = peerinfo.New(tcpNode, peerIDs, version.Version, enrsHash, gitHash, nil, false)
+
+	return tcpNode, func() {
+		_ = tcpNode.Close()
```
pinebit, line 191: Maybe it worth logging the error if `!errors.Is(err, context.Cancelled)`.

cmd/testpeers.go
```diff
@@ -63,7 +100,150 @@ func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersC
...
+	}, nil
+}
+
+func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.Result, error) {
+	select {
+	case <-ctx.Done():
+		return ping.Result{}, errors.New("context done")
```
pinebit, line 198: Just return `ctx.Err()` here and below

cmd/testpeers.go
```diff
@@ -63,7 +100,150 @@ func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersC
...
+				close(resCh)
+				return
+			}
+		default:
+			r, err := pingPeerOnce(ctx, tcpNode, peer)
+			if err != nil {
+				return
+			}
+			resCh <- r
+			nBig, err := rand.Int(rand.Reader, big.NewInt(100))
```
pinebit, line 232: Big integers seems overkill here, why not just https://pkg.go.dev/math/rand#Intn?

cmd/testpeers.go
```diff
@@ -144,50 +330,66 @@ func runTestPeers(ctx context.Context, w io.Writer, cfg testPeersConfig) (err er
 		}
 	}
 
+	done := make(chan os.Signal, 1)
+	signal.Notify(done, os.Interrupt)
```
pinebit, line 334: Maybe NotifyContext is better? https://github.com/pinebit/go-boot/blob/621b6b1a2df5d81ed9187219c08b9656cb2c0f05/boot/application.go#L39
(no need to deal with channels, just use context)

pinebit (commented):
Provided initial set of comments.

cmd/testpeers.go
KaloyanTanev: The cobra CLI works in a similar way as my duration structure, meaning the input can be `--load-test-duration=5s` and also `--load-test-duration=10m30s`. However, in my implementation I am starting a new continuous ping each second, so `--load-test-duration=10ms` does not make much sense for load testing. I can probably change the description to:
```suggestion
	cmd.Flags().DurationVar(&config.LoadTestDuration, "load-test-duration", 30*time.Second, "Time to keep running the load test. For each second a new continuous ping instance is spawned.")
```

cmd/testpeers.go
KaloyanTanev: Agree, that's how I started, but linter complained... wanted me to use `crypto/rand`.

`G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)`

I may override it for this instance here

pinebit: Yes I think for this instance it is better to override.

cmd/testpeers.go
```diff
@@ -258,46 +460,132 @@ func runSelfTest(ctx context.Context, queuedTestCases []testCaseName, allTestCas
...
+		case <-ctx.Done():
+			tr.Verdict = testVerdictFail
+			tr.Error = errTimeoutInterrupted.Error()
+
+			return tr
+		default:
+			result, err := pingPeerOnce(ctx, tcpNode, peer)
+			if err != nil {
+				tr.Verdict = testVerdictFail
+				tr.Error = err.Error()
```
pinebit, line 480: Looks like you always put Error's text as is to `tr.Error`. Because of this I suspect `tr.Error` shall be of type Error. Final conversion to string can be done in the output renderer (console | toml).

cmd/testpeers.go
```diff
@@ -258,46 +460,132 @@ func runSelfTest(ctx context.Context, queuedTestCases []testCaseName, allTestCas
...
-	case <-ctx.Done():
-		return testResult{Verdict: testVerdictFail}
-	default:
-		return testResult{
-			Verdict:     testVerdictFail,
-			Measurement: "10ms",
-			Error:       errors.New("pingMeasure not implemented").Error(),
-		}
+func peerPingMeasureTest(ctx context.Context, _ *testPeersConfig, tcpNode host.Host, peer p2p.Peer) testResult {
+	const thresholdAvg = 200 * time.Millisecond
```
pinebit, line 516: I suggest to move all timing constants to the global `const ()` section, for better visibility.

cmd/testpeers.go
```diff
@@ -258,46 +460,132 @@ func runSelfTest(ctx context.Context, queuedTestCases []testCaseName, allTestCas
...
+	resCh := make(chan ping.Result, math.MaxInt16)
+	pingCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for i := 0; i < s; i++ {
+		select {
+		case <-ctx.Done():
+			i = s
```
pinebit, line 565: lol, I've never seen this before :)

How about
```
deadlineC := time.After(cfg.LoadTestDuration)
for {
  select {
     case <-ctx.Done(): ....
     case <-ticker.C: ...
     case <-deadlineC: ....
  }
}
```

pinebit (changes requested):
This should be the final round.

cmd/testpeers_internal_test.go
```diff
@@ -36,64 +38,29 @@ func TestPeersTest(t *testing.T) {
...
 				CategoryName: "peers",
 				Targets: map[string][]testResult{
 					"self": {
 						{Name: "natOpen", Verdict: testVerdictFail, Measurement: "", Suggestion: "", Error: "natOpen not implemented"},
 					},
-					"enr:-1": {
-						{Name: "ping", Verdict: testVerdictFail, Measurement: "", Suggestion: "", Error: "ping not implemented"},
-						{Name: "pingMeasure", Verdict: testVerdictFail, Measurement: "", Suggestion: "", Error: "timeout"},
+					"frantic-colony - enr:-JG4QFI0llFYxSoTAHm24OrbgoVx77dL6Ehl1Ydys39JYoWcBhiHrRhtGXDTaygWNsEWFb1cL7a1Bk0klIdaNuXplKWGAYGv0Gt7gmlkgnY0gmlwhH8AAAGJc2VjcDI1NmsxoQL6bcis0tFXnbqG4KuywxT5BLhtmijPFApKCDJNl3mXFYN0Y3CCDhqDdWRwgg4u": {
+						{Name: "ping", Verdict: testVerdictFail, Measurement: "", Suggestion: "", Error: "timeout/interrupted"},
```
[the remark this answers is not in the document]
KaloyanTanev: Ah, missed the tests, thanks!

cmd/testpeers_internal_test.go
```diff
@@ -224,9 +199,15 @@ func TestPeersTest(t *testing.T) {
...
+			require.NoError(t, err)
+			cfg := test.config
+			cfg.DataDir = temp
+
 			var buf bytes.Buffer
 			ctx := context.Background()
-			err := runTestPeers(ctx, &buf, test.config)
+			err = runTestPeers(ctx, &buf, cfg)
 			if test.expectedErr != "" {
 				require.ErrorContains(t, err, test.expectedErr)
```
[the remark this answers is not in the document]
KaloyanTanev: It might be the case that errors are coming from external libraries and are wrapped, so I decided to keep it more generic. Currently there are mostly newly created errors, but don't want to force that for all test cases.

I plan to implement `natOpen` in my next PR, so not sure if necessary defining it now.

cmd/testpeers.go
```diff
@@ -110,6 +292,9 @@ func runTestPeers(ctx context.Context, w io.Writer, cfg testPeersConfig) (err er
 				break
 			}
 			maps.Copy(testResults, result)
+		case <-ctx.Done():
+			selfFinished = true
```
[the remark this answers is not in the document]
KaloyanTanev: That's really neat way of solving that! I have one concern though, reading `sync/errgroup.WithContext` I see:
```
// The derived Context is canceled the first time a function passed to Go
// returns a non-nil error or the first time Wait returns, whichever occurs
// first.
```
For my use case, that means if testing self fails, it will automatically stop my peers tests? If that's the case, it's not something I'd like to happen...

cmd/testpeers.go
```diff
@@ -258,46 +460,132 @@ func runSelfTest(ctx context.Context, queuedTestCases []testCaseName, allTestCas
...
-		return testResult{
-			Verdict: testVerdictFail,
-			Error:   errors.New("ping not implemented").Error(),
+func peerPingTest(ctx context.Context, _ *testPeersConfig, tcpNode host.Host, peer p2p.Peer) testResult {
+	tr := testResult{Name: "Ping"}
+
+	ticker := time.NewTicker(3 * time.Second)
+	defer ticker.Stop()
+
+	for ; true; <-ticker.C {
```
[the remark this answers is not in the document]
KaloyanTanev: Yap, true... The thing is if I do unconditional for loop, I always wait for the first tick, which gives me unnecessary wait time of 3 seconds. With `for ; true; <-ticker.C` I execute the first iteration of the loop immediately and then I wait on the tick. Not sure which is better, I might as well just go with `time.Sleep`...

cmd/testpeers.go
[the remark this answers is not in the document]
pinebit: If your testing routine got an error, then something is really bad and your test results will be inaccurate anyway, better to yield the error and shutdown.
Or, simply design test routines to never return an error if you wish. You don't have to deal with errors in this case.

cmd/testpeers.go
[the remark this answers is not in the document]
pinebit: You should be able to use [Reset](https://pkg.go.dev/time#Ticker.Reset).
Create the `Ticker` with minimal delay, e.g. 1ms, then upon each fire `Reset(3 * time.Second)`.

cmd/test.go
```diff
@@ -73,12 +77,24 @@ const (
 	categoryScoreC categoryScore = "C"
 )
 
+type testResultError struct{ error }
```
gsora, line 80: Even if it's unexported, can we have some godoc here explaining the purpose of this struct?

cmd/testpeers.go
```diff
@@ -4,22 +4,56 @@ package cmd
...
-	P2PRelays []string
+	ENRs             []string
+	P2P              p2p.Config
+	Log              log.Config
+	DataDir          string
+	KeepAlive        time.Duration
+	LoadTestDuration time.Duration
 }
 
+var (
```
gsora, line 44: Can those errors be re-used across tests? If yes, maybe we could move them in `cmd/test.go`?

cmd/testpeers.go
```diff
@@ -38,19 +72,30 @@ func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) e
...
-	return map[testCaseName]func(context.Context, *testPeersConfig, string) testResult{
+func bindTestLogFlags(flags *pflag.FlagSet, config *log.Config) {
+	flags.StringVar(&config.Format, "log-format", "console", "Log format; console, logfmt or json")
+	flags.StringVar(&config.Level, "log-level", "info", "Log level; debug, info, warn or error")
+	flags.StringVar(&config.Color, "log-color", "auto", "Log color; auto, force, disable.")
+	flags.StringVar(&config.LogOutputPath, "log-output-path", "", "Path in which to write on-disk logs.")
+}
+
+func supportedPeerTestCases() map[testCaseName]func(context.Context, *testPeersConfig, host.Host, p2p.Peer) testResult {
+	return map[testCaseName]func(context.Context, *testPeersConfig, host.Host, p2p.Peer) testResult{
```
gsora, line 98: Nit: maybe have a `testCase` type for `func(context.Context, *testPeersConfig, host.Host, p2p.Peer) testResult`? To increase readability.

cmd/testpeers.go
```diff
@@ -38,19 +72,30 @@ func newTestPeersCmd(runFunc func(context.Context, io.Writer, testPeersConfig) e
...
+	bindTestLogFlags(cmd.Flags(), &config.Log)
 
 	return cmd
 }
 
 func bindTestPeersFlags(cmd *cobra.Command, config *testPeersConfig) {
 	const enrs = "enrs"
-	cmd.Flags().StringSliceVar(&config.ENRs, "enrs", nil, "[REQUIRED] Comma-separated list of each peer ENR address.")
-	cmd.Flags().StringSliceVar(&config.P2PRelays, "p2p-relays", []string{"https://0.relay.obol.tech,https://2.relay.obol.tech"}, "Comma-separated list of each peer P2P relay address.")
+	cmd.Flags().StringSliceVar(&config.ENRs, enrs, nil, "[REQUIRED] Comma-separated list of each peer ENR address.")
```
gsora, line 84: We already have a way to enforcing required flags at runtime (see what we're doing with the `exit` commands). No need to explicitly write "required".

cmd/testpeers.go
```diff
@@ -63,7 +108,145 @@ func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersC
...
+	allENRs := cfg.ENRs
+	allENRs = append(allENRs, r.String())
+	slices.Sort(allENRs)
+	allENRsString := strings.Join(allENRs, ",")
+	allENRsHash := sha256.Sum256([]byte(allENRsString))
+
+	return setupP2P(ctx, key, cfg.P2P, peers, allENRsHash[:])
+}
+
+func setupP2P(ctx context.Context, key *k1.PrivateKey, cfg p2p.Config, peers []p2p.Peer, enrsHash []byte) (host.Host, func(), error) {
```
gsora, line 155: I wonder if we could export the setupP2P functions from `run` in a generic package and reuse everything... 

Just a random thought, we could explore this in the future.

cmd/testpeers.go
```diff
@@ -63,7 +108,145 @@ func supportedSelfTestCases() map[testCaseName]func(context.Context, *testPeersC
...
+
+	return tcpNode, func() {
+		err := tcpNode.Close()
+		if err != nil && !errors.Is(err, context.Canceled) {
+			log.Error(ctx, "Close TCP node", err)
+		}
+	}, nil
+}
+
+func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.Result, error) {
```
gsora, line 201: Wondering if we could reuse `PingService` here?

cmd/testpeers.go
KaloyanTanev: I also enforce them the same way afaict - `mustMarkFlagRequired`. This `[REQUIRED]` is only for UX, similar to the `create dkg` command.

cmd/testpeers.go
KaloyanTanev: Yap, probably we can make it a bit more generic if we want to export it.

cmd/testpeers.go
KaloyanTanev: If you refer to `NewPingService` in p2p package, it starts a long-lived pinging. What we can try to reuse is the `pingPeerOnce` function from the p2p package. We gotta export it alongside with some other functions in that package though.

pinebit (approved):
LGTM

cmd/testpeers.go
gsora: I'm not opposed to that. We could refactor `pingPeerOnce` as a standalone exported function and use it in both `PingService` and here. Not strictly necessary for this iteration, but would be a nice to have so that we use the same implementation in both.

cmd/testpeers.go
gsora: We can ticket that as a low-prio ticket for the future.

gsora (approved):

Issue #2544, opened 2023-08-14
add new libp2p metrics to relay


corverroos:
# 🎯 Problem to be solved
libp2p added a bunch of new circuit [relay metrics](https://github.com/libp2p/go-libp2p/pull/2154) as part of there v0.27.0 release. We should instrument these to make detecting issues with relay easier for operators.

# 🛠️ Proposed solution
Determine which metrics to enable and enable it.

gsora:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/2544?planning-poker) with Zenhub @KaloyanTanev @pinebit

PR #3040, opened 2024-04-16
cmd/relay: enabled new libp2p relay metrics

pinebit:
libp2p introduced set of new metrics for relay:
https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/metrics.go

This work integrates these metrics into our instrumentation flow.

Note: metrics reported by libp2p will be prefixed: `libp2p_relaysvc` whereas metrics reported by charon code are prefixed: `relay_p2p`. This distinction seems useful for clear separation and avoiding clashes. Therefore, `metrics.md` remains intact. This only enriches libp2p metrics with charon cluster labels.

category: feature
ticket: #2544

### DIFF
--- a/cmd/relay/p2p.go
+++ b/cmd/relay/p2p.go
@@ -16,36 +16,45 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
 	ma "github.com/multiformats/go-multiaddr"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/app/log"
 	"github.com/obolnetwork/charon/app/peerinfo"
+	"github.com/obolnetwork/charon/app/promauto"
 	"github.com/obolnetwork/charon/app/z"
 	"github.com/obolnetwork/charon/p2p"
 )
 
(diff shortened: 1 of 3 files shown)


gsora (approved):

cmd/relay/p2p.go
```diff
@@ -16,37 +16,46 @@ import (
...
 	if err != nil {
-		return nil, errors.Wrap(err, "new tcp node")
+		return nil, nil, errors.Wrap(err, "new tcp node")
 	}
 
 	p2p.RegisterConnectionLogger(ctx, tcpNode, nil)
 
+	labels := map[string]string{"relay_peer": p2p.PeerName(tcpNode.ID())}
+	log.SetLokiLabels(labels)
+	promRegistry, err := promauto.NewRegistry(labels)
```
KaloyanTanev, line 54: Isn't it a bit weird creating a prometheus registry in `startP2P` function? Does it make sense to create it outside of it (in `cmd/relay/relay.go:Run` afaict) and pass it as an argument, similarly to how the reporter is passed (another monitoring structure)?

KaloyanTanev (approved):

cmd/relay/p2p.go
pinebit: The problem here is in the dependency on `labels` which we can populate only after we started a tcpNode.
I would rework it in a different way to remove this dependency, but would be a drastic change, not for the scope of this PR for sure.

PR #3052, opened 2024-04-29
cmd: alpha test peers tests self

KaloyanTanev:
Add alpha peers tests for the TCP node started. Currently only port opened is tested. Test for whether NAT is open is skipped for now as it makes more sense to be added to the peers' tests.

category: feature
ticket: #3076

### DIFF
--- a/cmd/testpeers.go
+++ b/cmd/testpeers.go
@@ -570,8 +570,9 @@ func peerPingLoadTest(ctx context.Context, conf *testPeersConfig, tcpNode host.H
 	return testRes
 }
 
-func dialTCPIP(address string, timeout time.Duration) error {
-	conn, err := net.DialTimeout("tcp", address, timeout)
+func dialTCPIP(ctx context.Context, address string) error {
+	d := net.Dialer{Timeout: time.Second}
+	conn, err := d.DialContext(ctx, "tcp", address)
 	if err != nil {
 		return errors.Wrap(err, "net dial")
 	}
@@ -593,7 +594,7 @@ func libp2pTCPPortOpenTest(ctx context.Context, cfg *testPeersConfig) testResult
 
 	for _, addr := range cfg.P2P.TCPAddrs {
 		addrVal := addr
-		group.Go(func() error { return dialTCPIP(addrVal, time.Second) })
+		group.Go(func() error { return dialTCPIP(ctx, addrVal) })
 	}
 
 	err := group.Wait()
--- a/cmd/testpeers.go
+++ b/cmd/testpeers.go
@@ -570,12 +570,21 @@ func peerPingLoadTest(ctx context.Context, conf *testPeersConfig, tcpNode host.H
 	return testRes
 }
 
-func dialTCPIP(ctx context.Context, address string) error {
+func dialLibp2pTCPIP(ctx context.Context, address string) error {
 	d := net.Dialer{Timeout: time.Second}
 	conn, err := d.DialContext(ctx, "tcp", address)
 	if err != nil {
 		return errors.Wrap(err, "net dial")
 	}
+	buf := new(strings.Builder)
+	_, err = io.Copy(buf, conn)
+	if err != nil {
+		return errors.Wrap(err, "io copy")
+	}
+	if !strings.Contains(buf.String(), "/multistream/1.0.0") {
+		return errors.New("multistream not found", z.Any("found", buf.String()), z.Any("address", address))
+	}
+
 	err = conn.Close()
 	if err != nil {
 		return errors.Wrap(err, "close conn")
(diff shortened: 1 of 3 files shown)


cmd/testpeers.go
```diff
@@ -569,15 +570,41 @@ func peerPingLoadTest(ctx context.Context, conf *testPeersConfig, tcpNode host.H
...
-		return testResult{
-			Verdict: testVerdictFail,
-			Error:   errNotImplemented,
-		}
+func dialTCPIP(address string, timeout time.Duration) error {
+	conn, err := net.DialTimeout("tcp", address, timeout)
+	if err != nil {
+		return errors.Wrap(err, "net dial")
+	}
+	if conn == nil {
```
pinebit, line 578: I am pretty sure this is impossible case when err and conn are nil at the same time.

cmd/testpeers.go
```diff
@@ -569,15 +570,41 @@ func peerPingLoadTest(ctx context.Context, conf *testPeersConfig, tcpNode host.H
...
-	select {
-	case <-ctx.Done():
-		return testResult{Verdict: testVerdictFail}
-	default:
-		return testResult{
-			Verdict: testVerdictFail,
-			Error:   errNotImplemented,
-		}
+func dialTCPIP(address string, timeout time.Duration) error {
+	conn, err := net.DialTimeout("tcp", address, timeout)
```
pinebit, line 574: This ignores parent context, better to use `DialContext` and add a timeout via context.

pinebit (changes requested):
A couple comments

cmd/testpeers.go
```diff
@@ -569,15 +570,39 @@ func peerPingLoadTest(ctx context.Context, conf *testPeersConfig, tcpNode host.H
...
+
+	return nil
+}
+
+func libp2pTCPPortOpenTest(ctx context.Context, cfg *testPeersConfig) testResult {
+	testRes := testResult{Name: "Libp2pTCPPortOpen"}
+
+	group, _ := errgroup.WithContext(ctx)
+
+	for _, addr := range cfg.P2P.TCPAddrs {
```
gsora, line 592: What IP does this command connects to?

KaloyanTanev: IPs specified by user: `"Comma-separated list of listening TCP addresses (ip and port) for libP2P traffic. Empty default doesn't bind to local port therefore only supports outgoing connections."`.

I'd say it is not the most useful of all tests we can have for peers, but it is a start for self tests. As we discussed last week, I believe it is best to move direct/indirect connection tests to the tests against peers, rather than keeping them as NAT test in the self test category.

gsora: Considering Charon doesn't bind to any address by default (as specified in the help string itself), how does this test behave in that case?

KaloyanTanev: If empty, it passes as OK. The test peers command inherits the p2p flags of the other charon commands (the `bindP2PFlags` function), where user can specify the tcp addresses, relays, etc.

gsora: I'm not sure I see the point of this test then? What are we trying to test? 

E.g. If I configure one of my network interface with my public IPv4 address, and point this test to it, the Linux kernel will shortcut this as a localhost connection.

@pinebit would love your opinion on this

pinebit: Yes, maybe this test alone is not very useful as it is. Maybe if implemented as a part of a comprehensive self-test?

KaloyanTanev: One thing I thought about is that I can check if the IP:Port it properly bound? For a potential scenario where the port specified is occupied by another process. Let's say I have postgres running on 5432, try to add flag `--p2p-tcp-address="127.0.0.1:5432"`, then I can check what I connect to on port `5432`? I am not sure if I'm not making up scenarios at this point though :)

gsora: Hmm at that point you'll receive an error as soon as you're binding to that port?

KaloyanTanev: As far as I can tell, multiport binding is not as uncommon as it is used for hole punching in libp2p. What might happen is that the specified port is already in use by another process and both turning on `SO_REUSEPORT`, so it won't result in error. In that case the routing might get messed up. I have tried to add test check if there are more than one processes running on the specified port, but with no luck for now. Some options from external libraries, but I'm not a big fan of adding external library only for this check.

cmd/testpeers.go
```diff
@@ -570,12 +570,21 @@ func peerPingLoadTest(ctx context.Context, conf *testPeersConfig, tcpNode host.H
...
 }
 
-func dialTCPIP(ctx context.Context, address string) error {
+func dialLibp2pTCPIP(ctx context.Context, address string) error {
 	d := net.Dialer{Timeout: time.Second}
 	conn, err := d.DialContext(ctx, "tcp", address)
 	if err != nil {
 		return errors.Wrap(err, "net dial")
 	}
+	buf := new(strings.Builder)
```
gsora, line 580: Hmm a malicious server could keep the connection open forever. I think we should read a fixed amount of bytes, and check if the string begins with `/multistream/1.0.0`. WDYT?

KaloyanTanev: Hmmm yap, that can happen. The server will always be our own though, so it shouldn't be an issue here. I will change it though, better be more agnostic.

gsora (approved):

Issue #3053, opened 2024-04-29
Ping peer continuously closed channel


KaloyanTanev:
# 🐞 Bug Report

On command `alpha test peers` sometimes during peer pinging there is panic because of a closed channel. This is caused because we close the channel `testResCh` in `peerPingLoadTest` function and try to send data on it afterwards in `pingPeerContinuously`.

## 🔬 Minimal Reproduction

Run ping peer tests with 2 or more peers. Not guaranteed to error every time as it is dependant on async calls.

## 🔥 Error

```
panic: send on closed channel

goroutine 303 [running]:
github.com/obolnetwork/charon/cmd.pingPeerContinuously({0x101da42a0, 0x140005c0dc0}, {0x101db5960, 0x140000ec300}, {{0x14000880090, 0x27}, {0x0, 0x0, 0x0}, 0x0, ...}, ...)
        github.com/obolnetwork/charon/cmd/testpeers.go:216 +0x6c
created by github.com/obolnetwork/charon/cmd.peerPingLoadTest in goroutine 194
        github.com/obolnetwork/charon/cmd/testpeers.go:545 +0x39c
```

KaloyanTanev:
Fixed by #3054

PR #3054, opened 2024-04-29
cmd: move ctx check before writing to channel

KaloyanTanev:
Fix writing to closed channel error.

category: bug 
ticket: #3053

### DIFF
--- a/cmd/testpeers.go
+++ b/cmd/testpeers.go
@@ -204,14 +204,14 @@ func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.R
 
 func pingPeerContinuously(ctx context.Context, tcpNode host.Host, peer p2p.Peer, resCh chan ping.Result) {
 	for {
+		r, err := pingPeerOnce(ctx, tcpNode, peer)
+		if err != nil {
+			return
+		}
 		select {
 		case <-ctx.Done():
 			return
 		default:
-			r, err := pingPeerOnce(ctx, tcpNode, peer)
-			if err != nil {
-				return
-			}
 			resCh <- r
 			awaitTime := rand.Intn(100) //nolint:gosec // weak generator is not an issue here
 			time.Sleep(time.Duration(awaitTime) * time.Millisecond)
--- a/cmd/testpeers.go
+++ b/cmd/testpeers.go
@@ -8,7 +8,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
-	"math"
 	"math/rand"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/testpeers.go
```diff
@@ -204,14 +204,14 @@ func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.R
 
 func pingPeerContinuously(ctx context.Context, tcpNode host.Host, peer p2p.Peer, resCh chan ping.Result) {
 	for {
+		r, err := pingPeerOnce(ctx, tcpNode, peer)
+		if err != nil {
+			return
+		}
 		select {
 		case <-ctx.Done():
```
pinebit, line 212: This appears to be useless given that `pingPeerOnce` must return when `ctx.Done()`.

cmd/testpeers.go
```diff
@@ -204,14 +204,14 @@ func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.R
...
 			return
 		default:
-			r, err := pingPeerOnce(ctx, tcpNode, peer)
-			if err != nil {
-				return
-			}
 			resCh <- r
```
pinebit, line 215: Now this write does not respect the context and can block the routine.

cmd/testpeers.go
KaloyanTanev: That's true and this context doneness check is triggered regularly on `pingPeerOnce` (I see it being triggered on every `pingLoad` test). However, I have observed the code panic-ing, because `resCh` was closed and it tried sending data to it. I think it is because there is canceling of the context and closing of the channel happening between `r, err := pingPeerOnce(ctx, tcpNode, peer)` and `resCh <- r` (which is the error check). I've observed it happening couple of times, but it's really rare to catch.

cmd/testpeers.go
KaloyanTanev: Why it wouldn't respect the context? I do:
```
		select {
		case <-ctx.Done():
			return
		default:
			resCh <- r
			awaitTime := rand.Intn(100) //nolint:gosec // weak generator is not an issue here
			time.Sleep(time.Duration(awaitTime) * time.Millisecond)
		}
```
I still have the context check

cmd/testpeers.go
pinebit: That's why the channel shall normally to be closed by Producer.

cmd/testpeers.go
pinebit: If `resCh <- r` is blocked then you are stuck, because this code executes not under `select`, but in the code block of the `default` case. It is not a `case` to be simple, therefore `select` does not work for this write.

pinebit: https://goplay.tools/snippet/wS9Sbd8kMu9

KaloyanTanev: Looking at it now... It seems to have been an issue even before this change, right?

cmd/testpeers.go
KaloyanTanev: Pushed new commit where I create separate channels for each `pingPeerContinuously` call and close it upon exiting `pingPeerContinuously` function.

cmd/testpeers.go
pinebit: Right

pinebit (approved):

cmd/testpeers.go
pinebit: In this case you could simply use `sync.WaitGroup` to ensure all spawned goroutines will stop writing and then close the single `testResCh` after `.Wait()`. 
A few other observations:
* You may not need `deadlineC` at all, if you use a context with deadline with will give you a cancel function.
* You may not need `finished` flag at all, if you simply exit the infinite loop and then wait on the mentioned `WaitGroup`.

cmd/testpeers.go
```diff
@@ -203,6 +202,7 @@ func pingPeerOnce(ctx context.Context, tcpNode host.Host, peer p2p.Peer) (ping.R
 }
 
 func pingPeerContinuously(ctx context.Context, tcpNode host.Host, peer p2p.Peer, resCh chan ping.Result) {
+	defer close(resCh)
```
pinebit, line 205: This change alone does not solve the problem of the blocking write. I know this was introduced previously, not in this PR, but this is something I overlooked in past..

gsora (approved):

PR #3077, opened 2024-05-08
app/version: v1.0-dev

gsora:
Instead of going with version v0.20, use v1.0.

category: misc
ticket: none

### DIFF
--- a/app/peerinfo/peerinfo_test.go
+++ b/app/peerinfo/peerinfo_test.go
@@ -23,6 +23,12 @@ func TestPeerInfo(t *testing.T) {
 	now := time.Now()
 	const gitCommit = "1234567"
 
+	// when a major release happens, charon will only be compatible with a single version
+	versions := version.Supported()
+	if len(versions) < 2 {
+		versions = append(versions, versions...)
+	}
+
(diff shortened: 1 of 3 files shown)


pinebit (approved):

KaloyanTanev (approved):

app/peerinfo/peerinfo_test.go
```diff
@@ -23,22 +23,28 @@ func TestPeerInfo(t *testing.T) {
 	now := time.Now()
 	const gitCommit = "1234567"
 
+	// when a major release happens, charon will only be compatible with a single version
+	versions := version.Supported()
+	if len(versions) < 2 {
+		versions = append(versions, versions...)
+	}
```
KaloyanTanev, line 30: If there is a single version supported we... duplicate it?

gsora: yes, as a corner case in this test, in order to keep the test working when there will be more than one supported

KaloyanTanev: Not the prettiest, but fine by me, beauty is in the eye of the beholder 😄

PR #3079, opened 2024-05-08
cmd: increase timeout of timeout test

KaloyanTanev:
Sometimes timeout is so little that the test doesn't even reach the second instance, as observed in the pipeline [here](https://github.com/ObolNetwork/charon/actions/runs/9002400755/job/24730617672?pr=3078).

category: bug
ticket: none

### DIFF
--- a/cmd/testbeacon_internal_test.go
+++ b/cmd/testbeacon_internal_test.go
@@ -96,7 +96,7 @@ func TestBeaconTest(t *testing.T) {
 					OutputToml: "",
 					Quiet:      false,
 					TestCases:  nil,
-					Timeout:    time.Nanosecond,
+					Timeout:    100 * time.Nanosecond,
 				},
 				Endpoints: []string{endpoint1, endpoint2},
 			},


cmd/testbeacon_internal_test.go
```diff
@@ -96,7 +96,7 @@ func TestBeaconTest(t *testing.T) {
 					OutputToml: "",
 					Quiet:      false,
 					TestCases:  nil,
-					Timeout:    time.Nanosecond,
+					Timeout:    5 * time.Nanosecond,
```
pinebit, line 99: It's your business but to me this is the first indicator of potentially flakey test. :)

pinebit (approved):

cmd/testbeacon_internal_test.go
KaloyanTanev: Looks a bit weird indeed, but I want to have some timeout that is triggered before my HTTP call succeeds... Currently this test will falsely fail on 2 occasions:
- If the http call is blazing fast (<100ns), but that is impossible I believe
- If the code executes slower in some go routines and it doesn't reach the step of registering all the targets it needs to call (what happened before). I've bumped it to 100ns now. Seems highly unlikely the code doesn't execute for 2 targets for 100ns, as it was executing for 1ns for 2 targets in all of the runs except this last one up until now.

PR #3087, opened 2024-05-16
cmd: add available tests per test category in help

KaloyanTanev:
Some time ago @gsora spotted that user can run specific tests, however, the user cannot see which tests are available. 

Adding the available tests in the help command.
I've added also consts of the different categories.

category: feature
ticket: none


cmd/test.go
```diff
@@ -16,16 +16,24 @@ import (
 
 	"github.com/pelletier/go-toml/v2"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/maps"
```
gsora, line 19: We can bump the Go version in `go.mod` to 1.22 minimum and avoid importing an `x/exp` package, those utilities have been stabilized.

KaloyanTanev: Shall we do it in this PR or separate one? I'd advocate for separate

gsora: Agreed!

pinebit (approved):

cmd/test.go
KaloyanTanev: They are [still not included in go 1.22.3](https://github.com/golang/go/tree/go1.22.3/src/maps)

I see the functions I use are [in master](https://github.com/golang/go/tree/master/src/maps) though (under `iter.go`), so hopefully in go 1.23 we will have them

gsora: Woof, I thought they were in already :-/

gsora (approved):

PR #3098, opened 2024-05-23
app/obolapi: specify HTTP requests timeout

gsora:
Allow users of `app/obolapi` to specify HTTP requests timeout.

As a result of this feature, add the `--publish-timeout` CLI flag to `exit` and `dkg` commands.

category: feature
ticket: none

### DIFF
--- a/app/obolapi/api_internal_test.go
+++ b/app/obolapi/api_internal_test.go
@@ -1,0 +1,23 @@
+// Copyright © 2022-2024 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package obolapi
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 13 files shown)


app/obolapi/api_internal_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithTimeout(t *testing.T) {
+	// no timeout = 10s timeout
+	oapi, err := New("http://url.com")
+	require.NoError(t, err)
+	require.Equal(t, 10*time.Second, oapi.reqTimeout)
```
pinebit, line 16: You should have access to `defaultTimeout` const to use it instead of the explicit literal.

OisinKyne (approved):

KaloyanTanev (approved):

PR #3106, opened 2024-05-29
cmd: exit with validator index, allow BN URLs 

### LINKED ISSUE #3104: Presign Exit by validator index
# 🎯 Problem to be solved

Right now you must exit a validator by its `distributed_validator_public_key`, this can be confusing as it is not the public key of the partial nor is it the same name as the flag in this command `--validator-public-key`. Most humans can parse the 6/7 digit validator index much more effectively than the BLS pubkey for valdiators, and I think it would be more convenient to handle and reason about when using these exit commands. 

# 🛠️ Proposed solution

It might be a big ux improvement to also add `--validator-index` as an alternative to `--validator-public-key`. 

Technically we might even be able to skip the `beacon-node-url` flag from `charon exit sign` if it was only being used to retrieve the state/index anyways.  (While i write this, is there any chance we can edit this/alias the `beacon-node-url` flag to be `beacon-node-endpoints` like `charon run`? can imagine lots of people being caught by that gotcha otherwise, imo we should favour consistency. 

- [ ] Core team consensus on the proposed solution

Describe the solution to be implemented

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

gsora:
Allow specifying more than one beacon node to use with the `--beacon-node-endpoints` flag.

Users can now sign exits directly with a validator index rather than specifying its public key.

If both are specified, validator liveliness (i.e. whether or not the validator exists on the beacon chain) is not checked.

`sign` checks that the specified validator index exists in the cluster lock before proceeding.

category: feature
ticket: #3104 

Closes #3104

### DIFF
--- a/cmd/exit_sign.go
+++ b/cmd/exit_sign.go
@@ -61,7 +61,7 @@ func newSubmitPartialExitCmd(runFunc func(context.Context, exitConfig) error) *c
 
 		if strings.TrimSpace(config.ValidatorPubkey) == "" && !valIdxPresent {
 			//nolint:revive // we use our own version of the errors package.
-			return errors.New(fmt.Sprintf("%s or %s must be specified.", validatorIndex.String(), validatorPubkey.String()))
+			return errors.New(fmt.Sprintf("either %s or %s must be specified at least.", validatorIndex.String(), validatorPubkey.String()))
 		}
 
 		config.ValidatorIndexPresent = valIdxPresent
--- a/cmd/exit_sign.go
+++ b/cmd/exit_sign.go
@@ -57,12 +57,15 @@ func newSubmitPartialExitCmd(runFunc func(context.Context, exitConfig) error) *c
 
 	wrapPreRunE(cmd, func(cmd *cobra.Command, _ []string) error {
 		valIdxPresent := cmd.Flags().Lookup(validatorIndex.String()).Changed
+		valPubkPresent := cmd.Flags().Lookup(validatorPubkey.String()).Changed
+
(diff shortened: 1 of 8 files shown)


cmd/exit_sign.go
```diff
@@ -45,13 +47,29 @@ func newSubmitPartialExitCmd(runFunc func(context.Context, exitConfig) error) *c
...
 
 	bindLogFlags(cmd.Flags(), &config.Log)
 
+	wrapPreRunE(cmd, func(cmd *cobra.Command, _ []string) error {
+		valIdxPresent := cmd.Flags().Lookup(validatorIndex.String()).Changed
+		valPubkPresent := cmd.Flags().Lookup(validatorPubkey.String()).Changed
+
+		if strings.TrimSpace(config.ValidatorPubkey) == "" && !valIdxPresent {
+			//nolint:revive // we use our own version of the errors package.
+			return errors.New(fmt.Sprintf("%s or %s must be specified.", validatorIndex.String(), validatorPubkey.String()))
```
pinebit, line 64: Neat: prefix with "either".

pinebit (approved):

cmd/exit_sign.go
gsora: Thinking about it more, this error is wrong: the user must _at least_ provide X or Y, or both, but never none.

Thanks for pointing this out!

KaloyanTanev (approved):

Issue #3108, opened 2024-05-29
Feature flag to enable JSON-encoded block submission


boulder225:
## 🎯 Problem to be solved
The Nimbus beacon node requires JSON-encoded blocks to be submitted via the `/eth/v2/beacon/blocks` endpoint. However, the current implementation of the go-eth2-client library does not support this format for Nimbus, leading to issues when using Nimbus as a beacon node in Kurtosis.

## 🛠️ Proposed solution
- [ ] Introduce a feature flag (e.g., `--feature-set-enable=json_requests`) in Charon to enable JSON-encoded block submission for Nimbus.
- [ ] Document the requirement for enabling the JSON-encoded block submission feature when using Nimbus as a beacon node in the release notes.
- [ ] Plan for eventual removal of the feature flag once Nimbus and other beacon nodes support the SSZ-encoded block format natively.

boulder225:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/3108?planning-poker) with Zenhub @gsora @KaloyanTanev @pinebit

PR #3110, opened 2024-05-29
app: json_requests feature

pinebit:
Add `WithEnforceJSON(true)` parameter when `json_requests` feature is enabled in `featureset`.
This is needed for Nimbus BN to properly handle SubmitProposal requests.

category: feature
ticket: #3108
feature_flag: json_requests

### DIFF
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -80,10 +80,7 @@ func NewMultiHTTP(timeout time.Duration, addresses ...string) (Client, error) {
 			eth2http.WithAddress(address),
 			eth2http.WithTimeout(timeout),
 			eth2http.WithAllowDelayedStart(true),
-		}
-
-		if featureset.Enabled(featureset.JSONRequests) {
-			parameters = append(parameters, eth2http.WithEnforceJSON(true))
+			eth2http.WithEnforceJSON(featureset.Enabled(featureset.JSONRequests)),
 		}
 
 		cl := newLazy(func(ctx context.Context) (Client, error) {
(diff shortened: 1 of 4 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -74,13 +75,19 @@ func NewMultiHTTP(timeout time.Duration, addresses ...string) (Client, error) {
...
 		address := address // Capture range variable.
 
+		parameters := []eth2http.Parameter{
+			eth2http.WithLogLevel(zeroLogInfo),
+			eth2http.WithAddress(address),
+			eth2http.WithTimeout(timeout),
+			eth2http.WithAllowDelayedStart(true),
+		}
+
+		if featureset.Enabled(featureset.JSONRequests) {
```
gsora, line 85: I think you can pass the result of Enabled to WithEnforceJSON, avoiding the if statement entirely

pinebit: Indeed. Fixed.

app/featureset/status_string.go
```diff
@@ -10,19 +10,30 @@ func _() {
 	// An "invalid array index" compiler error signifies that the constant values have changed.
 	// Re-run the stringer command to generate them again.
 	var x [1]struct{}
+	_ = x[enable-9223372036854775807]
+	_ = x[disable-0]
```
KaloyanTanev, line 14: What is the purpose of those two? Disable I can understand - if we want to disable the feature later on, but enable?

gsora: That file is auto generated by stringer

gsora (approved):

PR #3114, opened 2024-06-03
app/eth2wrap: deep validator cache

gsora:
Implement a validator cache that caches the complete Validators endpoint request.

Short-cut the duty selection for the DV validator set to the known-active, cached list instead: this will reduce the load on the beacon node drastically, since before this change we were calling the BN once every slot.

Try to reduce at a minimum upstream BN requests when the VC calls for validator state.

All Validators() calls to the upstream BN, if greater than 200 validators, will have an exponential timeout calculated as `(50ms * validator amount)`.

category: feature
ticket: none


core/validatorapi/validatorapi.go
```diff
@@ -960,27 +960,66 @@ func (c Component) SyncCommitteeDuties(ctx context.Context, opts *eth2api.SyncCo
...
+	opts.PubKeys = pubkeys
+
+	cachedValidators, err := c.eth2Cl.CompleteValidators(ctx)
 	if err != nil {
 		return nil, err
 	}
-	vals := eth2Resp.Data
 
-	convertedVals, err := c.convertValidators(vals, len(opts.Indices) == 0)
+	cvMap := make(map[eth2p0.BLSPubKey]struct{})
```
pinebit, line 995: `cachedValidators` is already a map. Why not traverse `opts.Indices` instead of `opts.PubKeys` to avoid extra maps and loop?

app/eth2wrap/valcache.go
```diff
@@ -65,28 +71,40 @@ func (c *ValidatorCache) Trim() {
...
+
+	return c.complete, c.complete != nil
+}
+
+// Get returns the cached active validators, cached complete Validators response, or fetches them if not available populating the cache.
+func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteValidators, error) {
+	completeCached, completeOk := c.cached()
+	activeCached, activeOk := c.activeCached()
+
+	if completeOk && activeOk {
```
pinebit, line 98: This looks like "premature optimization" to me. If this is called once per slot, then acquiring write lock vs read lock shall not really matter. As the result, the code looks more complicated than it should be.

core/validatorapi/validatorapi.go
```diff
@@ -960,27 +960,66 @@ func (c Component) SyncCommitteeDuties(ctx context.Context, opts *eth2api.SyncCo
...
+	}
+
+	var nonCachedPubkeys []eth2p0.BLSPubKey
+	for _, ncVal := range opts.PubKeys {
+		if _, ok := cvMap[ncVal]; !ok {
+			nonCachedPubkeys = append(nonCachedPubkeys, ncVal)
+		}
+	}
+
+	if len(nonCachedPubkeys) != 0 {
```
pinebit, line 1007: This logic is complex enough to deserve new unit tests.

pinebit:
> All Validators() calls to the upstream BN, if greater than 200 validators, will have an exponential timeout calculated as (50ms * validator amount).

Is this done in this PR? I did not find changes implementing this logic.

core/validatorapi/validatorapi.go
gsora: `opts.Indices` is an optional field, as is `opts.PubKeys`: not providing them returns the list of all the validators on the network.

Thinking about it some more, we should also handle the caching layer for indices... though, no VC ever queries with it.

gsora:
@pinebit you can find it here: https://github.com/ObolNetwork/charon/pull/3114/files#diff-9496d7e5dde67e0e9b5aef94238c8031bb904688be029b2e7683d871a0b02320R115

app/eth2wrap/valcache.go
gsora: I see your point.

Previous maintainer added that code path, but I believe before the RWLock. Will re-check.

core/validatorapi/validatorapi.go
pinebit: Understood. Then why valcache uses indices for mapping? Maybe it needs to use pubkeys?

gsora: > Understood. Then why valcache uses indices for mapping? Maybe it needs to use pubkeys?

I'm caching both the active validator list and the raw Validators response. We cache by index because that's the most common way of looking for validators, although the semantics of the beacon REST API might suggest otherwise.

pinebit (approved):

KaloyanTanev (approved):

PR #3119, opened 2024-06-05
workflows: composite GitHub actions

KaloyanTanev:
Reuse `setup-go` step throughout the pipeline. This way it would be easier to update go version in the future and we ensure consistency. If this approach is mutually agreed to be better, we can apply it to other parts of the pipeline as well.

category: refactor
ticket: none


gsora:
This PR is a blessing for when we have to update to a newer Go version!

PR #3149, opened 2024-06-22
core/tracker: track blinded proposals

gsora:
Do the same type casting we do in core/bcast to switch on the type of block proposal.

Add a test to check for this property.

category: bug
ticket: none


KaloyanTanev (approved):
Approving now, so we potentially see if this will fix block proposals for during the weekend. We will jump into more details for `VersionedSignedProposal` and `VersionedSignedBlindedProposal` on Monday.

PR #3150, opened 2024-06-22
core/tracker: harden inclusion proposal checks

gsora:
Always recover when dealing with malformed proposals, and log some context to aid debugging.

category: refactor
ticket: none


KaloyanTanev (approved):
LGTM! In the future it makes sense to recover from anything coming from eth2wrap or make the code safer (i.e.: check if variables we try to access exist). What do you think?

PR #3155, opened 2024-06-25
core: deprecate VersionedSignedBlindedProposal

gsora:
Convert `VersionedSignedProposals` to their blinded counterpart (requested by go-eth2-client & Beacon API spec) on-demand, so that the execution flow is simpler to understand and debug.

category: refactor
ticket: none


core/signeddata.go
```diff
@@ -181,11 +178,57 @@ func NewPartialVersionedSignedProposal(proposal *eth2api.VersionedSignedProposal
 	}, nil
 }
 
+func NewVersionedSignedBlindedProposal(bp *eth2api.VersionedSignedBlindedProposal) (VersionedSignedProposal, error) {
```
pinebit, line 181: IMO these function names do not follow go idioms as they don't return *BlindedProposal and therefore may be confusing in future.
I would rename these to something like `NewVersionedSignedProposalFromBlindedProposal`...

core/tracker/inclusion.go
```diff
@@ -104,38 +104,32 @@ func (i *inclusionCore) Submitted(duty core.Duty, pubkey core.PubKey, data core.
...
 			}
 		}()
 
-		switch blinded {
+		switch block.Blinded {
 		case true:
-			if eth2wrap.IsSyntheticBlindedBlock(&blindedBlock.VersionedSignedBlindedProposal) {
+			blinded, err := block.ToBlinded()
+			if err != nil {
+				return errors.Wrap(err, "cannot broadcast, expected blinded proposal")
```
pinebit, line 129: "cannot broadcast" is wrong wording IMO, let it be just "expected blinded proposal"?

testutil/fuzz.go
```diff
@@ -26,11 +26,11 @@ import (
 func NewEth2Fuzzer(t *testing.T, seed int64) *fuzz.Fuzzer {
 	t.Helper()
 
-	blindedVersions := []eth2spec.DataVersion{
-		eth2spec.DataVersionBellatrix,
-		eth2spec.DataVersionCapella,
-		eth2spec.DataVersionDeneb,
-	}
+	// blindedVersions := []eth2spec.DataVersion{
```
pinebit, line 29: I think we can uncomment these, should be easy to fix the test

core/signeddata.go
gsora: > `NewVersionedSignedProposalFromBlindedProposal`

As painful as it looks, you're right :D Will rename.

core/testdata/TestJSONSerialisation_VersionedSignedProposal.json#01.golden
```diff
@@ -0,0 +1,350 @@
+{
```
KaloyanTanev, line 1: If we add this, do we still need to keep the golden blinded equivalent? Also, probably it would be nice to include `"blinded": false` in the non-`#01` version.

KaloyanTanev (approved):

core/testdata/TestJSONSerialisation_VersionedSignedProposal.json#01.golden
gsora: hmm well there's no value so it'll be false by default 

> keep the golden blinded equivalent

you mean the golden files for VersionedSignedBlindedProposal?

pinebit (approved):

PR #3169, opened 2024-07-11
app/stacksnipe: detect validator services

gsora:
Detect Ethereum validator stack components by reading `/proc` and filtering processes based on an allowlist.

To support containerized deployments, add `--proc-directory` flag to `charon run`: if not configured, Charon will log an `INFO` message saying it won't do anything.

Support detecting interpreted components like `lodestar`.

Expose the resulting data through the `app_validator_stack_params` Prometheus metric.

By default, poll the specified directory every 15 seconds.

category: feature
ticket: none


app/stacksnipe/stacksnipe.go
```diff
@@ -0,0 +1,227 @@
...
+		cmdLineStr := strings.Join(cmdLine, " ")
+
+		log.Debug(ctx, "Detected stack component", z.Str("name", vcName), z.U64("host_pid", hostPID), z.Str("cmdline", cmdLineStr))
+
+		wb <- StackComponent{
+			Name:      vcName,
+			CLIParams: cmdLineStr,
+		}
+
+		return nil
```
pinebit, line 225: Maybe to respect `ctx` state to some extent, we could return `ctx.Err()`?

pinebit (commented):
Provided a few minor comments

app/stacksnipe/stacksnipe.go
```diff
@@ -0,0 +1,227 @@
...
+	"strings"
+	"time"
+
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+const (
+	pollInterval = 15 * time.Second
```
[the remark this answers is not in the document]
gsora: On a restart it means there's a 1 minute lag before correct data is displayed though: even considering the outcome of this metric isn't supposed to change often, it's better to provide up-to-date data often.

Also considering it's `/proc`, reading often is not an issue performance-wise.

app/stacksnipe/stacksnipe.go
gsora: We could, but I figured it's not a big deal in this component, we just want to stop the servicce.

app/stacksnipe/stacksnipe.go
```diff
@@ -0,0 +1,227 @@
...
+// whose CLI parameters (also called cmdline) is read from a /proc-like filesystem.
+type StackComponent struct {
+	Name      string
+	CLIParams string
+}
+
+// Instance returns an instance of stacksnipe.
+type Instance struct {
+	procPath    string
+	metricsFunc func([]string, []string)
```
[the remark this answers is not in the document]
gsora: I see your point, but I didn't want to have `app` depend on this package more than it is already :-/

pinebit (approved):
This must be great feature!

cmd/run.go
```diff
@@ -91,6 +91,7 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
 	cmd.Flags().Uint64Var(&config.TestnetConfig.ChainID, "testnet-chain-id", 0, "Chain ID of the custom test network.")
 	cmd.Flags().Int64Var(&config.TestnetConfig.GenesisTimestamp, "testnet-genesis-timestamp", 0, "Genesis timestamp of the custom test network.")
 	cmd.Flags().StringVar(&config.TestnetConfig.CapellaHardFork, "testnet-capella-hard-fork", "", "Capella hard fork version of the custom test network.")
+	cmd.Flags().StringVar(&config.ProcDirectory, "proc-directory", "", "Directory to look into in order to detect other stack components running on the host.")
```
KaloyanTanev, line 94: Probably we can add an example of such dir?

KaloyanTanev (approved):

Issue #3171, opened 2024-07-12
Create a `charon test mev` command to check the performance and connectivity of MEV relays


boulder225:
## 🎯 Problem to be solved
We need a way to initially check the performance and connectivity of MEV relays for DV setups to help prevent issues with block production.

## 🛠️ Proposed solution
- [x] Create a `charon test mev` command
- [x] Implement an initial check for fast connection to MEV relays
- [x] Add the ability to measure basic response times from relays
- [x] Include this test in the charon setup process
- [x] Provide clear feedback to users about their MEV relay setup quality
- [x] Add documentation explaining the limitations of this test (e.g., performance may vary over time, which could also be related to general mev relay overload)

PR #3176, opened 2024-07-18
cmd: mev ping tests

KaloyanTanev:
Add `charon test mev` command for pings towards MEV relays. In the future more tests towards MEV relays can be added.

category: feature
ticket: #3171

### DIFF
--- a/cmd/testmev.go
+++ b/cmd/testmev.go
@@ -1,0 +1,263 @@
+// Copyright © 2022-2024 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptrace"
+	"time"
+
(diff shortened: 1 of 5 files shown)


cmd/testmev.go
```diff
@@ -0,0 +1,263 @@
...
+type testMEVConfig struct {
+	testConfig
+	Endpoints []string
+}
+
+type testCaseMEV func(context.Context, *testMEVConfig, string) testResult
+
+const (
+	thresholdMEVMeasureAvg = 40 * time.Millisecond
+	thresholdMEVMeasureBad = 100 * time.Millisecond
```
KaloyanTanev, line 30: @LukeHackett12 can I get your opinion on those 2 values? Those are the thresholds after which we consider average or bad performance for a RTT to a MEV relay.

KaloyanTanev:
golangci-lint should pass after #3178 is merged.

pinebit (approved):

PR #2850, opened 2024-02-02
testutil: replace math/rand deprecation

KaloyanTanev:
- Update global `rand.Seed(x)` with local `rand.New(rand.NewSource(x))`;
- Update test random functions that used global seed with local seed;
- Update `math/rand.Read(x)` with `crypto/rand.Read(x)` (not included in the issue, but deprecated as well).

category: refactor
ticket: #2831

PR #3178, opened 2024-07-18
fixbuild: bump linter version

KaloyanTanev:
golangci-lint was on older version that did not support go1.22 yet, which causes false alarms.

---

This update introduced a lot of new, deprecated and modified linters. There are a lot of changes in the files and I would advise for easier review to go over the commits, as I've tried to keep them concise, so there are only changes from one linter or if many, the changes are little. Most eye catching changes in the files are:

- `testifylint`: we had quite a lot of assertions inside of go routines, this is a bit of a red flag, because as observed in [this small example](https://go.dev/play/p/WoBGMiKQDEk), a failed assertion might or might not be caught
- The `golangci-lint` used in pre-commit was a default `golangci-lint` from [here](https://github.com/golangci/golangci-lint), using its default config, which made the linter pass locally all the time. This is something I've observed multiple times ([dating back to my very first PR](https://github.com/ObolNetwork/charon/pull/2850#issuecomment-1927408052) :)). Now the pre-commit hook is running the locally installed `golangci-lint` tool and it checks if the version is the same as the one in the pipelines.
- Added `max-same-issues=0` and `max-issues-per-linter=0` so there is no limit on what we see in the output. Previously the default of 3 was used, so if there were hundreds of errors with one linter, it would have displayed only 3, which made it quite a hassle to fix.
- Bump the `golangci-lint` action's version to v6. It now points where exactly the issue is, which is great!

category: fixbuild
ticket: #3179

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -19,7 +19,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.54.2
+          version: v1.59.1
       - name: notify failure
         if: failure() && github.ref == 'refs/heads/main'
         env:
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-go
       - name: golangci-lint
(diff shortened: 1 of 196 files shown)


KaloyanTanev:
golangci-lint should pass after #3178 is merged.

pinebit (approved):

gsora (approved):

pinebit (approved):

PR #3188, opened 2024-07-23
*: pin lighthouse version in compose tests

KaloyanTanev:
- Pin lighthouse version in Dockerfile for compose tests.
- Remove stdin flag (doesn't exist in new version).
- Add Docker images checks for dependabot.

category: bug
ticket: none


pinebit (approved):
I hope this works. Thank you!

gsora (approved):

PR #3198, opened 2024-07-26
*: flag refactors

KaloyanTanev:
- Change Obol's API default URL in flags to include `/v1`
- validator-pubkey to validator-public-key in flag description
- Rename test command measurement from Bad to Poor
- Increase test peer unit tests timeout
- bump ethdo version in compose tests to 1.35.2

category: misc
ticket: none

### DIFF
--- a/cmd/exit.go
+++ b/cmd/exit.go
@@ -115,7 +115,7 @@ func bindExitFlags(cmd *cobra.Command, config *exitConfig, flags []exitCLIFlag) 
 
 		switch flag {
 		case publishAddress:
-			cmd.Flags().StringVar(&config.PublishAddress, publishAddress.String(), "https://api.obol.tech", maybeRequired("The URL of the remote API."))
+			cmd.Flags().StringVar(&config.PublishAddress, publishAddress.String(), "https://api.obol.tech/v1", maybeRequired("The URL of the remote API."))
 		case beaconNodeEndpoints:
 			cmd.Flags().StringSliceVar(&config.BeaconNodeEndpoints, beaconNodeEndpoints.String(), nil, maybeRequired("Comma separated list of one or more beacon node endpoint URLs."))
 		case privateKeyPath:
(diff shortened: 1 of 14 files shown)


cmd/exit.go
```diff
@@ -137,7 +137,7 @@ func bindExitFlags(cmd *cobra.Command, config *exitConfig, flags []exitCLIFlag)
 		case publishTimeout:
 			cmd.Flags().DurationVar(&config.PublishTimeout, publishTimeout.String(), 30*time.Second, "Timeout for publishing a signed exit to the publish-address API.")
 		case validatorIndex:
-			cmd.Flags().Uint64Var(&config.ValidatorIndex, validatorIndex.String(), 0, "Validator index of the validator to exit, the associated public key must be present in the cluster lock manifest. If --validator-pubkey is also provided, validator liveliness won't be checked on the beacon chain.")
+			cmd.Flags().Uint64Var(&config.ValidatorIndex, validatorIndex.String(), 0, "Validator index of the validator to exit, the associated public key must be present in the cluster lock manifest. If --validator-public-key is also provided, validator liveliness won't be checked on the beacon chain.")
```
OisinKyne, line 140: Sorry for the small nit, would make the same tweak on the `validator-public-key` description. 'Liveliness' is a strange word to use. 

Sidenote: If you provide an index and no public key, will it fetch the validator from the chain, get its public key, check for its presence in the lockfile and continue gracefully? I assume yes. 

```suggestion
			cmd.Flags().Uint64Var(&config.ValidatorIndex, validatorIndex.String(), 0, "Validator index of the validator to exit, the associated public key must be present in the cluster lock manifest. If --validator-public-key is also provided, validator existence won't be checked on the beacon chain.")
```

cmd/test.go
```diff
@@ -105,7 +105,7 @@ const (
 	// measurement tests
 	testVerdictGood testVerdict = "Good"
 	testVerdictAvg  testVerdict = "Avg"
```
OisinKyne, line 107:
```suggestion
	testVerdictAvg  testVerdict = "Okay"
```

OisinKyne (approved):
My suggestion about preferring "Okay" instead of "Average" can be ignored until a future PR. Again its a nitpick, we don't really know what the average latency is to a relay, so okay is a more vague word than average. :)

cmd/exit.go
KaloyanTanev: Yes, that's correct!

cmd/test.go
KaloyanTanev: Let's revise those again when we improve the whole scoring system? Then we can have poor, average, okay, perfect, etc.

PR #3203, opened 2024-07-30
*: add fio tests for disk performance

KaloyanTanev:
- Change the existing disk write test to use fio
- Add disk write IOPS test, disk read speed test, disk read IOPS test
- Update Dockerfile to include fio

category: feature
ticket: #3207

### DIFF
--- a/cmd/testperformance.go
+++ b/cmd/testperformance.go
@@ -53,12 +53,12 @@ const (
 	diskOpsMBsTotal       = 4096 // split between number of jobs
 	diskWriteSpeedMBsAvg  = 1000
 	diskWriteSpeedMBsPoor = 500
-	diskWriteIOPSAvg      = 200
-	diskWriteIOPSPoor     = 100
+	diskWriteIOPSAvg      = 2000
+	diskWriteIOPSPoor     = 1000
 	diskReadSpeedMBsAvg   = 1000
 	diskReadSpeedMBsPoor  = 500
-	diskReadIOPSAvg       = 200
-	diskReadIOPSPoor      = 100
+	diskReadIOPSAvg       = 2000
+	diskReadIOPSPoor      = 1000
 
 	availableMemoryMBsAvg  = 4000
 	availableMemoryMBsPoor = 2000
--- a/cmd/testperformance.go
+++ b/cmd/testperformance.go
@@ -20,7 +20,6 @@ import (
 	"github.com/showwin/speedtest-go/speedtest"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
-	"golang.org/x/sys/unix"
 
 	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/app/log"
@@ -51,6 +50,7 @@ type fioResultSingle struct {
 
 const (
 	diskOpsNumOfJobs      = 8
+	diskOpsMBsTotal       = 4096 // split between number of jobs
 	diskWriteSpeedMBsAvg  = 1000
 	diskWriteSpeedMBsPoor = 500
 	diskWriteIOPSAvg      = 200
@@ -217,12 +217,12 @@ func testPerformance(ctx context.Context, queuedTests []testCaseName, allTests m
 	}
 }
 
-func fioCommand(ctx context.Context, filename string, fileSize int, blocksize int, operation string) ([]byte, error) {
+func fioCommand(ctx context.Context, filename string, blocksize int, operation string) ([]byte, error) {
 	//nolint:gosec
 	cmd, err := exec.CommandContext(ctx, "fio",
 		"--name=fioTest",
 		fmt.Sprintf("--filename=%v", filename),
-		fmt.Sprintf("--size=%vMb", fileSize),
+		fmt.Sprintf("--size=%vMb", diskOpsMBsTotal/diskOpsNumOfJobs),
 		fmt.Sprintf("--blocksize=%vk", blocksize),
 		fmt.Sprintf("--numjobs=%v", diskOpsNumOfJobs),
 		fmt.Sprintf("--rw=%v", operation),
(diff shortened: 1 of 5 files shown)


KaloyanTanev:
Sonarcloud code duplication does not pass, but I personally do not agree with it...

cmd/testperformance.go
```diff
@@ -28,19 +29,42 @@ import (
...
-	diskWriteMBsAvg               = 1000
-	diskWriteMBsPoor              = 500
-	availableMemoryMBsAvg         = 4000
-	availableMemoryMBsPoor        = 2000
-	totalMemoryMBsAvg             = 8000
-	totalMemoryMBsPoor            = 4000
+	diskOpsNumOfJobs      = 8
+	diskWriteSpeedMBsAvg  = 1000
+	diskWriteSpeedMBsPoor = 500
+	diskWriteIOPSAvg      = 200
```
gsora, line 56: 200 IOPS as ok/avg measure seems ill-advised, considering mechanical disks can reach 150 IOPS :)

My Linux desktop SSD reaches almost 80**k** IOPS in random 4KB block writes, more than 250**k** IOPS in the same class of reads.

[This](https://cloud.google.com/compute/docs/disks/benchmarking-pd-performance?hl=it) Google Cloud page has a good reference for `fio` command combos to test IOPS performance.

We should raise those numbers 10x at least IMO.

@OisinKyne thoughts?

cmd/testperformance.go
```diff
@@ -187,102 +217,241 @@ func testPerformance(ctx context.Context, queuedTests []testCaseName, allTests m
...
+	var err error
+	if conf.DiskIOTestFileDir == "" {
+		dir, err = os.UserHomeDir()
+		if err != nil {
+			return 0, "", errors.Wrap(err, "get user home dir")
+		}
+	} else {
+		dir = conf.DiskIOTestFileDir
+	}
+	err = unix.Statfs(dir, &stat)
```
gsora, line 253: Calling `unix` here effectively makes Charon *NIX only.

Can we avoid this?

cmd/testperformance.go
```diff
@@ -187,102 +217,241 @@ func testPerformance(ctx context.Context, queuedTests []testCaseName, allTests m
...
-	actualDiskWriteMB := conf.DiskWriteMB
+	diskWriteMB := 8192
 
-	for availableMB < actualDiskWriteMB {
-		log.Warn(ctx, fmt.Sprintf("Insufficient available disk space of %vMB, reducing the test size to %vMB. Note that this might result in slower write speed", availableMB, actualDiskWriteMB), nil)
-		actualDiskWriteMB /= actualDiskWriteMB
-		if actualDiskWriteMB == 0 {
-			return failedTestResult(testRes, errors.New("insufficient available disk space", z.Str("available_space", strconv.Itoa(availableMB)+"MB")))
+	for availableMB < diskWriteMB {
+		log.Warn(ctx, fmt.Sprintf("Insufficient available disk space of %vMB, reducing the test size of %vMB to %vMB. Note that this might result in lower performance", availableMB, diskWriteMB, diskWriteMB/2), nil)
```
gsora, line 263: I would leave this to `fio` to determine.

cmd/testperformance.go
```diff
@@ -187,102 +217,241 @@ func testPerformance(ctx context.Context, queuedTests []testCaseName, allTests m
...
-		if err != nil {
-			return failedTestResult(testRes, err)
-		}
-		diskWriteTotal += time
+	testFilePath := dir + "/fioTest"
+
+	return diskWriteMB, testFilePath, nil
+}
+
+func performanceDiskWriteSpeedTest(ctx context.Context, conf *testPerformanceConfig) testResult {
```
gsora, line 275: Wondering if we could run one `fio` test for read/write, and read disk speed and IOPS from the same output.

gsora (commented):
Left some comments, I have doubts about the test parameters.

That said, I think we should concentrate disk tests around a block size that makes sense for validator nodes, considering blobs etc.

I'd love @OisinKyne input on this.

gsora (commented):
Left some comments, I have doubts about the test parameters.

That said, I think we should concentrate disk tests around a block size that makes sense for validator nodes, considering blobs etc.

I'd love @OisinKyne input on this.

cmd/testperformance.go
KaloyanTanev: On my basic (but still latest) Macbook air with M3 chip I get ~1000 IOPS.

The numbers I have picked here are based only on my assumptions and I am definitely up for changing them if we can base them on some data!

gsora: > I get ~1000 IOPS

I think what we're seeing here is an unoptimized IO path (`fio`'s default on macOS) versus an highly-optimized one on Linux (`libaio`): the latter resembles more what a real-world disk can do!

cmd/testperformance.go
KaloyanTanev: I need to specify size, it is a required parameter when running `fio`. It crashes if the specified size is more than the allowed, so that's why I'm trying with lower and lower file size here.

cmd/testperformance.go
KaloyanTanev: Hmmm, I thought it is already 😄. I will think about another approach

cmd/testperformance.go
gsora: We should choose an appropriate size for the test that we need to run, not the other way around :D

I would go with 4GB, if a user is running this test on an unsuitable machine, It'll be apparent lol

cmd/testperformance.go
KaloyanTanev: Yes, that's definitely possible, we will save some time running the command. But on the other hand, I want to make the tests completely self sufficient and 1 test = 1 metric. Also easier to test 1 single thing by filtering the tests with the `--testcases` flag. EOD we are not wasting so much time running `fio`, compared to other performance tests here like the network connectivity.

gsora: I've seen several examples of `fio` commands that have a runtime defined in order to have trusted results.

Maybe we need  to run `fio` with `--runtime=30s`?

KaloyanTanev: I already have `--runtime=60s` in my `fio` command. This runtime though is not really a runtime, but more of a... [timeout](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-runtime).

If we want to have continuous tests over X period of time, we can use probably the [time_based parameter](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-time_based).

cmd/testperformance.go
KaloyanTanev: Might very much be! I had the idea in my head to include the OS in this test as well, but didn't find big enough reason to do it on this first iteration. Now probably there is :)

pinebit (approved):

gsora (approved):

Issue #3181, opened 2024-07-22
Fix Signature Verification Issue Causing Missed Blocks on Gnosis Chain


boulder225:
## 🎯 Problem to be solved
An issue in Charon is apparently causing validators on the Gnosis chain to miss block proposals. The issue appears to be related to a v1 endpoint failing to verify signatures, likely due to problems with the state root or version used by Charon for verification. This impacts the performance of Gnosis validators, with even a small set of 30 validators missing proposals every 20 minutes.

## 🛠️ Proposed solution
- [x] Add detailed logging to the problematic functions (verify eth2 sign data, verify partial sig) to gather more information
- [x] Investigate differences in signing or key mechanisms specific to Gnosis compared to other chains
- [x] Arrange a call with external experts (e.g., Dappline) to review the code and discuss potential solutions
- [x] Analyze the state root and version handling in Charon, focusing on Gnosis-specific implementations

`15:49:00.138 ERRO vapi Validator api 5xx response: signature not verified {"status_code": 500, "message": "Internal server error", "duration": "3.133976ms", "vapi_endpoint": "submit_proposal_v1"} tbls/herumi.go:297 .Verify tbls/tbls.go:122 .Verify eth2util/signing/signing.go:109 .Verify core/eth2signeddata.go:41 .VerifyEth2SignedData core/validatorapi/validatorapi.go:1110 .verifyPartialSig core/validatorapi/validatorapi.go:417 .SubmitProposal core/validatorapi/router.go:804 .func11
`

pinebit:
This ticket is used to track the work on temporary solution: relaxing partial signatures verification in the consensus flow.

PR #3238, opened 2024-08-22
core/validatorapi: workaround for Gnosis proposals

gsora:
If the "gnosis_block_hotfix" feature flag is enabled, use the Gnosis-compatible representation of the proposal when calculating its hash tree root.  This should fix Gnosis proposals failing due to bad signature.

category: feature
ticket: #3181 
feature_flag: gnosis_block_hotfix


core/signeddata.go
```diff
@@ -320,6 +322,12 @@ func (p VersionedSignedProposal) MessageRoot() ([32]byte, error) {
 			return p.DenebBlinded.Message.HashTreeRoot()
 		}
 
+		if p.IsGnosis {
```
pinebit, line 325: Wondering if start dealing with blinded blocks, shouldn't this `IsGnosis` come first, before `.Blinded` check?

pinebit: What if instead of this flag we simply test for `featureset.Enabled(featureset.GnosisBlockHotfix)`?

pinebit (approved):
Approved with comments.
Plus consider adding that "real block" unit test.

core/signeddata.go
gsora: Implemented!

go.mod
```diff
@@ -205,3 +205,7 @@ require (
 
 // We're replacing kryptology with our own fork, which fixes dependencies' security vulnerabilities.
 replace github.com/coinbase/kryptology => github.com/ObolNetwork/kryptology v0.0.0-20231016091344-eed023b6cac8
+
+// We're replacing go-eth2-client with a branch off our fork, at version v0.21.10.
+// This is needed to ensure Gnosis compatibility.
+replace github.com/attestantio/go-eth2-client => github.com/ObolNetwork/go-eth2-client v0.21.11-0.20240822135044-f0a5b21e02c6
```
KaloyanTanev, line 211: Wouldn't that mean we will have to manage this repo constantly and rebase it every time there is a new version of go-eth2-client?

gsora: Hopefully go-eth2-client will fix this upstream soon

KaloyanTanev (approved):

Issue #3241, opened 2024-08-28
Missing threshold parameter check for the create cluster command


aadomn:
# 🐞 Bug Report

### Description

The `create cluster` command does not prevent misconfigurations where the threshold exceeds the number of nodes (eg 4-out-of-3 setting). This can lead to erroneous secret sharings that will not be capable of threshold signing nor reconstructing the joint secret.

Similarly, the `create dkg` command allows to pass `--threshold=1` as input parameter but the subsequent `dkg` command will fail because the kryptology library returns an error if the threshold is smaller than 2.

### Has this worked before in a previous version?
Yes, in all previous versions.

## 🔬 Minimal Reproduction

```
charon create cluster \
    --nodes=3 \
    --network=holesky \
    --num-validators=1 \
    --name="Quickstart Guide Cluster" \
    --cluster-dir="cluster" \
    --fee-recipient-addresses=0x000000000000000000000000000000000000dead \
    --withdrawal-addresses=0x000000000000000000000000000000000000dead \
    --threshold=4
```

## 🔥 Error

When trying to recombine the joint secret, it prints the following:

```
charon combine --cluster-dir="./cluster" --output-dir="./combined"
07:47:15.008 INFO cmd        Recombining private key shares           {"input_dir": "/opt/charon/cluster", "output_dir": "/opt/charon/combined"}
07:47:15.044 INFO cmd        Loading keystore                         {"path": "/opt/charon/cluster/node0/validator_keys"}
07:47:15.109 INFO cmd        Loading keystore                         {"path": "/opt/charon/cluster/node1/validator_keys"}
07:47:15.173 INFO cmd        Loading keystore                         {"path": "/opt/charon/cluster/node2/validator_keys"}
07:47:15.237 ERRO cmd        Fatal error: insufficient private key shares found for validator {"validator_index": 0, "expected": 4, "actual": 3}
	cmd/combine/combine.go:99 .Combine
	cmd/combine.go:56 .newCombineFunc
	cmd/combine.go:31 .func1
	cmd/cmd.go:91 .func1
	main.go:19 .main
```

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 24.04 LTS
  </code>
</pre>

**What version of Charon are you running? (Which release)**

<pre>
  <code>
1.0.1
  </code>
</pre>

**Anything else relevant (validator index / public key)?**

PR #3242, opened 2024-08-28
cmd: hardening threshold parameter checks

aadomn:
Ensure the threshold parameter given as input to the `create dkg` and `create cluster` commands is sound.

category: bug
ticket: #3241

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/url"
 	"os"
 	"path"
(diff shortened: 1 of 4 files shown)


gsora:
@aadomn could you please add unit tests for those two edge cases?

aadomn:
> @aadomn could you please add unit tests for those two edge cases?

@gsora done :heavy_check_mark:

gsora (approved):

cmd/createcluster.go
```diff
@@ -51,6 +51,7 @@ const (
 	zeroAddress    = "0x0000000000000000000000000000000000000000"
 	defaultNetwork = "mainnet"
 	minNodes       = 3
+	minThreshold   = 2
```
pinebit, line 54: Shall this be `ceil(n*2/3)`?

cmd/createcluster.go
```diff
@@ -366,6 +368,15 @@ func validateCreateConfig(ctx context.Context, conf clusterConfig) error {
 		}
 	}
 
+	// check for threshold parameter
+	if conf.Threshold < minThreshold {
+		return errors.New("threshold cannot be smaller than 2", z.Int("threshold", conf.Threshold))
```
pinebit, line 373: Don't mention 2 in the message, use `z.Int` param similar to how `create dkg` is made.

pinebit (changes requested):
Need changes

cmd/createcluster.go
aadomn: done

cmd/createcluster.go
aadomn: I left `minThreshold=2` here for tests and to clearly indicate that is the minimum threshold charon supports (when using 3 nodes only).
But indeed when validating the configuration it checks that the threshold is at least `ceil(n*2/3)`.

PR #3246, opened 2024-09-03
cmd: unified create validation

pinebit:
Unified `create cluster` and `create dkg` commands validation:
* The minimum number of nodes (operators) is 3.
* The threshold cannot be greater than the number of nodes (operators).

category: refactor
ticket: none


pinebit:
Shall be closed in favor of https://github.com/ObolNetwork/charon/pull/3242/files

PR #3248, opened 2024-09-03
cmd: refactor exits

KaloyanTanev:
Refactor exits logic. There is no change in the logic anywhere. The main differences are:
- fetching the validator index and validator pub key are in separate functions, which makes it more readable what is the flow
- `ExpertMode` is renamed to `SkipBeaconNodeCheck` as this is essentially what it does

If others agree, I'm in favour of completely removing the `SkipBeaconNodeCheck` config, as it is unnecessary.

category: refactor
ticket: none

### DIFF
--- a/cmd/exit_sign.go
+++ b/cmd/exit_sign.go
@@ -60,7 +60,7 @@ func newSubmitPartialExitCmd(runFunc func(context.Context, exitConfig) error) *c
 		valIdxPresent := cmd.Flags().Lookup(validatorIndex.String()).Changed
 		valPubkPresent := cmd.Flags().Lookup(validatorPubkey.String()).Changed
 
-		if valIdxPresent && !valIdxPresent {
+		if !valPubkPresent && !valIdxPresent {
 			//nolint:revive // we use our own version of the errors package.
 			return errors.New(fmt.Sprintf("either %s or %s must be specified at least.", validatorIndex.String(), validatorPubkey.String()))
 		}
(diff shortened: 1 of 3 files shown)


pinebit (approved):

cmd/exit_sign.go
```diff
@@ -60,13 +60,13 @@ func newSubmitPartialExitCmd(runFunc func(context.Context, exitConfig) error) *c
 		valIdxPresent := cmd.Flags().Lookup(validatorIndex.String()).Changed
 		valPubkPresent := cmd.Flags().Lookup(validatorPubkey.String()).Changed
 
-		if strings.TrimSpace(config.ValidatorPubkey) == "" && !valIdxPresent {
+		if valIdxPresent && !valIdxPresent {
```
gsora, line 63: Search/replace error?

cmd/exit_sign.go
```diff
@@ -100,110 +100,124 @@ func runSignPartialExit(ctx context.Context, config exitConfig) error {
...
-		ctx = log.WithCtx(ctx, z.U64("validator_index", config.ValidatorIndex))
-	default:
-		ctx = log.WithCtx(ctx, z.Str("validator", validator.String()))
+	eth2Cl, err := eth2Client(ctx, config.BeaconNodeEndpoints, config.BeaconNodeTimeout, [4]byte(cl.GetForkVersion()))
+	if err != nil {
+		return errors.Wrap(err, "cannot create eth2 client for specified beacon node")
 	}
 
-	shareIdx, err := keystore.ShareIdxForCluster(cl, *identityKey.PubKey())
+	logFields := []z.Field{}
```
gsora, line 118: I don't see the point of creating a slice with a single element, just to pass it to `log.WithCtx`, I feel the old way is a slightly more readable.

cmd/exit_sign.go
```diff
@@ -100,110 +100,124 @@ func runSignPartialExit(ctx context.Context, config exitConfig) error {
...
+	if config.ValidatorIndexPresent {
+		logFields = append(logFields, z.U64("validator_index", config.ValidatorIndex))
+	}
+	if config.ValidatorPubkey != "" {
+		logFields = append(logFields, z.Str("validator_pubkey", config.ValidatorPubkey))
+	}
+	ctx = log.WithCtx(ctx, logFields...)
+
+	if config.SkipBeaconNodeCheck {
```
gsora, line 127: Is this conditional the only instance of usage of this flag?

gsora (commented):
Generally I like the idea, left a few comments that need clarification.

cmd/exit_sign.go
KaloyanTanev: Ouf...

cmd/exit_sign.go
KaloyanTanev: Yes, I'm in favour of totally removing it though.

cmd/exit_sign.go
KaloyanTanev: now when you say it... I think I have even better way :)

gsora (approved):

PR #3261, opened 2024-09-09
*: bump golang to 1.23.1

KaloyanTanev:
Security fixes...

category: misc
ticket: none


pinebit (approved):
Shall we update go.mod as well?
Also, maybe we don't need to update to every "minor" release?

Issue #3243, opened 2024-08-30
Add `--all` for exiting all active validators


KaloyanTanev:
# 🎯 Problem to be solved
Currently if we want to exit all active validators in the cluster, it is quite cumbersome process, having to specify every validator's PK or index. Having big number of validators (i.e.: 1000) makes this quite slow. It would be great if users can simply add `--all` flag and exit all active validators.

# 🛠️ Proposed solution
- [ ] Fetch current active validators
- [ ] Exit all upon passing `--all` flag

# 🧪 Tests
- [ ] Tested by new automated unit tests
- [ ] Manually tested on local Kurtosis devnet
- [ ] Manually tested on test Holesky clusters

# ❌ Out of Scope
Exiting all validators using the VC. This ticket is only about implementing this for charon exits.

PR #3272, opened 2024-09-11
*: add --all for exit sign command

KaloyanTanev:
Add `--all` command for signing partial exits. This PR is one of a couple incoming PRs that will be for the `--all` functionality. Mind that the CLI flag is not enabled until all of them are implemented and merged.

category: feature
ticket: #3243

### DIFF
--- a/cmd/exit_sign.go
+++ b/cmd/exit_sign.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	eth2api "github.com/attestantio/go-eth2-client/api"
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	libp2plog "github.com/ipfs/go-log/v2"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


pinebit (approved):

cmd/exit_sign.go
```diff
@@ -126,60 +133,119 @@ func runSignPartialExit(ctx context.Context, config exitConfig) error {
...
 	}
 
+	var exitBlobs []obolapi.ExitBlob
+	if config.All {
+		exitBlobs, err = signAllValidatorsExits(ctx, config, eth2Cl, shares)
+		if err != nil {
+			return errors.Wrap(err, "could not sign exits for all validators")
+		}
+	} else {
+		exitBlobs, err = signSingleValidatorExit(ctx, config, eth2Cl, shares)
```
gsora, line 143: Can we use signAllValidatorsExits here, and instead of passing all shares we pass a slice with the single share we're interested in?

KaloyanTanev: Yes, I've been thinking about that, but when we are signing for single validator we are checking the config if the user passed pubkey or index or both, in exit all scenario we don't have none of those. We can still use the same function for both cases, but I will have to add some flags and checks which to determine if we have to check the config or not. It will make the code less readable imo.

PR #3281, opened 2024-09-12
*: optimize Dockerfile

No0key:
Hello!

This PR optimizes the `Dockerfile`. 

Changes include:
- Combined RUN commands to reduce the number of layers
- Cached Go dependencies to speed up build times
- Cleaned up unnecessary apt files to reduce image size
- Consolidated user creation and permission setting commands

Reduced image size from 370MB to 302MB

category: refactor
ticket: none


Dockerfile
```diff
@@ -1,14 +1,22 @@
...
+
 # Install dependencies
-RUN apt-get update && apt-get install -y build-essential git
+RUN apt-get update && apt-get install -y --no-install-recommends build-essential git \
+  && rm -rf /var/lib/apt/lists/*
+
 # Prep and copy source
 WORKDIR /app/charon
+COPY go.mod go.sum ./
+RUN go mod download
```
KaloyanTanev, line 11: This is not necessary as go build will download the dependencies.

Dockerfile
```diff
@@ -1,14 +1,22 @@
 # Container for building Go binary.
 FROM golang:1.23.1-bookworm AS builder
+
 # Install dependencies
-RUN apt-get update && apt-get install -y build-essential git
+RUN apt-get update && apt-get install -y --no-install-recommends build-essential git \
+  && rm -rf /var/lib/apt/lists/*
```
KaloyanTanev, line 6: Not necessary, as this is multi-stage build and the image from the first stage is not used in the final image.

Dockerfile
```diff
@@ -18,33 +26,41 @@ RUN echo "Built charon version=$(./charon version)"
...
     --no-create-home \
     --uid "$UID" \
-    "$USER"
-RUN chown charon /usr/local/bin/charon
-RUN chmod u+x /usr/local/bin/charon
+    "$USER" \
+  && chown "$USER" /usr/local/bin/charon \
+  && chmod u+x /usr/local/bin/charon \
+  && mkdir -p "/opt/$USER" \
+  && chown "$USER" "/opt/$USER"
```
KaloyanTanev, line 52: Not necessary, as the `adduser` command already created this directory, see `--home "/opt/$USER"`. Its owner is also the said user.

Dockerfile
No0key: I removed the cache deletion, but now the code quality test suggests that the cache must be cleared.

No0key:
@KaloyanTanev can you rerun tests? title name was fixed

pinebit:
I am good with this PR, will delegate final approval and merge to @KaloyanTanev

KaloyanTanev (approved):
Looks good! Sorry for the delay

PR #3282, opened 2024-09-12
*: disable intrange linter

gsora:
Go v1.23 allows to range over int slices without calling `len` first.

Since it's a low-importance change, we can disable intrange linter to avoid being blocked by it during pre-commit.

category: misc
ticket: none

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -151,6 +151,7 @@ linters:
   enable-all: true
   disable:
     # Keep disabled
+    - intrange
     - containedctx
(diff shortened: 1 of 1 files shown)


KaloyanTanev:
I think I have updated all the places where this happens, so linter doesn't complain on any of the old code?

gsora:
Weird, I have it complaining on my (unrelated) PR, complaining about validatorapi :thinking:

pinebit (approved):

PR #3283, opened 2024-09-12
core/parsigdb: improve key uniqueness

gsora:
Instead of only having `(duty, pubkey)` as parsigdb key, also use `core.ParSignedDataSet.Root()` to increase uniqueness factor.

This effectively eliminates the possibility that peers in the network could share duty data for a given slot and validator that doesn't match what was previously decided during consensus.

category: refactor
ticket: none


core/parsigdb/memory.go
```diff
@@ -224,6 +233,24 @@ func getThresholdMatching(typ core.DutyType, sigs []core.ParSignedData, threshol
...
+	// known limitation: sync message and contributions might have different hashes,
+	// ignore them for now.
+	// signature duties don't have a MessageRoot implementation at all, by design.
+	if duty.Type == core.DutySyncMessage || duty.Type == core.DutySyncContribution || duty.Type == core.DutySignature {
+		return [32]byte{}, nil
+	}
+
+	root, err := data.MessageRoot()
+	if err != nil {
+		return [32]byte{}, errors.Wrap(err, "cannot hash tree root ParSignedData")
```
pinebit, line 248: The error seems a duplicate for line 88.
Maybe here we can simply `return data.MessageRoot()`?

core/parsigdb/memory.go
```diff
@@ -224,6 +233,24 @@ func getThresholdMatching(typ core.DutyType, sigs []core.ParSignedData, threshol
 	return nil, false, nil
 }
 
+// rootFromParSigDataSet returns the MessageRoot of data only if the duty is supported,
+// and only if data is not a core.Signature.
+func rootFromParSigDataSet(duty core.Duty, data core.ParSignedData) ([32]byte, error) {
+	// known limitation: sync message and contributions might have different hashes,
+	// ignore them for now.
+	// signature duties don't have a MessageRoot implementation at all, by design.
+	if duty.Type == core.DutySyncMessage || duty.Type == core.DutySyncContribution || duty.Type == core.DutySignature {
```
pinebit, line 242: What if we convert this to "whitelist" and use the logic for ProposeDuty only?

pinebit (approved):

KaloyanTanev (approved):

pinebit (approved):

gsora:
Closing in favor of #3292

PR #3288, opened 2024-09-17
cmd: broadcast all exits

KaloyanTanev:
Second part of adding the `--all` flag for exits.

This PR is for the `charon exit broadcast` command. Additionally, a flag `exitFromDir` is introduced so point to a directory with exit files.

category: feature
ticket: #3243

### DIFF
--- a/cmd/exit_broadcast.go
+++ b/cmd/exit_broadcast.go
@@ -56,7 +56,7 @@ func newBcastFullExitCmd(runFunc func(context.Context, exitConfig) error) *cobra
 		{lockFilePath, false},
 		{validatorKeysDir, false},
 		{exitEpoch, false},
-		{validatorPubkey, true},
+		{validatorPubkey, false},
 		{beaconNodeEndpoints, true},
 		{exitFromFile, false},
 		{exitFromDir, false},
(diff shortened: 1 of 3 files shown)


cmd/exit_broadcast.go
```diff
@@ -75,66 +108,137 @@ func runBcastFullExit(ctx context.Context, config exitConfig) error {
...
+		)
+		if err != nil {
+			return errors.Wrap(err, "cannot calculate hash tree root for exit message for verification")
+		}
+
+		if err := tbls.Verify(pubkey, exitRoot[:], signature); err != nil {
+			return errors.Wrap(err, "exit message signature not verified")
+		}
+
+		if err := eth2Cl.SubmitVoluntaryExit(valCtx, &fullExit); err != nil {
```
KaloyanTanev, line 239: We may exhaust the BN here on some big clusters. I'm thinking of ways to prevent that and I'm open to suggestions.

cmd/exit_broadcast.go
```diff
@@ -52,15 +56,44 @@ func newBcastFullExitCmd(runFunc func(context.Context, exitConfig) error) *cobra
...
 	bindLogFlags(cmd.Flags(), &config.Log)
 
+	wrapPreRunE(cmd, func(cmd *cobra.Command, _ []string) error {
+		valPubkPresent := cmd.Flags().Lookup(validatorPubkey.String()).Changed
+		exitFilePresent := cmd.Flags().Lookup(exitFromFile.String()).Changed
+		exitDirPresent := cmd.Flags().Lookup(exitFromDir.String()).Changed
+
+		if !valPubkPresent && !config.All {
+			//nolint:revive,perfsprint // we use our own version of the errors package; keep consistency with other checks.
+			return errors.New(fmt.Sprintf("%s must be specified when exiting single validator.", validatorPubkey.String()))
```
pinebit, line 76: I think you should be using `errors.New` parameters instead of `Sprintf`, e.g.

```
return errors.New("validator pubkey must be specified when exiting single validator", z.Str("pubkey", validatorPubkey.String()))
```

Check other instances below.

cmd/exit_broadcast.go
```diff
@@ -75,66 +108,137 @@ func runBcastFullExit(ctx context.Context, config exitConfig) error {
...
+		if config.ExitFromFileDir != "" {
+			entries, err := os.ReadDir(config.ExitFromFileDir)
+			if err != nil {
+				return errors.Wrap(err, "could not read exits directory")
+			}
+			for _, entry := range entries {
+				if !strings.HasPrefix(entry.Name(), "exit-") {
+					continue
+				}
+				exit, err := fetchFullExit(ctx, config.ExitFromFileDir+"/"+entry.Name(), config, cl, identityKey, "")
```
pinebit, line 128: `config.ExitFromFileDir+"/"+entry.Name()` is not cross-platform compatible.
Please use `path.Join(config.ExitFromFileDir, entry.Name())` function.

cmd/exit_broadcast.go
```diff
@@ -75,66 +108,137 @@ func runBcastFullExit(ctx context.Context, config exitConfig) error {
...
+			valCtx,
+			*fullExit.Message,
+			eth2Cl,
+			fullExit.Message.Epoch,
+		)
+		if err != nil {
+			return errors.Wrap(err, "cannot calculate hash tree root for exit message for verification")
+		}
+
+		if err := tbls.Verify(pubkey, exitRoot[:], signature); err != nil {
```
pinebit, line 235: Maybe we should verify all signatures before submitting any of these?

cmd/exit_broadcast.go
pinebit: I think a simple delay (maybe 100ms) should be safe enough? Otherwise NOP shall tweak rate limiting on its BN.

cmd/exit_broadcast.go
KaloyanTanev: Hmmm I was using what we had for the already existing exit messages ([like here](https://github.com/ObolNetwork/charon/blob/8364cf20505a14a5715af6f225dc1203e7c98a6c/cmd/exit_sign.go#L67)). Here it's not exactly parameters as we do in other error messages, as this is a constant and somehow part of the message. I'd rather add its value here (in this case `"validator-public-key"`) than putting it as a parameter.

cmd/exit_broadcast.go
KaloyanTanev: Hmm good call, probably it's better to have the broadcast in separate loop, not only because of the verify, but for any of the other functions that might potentially result in an error.

cmd/exit_broadcast.go
KaloyanTanev: @gsora any view on that? I think of maybe adding a flag of type "be-gentle-with-bn" that adds some delay between broadcasts :)

pinebit (approved):

gsora (approved):

PR #3292, opened 2024-09-18
core/validatorapi: check proposal data against consensus one

gsora:
Instead of blindly accepting VC proposal data, check it for integrity against what we have in DutyDB: if they differ, stop the process.

category: refactor
ticket: none


pinebit (approved):

gsora:
Tested in kurtosis, seems to be working well.

Issue #3299, opened 2024-09-24
Refactor QBFT debugger


pinebit:
In consensus abstraction design, we cannot allow specific consensus implementations to leak its designated package. In this case it is QBFT messages debugger that is leaking qbft package. This needs to be refactored.

## Acceptance Criteria

 - [x] QBFT implementation shall not leak outside of the consensus package.
 - [x] Test that the debugger is not broken due to the refactoring.
 - [x] Ensure sufficient unit tests coverage.

PR #3301, opened 2024-09-25
app: generic consensus debugger

pinebit:
Towards larger consensus abstraction work: reworked QBFT messages debugger to be the generic consensus messages debugger (except for wire representation). This is to eliminate QBFT specific leaked from the consensus/qbft packages.

Breaking change: the debug endpoint `/debug/qbft` has been changed to `/debug/consensus` (and the received file name is changed from `qbft_messages.pb.gz` to `consensus_messages.pb.gz`).
We are not aware of any systems or users actively using or relying on that endpoint in production.

category: refactor
ticket: #3299

### DIFF
--- a/core/consensus/debugger.go
+++ b/core/consensus/debugger.go
@@ -1,6 +1,6 @@
 // Copyright © 2022-2024 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
 
-package app
+package consensus
 
 import (
 	"bytes"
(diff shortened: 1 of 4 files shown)


core/consensus/debugger.go
```diff
@@ -16,77 +16,81 @@ import (
...
-		return len(b), err == nil
+	getSize := func(instance *pbv1.SniffedConsensusInstance) int {
+		return proto.Size(instance)
 	}
 
-	size, ok := getSize(instance)
-	if !ok {
-		return // Just drop this if we cannot calculate the size
-	}
+	size := getSize(instance)
```
KaloyanTanev, line 57: Why did we skip the `ok` check?

pinebit: There is no `ok` anymore.

gsora (approved):

KaloyanTanev (approved):

Issue #3305, opened 2024-10-01
Consensus component refactoring


pinebit:
This work is a part of Consensus Abstraction epic. The goal is to get the consensus component code prepared for the new components.

PR #3306, opened 2024-10-01
*: initial protocols refactoring

pinebit:
Ongoing consensus code refactoring, removing non needed mutable config, etc.
The current infosync handler just prints the selected consensus protocol for debugging, there is no impact on the actual protocol logic, it is still using QBFT all the time.

category: refactor
ticket: #3305


testutil/integration/nightly_dkg_test.go
pinebit, line 1: This fix is to resolve linter's error: "nested context".

PR #3307, opened 2024-10-01
*: removing mutable config

pinebit:
Removing the odd mutable config and the connection with the infosync as a part of ongoing consensus logic refactoring.

category: refactor
ticket: #3305


testutil/integration/nightly_dkg_test.go
pinebit, line 1: this is to fix "nested context" linter error

KaloyanTanev (approved):

gsora (approved):

PR #3327, opened 2024-10-08
*: consensus abstraction

pinebit:
Summary of changes:
1. Introduced & implemented `ConsensusController` interface. 
2. Refactored the entire `consensus` package with increased test coverage.
3. Added `cluster run --consensus-protocol=xyz` flag.
4. Wired Priority protocol to change the current core consensus algorithm.
5. Support to run two consensus protocols at a time (one for Priority - QBFT and another is for Core).
6. Added `docs/consensus.md`

category: feature
ticket: #3304


cmd/createcluster.go
```diff
@@ -393,7 +393,7 @@ func validateCreateConfig(ctx context.Context, conf clusterConfig) error {
 		return errors.New("number of operators is below minimum", z.Int("operators", conf.NumNodes), z.Int("min", minNodes))
 	}
 
-	if len(conf.ConsensusProtocol) > 0 && !consensus.IsSupportedProtocolName(conf.ConsensusProtocol) {
+	if len(conf.ConsensusProtocol) > 0 && !protocols.IsSupportedProtocolName(conf.ConsensusProtocol) {
```
gsora, line 396: Do we error out if no consensus protocol has been chosen? Should we? Or it's better to leave a sensible default?

core/consensus/metrics/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+		Name:      "error_total",
+		Help:      "Total count of consensus errors by protocol",
+	}, []string{"protocol"})
+)
+
+type consensusMetrics struct {
+	protocolID string
+}
+
+func NewConsensusMetrics(protocolID string) ConsensusMetrics {
```
gsora, line 53: Weird that golangci-lint didn't complain for missing godoc?

core/consensus/protocols/protocols.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+
+	return QBFTv2ProtocolID
+}
+
+// IsSupportedProtocolName returns true if the protocol name is supported.
+func IsSupportedProtocolName(name string) bool {
+	for _, p := range Protocols() {
+		nameAndVersion := strings.TrimPrefix(string(p), protocolIDPrefix)
```
gsora, line 36: Wondering if this logic could be attached to the `protocol.ID` struct as a method.

gsora (commented):
Mostly minor comments, generally looks great

cmd/createcluster.go
pinebit: The default, or "the last resort" is always QBFT v2.0. We error out here only if user specified something unknown, something we cannot find among the protocols supported by present implementation.

core/consensus/metrics/metrics.go
pinebit: It does not complain indeed. I will go over every exported function and add godoc.

core/consensus/protocols/protocols.go
pinebit: That was my original thought as well, but `protocol.ID` is the inner type of libp2p which we should not extend.. I could have introduce a derived type, but this seemed too much for this simple case.

pinebit:
Added @OisinKyne to review at least:
- docs/consensus.md
- CLI flags
- Metrics changes

go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.19.0
 	github.com/stretchr/testify v1.9.0
+	github.com/test-go/testify v1.1.4
```
KaloyanTanev, line 34: Up until now we managed to do our mocks and overall testing without testify. Shall we bring it now? It will create some discrepancies I think.

app/app.go
```diff
@@ -526,14 +528,20 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
...
-		deadlinerFunc("consensus"), gaterFunc, consensusDebugger.AddInstance)
+	consensusFactory, err := consensus.NewConsensusFactory(tcpNode, sender, peers, p2pKey, deadlinerFunc, gaterFunc, consensusDebugger)
 	if err != nil {
 		return err
 	}
 
+	defaultConsensus := consensusFactory.DefaultConsensus()
+	startDefaultConsensus := lifecycle.HookFuncCtx(defaultConsensus.Start)
+
+	coreConsensus := consensusFactory.CurrentConsensus() // points to DefaultConsensus() initially
```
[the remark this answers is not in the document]
pinebit: `defaultConsensus` is always used by Priority/InfoSync and cannot be changed for this purpose. This is because Priority is reaching consensus on what "core" consensus protocol must be running.
`coreConsensus` is dynamic and can be switched at runtime; the exact protocol is selected by Priority. Before that selection happens (this takes time), "core" will continue using the default consensus.

app/app.go
```diff
@@ -585,8 +594,9 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, tcpNode host.Host,
 	peers []peer.ID, threshold int, sendFunc p2p.SendReceiveFunc, coreCons core.Consensus,
 	sched core.Scheduler, p2pKey *k1.PrivateKey, deadlineFunc func(duty core.Duty) (time.Time, bool),
+	consensusFactory core.ConsensusFactory, clusterPreferredProtocol string,
 ) error {
-	cons, ok := coreCons.(*consensus.Component)
+	cons, ok := coreCons.(*qbft.Consensus)
```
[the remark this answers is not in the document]
pinebit: Yes, because this is for Priority only. By design Priority uses and will continue to use QBFT v2.0.

app/app.go
```diff
@@ -602,9 +612,22 @@ func wirePrioritise(ctx context.Context, conf Config, life *lifecycle.Manager, t
...
+	// 2. Prioritizing the protocol specified by CLI flag (cluster run) to the top.
+	// In all cases this prioritizes all versions of the protocol identified by name.
+	// The order of all these operations are important.
+	allProtocols := Protocols()
+	if clusterPreferredProtocol != "" {
+		allProtocols = protocols.PrioritizeProtocolsByName(clusterPreferredProtocol, allProtocols)
+	}
+	if conf.ConsensusProtocol != "" {
+		allProtocols = protocols.PrioritizeProtocolsByName(conf.ConsensusProtocol, allProtocols)
+	}
```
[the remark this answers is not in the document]
pinebit: The default order is specified by implementation (see `protocols.Protocols()` function) in order of precedence. For example, if we have "old" cluster lock and want to enable HotStuff, we will upgrade the entire cluster (all nodes) to a SW version supporting HotStuff and giving it higher priority (in implementation) than QBFT.

docs/consensus.md
```diff
@@ -0,0 +1,68 @@
...
+
+The Priority protocol runs once per epoch and changes its output depending on the inputs. If another protocol started to appear at the top of the list, Charon would switch the consensus protocol to that one with the next epoch.
+
+### Changing Consensus Protocol Preference
+
+A cluster creator can specify the preferred consensus protocol in the cluster configuration file. This new field `consensus_protocol` appeared in the cluster definition file from v1.9 onwards. The field is optional and if not specified, the cluster definition will not impact the consensus protocol selection.
+
+A node operator can also specify the preferred consensus protocol using the new CLI flag `--consensus-protocol` which has the same effect as the cluster configuration file, but it has a higher precedence. The flag is also optional.
+
+In both cases, a user is supposed to specify the protocol family name, e.g. `abft` string and not a fully-qualified ID. The precise version of the protocol is to be determined by the Priority protocol.
```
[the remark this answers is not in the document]
pinebit: This is what we discussed in the design review. We agreed to reduce cognitive overhead for the users and make it as simple as this. Good that we can always improve it later and allow specifying exact versions.

go.mod
pinebit: Pardon, I did not catch we already have `github.com/stretchr/testify`... fixing now.

app/app.go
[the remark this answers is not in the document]
pinebit: Updated `consensus.md` with more details on the topic.

cmd/createcluster.go
gsora: Noted!

core/consensus/protocols/protocols.go
gsora: Ah right, no need to over-abstract then, let's keep it this way.

app/app.go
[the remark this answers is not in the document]
KaloyanTanev: Got it, thank you!

KaloyanTanev (approved):

gsora (approved):

PR #3334, opened 2024-10-14
core/consensus: logging leader index

pinebit:
Added `core_consensus_decided_leader_index` gauge to reflect the leader index of QBFT decision round. Also, logging essential decision round data as a debug message.

category: feature
ticket: none


OisinKyne:
We also log number of rounds right? we should make sure we are accurate on who finished with the winning suggestion too, particularly in panels.

pinebit:
> We also log number of rounds right? we should make sure we are accurate on who finished with the winning suggestion too, particularly in panels.

Good point, now changed the implementation.
On the panel we can render the new metric together with "rounds" graph maybe. Also, in the logs we will get the comprehensive output.

KaloyanTanev (approved):

core/consensus/component.go
```diff
@@ -445,15 +445,31 @@ func (c *Component) runInstance(ctx context.Context, duty core.Duty) (err error)
...
 	decideCallback := func(qcommit []qbft.Msg[core.Duty, [32]byte]) {
+		round := qcommit[0].Round()
 		decided = true
-		decidedRoundsGauge.WithLabelValues(duty.Type.String(), string(roundTimer.Type())).Set(float64(qcommit[0].Round()))
+		decidedRoundsGauge.WithLabelValues(duty.Type.String(), string(roundTimer.Type())).Set(float64(round))
 		inst.decidedAtCh <- time.Now()
+
+		leaderIndex := leader(duty, round, nodes)
+		leaderName := c.peers[leaderIndex].Name
+		log.Debug(ctx, "QBFT consensus decided",
```
gsora, line 461: Shouldn't we say consensus is decided only if  `decided` is true?

pinebit: This is inside the callback when `decided` is `true`.

gsora (approved):

Issue #3328, opened 2024-10-09
Implement Fallback Beacon Node Selection for Charon


boulder225:
## 🎯 Problem to be solved
Charon currently doesn't have a reliable way to handle beacon node failures, which could lead to interruptions. We need to improve Charon's ability to switch to backup beacon nodes when the primary node fails.

## 🛠️ Proposed solution
- [ ] Implement a "pick list" feature in Charon for fallback beacon nodes
- [ ] Allow users to specify a list of primary and fallback beacon nodes through configuration
- [ ] When a call to a primary beacon node fails, automatically try the next available node from the fallback list
- [ ] If all nodes in the list fail, report the failure to Charon
- [ ] Ensure this feature works with the existing "multi" component that handles multiple beacon node connections
- [ ] Add error handling and logging for fallback node selection
- [ ] Update documentation to explain how to use and configure the fallback beacon node feature
- [ ] Create tests to verify the fallback mechanism works as expected

PR #3342, opened 2024-10-15
app/eth2wrap: fallback beacon nodes

gsora:
Implement a way to provide eth2wrap with two classes of beacon nodes addresses: standard and fallback beacon nodes.

When one of the multi BN calls fails, eth2wrap wrappers will try to get an available fallback BN from a list and re-do the call on that.

If no fallback BNs is specified, return the original error.

If the fallback BN call fails, return fallback error instead.

This PR firstly introduces concepts and code, will introduce CLI parameters and initialization code later.

category: feature
ticket: #3328

### DIFF
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -57,7 +57,21 @@ func Instrument(clients ...Client) (Client, error) {
 		return nil, errors.New("clients empty")
 	}
 
-	return newMulti(clients), nil
+	// TODO(gsora): remove once the implementation is agreed upon and
+	// wiring is complete.
+	fb := NewFallbackClient(0, [4]byte{}, nil)
+
+	return newMulti(clients, fb), nil
+}
+
+// InstrumentWithFallback returns a new multi instrumented client using the provided clients as backends and fallback
+// respectively.
+func InstrumentWithFallback(fallback *FallbackClient, clients ...Client) (Client, error) {
+	if len(clients) == 0 {
+		return nil, errors.New("clients empty")
+	}
+
+	return newMulti(clients, fallback), nil
 }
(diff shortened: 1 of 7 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -57,7 +57,21 @@ func Instrument(clients ...Client) (Client, error) {
 		return nil, errors.New("clients empty")
 	}
 
-	return newMulti(clients), nil
+	// TODO(gsora): remove once the implementation is agreed upon and
```
pinebit, line 60: Is this still valid?

gsora: Yes, the code does not have any wiring to the CLI interface, hence it acts as if it was a non-fallback client.

pinebit (approved):

app/eth2wrap/multi.go
```diff
@@ -83,9 +94,9 @@ func (m multi) ActiveValidators(ctx context.Context) (ActiveValidators, error) {
 	const label = "active_validators"
 	// No latency since this is a cached endpoint.
 
-	res0, err := provide(ctx, m.clients,
-		func(ctx context.Context, cl Client) (ActiveValidators, error) {
-			return cl.ActiveValidators(ctx)
+	res0, err := provide(ctx, m.clients, m.fallback,
+		func(ctx context.Context, args provideArgs) (ActiveValidators, error) {
+			return args.client.ActiveValidators(ctx)
```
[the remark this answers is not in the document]
gsora: ActiveValidators is a cached endpoint, which is handled internally, so whatever client we use is fine.

KaloyanTanev (approved):

Issue #3136, opened 2024-06-12
Improve error messages and logging in Charon when processing validator exits


boulder225:
## 🎯 Problem to be solved
When processing validator exits, Charon's error messages and logging do not provide enough detail to easily identify the specific causes of failures. This makes it difficult for developers to diagnose and resolve issues.

## 🛠️ Proposed solution
- [x] Review the current error handling and logging code paths related to processing validator exits
- [x] Update error messages to include relevant details such as validator IDs, error codes, or other contextual info
- [x] Add logging statements to capture important state or inputs/outputs at different steps of the exit process
- [x] Test changes by intentionally triggering various exit failure scenario
- [x] Update relevant documentation or runbooks

boulder225:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/3136?planning-poker) with Zenhub @gsora @KaloyanTanev

boulder225:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Obol-Workspace-648c1170ee94656741353327/issues/ObolNetwork/charon/3136?planning-poker) with Zenhub @LukeHackett12

KaloyanTanev:
Completed in #3347

PR #3347, opened 2024-10-23
*: improve logging and error handling for exits

KaloyanTanev:
Add more logging and more detailed errors for exits. Given that the exits are fire and forget operation and not a long running process, we can afford more details without the issue of polluting with too much.

category: feature
ticket: #3136

### DIFF
--- a/app/obolapi/api.go
+++ b/app/obolapi/api.go
@@ -83,7 +83,7 @@ func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
 	ctx, cancel := context.WithTimeout(ctx, c.reqTimeout)
 	defer cancel()
 
-	err = httpPost(ctx, addr, b)
+	err = httpPost(ctx, addr, b, nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


app/obolapi/api.go
```diff
@@ -105,27 +105,58 @@ func launchpadURLPath(lock cluster.Lock) string {
...
-		return errors.Wrap(err, "failed to call POST endpoint")
+		return errors.Wrap(err, "call POST endpoint")
 	}
 	defer res.Body.Close()
 
-	data, err := io.ReadAll(res.Body)
+	if res.StatusCode/100 != 2 {
+		data, err := io.ReadAll(res.Body)
+		if err != nil {
+			return errors.Wrap(err, "read POST response")
```
pinebit, line 127: It would be great to catch the status code as well as headers when an error occurs, or maybe log them as debug. Also, verbose logging may be printing the body as base64 or json (depending on content-type). I remember I needed this a few times and patched this code locally to do so.

KaloyanTanev: > It would be great to catch the status code as well as headers when an error occurs, or maybe log them as debug.
Indeed! Added it.

> Also, verbose logging may be printing the body as base64 or json (depending on content-type).
What do you mean here?

pinebit: I realized we don't have verbose logging level, debug is the minimum.
I meant the ability to intercept HTTP requests and responses (bodies) when we need to intercept traffic between stack components.. in addition to HTTP status code, URL and Headers. To see the full picture.

KaloyanTanev: Ah, I see now. Yes having some sort of a "trace" logging would help in a lot of scenarios I believe. We can probably add it on the roadmap to add it throughout the project?

pinebit: Maybe if your Wireshark experience was successful then we don't need to add anything here. Anyway this out of scope for the current work.

cmd/exit_fetch.go
```diff
@@ -128,7 +128,7 @@ func runFetchExit(ctx context.Context, config exitConfig) error {
 
 			fullExit, err := oAPI.GetFullExit(valCtx, validatorPubKeyHex, cl.GetInitialMutationHash(), shareIdx, identityKey)
 			if err != nil {
-				return errors.Wrap(err, "could not load full exit data from Obol API")
+				return errors.Wrap(err, "load full exit data from Obol API", z.Str("validator_public_key", validatorPubKeyHex))
```
gsora, line 131: Isn't the validator hex already in the context at this point?

KaloyanTanev: It is in the `valCtx`, but this error is unaware of it

gsora: Ah right!

gsora (approved):

PR #3348, opened 2024-10-23
*: bump golangci-lint to v1.61.0

KaloyanTanev:
Bump linter

category: misc
ticket: none


pinebit:
Is it possible to do this with dependabot?

pinebit (approved):

KaloyanTanev:
I looked into it couple of months ago and it isn't... It's not as smart :(

PR #3349, opened 2024-10-23
*: update bls-eth-go-binary

gsora:
Recent DKG issues may be due to an issue in bls-eth-go-binary, caused by an upstream bug in their dependencies.

Update to v1.36.1 which solves the issue.

Reference: https://github.com/herumi/bls-eth-go-binary/issues/63

category: bug
ticket: none

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/golang/snappy v0.0.4
 	github.com/google/gofuzz v1.2.0
 	github.com/gorilla/mux v1.8.1
-	github.com/herumi/bls-eth-go-binary v1.35.0
+	github.com/herumi/bls-eth-go-binary v1.36.1
 	github.com/holiman/uint256 v1.3.1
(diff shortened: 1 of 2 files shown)


KaloyanTanev (approved):

pinebit:
Looks like we will need to create a release-v1.1.2 branch, update app version there and then PR into that branch instead of main-v1.1? or is it fine to directly PR to main-v1.1? I am confused what we allow and what not :)

pinebit (approved):

gsora:
If you look at the history of `main-v1.1`, when we released v1.1.1 we simply pushed a new commit on that branch.

Github actions will take care of fixing the version string at compile time.

pinebit:
> If you look at the history of `main-v1.1`, when we released v1.1.1 we simply pushed a new commit on that branch.
> 
> Github actions will take care of fixing the version string at compile time.

Thank you for clarifying!

gsora:
Manually merging to kick the v1.1.2 tag down the line.

PR #3361, opened 2024-11-07
cmd: beacon node simulation

KaloyanTanev:
Beacon node simulation test. This test aims to mimc the amount and types of requests a charon node sends to a beacon node.

Requests are split into general cluster requests (independent of the amount of validators the charon node is servicing) and validator requests (requests are scaling based on the amount of validators).

For lower number of validators (1 and 10) more aggressive approach is taken - you would see more frequent requests related to proposing and sync committee duties. As the validator number grows, the distribution evens out and gets closer to a real scenario.

The time calculated for some duties is a sum of the RTT of multiple requests, because the duty requires that (i.e.: produce block + submit signed block), while others are the result of the RTT of a single request (i.e.: submit sync message).

The tests performed are a best-estimate, it should be taken into an account that a lot of duties cannot be mimicked 1:1 - we cannot propose an actual block that will put an actual load on the beacon node. On submitting new signed block, the beacon node will quickly return 4XX error. In the future this can be optimised, to make as truthful of a request as possible, so that the BN is doing more work, before it recognises an issue.

Running this test **should NOT** be considered an alternative to running a cluster on testnet to evaluate the performance of the Execution Layer and Consensus Layer nodes. Running an actual cluster on a network will always give you better perspective.

category: feature
ticket: none


cmd/testbeacon.go
```diff
@@ -451,3 +577,1286 @@ func beaconPeerCountTest(ctx context.Context, _ *testBeaconConfig, target string
...
+	log.Info(ctx, "Validators simulation finished",
+		z.Any("validators_count", params.TotalValidatorsCount),
+		z.Any("target", target),
+	)
+
+	return testRes
+}
+
+func singleClusterSimulation(ctx context.Context, simulationDuration time.Duration, target string, resultCh chan SimulationCluster, wg *sync.WaitGroup) {
+	defer wg.Done()
```
pinebit, line 812: Instead of passing `wg *sync.WaitGroup` you could just pass `func()` `wg.Done`

KaloyanTanev: Good one, thanks!

pinebit (approved):

PR #3379, opened 2024-11-14
cmd: add version check to beacon node tests

KaloyanTanev:
A sanity check for the user if they are testing against a beacon node they expect. It's not a test at its original sense per se, as we always return Ok, it's more for information of the user.

category: feature
ticket: none


pinebit (approved):

KaloyanTanev:
Weird how protoc-gen-go fails [and requires version to be v1.35.2](https://github.com/ObolNetwork/charon/actions/runs/11834356353/job/32974870144?pr=3379#step:6:408), while the [latest on github is v1.35.1](https://github.com/protocolbuffers/protobuf-go/releases)...

PR #3390, opened 2024-11-25
cmd: test command flags update

KaloyanTanev:
After QA testing and review by Oisin, some changes were made.

- Rename `--cluster-lock-file-path` to `--lock-file` and `--cluster-definition-file-path` to `--definition-file` in test peers command. This is to be aligned with the rest of the CLI.
- Allow passing of cluster definition by URL to `--definition-file` flag.
- Remove [REQUIRED] from the MEV test `--beacon-node-endpoint` flag description, as this flag is required only if `--load-test` is enabled.
- Do single MEV block creation test, to which users can specify number of blocks to smoothen the result. Previously we had 2 test, one with single block and second with multiple (by default 3). Now it is only the second test with a default of 1.
- Rename the `--load-test-blocks` flag in MEV test to `--number-of-payloads`.
- Round the measurements - > 1s to 2 digits after the decimal, < 1s to whole number.
- Delete the fio command created test file.
- Add a missed return statement.
- Bring back the test command to alpha state 😔.

category: feature
ticket: none


cmd/duration_test.go
```diff
@@ -294,3 +294,43 @@ func TestDurationUnmarshalText(t *testing.T) {
...
+		},
+		{
+			name:     "15.151515 milliseconds",
+			in:       cmd.Duration{15151515 * time.Nanosecond},
+			expected: cmd.Duration{15 * time.Millisecond},
+		},
+		{
+			name:     "2.344444 seconds",
+			in:       cmd.Duration{2344444 * time.Microsecond},
+			expected: cmd.Duration{2340 * time.Millisecond},
```
pinebit, line 317: why is this rounded? it should be "2344" no?

pinebit (approved):

cmd/duration_test.go
KaloyanTanev: I'm expecting it to be `2.34` seconds. However, time uses only whole numbers, meaning I should test with the lower magnitude, in this case ms. So, to test 2.34s I need 2340ms. If I wanted let's say 2.3s, I would need 2300ms, etc.

PR #3414, opened 2024-12-11
Fix typo

detrina:
Description:

Fixed grammatical error: "participants have" changed to "participants has" to ensure subject-verb agreement.
Corrected capitalization: "prometheus" changed to "Prometheus" to reflect proper noun formatting.


pinebit:
The subject "participants" is plural, so it requires the plural verb "have." Changing it to "participants has" would break the subject-verb agreement.

PR #3416, opened 2024-12-12
core/consensus: fixed instance_io race

pinebit:
Observed a bug in our consensus object `instance_io` where we have the race condition: `panic: close of closed channel`
Fixed the race condition using atomic operations instead of channels.

category: bug
ticket: none

### DIFF
--- a/core/consensus/utils/instance_io.go
+++ b/core/consensus/utils/instance_io.go
@@ -3,6 +3,7 @@
 package utils
 
 import (
+	"sync/atomic"
 	"time"
 
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


core/consensus/utils/instance_io.go
```diff
@@ -17,23 +18,20 @@ const (
...
 	}
 }
 
 // InstanceIO defines the async input and output channels of a
 // single consensus instance in the Component.
 type InstanceIO[T any] struct {
-	Participated chan struct{}      // Closed when Participate was called for this instance.
-	Proposed     chan struct{}      // Closed when Propose was called for this instance.
-	Running      chan struct{}      // Closed when runInstance was already called.
+	Participated int32              // Closed when Participate was called for this instance.
```
gsora, line 32: We should change those comments, there are no channels involved in this anymore.

gsora: Why not an `atomic.Bool`?

pinebit: Indeed, changed as suggested.

gsora (approved):

PR #3417, opened 2024-12-12
app: eth2wrap latency logging

KaloyanTanev:
Adding logs for some beacon node calls RTT. In order to help tracking issues with underperforming beacon nodes, caused either by slow infrastructure/networking or issues from the beacon node itself.

Currently logs have been added only on the calls for:
- Proposal (GET /eth/v3/validator/blocks/{slot}) - fetches a proposal for signing and is known to be the bottleneck call to the BNs on proposals.
- Validators (POST /eth/v1/beacon/states/{state_id}/validators) - on some BNs this call is taking long time and is blocking the whole BN reachability and any calls to it.

I'm open to including more endpoints.

Fixing linter on one place.

category: feature
ticket: none

### DIFF
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -181,10 +181,10 @@ func latency(ctx context.Context, endpoint string, enableLogs bool) func() {
 		latencyHist.WithLabelValues(endpoint).Observe(rtt.Seconds())
 		if enableLogs {
 			log.Debug(ctx, "Beacon node call finished", z.Str("endpoint", endpoint))
-			// If BN call took more than 1 second, send WARN log
-			if rtt > time.Second {
-				log.Warn(ctx, "Beacon node call took longer than expected", nil, z.Str("endpoint", endpoint), z.Str("rtt", rtt.String()))
-			}
+		}
+		// If BN call took more than 1 second, send WARN log
+		if rtt > time.Second {
+			log.Warn(ctx, "Beacon node call took longer than expected", nil, z.Str("endpoint", endpoint), z.Str("rtt", rtt.String()))
 		}
 	}
 }
--- a/app/eth2wrap/eth2wrap.go
+++ b/app/eth2wrap/eth2wrap.go
@@ -177,13 +177,13 @@ func latency(ctx context.Context, endpoint string, enableLogs bool) func() {
 	t0 := time.Now()
 
 	return func() {
-		rttSeconds := time.Since(t0).Seconds()
-		latencyHist.WithLabelValues(endpoint).Observe(rttSeconds)
+		rtt := time.Since(t0)
+		latencyHist.WithLabelValues(endpoint).Observe(rtt.Seconds())
 		if enableLogs {
 			log.Debug(ctx, "Beacon node call finished", z.Str("endpoint", endpoint))
 			// If BN call took more than 1 second, send WARN log
-			if rttSeconds > 1 {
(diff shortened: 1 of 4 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -165,14 +166,26 @@ func submit(ctx context.Context, clients []Client, work func(context.Context, Cl
...
+
 	return func() {
-		latencyHist.WithLabelValues(endpoint).Observe(time.Since(t0).Seconds())
+		rttSeconds := time.Since(t0).Seconds()
+		latencyHist.WithLabelValues(endpoint).Observe(rttSeconds)
+		if enableLogs {
+			log.Debug(ctx, "Beacon node call finished", z.Str("endpoint", endpoint))
+			// If BN call took more than 1 second, send WARN log
+			if rttSeconds > 1 {
+				log.Warn(ctx, "Beacon node call took longer than expected", nil, z.Str("endpoint", endpoint), z.F64("rtt_seconds", rttSeconds))
```
pinebit, line 186: `time.Duration` has nice String() function: https://pkg.go.dev/time#example-Duration.String
Better to print that instead of seconds?

KaloyanTanev: Yeah, might be better

app/eth2wrap/eth2wrap.go
```diff
@@ -165,14 +166,26 @@ func submit(ctx context.Context, clients []Client, work func(context.Context, Cl
...
+
 	return func() {
-		latencyHist.WithLabelValues(endpoint).Observe(time.Since(t0).Seconds())
+		rtt := time.Since(t0)
+		latencyHist.WithLabelValues(endpoint).Observe(rtt.Seconds())
+		if enableLogs {
+			log.Debug(ctx, "Beacon node call finished", z.Str("endpoint", endpoint))
+			// If BN call took more than 1 second, send WARN log
+			if rtt > time.Second {
+				log.Warn(ctx, "Beacon node call took longer than expected", nil, z.Str("endpoint", endpoint), z.Str("rtt", rtt.String()))
```
pinebit, line 186: I suggest we see this warning even when `enableLogs` is disabled.

pinebit (changes requested):
We shall probably discard this work entirely in favor of https://github.com/ObolNetwork/charon/issues/3429

KaloyanTanev:
@pinebit now that we've discarded #3429 from v1.3, should we still move forward with this one at least? I'd say yes

pinebit (approved):
Agreed @KaloyanTanev let's use this solution for now

app/eth2wrap/eth2wrap.go
KaloyanTanev: Agree

app/eth2wrap/eth2wrap.go
```diff
@@ -165,14 +166,26 @@ func submit(ctx context.Context, clients []Client, work func(context.Context, Cl
 	return err
 }
 
-// latency measures endpoint latency.
+// latency measures endpoint latency and writes metrics and logs results.
```
DiogoSantoss, line 169:
```suggestion
// latency measures endpoint latency, writes metrics and logs results.
```

DiogoSantoss (approved):

app/eth2wrap/eth2wrap.go
KaloyanTanev: The two "and"s are conjunctions of two different logical structures.
The first "and" is conjunction between two simple sentences, the second "and" is part of the second simple sentence is in the role of enumeration inside it.

PR #3427, opened 2024-12-16
fix: typos in documentation files

zeevick10:
Corrected `Deply` to `Deploy`
Corrected `Chck` to `Check`


pinebit:
We have an internal effort fixing the same issues.

Issue #1655, opened 2023-01-18
Add a `nickname` field to `peerinfo` protocol


OisinKyne:
## Problem to be solved

We have a hard time giving useful names to peers in a cluster. It would be great if we could give people an easier way to identify their node places. We could have people label their prometheus writes, but that doesn't get shared with peers, so might help us more than it helps users, which might not be what we are after. Instead, what if we included an optional field in the peerinfo protocol, or another protocol that allows us to share short strings with one another along with peerId/nodeId information. That way, we could (hopefully safely), use it in logs and monitoring on the connected person's side, to improve context and understanding of issues by 'making them human'. 

## Proposed solution

  Add a optional --nickname string flag to `charon run` (and optionally `charon dkg`)
  Share that with peers via peerinfo protocol
  Add app_peerinfo_nickname constant metric with nickname label (including own)

  (And even lazier MVP is adding the flag and just creating the prom metric for it, that would mean we could show it in obol's shared grafana but not the charon peer logs)	

### In Scope

Add a field to the grafana dashboard for these nicknames. Update the dashboard in charon and [cdvn](https://github.com/ObolNetwork/charon-distributed-validator-node), no need for cdvc.

<img width="706" alt="image" src="https://github.com/user-attachments/assets/64b71034-c7fd-4546-acc7-b305ecdbea2b" />

Asserting the maximum string is enforced at CLI and wire, and that it is safe from weird attacks such as terminal takeovers and stuff. It would also be great to prevent setting nickname equal to any peer name in the cluster.

### Out of Scope

Consensus on these nicknames across parties

corverroos:
Why don't we add it to the lock file rather?

Also wiring the result from peerinfo throughout the codebase is going to be hard to implement. It would require a big refactor switching from immutable types to mutable types which is a pain and something I want to avoid.

corverroos:
100 chars is also too long imo, it doesn't display nicely in grafana and people will write sentences. Suggest 32 chars.

corverroos:
another term could also be `nickname`, feels a bit simpler and more approachable to non-english speakers.

corverroos:
I'd also suggest we not try to "replace" existing `cluster_peer` names, as that will just result in a mess in grafana as timeseries chop and change. Suggest we only include the new nicks/monikers in the Peer panel, but keep on using deterministic peer names for everything else.

corverroos:
Note my suggested solution:
 - Add a optional `--nickname` string flag
[...]

PR #3428, opened 2024-12-17
app/peerinfo: add nickname field to peerinfo protocol

DiogoSantoss:
Add nickname flag to `run` command, exchange nicknames using peerinfo protocol and create nickname metric to display on Grafana.

Preview:
Screenshot from 2024-12-17 09-11-26

category: feature
ticket: #1655


gsora:
Do we need to handle the case in which two or more peer have the same nickname?

DiogoSantoss:
We probably should, just not sure what to do in that case. We could change our own by adding some suffix to the nickname. I don't know if we consider malicious nodes in this case.

cmd/run.go
```diff
@@ -93,11 +93,15 @@ func bindRunFlags(cmd *cobra.Command, config *app.Config) {
 	cmd.Flags().StringVar(&config.TestnetConfig.CapellaHardFork, "testnet-capella-hard-fork", "", "Capella hard fork version of the custom test network.")
 	cmd.Flags().StringVar(&config.ProcDirectory, "proc-directory", "", "Directory to look into in order to detect other stack components running on the host.")
 	cmd.Flags().StringVar(&config.ConsensusProtocol, "consensus-protocol", "", "Preferred consensus protocol name for the node. Selected automatically when not specified.")
+	cmd.Flags().StringVar(&config.Nickname,"nickname","","Human readable peer nickname. Maximum 32 characters.")
```
pinebit, line 96: Let's say "human friendly" maybe instead of "readable".
Also, I see no spacing around commas, see other lines above.
I suspect your golang code formatting is not set properly for IDE.

pinebit:
> Do we need to handle the case in which two or more peer have the same nickname?

I think we can't do much about this.. just allow them to do that.

pinebit (approved):

app/peerinfo/peerinfo.go
```diff
@@ -39,17 +39,18 @@ func Protocols() []protocol.ID {
...
+	metricSubmitter func(peerID peer.ID, clockOffset time.Duration, version, gitHash string, startTime time.Time, builderAPIEnabled bool, nickname string)
 )
 
 // New returns a new peer info protocol instance.
 func New(tcpNode host.Host, peers []peer.ID, version version.SemVer, lockHash []byte, gitHash string,
-	sendFunc p2p.SendReceiveFunc, builderEnabled bool,
+	sendFunc p2p.SendReceiveFunc, builderEnabled bool, nickname string,
 ) *PeerInfo {
-	// Set own version and git hash and start time metrics.
+	// Set own version and git hash and nickname and start time and metrics.
```
KaloyanTanev, line 49:
```suggestion
	// Set own version, git hash and nickname and start time and metrics.
```

KaloyanTanev (approved):

Issue #3430, opened 2024-12-19
Improve consensus timings


pinebit:
# 🎯 Problem to be solved

To prevent missing proposals we shall revisit the timings used by consensus layer.
Once the final values confirmed, we need to introduce a new timer type and a new version of QBFT protocol (see Consensus Abstraction design) that is using this new timer.

# 🛠️ Proposed solution
- [ ] Review the statistics on timings reported by Data.
- [ ] Implement new timer type and configure the new QBFT protocol version, say `qbft/v2.1.0`.
- [ ] Possibly change `--consensus-protocol` flag behavior to allow fully qualified protocol strings (so we can override with the precise qbft version).

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

KaloyanTanev:
If we get this PR merged in v1.3.0, I believe we will need data from at least a month from couple of clusters, in order be able to advance with this request. That said, v1.5.0 seems like a reasonable milestone for this one.

PR #3440, opened 2025-01-07
core/consensus: add new consensus round timer

DiogoSantoss:
Add a third consensus round timer which implements the following behavior:
- First consensus round has 1 second to complete
- Subsequent rounds have exponentially more time starting at 200ms (400ms, 800ms, 1.6s, etc...)

category: feature
ticket: #3430
feature_flag: exponential


OisinKyne:
Screenshot 2025-01-07 at 13 18 08
Nit: something that doubles every time is not exponential. For exponential it needs to be raised to the power of x. 

How about we call this `quadratic`?

pinebit (changes requested):
I can't see changes to `GetTimerFunc()` implementation that actually creates the new timer.

pinebit:
Please also update `docs/consensus.md` and describe all the timers and their behavior.

DiogoSantoss:
> Screenshot 2025-01-07 at 13 18 08 Nit: something that doubles every time is not exponential. For exponential it needs to be raised to the power of x.
> How about we call this `quadratic`?

This timer is exponential because the argument appears in an exponent (e.g. y=ab^x) and follows the formula (2^(x-1))*100, where x is the round number (and is only applicable after round 1).
So for x=2, y=200; x=3, y=400; etc...
However I'm free to change it to a better name instead of just exponential

app/featureset/featureset.go
```diff
@@ -45,6 +45,9 @@ const (
 	// The feature gets automatically enabled when the current network is gnosis|chiado,
 	// unless the user disabled this feature explicitly.
 	GnosisBlockHotfix Feature = "gnosis_block_hotfix"
+
+	// Exponential enables Exponential round timer for consensus rounds.
```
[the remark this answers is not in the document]
DiogoSantoss: Added "When active has precedence over EagerDoubleLinear round timer."

docs/consensus.md
```diff
@@ -53,6 +53,22 @@ To list all available consensus protocols (with versions), a user can run the co
 
 When a node starts, it sequentially mutates the list of preferred consensus protocols by processing the cluster configuration file and then the mentioned CLI flag. The final list of preferred protocols is then passed to the Priority protocol for cluster-wide consensus. Until the Priority protocol reaches consensus, the cluster will use the default QBFT v2.0 protocol for any duties.
 
+## Consensus Round Duration
+
+There are three diferent round timer implementations. These timers define the duration of each consensus round and how the timing adjusts for subsequent rounds.
```
pinebit, line 58: typo: diferent

docs/consensus.md
```diff
@@ -53,6 +53,22 @@ To list all available consensus protocols (with versions), a user can run the co
...
+
+### Increasing Round Timer
+
+The `IncreasingRoundTimer` uses a linear increment strategy for round durations. It starts with a base duration and increases by a fixed increment for each subsequend round. The formula for a given round `n` is `Duration = IncRoundStart + (IncRoundIncrease * n)`.
+
+### Eager Double Linear Round Timer
+
+The `EagerDoubleLinearRoundTimer` aligns start times across participants by starting at an absolute time and doubling the round duration when the leader is active. The round duration increases linearly according to `LinearRoundInc`. This aims to fix an issue with the original solution where the leader resets the timer at the start of the round while others reset when they receive the justified pre-prepare which leads to leaders getting out of sync with the rest.
+
+### Exponential Round Timer
```
pinebit, line 68: Could you please add CLI flag example of how to enable the specific timer?
Also, mention which one is default and that all nodes of the cluster must use the same consistent setting, otherwise the consensus would fail.

pinebit: Users would likely question themselves what timer to use, would be good to explain a bit what we recommend (the default one) and what difference makes the Exponential timer. Soon when we release this, NOPs will 100% ask this question and we can simply point them to this documentation.

pinebit (approved):

docs/consensus.md
```diff
@@ -53,6 +53,24 @@ To list all available consensus protocols (with versions), a user can run the co
...
+
+The `ExponentialRoundTimer` increases round durations exponentially. It provides a sufficient timeout for the initial round and grows from a smaller base timeout for subsequent rounds. The idea behind this timer is, after the first timeout, the remaninig nodes already had time to fetch their proposals and therefore won't need as much time to reach consensus as for the first round. The shorter subsequent rounds allows us to more quickly skip faulty leader when compared to both `IncreasingRoundTimer` and `EagerDoubleLinearRoundTimer`, giving more leaders a chance to advance the protocol before its too late. To enable this timer, use the flag `--feature-set-enable "exponential"`. Since this timer has precendence over the `EagerDoubleLinearRoundTimer` there is no need to disable the default timer.
```
KaloyanTanev, line 72: As we are doing this for proposals, it will apply for any duty. Can we make it a bit more generic here:

```suggestion
The `ExponentialRoundTimer` increases round durations exponentially. It provides a sufficient timeout for the initial round and grows from a smaller base timeout for subsequent rounds. The idea behind this timer is, that the consensus timeout includes fetching the signing data. As all nodes do that at the start, irregardless if they are leader or not, after the first timeout, the remaining nodes already had time to fetch their signing data. Therefore they won't need as much time to reach consensus as the leader for the first round did. The shorter subsequent rounds allow us to more quickly skip underperforming leader when compared to both `IncreasingRoundTimer` and `EagerDoubleLinearRoundTimer`, giving more leaders a chance to advance the protocol before its too late. To enable this timer, use the flag `--feature-set-enable "exponential"`. Since this timer has precedence over the `EagerDoubleLinearRoundTimer` there is no need to disable the default timer.
```

docs/consensus.md
```diff
@@ -53,6 +53,24 @@ To list all available consensus protocols (with versions), a user can run the co
...
 
+## Consensus Round Duration
+
+There are three different round timer implementations. These timers define the duration of each consensus round and how the timing adjusts for subsequent rounds. All nodes in a cluster must use the same timer implementation to ensure proper consensus operation.
+
+The default implementation is the `EagerDoubleLinearRoundTimer`, which is recommended for most deployments. Other round timers can be enabled or disabled by using the `--feature-set-enable <timer-name>` and `--feature-set-disable <timer-name>` flags.
+
+### Increasing Round Timer
+
+The `IncreasingRoundTimer` uses a linear increment strategy for round durations. It starts with a base duration and increases by a fixed increment for each subsequend round. The formula for a given round `n` is `Duration = IncRoundStart + (IncRoundIncrease * n)`. To enable this timer, disable the default timer by using the flag `--feature-set-disable "eager_double_linear"`.
```
KaloyanTanev, line 64:
```suggestion
The `IncreasingRoundTimer` uses a linear increment strategy for round durations. It starts with a base duration and increases by a fixed increment for each subsequent round. The formula for a given round `n` is `Duration = IncRoundStart + (IncRoundIncrease * n)`. To enable this timer, disable the default timer by using the flag `--feature-set-disable "eager_double_linear"`.
```

KaloyanTanev (approved):

PR #3454, opened 2025-01-13
Fix/typos in ssz_test.go and exchanger.go

teenager-ETH:
Corrected "doesnt'" to "doesn't"
- doesnt' support fuzzing
+ doesn't support fuzzing

Corrected "exhanges" to "exchanges"
- exchange exhanges partial signatures
+ exchange exchanges partial signatures


pinebit:
https://github.com/ObolNetwork/charon/blob/main/docs/contributing.md#chore-prs

Issue #3419, opened 2024-12-12
Add `targetGasLimit` to the `cluster_definition` to allow the creation of clusters with higher gas targets


OisinKyne:
# 🎯 Problem to be solved

The community wants to raise the target gas limit. Currently that is not easy to do in Charon. 

# 🛠️ Proposed solution

The challenge here, is trying to coordinate all VCs to sign a higher limit. This is something we attempted early in charon development, and disfavoured over signing a single registration at the point of dkg or key splitting. 

To begin, I suggest we stick to the key splitting/dkg moment, but further more realtime approaches may be considered in future. 

The suggestion is to include a `target_gas_limit` field in a new version of the cluster_definition object. With a default of 36m. This could be overridden with the flag `--target-gas-limit int          The target block gas limit for the cluster (Default: 36000000)` 

This const https://github.com/ObolNetwork/charon/blob/main/core/validatorapi/validatorapi.go#L36 is what should be parameterized. 

- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Describe the solution to be implemented

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

KaloyanTanev:
Completed by #3456, part of v1.3.0 release.

PR #3456, opened 2025-01-13
*: add custom target gas limit in cluster lock v1.10

KaloyanTanev:
Up until now the gas limit was hardcoded to 30M. However, the community is pushing towards bumping those numbers and you can't really customise that in Charon.

With those updates a `targetGasLimit` variable is introduced to the cluster definition and a new version of cluster definition is created, v1.10. The previous default version was v1.8, now the default is set to v1.10, including also the changes of consensus protocol introduced in v1.9. However, good to keep in mind there is only 1 consensus protocol enabled in charon, meaning the consensus protocol configuration in v1.9 isn't making much difference.

Previously the gas limit was hardcoded to 30M, now this can be configured by supplying `--target-gas-limit` flag to both CreateDKG and CreateCluster commands. The variable is not compulsory and its default is set to 36M (this is the new value commonly used in the ecosystem). If a new charon version (i.e.: v1.3) uses old definition (and lock) file, its default will still be the previous one of 30M. To bump the number, unfortunately, a new cluster needs to be created with the updated definition version.

category: feature
ticket: #3419


cluster/definition.go
```diff
@@ -160,6 +169,9 @@ type Definition struct {
 	// ConsensusProtocol is the consensus protocol name preferred by the cluster, e.g. "abft".
 	ConsensusProtocol string `config_hash:"12" definition_hash:"12" json:"consensus_protocol,omitempty" ssz:"ByteList[256]"`
 
+	// TargetGasLimit is the target block gas limit for the cluster.
+	TargetGasLimit uint `config_hash:"13" definition_hash:"13" json:"target_gas_limit" ssz:"uint64"`
+
 	// ConfigHash uniquely identifies a cluster definition excluding operator ENRs and signatures.
 	ConfigHash []byte `json:"config_hash,0xhex" ssz:"Bytes32" config_hash:"-" definition_hash:"13"`
```
pinebit, line 176: `definition_hash:"13"` is not updated, it must be 14 now.

pinebit (changes requested):

DiogoSantoss (approved):

pinebit (approved):

PR #3459, opened 2025-01-14
app/eth2wrap: add CLI support for fallback beacon node

DiogoSantoss:
Add `--fallback-beacon-node-endpoints` CLI flag to specify a list of beacon nodes to use if the primary list (specified using `--beacon-node-endpoints`) fails.

This is the continuation of a previous PR (#3342) which added the fallback logic.

category: feature
ticket: #3328


pinebit (approved):

app/eth2wrap/eth2wrap_test.go
```diff
@@ -146,6 +146,77 @@ func TestMulti(t *testing.T) {
...
+
+			// Create fallback client
+			fallback, err := beaconmock.New()
+			require.NoError(t, err)
+			fallback.SlotsPerEpochFunc = func(context.Context) (uint64, error) {
+				fallbackCalled = true
+				return 42, nil
+			}
+
+			eth2Cl, _ := eth2wrap.Instrument(primaryClients, []eth2wrap.Client{fallback})
```
KaloyanTanev, line 201: nitpick: can we make sure in the tests that 2 fallback nodes are called (meaning they work in a similar fan-out approach as the main nodes)?

KaloyanTanev (approved):

app/eth2wrap/eth2wrap_test.go
DiogoSantoss: updated

Issue #3278, opened 2024-09-12
Allow deposits greater than 32ETH


pinebit:
# 🎯 Problem to be solved

Currently the maximum allowed deposit is 32 ETH both in create cluster and in dkg. We want to allow deposits larger than 32 ETH per validator. The maximum is to be set to 2048 ETH. 

# 🛠️ Proposed solution
- [x] Allow deposits up to 2048 ETH, while keeping "default" value to be 32 ETH.
- [x] Partial deposits feature to follow the same rules.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #3464, opened 2025-01-16
*: allow deposits greater than 32eth

pinebit:
To support [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) we extend "partial deposits" feature to allow deposits up to 2048ETH (with 0x02 prefix) when `pectra` feature flag is enabled. If this feature flag is disabled, the max amount remains 32ETH (with 0x01 prefix).

category: feature
ticket: #3278
feature_flag: pectra


pinebit:
New PR: https://github.com/ObolNetwork/charon/pull/3466

PR #3466, opened 2025-01-16
*: allow deposits greater than 32eth

pinebit:
To support [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) we extend "partial deposits" feature to allow deposits up to 2048ETH (with 0x02 prefix).

category: feature
ticket: #3278


eth2util/deposit/deposit_test.go
```diff
@@ -103,15 +110,15 @@ func TestVerifyDepositAmounts(t *testing.T) {
 		require.ErrorContains(t, err, "each partial deposit amount must be greater than 1ETH")
 	})
 
-	t.Run("total sum is 32ETH", func(t *testing.T) {
+	t.Run("total sum is at least 32ETH", func(t *testing.T) {
 		amounts := []eth2p0.Gwei{
 			eth2p0.Gwei(1000000000),
```
KaloyanTanev, line 115:
```suggestion
			deposit.MinDepositAmount,
```

KaloyanTanev (approved):

eth2util/deposit/deposit.go
```diff
@@ -188,8 +191,8 @@ func withdrawalCredsFromAddr(addr string) ([32]byte, error) {
 	}
 
 	var creds [32]byte
-	copy(creds[0:], eth1AddressWithdrawalPrefix) // Add 1 byte prefix.
-	copy(creds[12:], addrBytes)                  // Add 20 bytes of ethereum address suffix.
+	copy(creds[0:], eip7251AddressWithdrawalPrefix) // Add 1 byte prefix.
```
KaloyanTanev, line 194: Will validator manually migrate to 0x02 credentials? If so, migration from 0x01 to 0x02 might take some time, meaning credentials for most of validators in the beginning will stay at `0x01...`. Have you double checked if this change won't break some stuff?

Also probably good to change the function description on top from `0x01` to `0x02`.

pinebit: AFAIK this code only applies to new cluster creations. It should be fine running a mix of 0x01 and 0x02 validators. For the old ones the users will not be able to use new features like partial withdrawals until the migration is complete.

pinebit: > Also probably good to change the function description on top from 0x01 to 0x02.

updated.

dkg/dkg_test.go
```diff
@@ -386,7 +386,11 @@ func verifyDistValidators(t *testing.T, lock cluster.Lock, def cluster.Definitio
 		// Assert Deposit Data
 		depositAmounts := deposit.DedupAmounts(def.DepositAmounts)
 		if len(depositAmounts) == 0 {
-			depositAmounts = []eth2p0.Gwei{deposit.MaxDepositAmount}
+			if !cluster.SupportPartialDeposits(def.Version) {
+				depositAmounts = []eth2p0.Gwei{deposit.DefaultDepositAmount}
+			} else {
+				depositAmounts = deposit.DefaultDepositAmounts()
+			}
```
DiogoSantoss, line 393: For consistency with `dkg.go`
```suggestion
			if cluster.SupportPartialDeposits(def.Version) {
				depositAmounts = deposit.DefaultDepositAmounts()
			} else {
				depositAmounts = []eth2p0.Gwei{deposit.DefaultDepositAmount}
			}
```

DiogoSantoss (approved):

PR #3469, opened 2025-01-16
feat: Implement weighted scoring system for test results

VolodymyrBg:
- Add weights for different test verdicts
- Introduce percentage-based score calculation
- Implement more granular A/B/C grade assignment
- Remove TODO comment as the feature is now implemented


KaloyanTanev:
Thank you for your contribution, however, the plan was that different tests have different weight. Meaning, the test weight is per test basis and not based on the blind result.
i.e.: a failing direct connection test has lower impact in the final score, compared to terribly performing ping latency test.

PR #3470, opened 2025-01-17
core/consensus: change exponential timer for linear timer

DiogoSantoss:
Change exponential timer (implemented in #3440) for linear timer.

- First consensus round has 1 second to complete
- Subsequent rounds have linearly more time starting at 400ms (600ms, 800ms, 1s, etc...)

category: refactor
ticket: #3430 
feature_flag: linear


KaloyanTanev (approved):
Approving, but we might want to come up with better name than `linear`. Something that describes the feature better.

pinebit (approved):

PR #3477, opened 2025-01-20
feat(core/fetcher): make graffiti configurable per validator

crStiv:
Replace hardcoded graffiti with configurable function in Fetcher struct.
This allows operators to set custom graffiti for each validator while
maintaining the default version+commit behavior.

Changes:
- Add graffitiFunc field to Fetcher struct
- Add SetGraffitiFunc method
- Move default graffiti logic to defaultGraffiti function

### DIFF
--- a/core/fetcher/fetcher.go
+++ b/core/fetcher/fetcher.go
@@ -23,10 +23,23 @@ import (
 
 // New returns a new fetcher instance.
 func New(eth2Cl eth2wrap.Client, feeRecipientFunc func(core.PubKey) string, builderEnabled bool) (*Fetcher, error) {
+	if eth2Cl == nil {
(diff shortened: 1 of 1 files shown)


KaloyanTanev:
Hi @crStiv, thanks for your contribution. However, the graffiti field has to be agreed upon from all validators. Easiest way to do so is from the cluster lock on cluster creation. This would require a new cluster version as well.

KaloyanTanev:
Thank you for your contribution @crStiv. We'd like to focus on stability before Pectra fork and we try to keep new features to minimum. Even though we have seen this feature being requested before, it's not something urgent we'd like to prioritise for now. We will keep you updated once we have it on our agenda.

PR #3482, opened 2025-01-22
*: electra versioned aggregations

KaloyanTanev:
Change the regular aggregations to versioned aggregations. Given that the attestation structure is changing, the one for aggregations does as well.

Some other small things included in this PR:
- minor fixes from the previous attestations PR
- bumping go-eth2-client to its newest electra version
- fix teku integration tests, using its mekong branch
- add electra version to more switch cases (not exhaustive, will open separate PR for that)

category: feature
ticket: none


core/validatorapi/router.go
```diff
@@ -453,13 +460,42 @@ func attestationData(p eth2client.AttestationDataProvider) handlerFunc {
...
-		err := unmarshal(typ, body, &atts)
-		if err != nil {
-			return nil, nil, errors.Wrap(err, "unmarshal attestations")
+		atts := []*eth2spec.VersionedAttestation{}
+
+		denebAtts := new([]eth2p0.Attestation)
+		err := unmarshal(typ, body, denebAtts)
+		if err == nil {
+			for _, att := range *denebAtts {
+				// TODO: Data version is not Deneb, it might be anything between Phase0 and Deneb
```
pinebit, line 469: Flagging this TODO, maybe you forgot to resolve it.

pinebit (approved):
Great job

DiogoSantoss (approved):

core/validatorapi/router.go
KaloyanTanev: Unfortunately, I haven't found a nice way yet... It's in my list for things to fix.

With other versioned structures (like blocks), we have a clunky but functional way of resolving the version - we simply try to parse all versions and whichever succeeds, we deem it is that one. However, for attestations, anything between phase0 and Deneb has the exact same structure, so this method won't really work.

PR #3484, opened 2025-01-27
ci: build-release-binaries workflow

bussyjd:
This workflow handles the automated build and release process for Charon binaries. Key features include:

1. Cross-Platform Build Support:
   - Builds for both AMD64 (x86_64) and ARM64 (aarch64) architectures
   - Uses Docker-based build environment for consistent compilation
   - Enables CGO with appropriate cross-compilation toolchains
   - Includes netgo and osusergo tags for better portability

2. Version Management:
   - Extracts version information from code or git tags
   - Injects version information during compilation via ldflags
   - Maintains consistent version naming across artifacts

3. Release Artifacts:
   - Generates compressed archives (.tar.gz) for each architecture
   - Creates a detailed checksums.txt with architecture-specific labels
   - Preserves both compressed and uncompressed binaries as build artifacts

4. Release Process:
   - Integrates with the main release workflow
   - Attaches binaries to draft releases for review
   - Uses Ubuntu 24.04 for stable and modern build environment

The workflow ensures reliable, reproducible builds while maintaining a clear and organized release process.

category: fixbuild
ticket: none


.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: Build Cross-Platform Binaries
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.23.4'
```
pinebit, line 25: Can we reference https://github.com/ObolNetwork/charon/blob/main/.github/actions/setup-go/action.yml to we remain consistent on golang version used across all Actions?

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,104 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y build-essential gcc-aarch64-linux-gnu
+
+    - name: Build x86_64 binary
+      run: |
+        mkdir -p dist
+        docker run --rm --platform linux/amd64 \
+          -v $(pwd):/workspace -w /workspace \
+          golang:1.23.4-bookworm \
```
pinebit, line 56: Hope we can make this `golang:1.23.4-bookworm` matching our globally set golang version..
We typically bump it in one place.

build-all.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+set -e
+
+# Create dist directory if it doesn't exist
+mkdir -p dist
+
+# Build for linux/amd64
+echo "Building for linux/amd64..."
```
pinebit, line 8: Do we know for sure we don't need a build for Mac ARM?

pinebit:
FYI. pre-commit fails because CI already bumped `buf` version:
```
-// 	protoc-gen-go v1.36.3
+// 	protoc-gen-go v1.36.4
```

We will need to update our code and PR separately to get you unblocked. Then you will need to rebase.
cc @KaloyanTanev

pinebit (approved):

PR #3493, opened 2025-02-03
ci: arm64 releases

bussyjd:
This workflow handles the automated build and release process for Charon binaries. Key features include:

1. Cross-Platform Build Support:
   - Builds for both AMD64 (x86_64) and ARM64 (aarch64) architectures
   - Uses Docker-based build environment for consistent compilation
   - Enables CGO with appropriate cross-compilation toolchains
   - Includes netgo and osusergo tags for better portability

2. Version Management:
   - Extracts version information from code or git tags
   - Injects version information during compilation via ldflags
   - Maintains consistent version naming across artifacts

3. Release Artifacts:
   - Generates compressed archives (.tar.gz) for each architecture
   - Creates a detailed checksums.txt with architecture-specific labels
   - Preserves both compressed and uncompressed binaries as build artifacts

4. Release Process:
   - Integrates with the main release workflow
   - Attaches binaries to draft releases for review
   - Uses Ubuntu 24.04 for stable and modern build environment

The workflow ensures reliable, reproducible builds while maintaining a clear and organized release process.

category: fixbuild
ticket: none

### DIFF
--- a/.github/workflows/build-release-binaries.yml
+++ b/.github/workflows/build-release-binaries.yml
@@ -65,8 +65,6 @@ jobs:
       with:
         name: charon-${{ env.RELEASE_VERSION }}-binaries
         path: |
-          dist/charon-${{ env.RELEASE_VERSION }}-linux-amd64
-          dist/charon-${{ env.RELEASE_VERSION }}-linux-arm64
           dist/charon-${{ env.RELEASE_VERSION }}-linux-amd64.tar.gz
           dist/charon-${{ env.RELEASE_VERSION }}-linux-arm64.tar.gz
           dist/checksums.txt


pinebit (approved):

.github/actions/setup-go/action.yml
```diff
@@ -1,9 +1,20 @@
...
+  go-version:
+    description: "The Go version that was set up"
+    value: ${{ steps.setup-go.outputs.go-version }}
+
 runs:
   using: 'composite'
   steps:
     - name: Setup go
+      id: setup-go
       uses: actions/setup-go@v4
```
KaloyanTanev, line 13: As we are on it, we can probably bump this version as well

```suggestion
      uses: actions/setup-go@v5
```

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,102 @@
...
+        cd dist
+        tar czf charon-${{ env.RELEASE_VERSION }}-linux-amd64.tar.gz charon-${{ env.RELEASE_VERSION }}-linux-amd64
+        tar czf charon-${{ env.RELEASE_VERSION }}-linux-arm64.tar.gz charon-${{ env.RELEASE_VERSION }}-linux-arm64
+        echo "# Linux AMD64 (x86_64)" > checksums.txt
+        sha256sum charon-${{ env.RELEASE_VERSION }}-linux-amd64.tar.gz >> checksums.txt
+        echo -e "\n# Linux ARM64 (aarch64)" >> checksums.txt
+        sha256sum charon-${{ env.RELEASE_VERSION }}-linux-arm64.tar.gz >> checksums.txt
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v3
```
KaloyanTanev, line 81:
```suggestion
      uses: actions/upload-artifact@v4
```

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,102 @@
...
+        path: |
+          dist/charon-${{ env.RELEASE_VERSION }}-linux-amd64
+          dist/charon-${{ env.RELEASE_VERSION }}-linux-arm64
+          dist/charon-${{ env.RELEASE_VERSION }}-linux-amd64.tar.gz
+          dist/charon-${{ env.RELEASE_VERSION }}-linux-arm64.tar.gz
+          dist/checksums.txt
+
+    - name: Attach to GitHub Release
+      if: github.event.workflow_run.conclusion == 'success'
+      uses: softprops/action-gh-release@v1
```
KaloyanTanev, line 93:
```suggestion
      uses: softprops/action-gh-release@v2
```

KaloyanTanev (approved):
Have you considered using `goreleaser`, I find it easier to maintain and straight forward? We've discussed it before, but we had some incompatibilities with it, which now are fixed afaik.

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,102 @@
...
+    - name: Build x86_64 binary
+      run: |
+        mkdir -p dist
+        docker run --rm --platform linux/amd64 \
+          -v $(pwd):/workspace -w /workspace \
+          golang:${{ steps.setup-go.outputs.go-version }}-bookworm \
+          bash -c "apt-get update && apt-get install -y build-essential && \
+          env CGO_ENABLED=1 GOOS=linux GOARCH=amd64 \
+          go build -tags='netgo osusergo' ${{ env.GO_BUILD_FLAG }} \
+          -o dist/charon-${{ env.RELEASE_VERSION }}-linux-amd64"
```
KaloyanTanev, line 58: Why are we using Docker here instead of simply go?

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,102 @@
...
+
+    - name: Build ARM64 binary
+      run: |
+        docker run --rm --platform linux/amd64 \
```
KaloyanTanev, line 62: If we want to use docker... why is it amd64 and not arm64?

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,101 @@
...
+    - uses: ./.github/actions/setup-go
+      id: setup-go
+
+    - name: Get app/version.Version from the code
+      if: github.ref_type == 'branch'
+      run: |
+        VERSION=$(grep 'var version' app/version/version.go | cut -d'"' -f2)
+        echo "APP_VERSION=$VERSION" >> $GITHUB_ENV
+        echo "RELEASE_VERSION=$VERSION" >> $GITHUB_ENV
+
```
KaloyanTanev, line 31: We don't need that as we will trigger this workflow only on tags.

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,101 @@
...
+      if: github.ref_type == 'tag'
+      run: |
+        echo "APP_VERSION=${{ github.ref_name }}" >> $GITHUB_ENV
+        echo "RELEASE_VERSION=${{ github.ref_name }}" >> $GITHUB_ENV
+        echo "LDFLAGS_VALUE=-X github.com/obolnetwork/charon/app/version.version=${{ github.ref_name }}" >> $GITHUB_ENV
+
+    - name: Set build flags for branch
+      if: github.ref_type == 'branch'
+      run: |
+        echo "LDFLAGS_VALUE=-X github.com/obolnetwork/charon/app/version.version=${{ env.APP_VERSION }}" >> $GITHUB_ENV
```
KaloyanTanev, line 42: We don't need that as we will trigger this workflow only on tags.

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,101 @@
...
+
+    - name: Build x86_64 binary
+      run: |
+        mkdir -p dist
```
KaloyanTanev, line 46: nitpick: we can have that in a separate action.

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,101 @@
...
+
+    - name: Create release archives
+      run: |
+        cd dist
+        tar czf charon-${{ env.RELEASE_VERSION }}-linux-amd64.tar.gz charon-${{ env.RELEASE_VERSION }}-linux-amd64
+        tar czf charon-${{ env.RELEASE_VERSION }}-linux-arm64.tar.gz charon-${{ env.RELEASE_VERSION }}-linux-arm64
+        echo "# Linux AMD64 (x86_64)" > checksums.txt
+        sha256sum charon-${{ env.RELEASE_VERSION }}-linux-amd64.tar.gz >> checksums.txt
+        echo -e "\n# Linux ARM64 (aarch64)" >> checksums.txt
+        sha256sum charon-${{ env.RELEASE_VERSION }}-linux-arm64.tar.gz >> checksums.txt
```
KaloyanTanev, line 74: Have we checked if that checksums.txt is properly formatted?

.github/workflows/build-release-binaries.yml
```diff
@@ -0,0 +1,101 @@
...
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: charon-${{ env.RELEASE_VERSION }}-binaries
+        path: |
+          dist/charon-${{ env.RELEASE_VERSION }}-linux-amd64
+          dist/charon-${{ env.RELEASE_VERSION }}-linux-arm64
```
KaloyanTanev, line 82: Do we want those uploaded or only the tarballs? I'd say only the tarballs. If we include those as downloads, we should add their checksums as well (but I don't see a reason why we would want them).

PR #3500, opened 2025-02-05
fix: refresh validator by slot instead of by head

DiogoSantoss:
Add method `GetBySlot` to fetch active/complete validator by slot.
In cases where the beacon node is slow, fetching validators by `head` would lead to outdated information (specifically regarding its active status). By fetching by slot we guarantee to get the up-to-date information.
The downside here is that this request is slower because it might take multiple requests until the beacon node has the information regarding the slot. 

category: bug
ticket: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -467,10 +467,8 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 
 		ctx = log.WithCtx(ctx, z.Bool("first_refresh", firstValCacheRefresh))
 
-		log.Debug(ctx, "Refreshing validator cache")
-
 		valCache.Trim()
-		_, _, err := valCache.Get(ctx)
+		_, _, err := valCache.GetBySlot(ctx, slot.Slot)
 		if err != nil {
 			log.Error(ctx, "Cannot refresh validator cache", err)
 			return err
--- a/app/app.go
+++ b/app/app.go
@@ -467,6 +467,8 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 
 		ctx = log.WithCtx(ctx, z.Bool("first_refresh", firstValCacheRefresh))
 
+		log.Info(ctx, "Refreshing validator cache")
+
 		valCache.Trim()
 		_, _, err := valCache.GetBySlot(ctx, slot.Slot)
 		if err != nil {
--- a/app/eth2wrap/valcache.go
+++ b/app/eth2wrap/valcache.go
@@ -148,7 +148,7 @@ func (c *ValidatorCache) GetBySlot(ctx context.Context, slot uint64) (ActiveVali
 	defer c.mu.Unlock()
 
 	opts := &eth2api.ValidatorsOpts{
-		State:   strconv.Itoa(int(slot)),
+		State:   strconv.FormatUint(slot, 10),
 		PubKeys: c.pubkeys,
 	}
 
--- a/app/eth2wrap/valcache.go
+++ b/app/eth2wrap/valcache.go
@@ -4,13 +4,17 @@ package eth2wrap
 
 import (
 	"context"
+	"strconv"
 	"sync"
+	"time"
 
 	eth2api "github.com/attestantio/go-eth2-client/api"
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
 )
 
 // ActiveValidators is a map of active validator indices to pubkeys.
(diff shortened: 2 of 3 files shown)


app/eth2wrap/valcache.go
```diff
@@ -132,3 +136,57 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
 
 	return resp, eth2Resp.Data, nil
 }
+
+// GetBySlot fetches active and complete validator by slot populating the cache.
+func (c *ValidatorCache) GetBySlot(ctx context.Context, slot uint64) (ActiveValidators, CompleteValidators, error) {
+
```
pinebit, line 142: We don't put an empty line at the beginning of a function. See other functions.

app/eth2wrap/valcache.go
```diff
@@ -132,3 +136,57 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+	defer c.mu.Unlock()
+
+	opts := &eth2api.ValidatorsOpts{
+		State:   strconv.Itoa(int(slot)),
+		PubKeys: c.pubkeys,
+	}
+
+	var eth2Resp *eth2api.Response[map[eth2p0.ValidatorIndex]*eth2v1.Validator]
+	var err error
+	maxRetries := 20
```
pinebit, line 153: I suggest we always put any constants like this one and the one below to file-level constants.

app/eth2wrap/valcache.go
```diff
@@ -132,3 +136,57 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+	retryDelay := 100 * time.Millisecond
+
+	for retryCount := 0; retryCount < maxRetries; retryCount++ {
+		eth2Resp, err = c.eth2Cl.Validators(ctx, opts)
+		if err == nil {
```
pinebit, line 158: Wondering if there are specific errors like `errors.Is(err, context.Canceled)` that shall stop typing immediately?

pinebit (commented):
Added my comments

app/eth2wrap/valcache.go
```diff
@@ -132,3 +136,57 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+	}
+
+	if err != nil {
+		log.Error(ctx, "Failed to fetch validators by slot after maximum retries", err, z.U64("slot", slot))
+		return nil, nil, wrapError(ctx, err, "Failed to fetch validators by slot after maximum retries")
+	}
+
+	vals := eth2Resp.Data
+
+	resp := make(ActiveValidators)
```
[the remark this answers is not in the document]
DiogoSantoss: This was to stay consistent with the `Get` method above, however I agree with the improved readability. Should I also change the variable names in `Get` method ?

app/eth2wrap/valcache.go
```diff
@@ -4,13 +4,22 @@ package eth2wrap
...
 	eth2api "github.com/attestantio/go-eth2-client/api"
 	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/z"
+)
+
+var (
```
pinebit, line 20: `var` => `const`

app/eth2wrap/valcache.go
```diff
@@ -132,3 +141,54 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+
+		if !val.Status.IsActive() {
+			continue
+		}
+
+		active[val.Index] = val.Validator.PublicKey
+	}
+
+	c.active = active
+	c.complete = eth2Resp.Data
```
pinebit, line 191:
```suggestion
	c.complete = complete
```

app/eth2wrap/valcache.go
pinebit, line 20:
```suggestion
const (
```

app/eth2wrap/valcache.go
```diff
@@ -132,3 +141,54 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+
+	return active, eth2Resp.Data, nil
```
pinebit, line 193:
```suggestion
	return active, complete, nil
```

app/eth2wrap/valcache_test.go
```diff
@@ -88,3 +88,70 @@ func TestValidatorCache(t *testing.T) {
...
+			}, nil
+		default:
+			return beaconmock.ValidatorSet{}, nil
+		}
+	}
+
+	valCache := eth2wrap.NewValidatorCache(eth2Cl, pubkeys)
+	ctx := context.Background()
+
+	active, complete, err := valCache.GetBySlot(ctx, 1)
```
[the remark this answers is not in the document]
DiogoSantoss: This triggers the backoff mechanism which causes the test to take 21s to finish. Is it a problem ?

app/app.go
```diff
@@ -467,10 +467,8 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
 
 		ctx = log.WithCtx(ctx, z.Bool("first_refresh", firstValCacheRefresh))
 
-		log.Debug(ctx, "Refreshing validator cache")
```
KaloyanTanev, line 470: Can we keep this debug log? I've found it useful before

app/eth2wrap/valcache.go
```diff
@@ -132,3 +141,54 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+
+	opts := &eth2api.ValidatorsOpts{
+		State:   strconv.FormatUint(slot, 10),
+		PubKeys: c.pubkeys,
+	}
+
+	var eth2Resp *eth2api.Response[map[eth2p0.ValidatorIndex]*eth2v1.Validator]
+	var err error
+
+	for retryCount := 0; retryCount < maxRetries; retryCount++ {
```
KaloyanTanev, line 158: Do you think we can utilise the `retry` package we already have for such kind of scenarios? We can probably add it as part of the `ValidatorCache` structure. You can have a look at it here `app/retry/retry.go`. It's pretty well tested already and works well.

app/eth2wrap/valcache.go
```diff
@@ -132,3 +141,54 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
...
+		}
+
+		sleepDuration := retryDelay * time.Duration(retryCount+1)
+		time.Sleep(sleepDuration)
+
+		log.Info(ctx, "Retrying fetching validators by slot", z.U64("slot", slot), z.Int("retryCount", retryCount+1), z.Err(err))
+	}
+
+	if err != nil {
+		log.Error(ctx, "Failed to fetch validators by slot after maximum retries", err, z.U64("slot", slot))
```
KaloyanTanev, line 171: No need to log here when we are returning an error. This message will be propagated to the logs either way.

app/eth2wrap/valcache.go
DiogoSantoss: I think we can use it but requires some kind of deadline function for the slot right ? So that it stops retrying when the slot is over

app/eth2wrap/valcache.go
DiogoSantoss: Removed

pinebit (approved):

PR #3511, opened 2025-02-06
app/eth2wrap: fix faulty fallback test

DiogoSantoss:
Faulty fallback unit test.
Test expects every client to be called when in reality if one client responds quickly the other might never be evoked.

category: bug 
ticket: none


KaloyanTanev (approved):
You can remove the protobuf changes from the PR. Usually we want to keep those in separate PRs, so in the case we cherry pick what to include in a release it's easier to do so.

pinebit (approved):

PR #3525, opened 2025-02-13
*: bump go-eth2-client and update atts and aggs

KaloyanTanev:
Previously we have had `Attestation` object sent to the `/eth/v1/beacon/pool/attestations`, however, now in electra, those are `SingleAttestation`s. [Link to the BN API](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/submitPoolAttestationsV2). This required some changes...

Change of the key used for recognising attestations in our memory. Previously part of the key was the **validator's index inside the committee**, however, now as it is `SingleAttestation` there are no aggregation bits from which we can derive that. We use now the new field inside the `VersionedAttestation` - `ValidatorIndex`. This allows us to uniquely identify the validator alongside with the other keys. Note that the current changes **are not** compatible with deneb. There are multiple places where we are not compatible with deneb, one prominent one is not supporting v1 endpoints. This is something to be addressed in separate PR.

category: feature
ticket: none

### DIFF
--- a/app/eth2wrap/eth2wrap_test.go
+++ b/app/eth2wrap/eth2wrap_test.go
@@ -237,6 +237,7 @@ func TestFallback(t *testing.T) {
 						return true
 					}
 				}
+
 				return false
 			}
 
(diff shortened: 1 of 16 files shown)


pinebit (approved):
trust your test results

DiogoSantoss (approved):

PR #3536, opened 2025-02-17
*: support v1 endpoints on electra

KaloyanTanev:
Bring back support to v1 endpoints regarding attestations and aggregations. With this PR the electra branch works on both deneb and electra.

category: feature
ticket: none

### DIFF
--- a/core/dutydb/memory.go
+++ b/core/dutydb/memory.go
@@ -16,6 +16,11 @@ import (
 	"github.com/obolnetwork/charon/core"
 )
 
+const (
+	attV1 = iota + 1
+	attV2 = iota + 1
+)
+
 // NewMemDB returns a new in-memory dutyDB instance.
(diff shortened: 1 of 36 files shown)


core/dutydb/memory.go
```diff
@@ -16,14 +16,22 @@ import (
 	"github.com/obolnetwork/charon/core"
 )
 
+const (
+	attV1 = iota + 1
+	attV2 = iota + 1
```
pinebit, line 21: Hmm, this looks weird. I used to see just single `iota`, or we need to increment the second one?

pinebit (approved):

DiogoSantoss (approved):

core/dutydb/memory.go
KaloyanTanev: Yes, this is better probably, changed it.

PR #3538, opened 2025-02-19
feat: add fallback beacon node metric

DiogoSantoss:
Add `app_eth2_using_fallback` gauge metric to track how often nodes rely on fallback  beacon nodes.  This can be queried, for example, using promql `min_over_time(app_eth2_using_fallback[30s]) == 1` to see if any node has been only using the fallback beacon node for the last 30 seconds.

category: feature
ticket: none


pinebit (approved):
Every time you add/patch a metric, you need to regenerate `metrics.md` reference. See `TestMetricReference` test.

KaloyanTanev (approved):

Issue #3518, opened 2025-02-10
`charon create dkg` can return a launchpad invitation to a DKG


OisinKyne:
# 🎯 Problem to be solved

`charon create dkg` creates a local cluster-definition with operator-enrs only. There is no easy CLI way to create a cluster invitation for someone else. 

# 🛠️ Proposed solution

Add a `--publish` (and `--publish-address`) flag to `charon create dkg`, which if present, outputs a text prompt saying something like:

```
Cluster Invitation Prepared:

Direct the Node Operators to: https://launchpad.obol.org/invite/<cluster-defintion> to review the cluster configuration and begin the distributed key generation ceremony. 
```

Complexity:
- If we want to add addresses rather than ENRs, we probably need the `creator` field of a definition to be populated. We can probably use the --charon-private-key if present? or we can generate a temp one if none is present? its not generally that crucial an address right now
- Whatever `creator` address we decide to use, we likely have to first call the /signTermsandConditions api endpoint or the request to POST a defintion to /dv will be rejected. 

- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Describe the solution to be implemented

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #3583, opened 2025-03-12
cluster: add support for cluster invitation link

DiogoSantoss:
Add support for cluster creator to generate a partial cluster definition and submit it to Obol's API to receive a cluster invitation link that can be shares with remaining operators.

Design doc: [link](https://docs.google.com/document/d/1OUTDAnWlqMw-edEQCb2-n-9FV4lMzlBQsqXPV5l0Gbc/edit?tab=t.0)

category: feature
ticket: #3518

### DIFF
--- a/app/obolapi/api.go
+++ b/app/obolapi/api.go
@@ -21,8 +21,11 @@ const (
 	// launchpadReturnPathFmt is the URL path format string at which one can find details for a given cluster lock hash.
 	launchpadReturnPathFmt = "/lock/0x%X/launchpad"
 
-	// defaultTimeout is the default HTTP request timeout if not specified
+	// defaultTimeout is the default HTTP request timeout if not specified.
 	defaultTimeout = 10 * time.Second
+
+	// termsAndConditionsHash is the hash of the terms and conditions that the user must sign.
+	termsAndConditionsHash = "0xd33721644e8f3afab1495a74abe3523cec12d48b8da6cb760972492ca3f1a273"
 )
 
 // New returns a new Client.
@@ -118,38 +121,6 @@ func (c Client) PublishDefinition(ctx context.Context, def cluster.Definition, s
 	return nil
 }
 
-// VeriftySignedTermsAndConditions verifies if the user address has previously signed Obol's Terms and Conditions.
-func (c Client) VerifySignedTermsAndConditions(ctx context.Context, userAddr string) (bool, error) {
-	type response struct {
-		Signed bool `json:"isTermsAndConditionsSigned"`
-	}
-
-	addr := c.url()
-	addr.Path = "v1/termsAndConditions/" + userAddr
-
-	ctx, cancel := context.WithTimeout(ctx, c.reqTimeout)
-	defer cancel()
-
-	resp, err := httpGet(ctx, addr, nil)
-	if err != nil {
-		return false, err
-	}
-
-	defer resp.Close()
-
-	buf, err := io.ReadAll(resp)
-	if err != nil {
-		return false, errors.Wrap(err, "read response body")
-	}
-
-	var res response
-	if err := json.Unmarshal(buf, &res); err != nil {
-		return false, errors.Wrap(err, "unmarshal response")
-	}
-
-	return res.Signed, nil
-}
-
 // SignTermsAndConditions submits the user's signature of Obol's Terms and Conditions to obol-api.
 func (c Client) SignTermsAndConditions(ctx context.Context, userAddr string, forkVersion []byte, sig []byte) error {
 	type request struct {
@@ -165,7 +136,7 @@ func (c Client) SignTermsAndConditions(ctx context.Context, userAddr string, for
 	req := request{
 		Address:                userAddr,
 		Version:                1,
-		TermsAndConditionsHash: "0xd33721644e8f3afab1495a74abe3523cec12d48b8da6cb760972492ca3f1a273",
+		TermsAndConditionsHash: termsAndConditionsHash,
 		ForkVersion:            fmt.Sprintf("0x%x", forkVersion),
 	}
 
--- a/app/obolapi/api.go
+++ b/app/obolapi/api.go
@@ -114,11 +114,6 @@ func (c Client) PublishDefinition(ctx context.Context, def cluster.Definition, s
 	defer cancel()
 
 	return httpPost(ctx, addr, b, headers)
-	if err != nil {
-		return err
-	}
-
-	return nil
 }
 
 // SignTermsAndConditions submits the user's signature of Obol's Terms and Conditions to obol-api.
--- a/app/obolapi/api.go
+++ b/app/obolapi/api.go
@@ -99,7 +99,7 @@ func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
 // It requires the cluster creator to previously sign Obol's Terms and Conditions.
 func (c Client) PublishDefinition(ctx context.Context, def cluster.Definition, sig []byte) error {
 	addr := c.url()
-	addr.Path = "v1/definition"
+	addr.Path = "definition"
 
 	b, err := def.MarshalJSONAPI()
 	if err != nil {
@@ -107,7 +107,7 @@ func (c Client) PublishDefinition(ctx context.Context, def cluster.Definition, s
 	}
 
 	headers := map[string]string{
-		"authorization": fmt.Sprintf("Bearer 0x%x", sig),
+		"authorization": bearerString(sig),
 	}
 
 	ctx, cancel := context.WithTimeout(ctx, c.reqTimeout)
@@ -126,7 +126,7 @@ type RequestSignTermsAndConditions struct {
 // SignTermsAndConditions submits the user's signature of Obol's Terms and Conditions to obol-api.
 func (c Client) SignTermsAndConditions(ctx context.Context, userAddr string, forkVersion []byte, sig []byte) error {
 	addr := c.url()
-	addr.Path = "v1/termsAndConditions"
+	addr.Path = "termsAndConditions"
 
 	req := RequestSignTermsAndConditions{
 		Address:                userAddr,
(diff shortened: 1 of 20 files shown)


app/obolapi/api.go
```diff
@@ -91,6 +92,103 @@ func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
...
+	}
+
+	headers := map[string]string{
+		"authorization": fmt.Sprintf("Bearer 0x%x", sig),
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, c.reqTimeout)
+	defer cancel()
+
+	err = httpPost(ctx, addr, b, headers)
```
pinebit, line 113:
```suggestion
	return httpPost(ctx, addr, b, headers)
```

pinebit (commented):
May have more comments later

pinebit:
This needs a rebase

cluster/definition.go
```diff
@@ -1112,6 +1161,26 @@ type definitionJSONv1x10 struct {
 	DefinitionHash     ethHex                    `json:"definition_hash"`
 }
 
+// definitionJSONAPI is the json formatter of Definition for Obol API.
+type definitionJSONAPI struct {
```
KaloyanTanev, line 1165: Can't we use the `Definition` struct from the `cluster` package? It seems more or less the same and adding this is yet another place to update every time we add something to the definition in future versions.

cmd/createdkg.go
```diff
@@ -68,15 +75,24 @@ func newCreateDKGCmd(runFunc func(context.Context, createDKGConfig) error) *cobr
 			}
 		}
 
+		if config.Publish {
+			mustMarkFlagRequired(cmd, "publish-address")
```
KaloyanTanev, line 79: This one should have default, so you will never hit this guard.

cmd/createdkg.go
```diff
@@ -85,14 +101,15 @@ func bindCreateDKGFlags(cmd *cobra.Command, config *createDKGConfig) {
 	cmd.Flags().StringSliceVar(&config.WithdrawalAddrs, "withdrawal-addresses", nil, "Comma separated list of Ethereum addresses to receive the returned stake and accrued rewards for each validator. Either provide a single withdrawal address or withdrawal addresses for each validator.")
 	cmd.Flags().StringVar(&config.Network, "network", defaultNetwork, "Ethereum network to create validators for. Options: mainnet, goerli, sepolia, hoodi, holesky, gnosis, chiado.")
 	cmd.Flags().StringVar(&config.DKGAlgo, "dkg-algorithm", "default", "DKG algorithm to use; default, frost")
-	cmd.Flags().IntSliceVar(&config.DepositAmounts, "deposit-amounts", nil, "List of partial deposit amounts (integers) in ETH. Values must sum up to at least 32ETH.")
-	cmd.Flags().StringSliceVar(&config.OperatorENRs, operatorENRs, nil, "[REQUIRED] Comma-separated list of each operator's Charon ENR address.")
+	cmd.Flags().IntSliceVar(&config.DepositAmounts, "deposit-amounts", nil, "List of partial deposit amounts (integers) in ETH. Values must sum up to exactly 32ETH.")
```
KaloyanTanev, line 104: Why is that change in this PR?

cmd/createdkg.go
```diff
@@ -149,29 +187,81 @@ func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
 		return err
 	}
 
+	var privKey *k1.PrivateKey
+	creator := cluster.Creator{}
+
+	// Populate creator field
+	if conf.Publish {
+		// TODO(diogo): Should we store this private key in the disk?
```
KaloyanTanev, line 195: 1. Do we use that creator data somewhere? If we aren't using it we can simply continue passing empty cluster creator.
2. Do we use the private/public key generated here in some other flow? If not, we can keep this key in memory instead of disk.

cmd/createdkg.go
```diff
@@ -149,29 +187,81 @@ func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
...
+		if err != nil {
+			return errors.Wrap(err, "sign cluster definition")
+		}
+	}
+
 	eth1Cl := eth1wrap.NewDefaultEthClientRunner(conf.ExecutionEngineAddr)
 	go eth1Cl.Run(ctx)
 
-	if err := def.VerifySignatures(eth1Cl); err != nil {
+	if err := def.VerifySignatures(eth1Cl, conf.Publish); err != nil {
```
KaloyanTanev, line 232: Do we even need to call that if we have `conf.Publish`? Aren't we verifying the signature we have just created 10 lines above?

cmd/createdkg.go
```diff
@@ -149,29 +187,81 @@ func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
...
 		return err
 	}
 
+	if conf.Publish {
```
KaloyanTanev, line 236: The code below is probably a good candidate for its own separate function 😄

cmd/createdkg.go
```diff
@@ -149,29 +187,81 @@ func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
...
+
+		log.Info(ctx, "Creator successfully signed Obol's terms and conditions")
+
+		err = apiClient.PublishDefinition(ctx, def, def.Creator.ConfigSignature)
+		if err != nil {
+			return errors.Wrap(err, "publish cluster definition")
+		}
+
+		log.Info(ctx, "Cluster Invitation Prepared")
+		log.Info(ctx, "Direct the Node Operators to: https://launchpad.obol.org/dv#"+fmt.Sprintf("%#x", def.ConfigHash)+" to review the cluster configuration and begin the distributed key generation ceremony.\n")
```
KaloyanTanev, line 260: Probably no need of the `\n`?

cmd/createdkg.go
```diff
@@ -149,29 +187,81 @@ func runCreateDKG(ctx context.Context, conf createDKGConfig) (err error) {
...
+
+		return nil
```
KaloyanTanev, line 262: Why do we want to skip cluster definition being written to disk?

eth2util/eip712/eip712.go
```diff
@@ -64,7 +65,14 @@ func domainToType(domain Domain) Type {
 
 // HashTypedData returns the hash of the typed data.
 func HashTypedData(data TypedData) ([]byte, error) {
-	domainHash, err := hashData(domainToType(data.Domain))
+	domainType := domainToType(data.Domain)
+
+	// TODO(diogo): any better ideas?
```
KaloyanTanev, line 70: Having the type names as consts is slightly neater.

I don't get why are we deleting those fields?

app/obolapi/api.go
```diff
@@ -91,6 +95,66 @@ func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
 	return nil
 }
 
+// PublishDefinition posts the cluster definition to obol-api.
+// It requires the cluster creator to previously sign Obol's Terms and Conditions.
+func (c Client) PublishDefinition(ctx context.Context, def cluster.Definition, sig []byte) error {
+	addr := c.url()
+	addr.Path = "v1/definition"
```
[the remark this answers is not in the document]
DiogoSantoss: will correct the "v1" part
`PublishLock` also does this `addr.Path="lock"` should we also extract to const?

cluster/definition.go
DiogoSantoss: `Definition` contains extra fields which api refuses to accept

cmd/createdkg.go
DiogoSantoss: Probably my mistake during rebase

cmd/createdkg.go
DiogoSantoss: I don't understand the comment
I placed this here to prevent misuse of the flags

cmd/createdkg.go
DiogoSantoss: The creator is necessary because someone has to sign the eip712 messages. The issue with not storing it in disk is that later on no one will be able to verify who created the partial definition

cmd/createdkg.go
DiogoSantoss: Because this is a partial definition so I assumed operators would get the necessary artifacts from launchpad

app/obolapi/api.go
```diff
@@ -91,6 +95,66 @@ func (c Client) PublishLock(ctx context.Context, lock cluster.Lock) error {
 	return nil
 }
 
+// PublishDefinition posts the cluster definition to obol-api.
+// It requires the cluster creator to previously sign Obol's Terms and Conditions.
+func (c Client) PublishDefinition(ctx context.Context, def cluster.Definition, sig []byte) error {
```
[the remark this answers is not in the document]
DiogoSantoss: Are you saying all publish related functions or also other calls to api ?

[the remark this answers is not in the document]
KaloyanTanev: The functions you've added + the already existing `PublishLock` one

app/obolapi/api.go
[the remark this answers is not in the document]
KaloyanTanev: Yes, let's extract that as well, you can check how we have them on top as consts in exits.

cmd/createdkg.go
KaloyanTanev: As we have default value for the flag `--publish-address`, there is no need to mark it as required. If the user doesn't specify it, it will use the default. If the user specifies it, it will use the one from the user. Hence, this check will never be triggered.

cmd/createdkg.go
KaloyanTanev: Alright, I see. And with the currently existing flow, do we have this option - to verify who created definition?

cmd/createdkg.go
KaloyanTanev: I see, makes sense!

eth2util/eip712/eip712.go
KaloyanTanev: Weird. Probably add a comment to that function and explain the reasoning... And check with API team why is that.

cluster/definition.go
DiogoSantoss: For example, `{"status": 400, "body": "{\"message\":[\"property definition_hash should not exist\"],\"error\":\"Bad Request\",\"statusCode\":400}"}`
I'm guessing it's anything not in the DTO

DiogoSantoss: Since api is going to accept this extra field I'll be removing this struct and going back to the regular ones

pinebit (approved):

PR #3599, opened 2025-03-24
vapi: return 415 on ssz unmarshaling failures

KaloyanTanev:
Some VCs can force SSZ on all requests, even though they are not officially in the spec (i.e.: /eth/v1/validator/duties/attester/{epoch}). On 415 they fallback to JSON. However, those 415 fallbacks fail on our end because we officially support SSZ, but then on parsing the payload to SSZ we return 500.

category: bug
ticket: none

### DIFF
--- a/core/validatorapi/router.go
+++ b/core/validatorapi/router.go
@@ -1448,7 +1448,7 @@ func unmarshal(typ contentType, body []byte, v any) error {
 		unmarshaller, ok := v.(ssz.Unmarshaler)
 		if !ok {
 			return apiError{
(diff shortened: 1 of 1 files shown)


pinebit (approved):

nflaig (commented):
Thanks for looking into this! This is the behavior as [noted in the beacon-api spec](https://github.com/ethereum/beacon-APIs/blob/master/beacon-node-oapi.yaml#L14-L17)

DiogoSantoss (approved):

PR #3601, opened 2025-03-24
cmd: increase test command timeout tests x1000

KaloyanTanev:
Sometimes we see those tests failing, because the timeout of the context is so short, the program cannot even reach the point where all processes are started (i.e.: when testing with 2 MEV relays, we get output of only 1).

category: test
ticket: none


pinebit (approved):

KaloyanTanev:
Seems like this timeout is too long now and some tests pass... I will revise those tests later

Issue #3596, opened 2025-03-20
Bump golang version to 1.24


pinebit:
# 🎯 Problem to be solved

Bump project's golang version to 1.24. There are new features we would like to use in that release.

# 🛠️ Proposed solution
- [ ] Bump golang version for the project.

# 🧪 Tests
- [ ] Tested by automated unit/integration/smoke tests
- [ ] Manually tested with kurtosis

KaloyanTanev:
Finished by #3605

PR #3605, opened 2025-03-25
*: bump go to 1.24

KaloyanTanev:
- bump go to v1.24
- bump linter from v1 to v2
- some autoformatting of yaml files
- linting
  - swap deprecated functions
  - accommodate new linters
  - remove unused linters

category: misc
ticket: #3596

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -1,9 +1,9 @@
 name: Setup go
 
 runs:
-  using: 'composite'
+  using: "composite"
   steps:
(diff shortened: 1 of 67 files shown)


pinebit (approved):

pinebit (approved):

DiogoSantoss (approved):
nit: technically in go 1.24 `Read` from `crypto/rand` will always return nil error [source](https://pkg.go.dev/crypto/rand@master#Read) so we could remove some checks

PR #3609, opened 2025-03-27
*: tracing refactoring

pinebit:
Improved tracing. Removed odd spans.
Approximate traffic: ~30Kb/sec per node.

category: refactor
ticket: none


core/parsigex/parsigex.go
```diff
@@ -87,8 +88,11 @@ func (m *ParSigEx) handle(ctx context.Context, _ peer.ID, req proto.Message) (pr
 		return nil, false, errors.Wrap(err, "convert parsigex proto")
 	}
 
-	ctx, span := core.StartDutyTrace(ctx, duty, "core/parsigex.Handle")
-	defer span.End()
+	if duty.Type == core.DutyProposer || duty.Type == core.DutyBuilderProposer {
```
KaloyanTanev, line 91: I think we can safely assume `core.DutyBuilderProposer` isn't used anymore. I believe any version that uses it won't be compatible with the Charon version that will have this code, as it'll be before v1.

KaloyanTanev (approved):

core/parsigex/parsigex.go
pinebit: Removed, thank you

Issue #3350, opened 2024-10-25
Add optional --publish flags to our test commands


OisinKyne:
# 🎯 Problem to be solved

People are open to giving anonymous and maybe pseudonymous performance data, to get a better understanding of their node's health, and how best to optimise it. 

Right now they fill out forms by hand, translating what their CLI test output says. We then manually translate that into charts and reports. We will need to simplify this to scale it. 

# 🛠️ Proposed solution

We could add an optional `--publish` flag (and `--publish-address`), like the rest of our commands, that pushes to our API. 

- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Describe the solution to be implemented

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

# 👐 Additional acceptance criteria
- [ ] All of the related impact on Launchpad/API team 

# ❌ Out of Scope

In future we could also return a launchpad URI like `https://launchpad.obol.org/operator/<address>/<enr>/` that shows a more fancy test report.

LukeHackett12:
sample output here would be great 👍🏼

LukeHackett12:
{
"charon_peers": {
"category_name": "peers",
"targets": {
"peer anxious-pencil enr:-HW4QDwUF...vKDw": [
{
"name": "ping",
"verdict": "Fail",
"error": "timeout/interrupted"
}
],
"peer inexpensive-farm enr:-HW4QBHlc...rx6o": [
{
"name": "ping",
"verdict": "Fail",
"error": "timeout/interrupted"
}
],
"relay https://0.relay.obol.tech": [
{
"name": "pingRelay",
"verdict": "OK",
"error": ""
},
{
"name": "pingMeasureRelay",
"verdict": "Poor",
"measurement": "290.590325ms",
"error": ""
}
],
"relay https://1.relay.obol.tech": [
{
"name": "pingRelay",
"verdict": "OK",
"error": ""
},
{
"name": "pingMeasureRelay",
"verdict": "Poor",
"measurement": "294.153029ms",
"error": ""
}
],
"relay https://2.relay.obol.dev": [
{
"name": "pingRelay",
"verdict": "OK",
"error": ""
},
{
"name": "pingMeasureRelay",
"verdict": "Avg",
"measurement": "70.054395ms",
"error": ""
}
],
"self": [
{
[...]

PR #3631, opened 2025-04-07
*: add test command publish

KaloyanTanev:
Add publish option for test commands.

category: feature
ticket: #3350

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -20,12 +20,16 @@ import (
 	"time"
 	"unicode/utf8"
 
+	ssz "github.com/ferranbt/fastssz"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/k1util"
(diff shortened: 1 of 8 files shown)


DiogoSantoss (approved):

cmd/test.go
```diff
@@ -213,12 +231,67 @@ func appendScore(cat []string, score []string) []string {
...
+		Sig:  sig,
+		Data: data,
+	}
+
+	obolAPIJSON, err := json.Marshal(res)
+	if err != nil {
+		return errors.Wrap(err, "marshal Obol API test struct")
+	}
+
+	err = obolAPI.PostTestResult(ctx, obolAPIJSON)
```
pinebit, line 289: Could be just `return obolAPI.PostTestResult(ctx, obolAPIJSON)`

pinebit (approved):

cmd/test.go
KaloyanTanev: I usually prefer returning nils in longer functions like that one... no big difference, just a bit more clear for the reader IMO and easier to add more functionalities/debug logs later on.

Issue #3640, opened 2025-04-09
New default partial deposits in compounding mode


pinebit:
# 🎯 Problem to be solved

Apply default partial deposits `[1,8,32,256]` when `--compounding` flag is enabled. Otherwise, the default values remain `[1,32]`. Also, increase lock file publishing timeout to handle large lock files.

# 🛠️ Proposed solution
- [ ] Change default partial deposit values
- [ ] Change POST timeout logic

Describe the solution to be implemented

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #3641, opened 2025-04-09
*: new default partial deposits

pinebit:
New default partial deposits `[1,8,32,256]` when `--compounding` flag is enabled. Otherwise, the default values are `[1,32]`.

category: feature
ticket: #3640

### DIFF
--- a/eth2util/deposit/deposit.go
+++ b/eth2util/deposit/deposit.go
@@ -291,8 +291,14 @@ func DedupAmounts(amounts []eth2p0.Gwei) []eth2p0.Gwei {
 	return result
 }
 
-// DefaultDepositAmounts returns the default deposit amounts: 1ETH and 32ETH.
-func DefaultDepositAmounts() []eth2p0.Gwei {
+// DefaultDepositAmounts returns the default deposit amounts:
+// --compounding=false: [1,32] ETH,
+// --compounding=true: [1,8,32,256] ETH.
+func DefaultDepositAmounts(compounding bool) []eth2p0.Gwei {
+	if compounding {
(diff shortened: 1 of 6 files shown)


DiogoSantoss (approved):

eth2util/deposit/deposit.go
```diff
@@ -291,8 +291,14 @@ func DedupAmounts(amounts []eth2p0.Gwei) []eth2p0.Gwei {
...
 }
 
-// DefaultDepositAmounts returns the default deposit amounts: 1ETH and 32ETH.
-func DefaultDepositAmounts() []eth2p0.Gwei {
+// DefaultDepositAmounts returns the default deposit amounts:
+// --compounding=false: [1,32] ETH,
+// --compounding=true: [1,8,32,256] ETH.
+func DefaultDepositAmounts(compounding bool) []eth2p0.Gwei {
+	if compounding {
+		return []eth2p0.Gwei{MinDepositAmount, 8 * OneEthInGwei, 32 * OneEthInGwei, 256 * OneEthInGwei}
```
KaloyanTanev, line 299: Does it make sense to make it like that?

```suggestion
		return []eth2p0.Gwei{MinDepositAmount, 8 * MinDepositAmount, 32 * MinDepositAmount, 256 * MinDepositAmount}
```

KaloyanTanev (approved):

eth2util/deposit/deposit.go
pinebit: "MinDepositAmount" looks inconvenient here, because the reader needs to go and check what value it is. `OneEthInGwei` looks self-explanatory.

Issue #3636, opened 2025-04-08
Review fetch active/complete activators by slot


DiogoSantoss:
# 🎯 Problem to be solved
We updated the validator cache refresh logic to fetch active/complete validators by slot rather than by the head of beacon node. Since some beacon nodes could be slow, we introduced a retry mechanism.

However this retry mechanism has a high retry count, doesn't use our `expbackoff` implementation and doesn't default to fetching by `head` after N failed attempts. This lead to some beacon nodes refusing to reply due to our excessive retries leading to empty response which caused further issues.

# 🛠️ Proposed solution
- [ ] Core team consensus on the proposed solution

Try to fetch once by slot (N).
If it fails, fetch by head and record that the previous fetched failed.
In the next slot (N+1), we check if the previous failed and retry fetching by slot (N and not N+1).

Add grafana panel to display which peers are connected to possibly slow beacon nodes by recording when we are able to fetch by slot versus when we default to head.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #3642, opened 2025-04-09
*: fix validator cache behavior

DiogoSantoss:
Validator cache is refreshed at the start of the epoch. Previous behavior would request too many times causing issues with beacon node and had no fallback in case of complete failure.

This PR fixes this behavior according to what's described in ticket.

category: bug
ticket: #3636

### DIFF
--- a/app/eth2wrap/valcache.go
+++ b/app/eth2wrap/valcache.go
@@ -92,7 +92,7 @@ func (c *ValidatorCache) cached() (CompleteValidators, bool) {
 }
 
 // Get returns the cached active validators, cached complete Validators response, or fetches them if not available populating the cache.
-func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteValidators, error) {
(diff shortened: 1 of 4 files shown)


app/eth2wrap/valcache.go
```diff
@@ -143,40 +135,35 @@ func (c *ValidatorCache) Get(ctx context.Context) (ActiveValidators, CompleteVal
 }
 
 // GetBySlot fetches active and complete validator by slot populating the cache.
-func (c *ValidatorCache) GetBySlot(ctx context.Context, slot uint64) (ActiveValidators, CompleteValidators, error) {
+// If it fails to fetch by slot, it falls back to head state and retries to fetch by slot next slot.
+func (c *ValidatorCache) GetBySlot(ctx context.Context, slot uint64) (ActiveValidators, CompleteValidators, bool, error) {
```
KaloyanTanev, line 139: Can we rename the other function to `GetByHead`?

KaloyanTanev (approved):

PR #3643, opened 2025-04-10
*: improving spans

pinebit:
Improved tracing spans data: added status codes, recording errors, etc. This is in particular useful with spanmetrics connector and raw traces rendering with grafana, as it gives more context. This barely increases network traffic, because we only track spans for Propose duty.

Fixed a bug in QBFT that prevented QBFT instance to stop promptly upon Decide. The instance was eventually terminating after the timeout - 60s, but this retained unnecessary memory and system handles. Now this finished QBFT protocol as soon as Decided is triggered. Also, this resulted in proper tracing span duration evaluation.

category: refactor
ticket: none

### DIFF
--- a/core/parsigdb/memory.go
+++ b/core/parsigdb/memory.go
@@ -92,10 +92,6 @@ func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet co
 			continue
 		}
 
-		log.Debug(ctx, "Partial signed data stored",
-			z.Int("count", len(sigs)),
-			z.Any("pubkey", pubkey))
-
 		// Check if sufficient matching partial signed data has been received.
 		psigs, ok, err := getThresholdMatching(duty.Type, sigs, db.threshold)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


core/parsigdb/memory.go
```diff
@@ -92,10 +92,6 @@ func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet co
 			continue
 		}
 
-		log.Debug(ctx, "Partial signed data stored",
```
pinebit, line 95: This log is annoying, generates a lot of entries, yet barely useful. Hence it was removed.

pinebit: Debugging is one time event, while it generates tons of traffic we pay for. I think we shall ensure we log sufficient warnings+errors to investigate in prod.

DiogoSantoss (approved):

core/parsigdb/memory.go
KaloyanTanev: We shouldn't run Debug mode in prod either way, right?
Can we at least move it to `StoreInternal`? This way we won't be logging on partial signatures received from peers and VC, but only from VC.
Another option is to move it to the more granular `Trace` which we currently aren't making use of.

Issue #3580, opened 2025-03-10
Listen to the beacon node SSE endpoint and expose via prom metrics


OisinKyne:
# 🎯 Problem to be solved

During the holesky fork incident, we did not have much visibility into what chain each beacon node was on. We can improve this observability. 

We also hit the BN with many requests every slot, we could potentially change charon's behaviour to be more efficient by leveraging the beacon node server sent events feature. 

# 🛠️ Proposed solution

I believe we can progress both of these problems, with a low impact feature which listens to BN events and debug logs + monitors them. 

First we should create a design doc, outlining exactly what events we will listen to, what we log, and what of them we expose to monitoring and why. We should consider scraping every type of BN for the [standard-ish metrics](https://github.com/ethereum/beacon-metrics/blob/master/metrics.md) in every config charon is deployed in, and we should consider a charon-based client, that either outbounds requests to BNs, or subscribes to the SSE endpoint. 

We Register a handler to the BN SSE endpoints. 

When events come in. Debug log them, and update prom gauges as appropriate. 

Create grafana panels that display this info. 

- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

# 👐 Additional acceptance criteria

We can plan a future feature where we share something like the parentRoot observed on peerInfo, such that we can warn if it looks like peers are on different forks (we can spot this on our central prom before shipping this). 

# ❌ Out of Scope

Making changes to scheduler or triggering retries etc if we detect a re-org. (Though we should plan to make optimisations here). 

Pushing this code into an eth2-client library/package (though this would be a good candidate for something to abstract into a package early in the development of one)

OisinKyne:
Feedback from @anadi2311 in retro today that being able to diagnose incorrect head votes is difficult because there are no metrics for what a peer thinks HEAD is. I think this might be a solid user/use case to test this ticket with, as it nicely incorporates metrics + logs. 

[...]

PR #3666, opened 2025-04-29
app: add SSE beacon node metrics

KaloyanTanev:
Add SSE polling for monitoring purposes.

SSE is started at charon run, opening simultaneous connection to all configured beacon nodes (but not to fallback nodes). Once head is updated or reorg occurs, they are sent to the charon's metrics.

Currently gauges are set, however, those will most likely change after discussion with Luke.

category: feature
ticket: #3580


app/sse.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/app/log"
+	"github.com/obolnetwork/charon/app/sseclient"
+	"github.com/obolnetwork/charon/app/z"
+	"github.com/obolnetwork/charon/eth2util"
+)
+
+const (
+	sseHead       = "head"
+	sseChainReorg = "chain_reorg"
```
pinebit, line 19:
```suggestion
	sseHeadEvent       = "head"
	sseChainReorgEvent = "chain_reorg"
```

app/sse.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type SSEHead struct {
+	Slot                      string `json:"slot"`
```
pinebit, line 23: Can this be uint64 to avoid str->int conversion after json.Unmarshal?

app/sseclient/client.go
```diff
@@ -0,0 +1,214 @@
...
+	return &Client{
+		URL:        url,
+		HTTPClient: &http.Client{},
+		Headers:    make(http.Header),
+	}
+}
+
+// Start connects to the SSE stream. This function will block until SSE stream is stopped.
+func (c *Client) Start(ctx context.Context, eventFn EventHandler, errorFn ErrorHandler) error {
+	timeout := c.Retry / 32
```
pinebit, line 48: Please check if `app/expbackoff` is sufficient to do the same in a nicer way.

app/sseclient/client.go
```diff
@@ -0,0 +1,214 @@
...
+
+		// Check response type.
+		switch string(parts[0]) {
+		case "retry":
+			ms, err := strconv.Atoi(string(parts[1]))
+			if err != nil {
+				continue
+			}
+
+			c.Retry = time.Duration(ms) * time.Millisecond
```
pinebit, line 160: shall we disconnect+reconnect after this?

app/sse.go
KaloyanTanev: By the spec we are receiving a string...

pinebit: OMG

app/sseclient/client.go
KaloyanTanev: I forgot about our expbackoff! Will utilise it here.

app/sseclient/client.go
KaloyanTanev: I've looked into BNs implementations and it doesn't seem they utilise this "retry" call. However, I've kept the implementation of it to be compliant with the [official docs](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface) of event sourcing. The official docs don't really give details about what is expected from "retry", so I assume it's up to the implementer (as some other [parts of the doc](https://html.spec.whatwg.org/multipage/server-sent-events.html#concept-event-stream-reconnection-time)). Given that BNs don't utilise that, I don't think it's worth putting more effort into that...

PR #3667, opened 2025-04-29
app/health: add fallback beacon node usage health check

DiogoSantoss:
Add health check which monitor increase in usage of fallback beacon nodes.

category: feature
ticket: #3328


KaloyanTanev (approved):

pinebit (changes requested):
Please add test(s) for this warning (see other tests in `checks_internal_test.go`

Issue #3671, opened 2025-05-02
Revisiting BN calls and retry settings


pinebit:
# 🎯 Problem to be solved

We shall revisit our calls to BN, retry policies and improve any inefficient occurrences. This is to reduce BN load which performance may suffer from unnecessary charon calls, especially retries that don't make sense.

# 🛠️ Proposed solution
- [ ] Reivist all BN calls and the corresponding retry settings.
- [ ] Remove/adjust odd calls and/or fix retry settings.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

pinebit:
Design proposal is under review.

PR #3685, opened 2025-05-07
*: introduced duty submission deadline

pinebit:
The proposal to avoid hammering BNs when executing our standard retrying logic: for specific duties, we evaluate different deadlines specifically for submission (applied to bcast.Broadcast):

```
        case DutyProposer:
			end = start.Add(slotDuration / 3)
		case DutyAttester:
			end = start.Add(slotDuration * 2 / 3)
		case DutySyncMessage:
			end = start.Add(slotDuration * 2 / 3)
		case DutyAggregator:
			end = start.Add(slotDuration)
		case DutySyncContribution:
			end = start.Add(slotDuration)
```

for others, we remain the deadline intact.

category: feature
ticket: #3671

### DIFF
--- a/core/deadline.go
+++ b/core/deadline.go
@@ -26,7 +26,12 @@ const lateFactor = 5
 const lateMin = time.Second * 30
 
 // DeadlineFunc is a function that returns the deadline for a duty.
+// The second return value indicates whether the duty has a deadline.
 type DeadlineFunc func(Duty) (time.Time, bool)
+
+// DeadlineFunc is a function that returns the submission deadline for a duty.
+// The second return value indicates whether the duty has a deadline.
+type SubmitDeadlineFunc func(Duty) (time.Time, bool)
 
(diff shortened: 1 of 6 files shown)


core/deadline.go
```diff
@@ -197,37 +202,75 @@ func getCurrDuty(duties map[Duty]bool, deadlineFunc DeadlineFunc) (Duty, time.Ti
...
-		}
+		start := genesisTime.Add(slotDuration * time.Duration(duty.Slot))
 		end := start.Add(delta)
 
 		return end, true
-	}, nil
+	}
+
+	submitDutyDeadline := func(duty Duty) (time.Time, bool) {
+		if duty.Type == DutyExit || duty.Type == DutyBuilderRegistration {
```
pinebit, line 245: I think to refactor this entirely: to make an object for each duty that is responsible for tracking its own deadlines. Having common mockable interface of course.

PR #3696, opened 2025-05-07
tracker: check inclusion by validx, instead of agg bits

KaloyanTanev:
Check attestation inclusion in an easier way, by validator index, instead of aggregation bits, which was throwing false positives before.

category: bug
ticket: none

### DIFF
--- a/core/tracker/inclusion.go
+++ b/core/tracker/inclusion.go
@@ -656,6 +656,9 @@ func (a *InclusionChecker) Run(ctx context.Context) {
 				if !ok {
 					continue
 				}
+				if att.ValidatorIndex == nil {
+					continue
+				}
 				indices = append(indices, *att.ValidatorIndex)
 			}
 
--- a/core/tracker/inclusion.go
+++ b/core/tracker/inclusion.go
@@ -5,11 +5,13 @@ package tracker
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strconv"
 	"sync"
 	"time"
 
 	eth2api "github.com/attestantio/go-eth2-client/api"
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	eth2spec "github.com/attestantio/go-eth2-client/spec"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/prysmaticlabs/go-bitfield"
@@ -59,6 +61,7 @@ type block struct {
 // blockV2 is a simplified block with its v2 attestations.
 type blockV2 struct {
 	Slot                   uint64
+	AttDuties              []*eth2v1.AttesterDuty
 	AttestationsByDataRoot map[eth2p0.Root]*eth2spec.VersionedAttestation
 	BeaconCommitees        []*statecomm.StateCommittee
 }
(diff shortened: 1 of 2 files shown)


core/tracker/inclusion.go
```diff
@@ -423,27 +425,33 @@ func checkAttestationInclusion(sub submission, block block) (bool, error) {
 
 // checkAttestationV2Inclusion checks whether the attestation is included in the block.
 func checkAttestationV2Inclusion(sub submission, block blockV2) (bool, error) {
+	subData, ok := sub.Data.(core.VersionedAttestation)
+	if !ok {
+		return false, errors.New("invalid attestation")
```
pinebit, line 430: I would re-phrase the error: "not an attestation block"

core/tracker/inclusion.go
```diff
@@ -423,27 +425,33 @@ func checkAttestationInclusion(sub submission, block block) (bool, error) {
 
 // checkAttestationV2Inclusion checks whether the attestation is included in the block.
 func checkAttestationV2Inclusion(sub submission, block blockV2) (bool, error) {
+	subData, ok := sub.Data.(core.VersionedAttestation)
+	if !ok {
```
pinebit, line 429: let's check `subData` is not `nil` just in case

core/tracker/inclusion.go
```diff
@@ -499,11 +507,6 @@ func reportMissed(ctx context.Context, sub submission) {
 // reportAttV2Inclusion reports attestations that were included in a block.
 func reportAttV2Inclusion(ctx context.Context, sub submission, block blockV2) {
 	att := block.AttestationsByDataRoot[sub.AttDataRoot]
```
pinebit, line 509: let's do the same check: `att, ok := ...` + `if !ok ...`

core/tracker/inclusion.go
```diff
@@ -423,27 +425,33 @@ func checkAttestationInclusion(sub submission, block block) (bool, error) {
...
+			attesterDutyData = ad
+			break
+		}
 	}
 
-	subData, ok := sub.Data.(core.VersionedAttestation)
-	if !ok {
-		return false, errors.New("invalid attestation")
+	if attesterDutyData == nil {
+		return false, nil
```
pinebit, line 447: can we add a debug output here that we did not find a matching validator as there should be no such the case I guess

core/tracker/inclusion.go
```diff
@@ -632,8 +647,37 @@ func (a *InclusionChecker) Run(ctx context.Context) {
 			if checkedSlot == slot {
 				continue
 			}
+			if checkedEpoch != slot/slotsPerEpoch {
+				epoch := slot / slotsPerEpoch
+				indices := []eth2p0.ValidatorIndex{}
+				a.core.mu.Lock()
+				subs := a.core.submissions
```
pinebit, line 654: This is wrong:
> You are still accessing the original map, not a real copy of its contents. If another goroutine writes to the map at the same time, this may cause: a) race conditions, b) runtime panic: concurrent map read and map write, if writes occur concurrently with reads

Maybe here we start using brand new `sync.Map` from golang 1.24? and avoid mutex entirely?

pinebit (changes requested):
`subs := a.core.submissions` is unsafe

core/tracker/inclusion.go
KaloyanTanev: You mean if it has filled at least one version in the structure? subData itself cannot be nil.

core/tracker/inclusion.go
KaloyanTanev: Ah, you are right! I've cloned it.

I'm a bit reluctant on refactoring for this bugfix... I know also the `sync.Map` is a bit slower. We can probably think about it later?

core/tracker/inclusion.go
```diff
@@ -632,8 +647,37 @@ func (a *InclusionChecker) Run(ctx context.Context) {
 			if checkedSlot == slot {
 				continue
 			}
+			if checkedEpoch != slot/slotsPerEpoch {
```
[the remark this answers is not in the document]
KaloyanTanev: Oh my :)

core/tracker/inclusion.go
KaloyanTanev: Practically we cannot end up in this scenario I believe. It's just a sanity check for the pointer. I will refactor it not to be a pointer

pinebit (approved):

DiogoSantoss (approved):

core/tracker/inclusion.go
```diff
@@ -632,8 +646,36 @@ func (a *InclusionChecker) Run(ctx context.Context) {
 			if checkedSlot == slot {
 				continue
 			}
+			indices := []eth2p0.ValidatorIndex{}
+			a.core.mu.Lock()
+			subs := maps.Clone(a.core.submissions)
```
pinebit, line 651: We shall refactor this after the fix. Now lock+clone is definitely much slower than the new `sync.Map`.

PR #3713, opened 2025-05-14
*: eliminated genesis and spec fetching

pinebit:
Eliminated genesis and spec fetching, to avoid extra BN calls and general code maintainability. 

category: refactor
ticket: none


app/eth2wrap/spec.go
```diff
@@ -0,0 +1,63 @@
...
+	spec, err := client.Spec(ctx, &api.SpecOpts{})
+	if err != nil {
+		return NetworkSpec{}, errors.Wrap(err, "failed to fetch network spec")
+	}
+
+	if spec == nil {
+		return NetworkSpec{}, errors.New("missing network spec")
+	}
+
+	genesisTime, ok := spec.Data["MIN_GENESIS_TIME"].(time.Time)
```
KaloyanTanev, line 43: Is this the same as `genesis_time` fetched from `/eth/v1/beacon/genesis`?

core/validatorapi/validatorapi_test.go
```diff
@@ -1626,7 +1626,7 @@ func TestComponent_SubmitValidatorRegistration(t *testing.T) {
 	allPubSharesByKey := map[core.PubKey]map[int]tbls.PublicKey{corePubKey: {shareIdx: pubkey}} // Maps self to self since not tbls
 
 	// Configure beacon mock
-	bmock, err := beaconmock.New()
+	bmock, err := beaconmock.New(beaconmock.WithGenesisTime(time.Now().Add(-time.Hour)))
```
KaloyanTanev, line 1629: Why is that?

KaloyanTanev (approved):

app/eth2wrap/spec.go
pinebit: Ah.. might be a big fault of mine. I compared values with kurtosis and they were the same, but now double checking this - these in fact might be different.

core/validatorapi/validatorapi_test.go
pinebit: Still need to double check why this test was failing otherwise.

Issue #2922, opened 2024-02-28
Keep a regex for minimum CL versions we support, and emit a startup/occasional warning log if its condition is not met


OisinKyne:
# 🎯 Problem to be solved

We want to keep people updating their CLs promptly. If they end up with unstashed changes in their template repo, they may be uploading charon only via the env and having their CL fall behind. We may discover incompatibilites between charon and older versions. 

# 🛠️ Proposed solution

We can emit a startup or e.g. hourly `warn` log that charon is not fully compatible with the connected CL. This should prompt people to keep their CL slightly more up to date. 

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

xenowits:
This is part of building the compatibility matrix, so that we know exactly which versions of CL we definitely support. The current ticket follows from there

KaloyanTanev:
@OisinKyne is this still relevant? I'd say people should stick to our Kurtosis tested versions. If we want people to update their stack on every new version of EL/CL/VC/MEV/etc., we should first improve our Kurtosis testing to test latest charon with latest said version.

PR #3723, opened 2025-05-19
*: add beacon node minimum version verification

DiogoSantoss:
Add warning log for connect beacon clients running below supported version

category: feature
ticket: #2922

### DIFF
--- a/app/eth2wrap/version.go
+++ b/app/eth2wrap/version.go
@@ -15,7 +15,7 @@ var (
 	minLighthouseVersion, _ = version.Parse("v7.0.0")
 	minTekuVersion, _       = version.Parse("v25.4.1")
 	minLodestarVersion, _   = version.Parse("v1.29.0")
-	minNimbusVersion, _     = version.Parse("v25.4.1")
+	minNimbusVersion, _     = version.Parse("v25.4.0")
 	minPrysmVersion, _      = version.Parse("v6.0.0")
 	minGrandineVersion, _   = version.Parse("v1.1.0")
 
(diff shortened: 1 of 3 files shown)


pinebit (approved):

app/eth2wrap/version.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+var versionExtractRegex = regexp.MustCompile(`^([^/]+)/v?([0-9]+\.[0-9]+\.[0-9]+)`)
+
+// CheckBeaconNodeVersionStatus checks the version of the beacon node client against the minimum required version.
+// It returns the status of the version check as an enum, the current version, and the minimum required version.
+func checkBeaconNodeVersionStatus(bnVersion string) (beaconNodeVersionStatus BeaconNodeVersionStatus, clVer string, minVer string) {
+	matches := versionExtractRegex.FindStringSubmatch(bnVersion)
+	if len(matches) != 3 {
+		return VersionFormatError, "", ""
```
KaloyanTanev, line 46: Can we return the `clVer` here, with the full provided string? This is so that when we log, we can see what the received version was, instead of blindly guessing what the issue was. Would be helpful especially for charons with multiple BNs. Possibly add it also on the further down bad weather path returns, not only this one.

KaloyanTanev (approved):

app/eth2wrap/version.go
DiogoSantoss: The `CheckBeaconNodeVersion` function catches the error and logs the original `bnVersion` so I don't think is necessary to also return it here

KaloyanTanev: Right, I missed that, pardon.

PR #3725, opened 2025-05-19
*: add missing entries to release template and stacksnipe

DiogoSantoss:
Add Grandine and Vouch to release template and Vouch to `stacksnipe`

category: feature
ticket: none


KaloyanTanev (approved):
Do you know if stack snipe succeeds with vouch? Can we check?

pinebit (approved):

DiogoSantoss:
> Do you know if stack snipe succeeds with vouch? Can we check?

I believe it should succeed because it looks for a process named `vouch` which is what we have when we run the command `./vouch`. Testing is a bit difficult because this assumed the entire stack is running on the same machine which is not the case for kurtosis

PR #3738, opened 2025-05-27
docs: replaced broken link intro

Olexandr88:
replaced broken link intro

category: docs
ticket: none

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ which contains a docker compose setup for running a full charon cluster on your 
 ## Documentation
 
 The [Obol Docs](https://docs.obol.tech/) website is the best place to get started.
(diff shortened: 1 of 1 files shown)


KaloyanTanev:
The new link sends to the SDK, not to the general docs landing page. You can make it to `https://docs.obol.org`, in order to be proper fix.

Fix PR's body as well, according to the guidelines. Thanks.

Olexandr88:
> @KaloyanTanev Thanks for the tip, but I'm not sure if this is the right link. https://docs.obol.org/sdk/index#getting-started

Olexandr88:
> @Olexandr88 what you've addressed is the SDK. SDK is separate from Charon itself and shouldn't be the main point of contact for Charon.

@KaloyanTanev check if I have selected the correct link

PR #3739, opened 2025-05-28
core/consensus: add longer proposer duty timeout

DiogoSantoss:
Increase first round timeout value for proposer duty to 1.5s across all timers

category: feature
ticket: #3430
feature_flag: proposal_timeout

### DIFF
--- a/app/featureset/featureset.go
+++ b/app/featureset/featureset.go
@@ -57,7 +57,7 @@ const (
 	// Previously this was the default behaviour, however, tracking on-chain inclusions post-electra is costly.
 	// The extra load that Charon puts the beacon node is deemed so high that it can throttle the completion of other duties.
 	AttestationInclusion Feature = "attestation_inclusion"
-	
+
 	// ProposalTimeout enables a longer first consensus round timeout of 1.5 seconds for proposal duty.
 	ProposalTimeout = "proposal_timeout"
 )


core/consensus/utils/roundtimer.go
```diff
@@ -82,6 +78,12 @@ type RoundTimer interface {
 	Type() TimerType
 }
 
+// proposalTimeoutOptimization returns true if ProposalTimeout feature is enabled, the duty is proposer and
+// we are in the first round
```
KaloyanTanev, line 82:
```suggestion
// we are in the first round.
```

core/consensus/utils/roundtimer.go
```diff
@@ -94,17 +96,38 @@ func NewIncreasingRoundTimerWithClock(clock clockwork.Clock) RoundTimer {
...
 }
 
 func (increasingRoundTimer) Type() TimerType {
 	return TimerIncreasing
 }
 
 func (t increasingRoundTimer) Timer(round int64) (<-chan time.Time, func()) {
-	timer := t.clock.NewTimer(increasingRoundTimeout(round))
+	timeout := increasingRoundTimeout(round)
+	if proposalTimeoutOptimization(t.duty, round) {
```
KaloyanTanev, line 125: Each participant would need to have this feature flag set, in order for the consensus to continue working, right? Otherwise the timings between them will cause chaos? Can we test it what happens when this is the case - 1/2 nodes having this flag set, the other 1/2 do not and it takes >1s for the first block proposal leader?

pinebit: > Each participant would need to have this feature flag set, in order for the consensus to continue working, right?
N-F

DiogoSantoss: Different timeouts don't cause chaos. The only issue is that if there are more nodes with a smaller timeout they can trigger a round change earlier (after 1s instead of 1.5s). Even nodes with this feature_set would still behave correctly  because they would see other nodes voting for a round change and would support this decision (with `UponFPlus1RoundChanges` rule from qbft).
In the specific situtation of:
- 1/2 timeout 1s
- 1/2 timeout 1.5s
- >1s proposal
The result is that we would get `>F+1` votes for a round change after 1s which would trigger other nodes to also broadcast a round change. Eventually they would gather `2F+1` and switch to the next round.

DiogoSantoss:
@pinebit do you mean keeping the same behavior from `linear` but having the first timeout be `1.5` or copying from `eager_dlinear` but with the `1.5`

core/consensus/utils/roundtimer.go
DiogoSantoss: That's fair, I've tested in kurtosis with 2 higher timeout + 1 regular timeout and 1 higher timeout + 2 regular timeout and everything seemed fine, duties were progressing within same number of rounds and consensus duration. We can possibly deploy it to canaries to test half/half and see if the bigger timeout reduces consensus rounds.

pinebit (approved):

PR #3756, opened 2025-06-02
*: scheduler handling reorgs

pinebit:
Scheduler to refresh all duties in case of a previous epoch reorg reported by sse component.

category: feature
ticket: #3580
feature_flag: sse_reorg_duties

### DIFF
--- a/app/featureset/featureset.go
+++ b/app/featureset/featureset.go
@@ -49,6 +49,9 @@ const (
 	// Linear enables Linear round timer for consensus rounds.
 	// When active has precedence over EagerDoubleLinear round timer.
 	Linear Feature = "linear"
+
+	// ReorgRefreshDuties enables Scheduler to refresh duties when reorg occurs.
+	ReorgRefreshDuties Feature = "reorg_refresh_duties"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


core/scheduler/scheduler.go
```diff
@@ -128,9 +129,25 @@ func (s *Scheduler) Run() error {
...
+				return
+			}
+
+			s.trimDuties(uint64(epoch))
+			if err = s.resolveDuties(ctx, slot); err != nil {
+				log.Error(ctx, "Resolving duties error after chain reorg", err, z.U64("reorg_epoch", uint64(epoch)))
+			}
+		}
+	} else {
+		log.Warn(ctx, "Chain reorg event ignored due to disabled ReorgRefreshDuties feature", nil, z.U64("reorg_epoch", uint64(epoch)))
```
KaloyanTanev, line 144: I support that warn :)

app/featureset/featureset.go
```diff
@@ -49,6 +49,9 @@ const (
 	// Linear enables Linear round timer for consensus rounds.
 	// When active has precedence over EagerDoubleLinear round timer.
 	Linear Feature = "linear"
+
+	// ReorgRefreshDuties enables Scheduler to refresh duties when reorg occurs.
```
KaloyanTanev: Can we add something here that it uses the SSE endpoint? I'm not sure how widely it's used and I want to avoid cases where it's not available at some BN deployments...

KaloyanTanev (approved):

app/featureset/featureset.go
pinebit: Renamed with SSE prefix.

DiogoSantoss (approved):

Issue #2978, opened 2024-03-19
Secure and Simplify Import of .charon Folders for DappNode Integration


boulder225:
## 🎯 Problem to be solved
Explore methods for importing a pre-created `.charon` folder into a DappNode. DappNode operates through a web UI, which makes it difficult to upload folders/directories directly. The current workaround involves zipping the `.charon` folder, uploading it, and then unzipping it on the server side. Concerns include the security of private keys during this process and the reliance on web UI over other methods like SSH.

## 🛠️ Proposed solution
- [ ] Consider implementing an optional flag (e.g., `charon create cluster --zipped`) to automatically output node subfolders as zipped files
- [ ] Confirm with the DappNode team the feasibility of uploading folders/directories through the web UI.
- [ ] Investigate the support of using SSH to transfer the `.charon` folder despite the general non-reliance on SSH for DappNode UI access.
- [ ] Address security concerns by exploring the addition of export/import commands that zip and encrypt the data

PR #3757, opened 2025-06-02
cmd: add zipped flag to create cluster

DiogoSantoss:
Add `--zipped` flag to bundle `create cluster` command output files and folders. This creates a `.tar.gz` file in the `conf.ClusterDir`.

category: feature
ticket: #2978

### DIFF
--- a/cmd/createcluster.go
+++ b/cmd/createcluster.go
@@ -1193,8 +1193,8 @@ func validateNetworkConfig(conf clusterConfig) error {
 	return errors.New("missing --network flag or testnet config flags")
 }
 
-// bundleOutput creates a gzipped tarball of the contents of targetDir named cluster.tar.gz and stores it in targetDir.
-// Deletes archived content by removing all node* folders.
+// bundleOutput archives all node directories (node0, node1, ..., nodeN) within targetDir into a gzipped tarball named "cluster.tar.gz" in targetDir.
+// After successfully creating the archive, it deletes the original node directories from disk.
 func bundleOutput(targetDir string, numNodes int) error {
 	file, err := os.Create(filepath.Join(targetDir, "cluster.tar.gz"))
 	if err != nil {
@@ -1208,9 +1208,12 @@ func bundleOutput(targetDir string, numNodes int) error {
 	tw := tar.NewWriter(gzw)
 	defer tw.Close()
 
+	dirs := []string{}
 	for i := range numNodes {
-		dir := filepath.Join(targetDir, fmt.Sprintf("node%d",i))
+		dirs = append(dirs, filepath.Join(targetDir, fmt.Sprintf("node%d", i)))
+	}
 
+	for _, dir := range dirs {
 		err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
 			if err != nil {
 				return errors.Wrap(err, "filepath walk")
@@ -1245,7 +1248,9 @@ func bundleOutput(targetDir string, numNodes int) error {
 		if err != nil {
 			return errors.Wrap(err, "filepath walk")
 		}
+	}
 
+	for _, dir := range dirs {
 		err := os.RemoveAll(dir)
 		if err != nil {
 			return errors.Wrap(err, "remove file")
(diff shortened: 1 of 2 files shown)


cmd/createcluster.go
```diff
@@ -153,6 +158,7 @@ func bindClusterFlags(flags *pflag.FlagSet, config *clusterConfig) {
 	flags.UintVar(&config.TargetGasLimit, "target-gas-limit", 36000000, "Preferred target gas limit for transactions.")
 	flags.BoolVar(&config.Compounding, "compounding", false, "Enable compounding rewards for validators by using 0x02 withdrawal credentials.")
 	flags.StringVar(&config.ExecutionEngineAddr, "execution-client-rpc-endpoint", "", "The address of the execution engine JSON-RPC API.")
+	flags.StringVar(&config.Zipped, "zipped", "", "Zip created files and folders into a zip or tar file")
```
pinebit: Missing dot at the end. Unclear what files and folder are meant here and what are the possible values.

cmd/createcluster.go
pinebit, line 1: Please add a test for this.

DiogoSantoss: Added

cmd/createcluster.go
DiogoSantoss: Switched to bool variable since only tar support is required. Updated the description

cmd/createcluster.go
```diff
@@ -338,6 +343,12 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
 		return err
 	}
 
+	if conf.Zipped {
```
KaloyanTanev, line 346: Do we delete the regular files once the tar is created?

cmd/createcluster.go
```diff
@@ -1178,3 +1189,62 @@ func validateNetworkConfig(conf clusterConfig) error {
...
+		}
+		f, err := os.Open(path)
+		if err != nil {
+			return errors.Wrap(err, "open file")
+		}
+		_, err = io.Copy(tw, f)
+		closeErr := f.Close()
+		if err != nil {
+			return errors.Wrap(err, "copy file", z.Str("filename", path))
+		}
```
KaloyanTanev: Probably this error needs to be a line upwards?

cmd/createcluster.go
```diff
@@ -1178,3 +1189,62 @@ func validateNetworkConfig(conf clusterConfig) error {
...
+
+	err = filepath.Walk(targetDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return errors.Wrap(err, "filepath walk")
+		}
+		// Ignore the tar.gz file itself
+		if strings.HasSuffix(path, ".tar.gz") {
+			return nil
+		}
+		if !info.Mode().IsRegular() {
```
KaloyanTanev: What does this one do?

cmd/createcluster.go
DiogoSantoss: No, do we want to do that ?

cmd/createcluster.go
DiogoSantoss: No because that would cause us to return before closing the file

cmd/createcluster.go
DiogoSantoss: Ignores files like symbolic links, named pipes, etc that we don't want to include in our zip. Mostly a sanity check

cmd/createcluster.go
KaloyanTanev: Wouldn't putting the `f.Close()` in a defer be slightly better? We can also avoid issues with program being abruptly stopped this way.

cmd/createcluster.go
KaloyanTanev: I'd say yes, we don't want duplicate files. And most likely if a person needs a zip, s/he won't need the non-zipped files.

cmd/createcluster.go
DiogoSantoss: I did it manually because usually with recursive functions we want to make sure we don't have too many files opened since they will only close after filepath walk ends.

cmd/createcluster_internal_test.go
```diff
@@ -956,6 +958,96 @@ func TestClusterCLI(t *testing.T) {
 	}
 }
 
+func TestZipping(t *testing.T) {
+	ctx := context.Background()
```
pinebit: Please use `t.Context()` in tests. From now on.

KaloyanTanev (approved):

PR #3760, opened 2025-06-04
eth2util: use the built-in max/min to simplify the code

findmyhappy:
use the built-in max/min to simplify the code


pinebit:
Please check https://github.com/ObolNetwork/charon/blob/main/docs/contributing.md#chore-prs

PR #3762, opened 2025-06-04
core/scheduler: bugfix

pinebit:
Bugfix is to address the following edge case:
- Scheduler is fetching/resolving duties for a slot.
- GetDutyDefinition is called for the same slot at the same time (from VAPI).
- This resulted in failed VAPI request, because the epoch in question did not appear "resolved".

The fix will cause GetDutyDefinition to wait if Scheduler is in process of resolving duties for the requested epoch. This would prevent VAPI request to fail, but apparently it will delay the request by the time needed to complete all duties resolution.

category: bug
ticket: none

### DIFF
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -5,6 +5,7 @@ package scheduler
 import (
 	"context"
 	"math"
+	"runtime"
 	"sort"
 	"sync"
 	"testing"
@@ -59,6 +60,7 @@ func New(pubkeys []core.PubKey, eth2Cl eth2wrap.Client, builderEnabled bool) (*S
 		},
 		metricSubmitter: newMetricSubmitter(),
 		resolvedEpoch:   math.MaxInt64,
+		resolvingEpoch:  math.MaxInt64,
 		builderEnabled:  builderEnabled,
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


core/scheduler/scheduler.go
```diff
@@ -170,6 +172,16 @@ func (s *Scheduler) GetDutyDefinition(ctx context.Context, duty core.Duty) (core
...
 
 	epoch := duty.Slot / slotsPerEpoch
+
+	// This has to be very rare event, when the requested epoch is being resolved.
+	// We wait for the epoch to be resolved before returning the duty definition.
+	for s.isResolvingEpoch(epoch) {
+		if ctx.Err() != nil {
+			return nil, errors.Wrap(ctx.Err(), "context cancelled while waiting for epoch to resolve")
+		}
+		time.Sleep(100 * time.Millisecond)
```
KaloyanTanev, line 182: Probably will be better if we wait on some channel, but not sure if it's worth the time implementing that?

KaloyanTanev (approved):

core/scheduler/scheduler.go
pinebit: I was thinking about the same first, but we need to compare against specific epoch number.. which adds lots of complexity when done via channels. Because this is an edge case, I think it is sufficient to do just busy-wait.

PR #3795, opened 2025-06-10
app: added sse_head_delay buckets

pinebit:
Added explicit buckets to `sse_head_delay` histogram.

category: refactor
ticket: none


OisinKyne (approved):
Kind of felt before/after t=0 was a cleaner good/bad delineation but willing to go with majority rules on how its implemented here with t=12 being the important threshold.

KaloyanTanev (approved):

PR #3807, opened 2025-06-16
cmd: test publish using temp enr key

pinebit:
For `alpha test` command, when `--publish` is specified and there is no private key present at `.charon/charon-enr-private-key`, the implementation will auto generate temporary private key that will not be matching any of real nodes on the backend, but still allows to proceed and publish.

category: feature
ticket: none

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -245,6 +245,8 @@ func publishResultToObolAPI(ctx context.Context, data allCategoriesResult, path 
 	)
 
 	if !fileExists(privateKeyFile) {
+		log.Warn(ctx, "Private key file does not exist, will generate a temporary key", nil)
+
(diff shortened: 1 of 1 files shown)


cmd/test.go
```diff
@@ -238,9 +239,21 @@ type obolAPIResult struct {
...
 func publishResultToObolAPI(ctx context.Context, data allCategoriesResult, path string, privateKeyFile string) error {
-	p2pPrivKey, err := k1util.Load(privateKeyFile)
-	if err != nil {
-		return err
+	var (
+		err        error
+		p2pPrivKey *k1.PrivateKey
+	)
+
+	if !fileExists(privateKeyFile) {
```
KaloyanTanev, line 247: Can we add a warning log here, just for sanity check?

pinebit: Added

PR #3812, opened 2025-06-19
app: fixed vapi monitoring code

pinebit:
Fixed VAPI monitoring code to avoid false-positive triggering "vc_missing_validators" error.

category: refactor
ticket: none


app/monitoringapi.go
```diff
@@ -128,17 +128,36 @@ func startReadyChecker(ctx context.Context, tcpNode host.Host, eth2Cl eth2wrap.C
...
+
 	go func() {
-		ticker := clock.NewTicker(10 * time.Second)
+		genesisTime, err := eth2wrap.FetchGenesisTime(ctx, eth2Cl)
+		if err != nil {
+			log.Error(ctx, "Failed to fetch genesis time", err)
+			return
+		}
+		slotDuration, slotsPerEpoch, err := eth2wrap.FetchSlotsConfig(ctx, eth2Cl)
+		if err != nil || slotDuration == 0 || slotsPerEpoch == 0 {
```
KaloyanTanev: nitpick:
Can we add those `slotDuration == 0 || slotsPerEpoch == 0` to the `eth2wrap.FetchSlotsConfig` function? I believe they are applicable in general for this function and not specific to the monitoring API.

KaloyanTanev (approved):

app/monitoringapi.go
pinebit: Good idea, done.

Issue #3811, opened 2025-06-16
Put QUIC behind a feature flag


OisinKyne:
# 🎯 Problem to be solved

Big clusters have high consensus duration without many proposer timeouts. This might be due to message sizes increasing beyond the TCP packet size. 

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Turn on quic as a supported/preferred libp2p protocol when a feature flag is enabled. Observe a change in consensus duration. 

Expose the protocol in use in debug logging. 

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #3814, opened 2025-06-23
*: add libp2p quic support

DiogoSantoss:
Add support for QUIC protocol in libp2p. 
To use this feature, enable the feature flag `quic` and provide a value to `--p2p-udp-address` flag.

Notes:
Relay nodes should be updated to listen to both TCP and QUIC(UDP) traffic.
Nodes with `quic` feature active will listen to both TCP and QUIC(UDP) traffic.

category: feature
ticket: #3811 
feature_flag: quic

### DIFF
--- a/p2p/p2p.go
+++ b/p2p/p2p.go
@@ -487,6 +487,7 @@ func UpgradeToQUICConnections(p2pNode host.Host, peerIDs []peer.ID) lifecycle.Ho
 			p2pNode.Peerstore().ClearAddrs(p)
 			p2pNode.Peerstore().AddAddrs(p, quicAddrs, peerstore.PermanentAddrTTL)
 
+			t0 := time.Now() // t0: Start closing connections.
 			// Close previous direct connections to ensure that we can establish a QUIC connection
 			for _, conn := range conns {
 				addr := conn.RemoteMultiaddr()
@@ -500,6 +501,7 @@ func UpgradeToQUICConnections(p2pNode host.Host, peerIDs []peer.ID) lifecycle.Ho
 				}
 			}
 
+			t1 := time.Now() // t1: Finished closing. Start connecting. From t1 to t2 node will be disconnected.
 			// Attempt to connect over QUIC
 			// In case of QUIC failure due to infrastructure error (e.g. UDP port closed) the Connect call
 			// may fail or succeed. It can succeed if another peer calls Connect to this peer with a working
@@ -507,6 +509,8 @@ func UpgradeToQUICConnections(p2pNode host.Host, peerIDs []peer.ID) lifecycle.Ho
 			// If it fails, we try to reconnect via any available address
 			// If it succeeds, we check whether we are actually connected via QUIC
 			err := p2pNode.Connect(network.WithForceDirectDial(ctx, ""), peer.AddrInfo{ID: p})
+			t2 := time.Now() // t2: Finished connecting.
+
 			if err != nil {
 				recordUpgradeFailure(p)
 				log.Debug(ctx, "Failed to connect to peer during QUIC upgrade. Reconnecting via any address", z.Str("peer", PeerName(p)), z.Any("peerstore", p2pNode.Peerstore().Addrs(p)), z.Err(err))
@@ -534,7 +538,7 @@ func UpgradeToQUICConnections(p2pNode host.Host, peerIDs []peer.ID) lifecycle.Ho
 					if isQUICAddr(addr) {
 						connectedViaQUIC = true
 
-						log.Debug(ctx, "Upgraded connection to QUIC", z.Str("peer", PeerName(p)), z.Any("addr", addr), z.Any("direction", conn.Stat().Direction))
+						log.Debug(ctx, "Upgraded connection to QUIC", z.Str("peer", PeerName(p)), z.Any("closing_period", t1.Sub(t0)), z.Any("connecting_period", t2.Sub(t1)), z.Any("addr", addr), z.Any("direction", conn.Stat().Direction))
 					} else if isTCPAddr(addr) {
 						log.Debug(ctx, "Connected via TCP after upgrade to QUIC connection", z.Str("peer", PeerName(p)), z.Any("addr", addr), z.Any("direction", conn.Stat().Direction))
 					}
@@ -542,7 +546,7 @@ func UpgradeToQUICConnections(p2pNode host.Host, peerIDs []peer.ID) lifecycle.Ho
 			}
 
 			if !connectedViaQUIC {
-				log.Debug(ctx, "Failed to establish direct connection via QUIC to peer", z.Str("peer", PeerName(p)), z.Any("conns", p2pNode.Network().ConnsToPeer(p)), z.Any("peerstore", p2pNode.Peerstore().Addrs(p)))
+				log.Debug(ctx, "Failed to establish direct connection via QUIC to peer", z.Str("peer", PeerName(p)), z.Any("closing_period", t1.Sub(t0)), z.Any("connecting_period", t2.Sub(t1)), z.Any("conns", p2pNode.Network().ConnsToPeer(p)), z.Any("peerstore", p2pNode.Peerstore().Addrs(p)))
 				recordUpgradeFailure(p)
 			} else {
 				clearUpgradeBackoff(p)


app/app.go
```diff
@@ -330,13 +330,13 @@ func Run(ctx context.Context, conf Config) (err error) {
...
-func wirePeerInfo(life *lifecycle.Manager, tcpNode host.Host, peers []peer.ID, lockHash []byte, sender *p2p.Sender, builderEnabled bool, nickname string) {
+func wirePeerInfo(life *lifecycle.Manager, p2pNode host.Host, peers []peer.ID, lockHash []byte, sender *p2p.Sender, builderEnabled bool, nickname string) {
 	gitHash, _ := version.GitCommit()
-	peerInfo := peerinfo.New(tcpNode, peers, version.Version, lockHash, gitHash, sender.SendReceive, builderEnabled, nickname)
+	peerInfo := peerinfo.New(p2pNode, peers, version.Version, lockHash, gitHash, sender.SendReceive, builderEnabled, nickname)
 	life.RegisterStart(lifecycle.AsyncAppCtx, lifecycle.StartPeerInfo, lifecycle.HookFuncCtx(peerInfo.Run))
 }
 
-// wireP2P constructs the p2p tcp (libp2p) and udp (discv5) nodes and registers it with the life cycle manager.
+// wireP2P constructs the p2p tcp or udp (libp2p) and udp (discv5) nodes and registers it with the life cycle manager.
```
KaloyanTanev: There is no `discv5`, no? Probably nobody bothered changing it here.

```suggestion
// wireP2P constructs the p2p tcp or udp (libp2p) node and registers it with the life cycle manager.
```

cmd/relay/relay.go
```diff
@@ -218,26 +219,88 @@ func newENRHandler(ctx context.Context, tcpNode host.Host, p2pKey *k1.PrivateKey
...
+			tcpPort, udpPort int
+		)
+
+		for _, addr := range addrs {
+			// Sanity check
+			if len(addr.Protocols()) == 0 {
+				continue
+			}
+
+			if tcpNetAddr == nil && addr.Protocols()[1].Name == "tcp" {
```
KaloyanTanev: We need to check if there are at least 2 protocols here, as otherwise this may panic.

cmd/relay/relay.go
```diff
@@ -218,26 +219,88 @@ func newENRHandler(ctx context.Context, tcpNode host.Host, p2pKey *k1.PrivateKey
...
+				return nil, errors.New("conflicting IP addresses", z.Any("udp IP", udpNetAddr.IP), z.Any("tcp IP", tcpNetAddr.IP))
+			}
+
+			ip = tcpNetAddr.IP
+			tcpPort = tcpNetAddr.Port
+			udpPort = udpNetAddr.Port
+		} else if tcpNetAddr != nil {
+			ip = tcpNetAddr.IP
+			tcpPort = tcpNetAddr.Port
+			udpPort = tcpNetAddr.Port
```
KaloyanTanev: Why do we set both to the TCP port? It's a bit misleading.

cmd/relay/relay.go
```diff
@@ -218,26 +219,88 @@ func newENRHandler(ctx context.Context, tcpNode host.Host, p2pKey *k1.PrivateKey
...
+			ip = udpNetAddr.IP
+			tcpPort = udpNetAddr.Port
+			udpPort = udpNetAddr.Port
+		} else {
+			return nil, errors.New("no udp or tcp addresses provided")
 		}
 
-		// Build the ENR
-		r, err := enr.New(p2pKey, enr.WithIP(tcpAddr.IP), enr.WithTCP(tcpAddr.Port), enr.WithUDP(9999)) // Include invalid dummy UDP port so v0.13 can parse the ENR.
+		r, err := enr.New(p2pKey, enr.WithIP(ip), enr.WithTCP(tcpPort), enr.WithUDP(udpPort))
```
KaloyanTanev, line 298: Could this break some already existing connections?

docs/metrics.md
```diff
@@ -81,7 +81,7 @@ when storing metrics from multiple nodes or clusters in one Prometheus instance.
 | `core_validatorapi_request_total` | Counter | The total number of requests per content-type and endpoint | `endpoint, content_type` |
 | `core_validatorapi_vc_user_agent` | Gauge | Gauge with label set to user agent string of requests made by VC | `user_agent` |
 | `p2p_peer_connection_total` | Counter | Total number of libp2p connections per peer. | `peer` |
-| `p2p_peer_connection_types` | Gauge | Current number of libp2p connections by peer and type (`direct` or `relay`). Note that peers may have multiple connections. | `peer, type` |
+| `p2p_peer_connection_types` | Gauge | Current number of libp2p connections by peer, type (`direct` or `relay`), and protocol (`tcp`, `quic`, etc.). Note that peers may have multiple connections. | `peer, type, protocol` |
```
KaloyanTanev:
```suggestion
| `p2p_peer_connection_types` | Gauge | Current number of libp2p connections by peer, type (`direct` or `relay`), and protocol (`tcp`, `quic`). Note that peers may have multiple connections. | `peer, type, protocol` |
```

p2p/bootnode.go
```diff
@@ -221,20 +221,36 @@ func multiAddrFromENRStr(enrStr string) (ma.Multiaddr, error) {
...
-	addr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d/p2p/%s", ip.String(), port, id))
-	if err != nil {
-		return nil, errors.Wrap(err, "create multiaddr")
+	var addrs []ma.Multiaddr
+
+	port, ok := r.UDP()
+	if ok {
+		addr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/udp/%d/quic-v1/p2p/%s", ip.String(), port, id))
+		if err != nil {
+			return nil, errors.Wrap(err, "create multiaddr")
```
KaloyanTanev:
```suggestion
			return nil, errors.Wrap(err, "create quic-v1 multiaddr")
```

p2p/bootnode.go
```diff
@@ -221,20 +221,36 @@ func multiAddrFromENRStr(enrStr string) (ma.Multiaddr, error) {
...
+		}
+
+		addrs = append(addrs, addr)
+	}
+
+	port, ok = r.TCP()
+	if ok {
+		addr, err := ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d/p2p/%s", ip.String(), port, id))
+		if err != nil {
+			return nil, errors.Wrap(err, "create multiaddr")
```
KaloyanTanev:
```suggestion
			return nil, errors.Wrap(err, "create tcp multiaddr")
```

p2p/config.go
```diff
@@ -61,7 +100,7 @@ func (c Config) Multiaddrs() ([]ma.Multiaddr, error) {
 	return res, nil
 }
 
-func resolveListenAddr(addr string) (*net.TCPAddr, error) {
+func resolveListenTCPAddr(addr string) (*net.TCPAddr, error) {
 	tcpAddr, err := net.ResolveTCPAddr("tcp", addr)
 	if err != nil {
 		return nil, errors.Wrap(err, "resolve P2P bind addr")
```
KaloyanTanev: I wanted to suggest for the error below as well, but github doesn't allow me :/. If you agree with this one, can you change the other as well?

```suggestion
		return nil, errors.Wrap(err, "resolve P2P TCP bind addr")
```

p2p/config.go
```diff
@@ -74,8 +113,42 @@ func resolveListenAddr(addr string) (*net.TCPAddr, error) {
 	return tcpAddr, nil
 }
 
-// multiAddrFromIPPort returns a multiaddr composed of the provided ip (v4 or v6) and tcp port.
-func multiAddrFromIPPort(ip net.IP, port int) (ma.Multiaddr, error) {
+func resolveListenUDPAddr(addr string) (*net.UDPAddr, error) {
+	udpAddr, err := net.ResolveUDPAddr("udp", addr)
+	if err != nil {
+		return nil, errors.Wrap(err, "resolve P2P bind addr")
```
KaloyanTanev:
```suggestion
		return nil, errors.Wrap(err, "resolve P2P UDP bind addr")
```

p2p/config.go
```diff
@@ -74,8 +113,42 @@ func resolveListenAddr(addr string) (*net.TCPAddr, error) {
...
+	}
+
+	if udpAddr.IP == nil {
+		return nil, errors.New("p2p bind IP not specified")
```
KaloyanTanev:
```suggestion
		return nil, errors.New("p2p bind UDP IP not specified")
```

p2p/config.go
```diff
@@ -74,8 +113,42 @@ func resolveListenAddr(addr string) (*net.TCPAddr, error) {
...
+	}
+
+	return udpAddr, nil
+}
+
+// multiAddrFromIPUDPPort returns a multiaddr composed of the provided ip (v4 or v6) and udp port.
+func multiAddrFromIPUDPPort(ip net.IP, port int) (ma.Multiaddr, error) {
+	if ip.To4() == nil && ip.To16() == nil {
+		return nil, errors.New("invalid ip address")
```
KaloyanTanev:
```suggestion
		return nil, errors.New("invalid UDP IP address")
```

p2p/config.go
```diff
@@ -74,8 +113,42 @@ func resolveListenAddr(addr string) (*net.TCPAddr, error) {
...
+	var typ string
+	if ip.To4() != nil {
+		typ = "ip4"
+	} else if ip.To16() != nil {
+		typ = "ip6"
+	}
+
+	maddr, err := ma.NewMultiaddr(fmt.Sprintf("/%s/%s/udp/%d/quic-v1", typ, ip.String(), port))
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid quic multiaddr")
```
KaloyanTanev: Here it's really strictly `quic-v1` specific, so this might be slightly better...
```suggestion
		return nil, errors.Wrap(err, "invalid quic-v1 multiaddr")
```

p2p/config.go
```diff
@@ -74,8 +113,42 @@ func resolveListenAddr(addr string) (*net.TCPAddr, error) {
...
+	}
+
+	return maddr, nil
+}
+
+// multiAddrFromIPTCPPort returns a multiaddr composed of the provided ip (v4 or v6) and tcp port.
+func multiAddrFromIPTCPPort(ip net.IP, port int) (ma.Multiaddr, error) {
 	if ip.To4() == nil && ip.To16() == nil {
 		return nil, errors.New("invalid ip address")
```
KaloyanTanev:
```suggestion
		return nil, errors.New("invalid TCP IP address")
```

p2p/metrics.go
```diff
@@ -17,6 +17,12 @@ import (
 const (
 	addrTypeRelay  = "relay"
 	addrTypeDirect = "direct"
+
+	protocolTCP     = "tcp"
+	protocolQUIC    = "quic"
+	protocolUDP     = "udp"
```
KaloyanTanev: Why do we need both?

p2p/gater.go
```diff
@@ -41,11 +46,13 @@ type ConnGater struct {
 }
 
 // InterceptPeerDial does nothing.
-func (ConnGater) InterceptPeerDial(_ peer.ID) (allow bool) {
+func (ConnGater) InterceptPeerDial(p peer.ID) (allow bool) {
+	log.Debug(log.WithTopic(context.Background(), "p2p"), "Dialing peer", z.Str("peer", PeerName(p)))
```
KaloyanTanev: Do we need those debug logs for the 3 gater functions?

p2p/p2p.go
```diff
@@ -32,40 +35,97 @@ import (
...
+		externalAddrs []ma.Multiaddr
+		transport     libp2p.Option
+		err           error
+	)
 
-	externalAddrs, err := externalMultiAddrs(cfg)
-	if err != nil {
-		return nil, err
-	}
+	switch nodeType {
```
KaloyanTanev: Purely code simplicity suggestion here:
What do you think of always executing the code that is in the `case NodeTypeTCP` and then having an if clause for the code under `NodeTypeQUIC`? This way there will be less repetitiveness.

Something like
```go
var transportOpts []libp2p.Option

tcpAddrs, err = cfg.TCPMultiaddrs()
if len(addrs) == 0 { log.Info(ctx, "LibP2P not accepting incoming TCP connections since --p2p-tcp-addresses is empty")}
externalTCPAddrs, err = externalTCPMultiAddrs(cfg)
transportOpts.append(libp2p.Transport(tcp.NewTCPTransport, libP2POpts...))

// Use both QUIC and TCP transport when QUIC is enabled, as we are accommodating both connections in the cases where QUIC is enabled.
if nodeType == NodeTypeQUIC {
	... same as above but for UDP ...
	transportOpts.append(libp2p.Transport(quic.NewTransport))
}

transport = libp2p.ChainOptions(transportOpts...)
```

p2p/p2p.go
```diff
@@ -76,18 +136,17 @@ func NewTCPNode(ctx context.Context, cfg Config, key *k1.PrivateKey, connGater C
 		libp2p.AddrsFactory(func(internalAddrs []ma.Multiaddr) []ma.Multiaddr {
 			return filterAdvertisedAddrs(externalAddrs, internalAddrs, filterPrivateAddrs)
 		}),
-		libp2p.Transport(tcp.NewTCPTransport, tcpOpts...),
-		libp2p.SwarmOpts(swarm.WithDialRanker(swarm.NoDelayDialRanker)),
```
KaloyanTanev, line 80: Why is this one deleted?

p2p/p2p.go
```diff
@@ -137,7 +196,47 @@ func externalMultiAddrs(cfg Config) ([]ma.Multiaddr, error) {
...
+
+			resp = append(resp, maddr)
+		}
+	}
+
+	if cfg.ExternalHost != "" {
+		for _, port := range ports {
+			maddr, err := ma.NewMultiaddr(fmt.Sprintf("/dns/%s/udp/%d/quic-v1", cfg.ExternalHost, port))
+			if err != nil {
+				return nil, errors.Wrap(err, "invalid dns multiaddr")
```
KaloyanTanev: If you agree, same applies for the TCP function, github doesn't allow me to suggest there.
```suggestion
				return nil, errors.Wrap(err, "invalid dns quic-v1 multiaddr")
```

p2p/p2p.go
```diff
@@ -175,16 +274,21 @@ func multiAddrsViaRelay(relayPeer Peer, peerID peer.ID) ([]ma.Multiaddr, error)
...
+		encapAddr := transportAddr.Encapsulate(relayAddr)
+		if isQUICAddr(encapAddr) {
+			quicAddrs = append(quicAddrs, encapAddr)
+		} else {
+			otherAddrs = append(otherAddrs, encapAddr)
+		}
 	}
 
-	return resp, nil
+	return append(quicAddrs, otherAddrs...), nil
```
KaloyanTanev: Why do we want quic addresses to be first in this list?

p2p/p2p.go
```diff
@@ -261,6 +365,21 @@ func ForceDirectConnections(tcpNode host.Host, peerIDs []peer.ID) lifecycle.Hook
...
+func isQUICEnabled(h host.Host) bool {
+	if slices.ContainsFunc(h.Network().ListenAddresses(), isQUICAddr) {
+		return true
+	}
+
+	if addrs := h.Addrs(); len(addrs) > 0 {
+		if slices.ContainsFunc(addrs, isQUICAddr) {
+			return true
+		}
+	}
```
KaloyanTanev, line 354: How do those 2 differ? Can we add a comment or two for explanation of what we are checking exactly

p2p/p2p.go
```diff
@@ -274,15 +393,264 @@ func isDirectConnAvailable(conns []network.Conn) bool {
...
+				continue // skip self
+			}
+
+			if shouldSkipUpgradeAttempt(ctx, p) {
+				continue // skip this peer due to active backoff
+			}
+
+			conns := p2pNode.Network().ConnsToPeer(p)
+			if len(conns) == 0 {
+				log.Debug(ctx, "No connection to peer", z.Str("peer", PeerName(p)))
```
KaloyanTanev, line 441: Let's be a bit more descriptive, as this will most likely pop in the logs alongside with other similar logs if no peer connnections.
```suggestion
				log.Debug(ctx, "No peer connection for QUIC connection attempt", z.Str("peer", PeerName(p)))
```

p2p/p2p.go
```diff
@@ -274,15 +393,264 @@ func isDirectConnAvailable(conns []network.Conn) bool {
...
+			var quicAddrs []ma.Multiaddr
+
+			for _, addr := range p2pNode.Peerstore().Addrs(p) {
+				if isQUICAddr(addr) && !isRelayAddr(addr) {
+					quicAddrs = append(quicAddrs, addr)
+				}
+			}
+
+			if len(quicAddrs) == 0 {
+				log.Debug(ctx, "No knonw QUIC addresses to peer", z.Str("peer", PeerName(p)), z.Any("conns", conns))
```
KaloyanTanev, line 479:
```suggestion
				log.Debug(ctx, "No known QUIC addresses to peer", z.Str("peer", PeerName(p)), z.Any("conns", conns))
```

p2p/p2p.go
```diff
@@ -274,15 +393,264 @@ func isDirectConnAvailable(conns []network.Conn) bool {
...
+				continue // no known QUIC addresses
+			}
+
+			log.Debug(ctx, "Trying to upgrade to QUIC connection with peer", z.Str("peer", PeerName(p)))
+
+			// To maximize change of connectia via QUIC clear peerstore of TCP addresses
```
KaloyanTanev, line 485:
```suggestion
			// To maximize chance of connection via QUIC clear peerstore of TCP addresses
```

p2p/p2p.go
```diff
@@ -274,15 +393,264 @@ func isDirectConnAvailable(conns []network.Conn) bool {
...
+				}
+			}
+
+			// Attempt to connect over QUIC
+			// In case of QUIC failure due to infrastructure error (e.g. UDP port closed) the Connect call
+			// may fail or succeed. It can succeed if another peer calls Connect to this peer with a working
+			// TCP address which is misleading to us
+			// If it fails, we try to reconnect via any available address
+			// If it succeeds, we check whether we are actually connected via QUIC
+			err := p2pNode.Connect(network.WithForceDirectDial(ctx, ""), peer.AddrInfo{ID: p})
```
KaloyanTanev, line 511: Do you know how long this call takes? Essentially this will be the time for which we are not connected to the given peer, so we'd like it to be as fast as possible.

cmd/relay/relay.go
DiogoSantoss: Only because it was like that previously. Someone had a comment saying "// Include invalid dummy UDP port so v0.13 can parse the ENR."

DiogoSantoss: But looking back maybe we can keep the dummy value `9999`

cmd/relay/relay.go
DiogoSantoss: To prevent that I'll keep the previous dummy value of `9999`. This way nodes without the QUIC feature will produce the exact same ENRs as before.

p2p/gater.go
DiogoSantoss: Possibly not but they are quite useful when debugging. However, it also true that they produce a lot of noise during startup, specially `InterceptAddrDial` as it will logs every address through which we will dial the peer.

p2p/metrics.go
DiogoSantoss: Forgot to remove, thanks

p2p/p2p.go
DiogoSantoss: I can measure but it's certainly less than a second

p2p/p2p.go
DiogoSantoss: We can probably just check one of them. The first is checking where the host is listenning to QUIC addresses and the second is checking for advertises addresses. However, in practice if its listenning then it should also be advertising QUIC. I'll update to only check for advertised addresses

p2p/p2p.go
DiogoSantoss: The dial ranker `NoDelayDialRanker` dials every address at the same time as opposed to the default which dials first QUIC addresses. Since now we are interested in guaranteeing that peers talk via QUIC rather than TCP (when available), we delete this ranker and use the default.

p2p/gater.go
KaloyanTanev: Yeah, my thought was... In the case of someone enabling QUIC, but not being able to connect via UDP, these logs will pollute a lot.

p2p/p2p.go
KaloyanTanev: It's fine checking for both, just a comment might be useful "Sanity check - if it's listening to QUIC addresses, it should also be advertising QUIC"

p2p/p2p.go
KaloyanTanev: Okay. A second is quite a lot though. Imagine we have a proposal at that time, a peer might miss its participation. Can we measure it just so we are aware? You can probably add debug logs with call measurement. Ideally, we'd like to test it between nodes deployed on different machines.

Issue #3805, opened 2025-06-13
Batch duplicate logs


KaloyanTanev:
# 🎯 Problem to be solved
We log on a lot of places the same line multiple times for distinct validators in a cluster.
```
08:03:32.005 DEBG fetcher    Attester not selected for aggregation duty {"pubkey": "a5a_d41", "duty": "632367/aggregator"}
08:03:32.005 DEBG fetcher    Attester not selected for aggregation duty {"pubkey": "a63_35a", "duty": "632367/aggregator"}
...
08:03:32.002 INFO fetcher    Resolved attester aggregation duty       {"pubkey": "b00_1dd", "duty": "632367/aggregator"}
08:03:32.005 DEBG fetcher    Attester not selected for aggregation duty {"pubkey": "96c_049", "duty": "632367/aggregator"}
```

We can simplify our logging and make it more readable for certain log lines.

# 🛠️ Proposed solution
Log for multiple validators, instead of a single one.

```
08:03:32.005 DEBG fetcher    Attester not selected for aggregation duty {"pubkeys": "a63_35a, a5a_d41, ..., 96c_049", "duty": "632367/aggregator"}
08:03:32.002 INFO fetcher    Resolved attester aggregation duty       {"pubkey": "b00_1dd", "duty": "632367/aggregator"}
```

pinebit:
Just to clarify this needs to be done at the application level, not in zap/logging layer.

PR #3835, opened 2025-07-07
core: improved fetcher logging

pinebit:
Implemented batching logs produced by fetcher.

category: refactor
ticket: #3805

### DIFF
--- a/core/fetcher/fetcher.go
+++ b/core/fetcher/fetcher.go
@@ -172,15 +172,8 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot uint64, defSet cor
 
 // fetchAggregatorData fetches the attestation aggregation data.
 func (f *Fetcher) fetchAggregatorData(ctx context.Context, slot uint64, defSet core.DutyDefinitionSet) (core.UnsignedDataSet, error) {
-	var notSelectedCount, resolvedCount int
-	defer func() {
-		if notSelectedCount > 0 {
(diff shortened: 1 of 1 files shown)


core/fetcher/fetcher.go
```diff
@@ -172,6 +172,16 @@ func (f *Fetcher) fetchAttesterData(ctx context.Context, slot uint64, defSet cor
 
 // fetchAggregatorData fetches the attestation aggregation data.
 func (f *Fetcher) fetchAggregatorData(ctx context.Context, slot uint64, defSet core.DutyDefinitionSet) (core.UnsignedDataSet, error) {
+	var notSelectedCount, resolvedCount int
+	defer func() {
+		if notSelectedCount > 0 {
+			log.Debug(ctx, "Attester(s) not selected for aggregation duty", z.Int("total_pubkeys", len(defSet)), z.Int("not_selected", notSelectedCount))
```
KaloyanTanev: Can we not sacrifice data when we do this batch logging? Instead of keeping count, can we keep slice of the pubkeys and log that slice. For both selected and the non-selected.

Same applies for sync aggregations.

pinebit: I think it might be problematic for 1000+ pubkeys. Such the long log line can be skipped by the logger.
Initially I thought we can log at least N pubkeys, but then I realized it may not be useful.. then I ended up logging just the counts.

pinebit: Updated the logic

Issue #3798, opened 2025-06-10
Support deleting partial exits


anadi2311:
# 🎯 Problem to be solved
Node Operators can sometime push exits with wrong exit data and accidentally sign exits for pubkeys that they don't want to exit. Currently they have no way to remove a previously-submitted partial exit signature (“share”) via Charon CLI. Charon utilize the delete partial exit endpoint to facilitate this.

# 🛠️ Proposed solution
- [ ] Utilize the [delete partial exit endpoint](https://github.com/ObolNetwork/obol-api/blob/main/src/exit/exit.controller.ts#L230-L250) : exit/:lockHash/:shareIdx/:validatorPubkey so as user can delete 
  -  partial exit share for a specific pubkey that it has signed
  - partial exit shares for all pubkeys of a lock hash that it has signed
- [ ]  Show user friendly output that API returns
  - Deleted partial exit signature on 200
  - Share not found on 404
  - Unknown error: <message> on 500

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #3838, opened 2025-07-09
*: delete partial exits

KaloyanTanev:
Delete partial exits from Obol API.

N.B.: The endpoint is `/exp/exit/{lockHash}/{shareIndex}/{valPubkey}`, while it should be `/exp/partial_exits/{lockHash}/{shareIndex}/{valPubkey}`. I've talked with Launchpad to change it, but in the meantime I submit the PR as it is currently.

category: feature
ticket: #3798

### DIFF
--- a/app/obolapi/api.go
+++ b/app/obolapi/api.go
@@ -118,8 +118,6 @@ func httpGet(ctx context.Context, url *url.URL, headers map[string]string) (io.R
 		return nil, errors.Wrap(err, "new GET request with ctx")
 	}
 
-	req.Header.Add("Content-Type", "application/json")
-
 	for key, val := range headers {
 		req.Header.Set(key, val)
 	}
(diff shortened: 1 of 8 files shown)


app/obolapi/api.go
```diff
@@ -140,3 +140,30 @@ func httpGet(ctx context.Context, url *url.URL, headers map[string]string) (io.R
 
 	return res.Body, nil
 }
+
+func httpDelete(ctx context.Context, url *url.URL, headers map[string]string) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, url.String(), nil)
+	if err != nil {
+		return errors.Wrap(err, "new DELETE request with ctx")
+	}
+	req.Header.Add("Content-Type", "application/json")
```
pinebit: if no body, then why content-type?

KaloyanTanev: My bad, copy-pasted most of the logic from the get request. Will remove it from the get request as well.

pinebit (approved):
LGTM, except for SonarQube report (code duplication threshold).

Issue #3825, opened 2025-06-30
cluster-lock merge script


OisinKyne:
# 🎯 Problem to be solved

We want to make it easier to add new keys to an existing cluster. 

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

Ideal add validators is a longer workstream that may have an onchain component. This is an idea for solo or quite centralised setups. 

Create a `./cluster_merge.sh` script, that takes two full `cluster/` folders, and merges one into the other. This command should take the `distributed_validators` object from the second cluster, and inject it into the former, and should copy the keystores into the appropriate cluster1/node*/ subfolders. The command should tell the user to add `CHARON_NO_VERIFY=true` or `--no-verify=true` to their charon run command, along with warning the user to take caution around getting slashed. 

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

OisinKyne:
Worth noting that lido want fresh keys to be DKG'd, but existing keys to be split. We'll need to gracefully handle both cases in this script

PR #3847, opened 2025-07-14
*: added cluster merge scripts

pinebit:
Added `node_merge.sh` and `cluster_merge.sh` scripts. See `README.md` for essential details.

category: feature
ticket: #3825

### DIFF
--- a/scripts/cluster_merge.sh
+++ b/scripts/cluster_merge.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -euo pipefail
+
 #
 # This script merges cluster-lock and key files from a source cluster (node)
 # into a destination cluster (node).
@@ -63,7 +65,7 @@ if [ ! -w "$DST_CLUSTER_LOCK_FILE" ]; then
 fi
 
 # Check if jq is installed
-if ! command -v jq &> /dev/null; then
+if ! command -v jq &>/dev/null; then
     echo "Error: 'jq' is not installed. Please install it to use this script."
     echo "  On Debian/Ubuntu: sudo apt-get install jq"
     echo "  On macOS (Homebrew): brew install jq"
@@ -79,8 +81,8 @@ echo "Merging '$CLUSTER_LOCK_FILENAME' from '$SRC_CLUSTER_LOCK_FILE' into '$DST_
 # of the $src_data variable.
 MERGED_JSON=$(
     jq \
-    --slurpfile src_data "$SRC_CLUSTER_LOCK_FILE" \
-    '
+        --slurpfile src_data "$SRC_CLUSTER_LOCK_FILE" \
+        '
     .cluster_definition.num_validators = (.cluster_definition.num_validators + ($src_data[0].cluster_definition.num_validators)) |
     .cluster_definition.validators = (.cluster_definition.validators + $src_data[0].cluster_definition.validators) |
     .distributed_validators = (.distributed_validators + $src_data[0].distributed_validators)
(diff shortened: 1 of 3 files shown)


scripts/node_merge.sh
```diff
@@ -0,0 +1,218 @@
...
+    .cluster_definition.num_validators = (.cluster_definition.num_validators + ($src_data[0].cluster_definition.num_validators)) |
+    .cluster_definition.validators = (.cluster_definition.validators + $src_data[0].cluster_definition.validators) |
+    .distributed_validators = (.distributed_validators + $src_data[0].distributed_validators)
+    ' "$DST_CLUSTER_LOCK_FILE"
+)
+
+# Write the merged JSON back to the destination file
+# It's safer to write to a temporary file first and then move it,
+# to prevent data loss if the write operation is interrupted.
+TEMP_DST_JSON_FILE=$(mktemp "${DST_CLUSTER_LOCK_FILE}.tmp.XXXXXX")
```
KaloyanTanev: Add unix timestamp to the file name be on the safe side, if it fails multiple times.

```suggestion
TEMP_DST_JSON_FILE=$(mktemp "${DST_CLUSTER_LOCK_FILE}.${date +%s%N}.tmp")
```

scripts/node_merge.sh
```diff
@@ -0,0 +1,218 @@
...
+
+# Check if the temporary file was written successfully
+if [ $? -ne 0 ]; then
+    echo "Error: Failed to write to temporary file '$TEMP_DST_JSON_FILE'."
+    rm -f "$TEMP_DST_JSON_FILE" # Clean up temp file
+    exit 1
+fi
+
+# Move the temporary file to overwrite the original destination file
+mv "$TEMP_DST_JSON_FILE" "$DST_CLUSTER_LOCK_FILE"
```
KaloyanTanev, line 107: Isn't it better to do this after successful part 2 as well?
Or we can create a trap function to bring back the original if something in part 2 fails?

scripts/cluster_merge.sh
```diff
@@ -0,0 +1,91 @@
...
+fi
+if [ ! -x "$NODE_MERGE_SCRIPT" ]; then
+    echo "Error: '$NODE_MERGE_SCRIPT' is not executable. Please run 'chmod +x $NODE_MERGE_SCRIPT'."
+    exit 1
+fi
+
+echo "--- Starting cluster merge from '$SRC_CLUSTER_ROOT' to '$DST_CLUSTER_ROOT' ---"
+echo ""
+
+# Find all nodeX folders in the source cluster
```
KaloyanTanev, line 53: Shouldn't we do the same for destination folder?
Also ensure the same amount of nodes are found?

pinebit: None of these were done for greater simplification and the expected short lifecycle of the script (until we add `addvalidators` command probably).
Checking the same nodes is barely useful, because vast majority of clusters is either 4 or 7 nodes.. All this script is insecure and must be used by advanced users only.
If you want me to add some specific (critical?) checks, let's do that.

scripts/node_merge.sh
pinebit: Might be a subtle improvements, yes. In according to the docs, a user must create a backup before executing the script and then ensure the script executed properly before using the merged cluster. 
There are many cases unfortunately that can leave the merged cluster in a "broken" state if something went wrong.

scripts/node_merge.sh
```diff
@@ -0,0 +1,218 @@
...
+    .cluster_definition.num_validators = (.cluster_definition.num_validators + ($src_data[0].cluster_definition.num_validators)) |
+    .cluster_definition.validators = (.cluster_definition.validators + $src_data[0].cluster_definition.validators) |
+    .distributed_validators = (.distributed_validators + $src_data[0].distributed_validators)
+    ' "$DST_CLUSTER_LOCK_FILE"
+)
+
+# Write the merged JSON back to the destination file
+# It's safer to write to a temporary file first and then move it,
+# to prevent data loss if the write operation is interrupted.
+TEMP_DST_JSON_FILE=$(mktemp "${DST_CLUSTER_LOCK_FILE}.${date +%s%N}.tmp")
```
KaloyanTanev: typo from myself :)

```suggestion
TEMP_DST_JSON_FILE=$(mktemp "${DST_CLUSTER_LOCK_FILE}.$(date +%s%N).tmp")
```

KaloyanTanev (approved):

Issue #3672, opened 2025-05-02
Research spike: cluster mutability


pinebit:
# 🎯 Problem to be solved

For a long time we have been discussing "cluster mutability" as the most demanded feature. This should allow at least adding new validators to an active cluster and maybe more - changing node operators set, etc. This ticket is to summarize the requirements, prior research done by other engineers and suggest implementation options for the future brainstorming.

# 🛠️ Proposed solution
- [ ] Approved design doc: *link*
- [ ] Core team consensus on the proposed solution

This is just to research and design, no implementation or PoC is assumed to be done under this ticket, but small experiments proving ideas can surely be implemented.

# 🧪 Tests

No tests are expected.

pinebit:
I think we can close this umbrella ticket with 1.9.0 release.

PR #3857, opened 2025-07-26
*: `add-validators` command

pinebit:
Introduced new command `charon alpha add-validators` that adds any number of new validators to an existing cluster.

Designed usage:
```
charon alpha add-validators --num-validators=3 --src-dir=existing/node0 --dst-dir=combined/node0 --fee-recipient-addresses=0xAA.. --withdrawal-addresses=0xBB..
```

All Node Operators must run the same command simultaneously, same as for DKG.

The commands reads the existing cluster data from `--src-dir`, runs DKG protocol for `--num-validators`, appends the newly generated data to the existing cluster data and writes everything to `--dst-dir`. The `--dst-dir` gets all cluster files needed for charon and a VC to operate. The previous cluster must be stopped and `--src-dir` can be deleted after the ceremony.

category: feature
ticket: #3672


app/disk.go
```diff
@@ -61,3 +63,42 @@ func FileExists(path string) bool {
...
+
+	sourceInfo, err := os.Stat(src)
+	if err != nil {
+		return errors.Wrap(err, "failed to get source file info", z.Str("file", src))
+	}
+
+	err = os.Chmod(dst, sourceInfo.Mode().Perm())
+	if err != nil {
+		return errors.Wrap(err, "failed to set destination file permissions", z.Str("file", dst))
+	}
```
KaloyanTanev, line 125: Why are we changing the source's permissions here?

app/disk_internal_test.go
```diff
@@ -3,100 +3,27 @@
...
-	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/stretchr/testify/require"
 
-	"github.com/obolnetwork/charon/app/errors"
-	"github.com/obolnetwork/charon/core"
-	"github.com/obolnetwork/charon/testutil/beaconmock"
+	"github.com/obolnetwork/charon/app/eth1wrap"
 )
 
-func TestCalculateTrackerDelay(t *testing.T) {
```
KaloyanTanev, line 19: Why are we deleting those tests?

cmd/addvalidators.go
```diff
@@ -0,0 +1,232 @@
...
+	if config.DKG.Publish {
+		return errors.New("add-validators does not support --publish flag yet")
+	}
+
+	config.FeeRecipientAddrs, config.WithdrawalAddrs, err = validateAddresses(config.NumValidators, config.FeeRecipientAddrs, config.WithdrawalAddrs)
+
+	return err
+}
+
+func createDstClusterDir(dir string) error {
```
KaloyanTanev: Probably we can make that generic and add it to `disk.go`?

dkg/dkg.go
```diff
@@ -252,6 +310,28 @@ func Run(ctx context.Context, conf Config) (err error) {
...
+		def.NumValidators = totalValidators
+		def.ValidatorAddresses = append(def.ValidatorAddresses, conf.AppendConfig.ValidatorAddresses...)
+
+		// We have to reset creator and operators signatures, as they are not valid anymore
+		def.Creator = cluster.Creator{}
+		for i := range def.Operators {
+			def.Operators[i] = cluster.Operator{
+				ENR: def.Operators[i].ENR,
+			}
+		}
```
KaloyanTanev: IIRC, the `address`, `config_signature` and `enr_signature` are populated during creation of cluster_definition from within the launchpad.
Would it be possible if `addvalidators` is done from within the launchpad, to keep those? Probably this will involve something like... creating a new cluster-definition. I'm cautious not to break anything for any external integrations.
Or at least we can prompt the user if he really wants to overwrite those with empty ones.

app/disk.go
pinebit: We keep the original file permissions. e.g. if this is copying 0o444 file then we want the target file to have the same permissions (likely). Hence we adjust the permissions.

app/disk_internal_test.go
pinebit: This (along with some others) has been moved to `app_internal_test.go`, because the tested code resides in `app.go`. No tests were deleted anywhere here.

dkg/dkg.go
pinebit: The biggest pain is interactivity - these signatures are produced using EL account's key. Therefore this would require all NOPs to go over re-signing the definition. And this warrants a new UI in Launchpad.. Maybe this does not make sense in view of larger work - the overall cluster mutability? But it's up to Product Owner to decide.
cc @HananINouman

cmd/addvalidators.go
pinebit: Done

dkg/dkg.go
KaloyanTanev: Indeed, that was my idea of the workload it would involve. If I have to cast a vote, I'd say we should introduce this as well, of course based on the availability and resources that the Launchpad team has.

app/disk_internal_test.go
KaloyanTanev: Oh, right...

app/disk.go
KaloyanTanev: Ah, right, I misread 😅

dkg/dkg.go
KaloyanTanev: @pinebit In the meantime, do you think it would be useful to put some prompt to the user?
Something along the lines of "This command will nullify the creator's address and signature and each operator's address, config signature and enr signature. It does not affect Charon, but this information might be used in other external toolings. Do you wish to continue?"

pinebit: That's the point - I couldn't find any other external tooling other than LP. Probably after posting the updated definition/lock to API, it can rely on node signatures.. but making any charon interactive is not the best idea. For now I will add more informational messages at the end to highlight this.

EDIT: because this command does not modify the existing cluster data "in-place", it shall be sufficient to warn the user at the end, WDYT?

testutil/relay/relay.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Wait up to 5s for bootnode to become available.
+	ctx, cancel := context.WithTimeout(parentCtx, 5*time.Second)
+	defer cancel()
+
+	isUp := make(chan struct{})
+
+	go func() {
+		for ctx.Err() == nil {
+			// #nosec G107
```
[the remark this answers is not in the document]
pinebit: Used only in tests, won't fix.

cmd/addvalidators.go
```diff
@@ -0,0 +1,262 @@
...
+	// copying ENR private key file to the temporary directory for DKG.
+	if err := app.CreateNewEmptyDir(conf.DstDir); err != nil {
+		return err
+	}
+
+	if err := app.CopyFile(filepath.Join(conf.SrcDir, enrPrivateKeyFile), filepath.Join(conf.DstDir, enrPrivateKeyFile)); err != nil {
+		return err
+	}
+
+	log.Info(ctx, "Starting add-validators ceremony", z.Int("numValidators", conf.NumValidators), z.Str("lockHash", app.Hex7(lock.LockHash)))
```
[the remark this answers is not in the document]
pinebit: This shall be safe as we log this shorten hash in many other places.

dkg/exchanger.go
```diff
@@ -114,14 +112,19 @@ func (e *exchanger) exchange(ctx context.Context, sigType sigType, set core.ParS
...
 		select {
 		case sigDatas, ok := <-e.sigDatasChan:
 			if !ok {
 				return nil, errors.New("sigdata channel has been closed")
 			}
-			// We are done when we have ParSignedData of all the DVs from all each peer
-			return sigDatas, nil
+
+			if len(sigDatas) == expectedSigs {
+				// We are done when we have ParSignedData of all the DVs from all each peer
```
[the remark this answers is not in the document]
pinebit: This is wrong, `pushPsigs` is connected to `SubscribeThreshold` and therefore is triggered only when threshold number of peers have contributed.

PR #3901, opened 2025-08-12
cmd: add hostname verification

DiogoSantoss:
Add external hostname verification to prevent users from passing `--p2p-external-hostname` with empty value followed by another flag.

category: feature
ticket: none

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -5,12 +5,12 @@ package cmd
 import (
 	"context"
 	"net/url"
-	"strings"
 	"time"
 
 	libp2plog "github.com/ipfs/go-log/v2"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"golang.org/x/net/idna"
 
 	"github.com/obolnetwork/charon/app"
 	"github.com/obolnetwork/charon/app/errors"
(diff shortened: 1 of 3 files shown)


KaloyanTanev (approved):

cmd/run.go
```diff
@@ -174,6 +175,10 @@ func bindP2PFlags(cmd *cobra.Command, config *p2p.Config) {
 	cmd.Flags().BoolVar(&config.DisableReuseport, "p2p-disable-reuseport", false, "Disables TCP port reuse for outgoing libp2p connections.")
 
 	wrapPreRunE(cmd, func(cmd *cobra.Command, args []string) error { //nolint:revive // keep args variable name for clarity
+		if _, err := idna.Lookup.ToASCII(config.ExternalHost); err != nil {
```
pinebit, line 178: I see we have config verification in bindRunFlags/wrapPreRunE, line 108. Shouldn't this check go there?

pinebit (approved):

cmd/run.go
DiogoSantoss: In line 108, we are validating `bindRunFlags`, here we are validating `bindP2PFlags`

Issue #3877, opened 2025-08-05
Track SSE events `block_gossip` and `block`


KaloyanTanev:
# 🎯 Problem to be solved
We currently track head updates, which occur once new head is chosen (>=2/3 slot has passed = a new head was chosen), however, we would like to also track how fast the beacon node received the initial block proposed (<= 1/3 slot has passed).

# 🛠️ Proposed solution
- [ ] Subscribe to `block_gossip` and `block` SSE events
- [ ] Log in the same way the received block

# 🧪 Tests
- [ ] Manually tested on local compose simnet

PR #3938, opened 2025-08-29
app/sse: add block and block gossip event

DiogoSantoss:
Track `block` and `block_gossip` event.
Fix event timestamp measurement.
Readjust head buckets.

category: feature
ticket: #3877

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -14,6 +14,8 @@ when storing metrics from multiple nodes or clusters in one Prometheus instance.
 | Name | Type | Help | Labels |
 |---|---|---|---|
 | `app_beacon_node_peers` | Gauge | Gauge set to the peer count of the upstream beacon node |  |
+| `app_beacon_node_sse_block` | Histogram | Block imported into fork choice delay, supplied by beacon node`s SSE endpoint. Values between 0s and 4s for Ethereum mainnet are considered safe | `addr` |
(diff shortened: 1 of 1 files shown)


app/sse/metrics.go
```diff
@@ -21,7 +21,7 @@ var (
 		Subsystem: "beacon_node",
 		Name:      "sse_head_delay",
 		Help:      "Delay in seconds between slot start and head update, supplied by beacon node's SSE endpoint. Values between 8s and 12s for Ethereum mainnet are considered safe.",
-		Buckets:   []float64{4, 6, 8, 10, 12, 16, 20},
+		Buckets:   []float64{2, 4, 6, 8, 10, 12},
```
pinebit, line 24: Shall we make these buckets matching the `sseBlock*` buckets?

app/sse/metrics.go
```diff
@@ -31,4 +31,20 @@ var (
 		Help:      "Chain reorg depth, supplied by beacon node's SSE endpoint",
 		Buckets:   []float64{1, 2, 4, 6, 8, 16},
 	}, []string{"addr"})
+
+	sseBlockGossipHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
pinebit, line 35: Need to update `TestMetricsReference`

app/sse/metrics.go
DiogoSantoss: I made `sseBlock*` buckets more fine grained since we want to analyze where in the 4s window they fall but if `head` also requires this I'm open to change it

KaloyanTanev (approved):

PR #3977, opened 2025-09-15
*: pedersen dkg

pinebit:
* Added `pedersen` DKG algorithm using kyber library.
* Added `alpha edit recreate-private-keys` command.
* Added `alpha edit add-operators` command.
* Added `alpha edit remove-operators` command.
* Moved `add-validators` to the `alpha edit` section.
* Ongoing code refactoring.

category: feature
ticket: #3672


dkg/pedersen/dkg.go
```diff
@@ -0,0 +1,203 @@
...
+		NewNodes:  nodes,
+		Threshold: config.Threshold,
+		Auth:      drandbls.NewSchemeOnG2(kbls.NewBLS12381Suite()),
+	}
+
+	for range numVals {
+		// TODO: This phaser implementation is odd, it makes pauses between rounds,
+		// relying on all other nodes to complete the round in that time.
+		// Unfortunately, kyber does not expose any fine-grained control over the protocol.
+		// A better solution would be a signal-based phaser that relies on Board progress.
```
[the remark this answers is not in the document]
pinebit: There will be another iteration with a different Phaser implementation.

dkg/pedersen/config.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/obolnetwork/charon/app/errors"
+	"github.com/obolnetwork/charon/cluster"
+)
+
+const (
+	phaseDuration = time.Second
+)
+
+var DefaultSuite = kbls.NewBLS12381Suite().G1().(kdkg.Suite)
+
```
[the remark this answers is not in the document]
pinebit: This was following the library examples, the suggested change does not add any value.

dkg/pedersen/reshare.go
```diff
@@ -0,0 +1,281 @@
...
+		if idx.PeerIdx == thisNodeIndex {
+			thisIsOldNode = true
+		}
+
+		newNodes = slices.DeleteFunc(newNodes, func(n kdkg.Node) bool {
+			return n.Index == kdkg.Index(idx.PeerIdx)
+		})
+	}
+
+	nonce := sha256.Sum256(config.SessionID)
```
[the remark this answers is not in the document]
pinebit: The nonce here is not a true none in the sense of cryptography. It serves as a unique session identifier.

dkg/protocol_removeoperators.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (*removeOperatorsProtocol) GetPeers(lock *cluster.Lock) ([]p2p.Peer, error) {
+	return lock.Peers()
+}
+
+func (p *removeOperatorsProtocol) PostInit(ctx context.Context, pctx *ProtocolContext) error {
+	newN := len(pctx.PeerIDs) - len(p.oldENRs)
+	newT := newN - (newN-1)/3
+
```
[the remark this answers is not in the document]
pinebit: Comments will be added in a different way

cmd/test.go
```diff
@@ -66,8 +66,8 @@ type testConfig struct {
 func newTestCmd(cmds ...*cobra.Command) *cobra.Command {
 	root := &cobra.Command{
 		Use:   "test",
-		Short: "Test subcommands provide test suite to evaluate current cluster setup",
-		Long:  `Test subcommands provide test suite to evaluate current cluster setup. The full validator stack can be tested - charon peers, consensus layer, validator client, MEV. Current machine's infra can be examined as well.`,
+		Short: "Subcommands provide test suites to evaluate a cluster setup",
+		Long:  `Subcommands provide test suites to evaluate a cluster setup. The full validator stack can be tested - charon peers, consensus layer, validator client, MEV. Current machine's infra can be examined as well.`,
```
DiogoSantoss:
```suggestion
		Long:  "Subcommands provide test suites to evaluate a cluster setup. The full validator stack can be tested - charon peers, consensus layer, validator client, MEV. Current machine's infra can be examined as well.",
```

cmd/addoperators.go
```diff
@@ -0,0 +1,161 @@
...
+			return runFunc(cmd.Context(), config, dkgConfig)
+		},
+	}
+
+	cmd.Flags().StringVar(&config.PrivateKeyPath, "private-key-file", ".charon/charon-enr-private-key", "The path to the charon enr private key file. ")
+	cmd.Flags().StringVar(&config.LockFilePath, "lock-file", ".charon/cluster-lock.json", "The path to the cluster lock file defining the distributed validator cluster.")
+	cmd.Flags().StringVar(&config.ValidatorKeysDir, "validator-keys-dir", ".charon/validator_keys", "Path to the directory containing the validator private key share files and passwords.")
+	cmd.Flags().StringVar(&config.OutputDir, "output-dir", "distributed_validator", "The destination folder for the new cluster data. Must be empty.")
+	cmd.Flags().StringSliceVar(&config.NewENRs, "new-operator-enrs", nil, "Comma-separated list of the new operators to be added (Charon ENR addresses).")
+	cmd.Flags().IntVar(&config.NewThreshold, "new-threshold", 0, "Optional override of the new threshold required for signature reconstruction. Defaults to ceil(n*2/3) if zero. Warning, non-default values decrease security. All operators must use the same value.")
```
aadomn: It seems to me that it does not make sense to allow for a greater threshold here: even though the threshold changes from `t` to `t'` with `t' > t`, nothing prevents `t` operators from the old set to collude and recover the secret.
As example, assume a `(3,4)` cluster where one wants to add `2` operators, as currently implemented we will now have a `(4,6)` but actually the old operator set have shares that allow reconstruction for `t=3`, which contradicts with the cluster definition.
Therefore, I think we should remove this `new-threshold` flag for this specific command and I'd suggest to let the threshold unchanged instead.

What do you think?

cmd/addoperators.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	newN := len(lock.Operators) + len(config.NewENRs)
+	newT := newN - (newN-1)/3
+
+	if config.NewThreshold != 0 {
+		if config.NewThreshold >= newN || config.NewThreshold < newT {
+			return errors.New("new-threshold is invalid", z.Int("recommendedThreshold", newT))
+		}
+	}
```
aadomn: See my comment on `new-threshold` flag for `addoperators` command

dkg/pedersen/dkg.go
```diff
@@ -0,0 +1,202 @@
...
+	if err := board.BroadcastNodePubKey(ctx, pkBytes); err != nil {
+		return nil, errors.Wrap(err, "broadcast node pubkey")
+	}
+
+	nodes, _, err := makeNodes(ctx, config, board)
+	if err != nil {
+		return nil, errors.Wrap(err, "make nodes")
+	}
+
+	nonce := sha256.Sum256(config.SessionID)
```
aadomn: Are we sure it is fine to derive the nonce this way? I could not find how is `SessionID` generated.
[According to the kyber library](https://github.com/drand/kyber/blob/94dae51d79b4b0c2d2a9b9cc382b864cf3537783/share/dkg/dkg.go#L103-L107), it should be treated as a cryptographic nonce and [their internal function to generate nonces relies on randomness](https://github.com/drand/kyber/blob/94dae51d79b4b0c2d2a9b9cc382b864cf3537783/share/dkg/dkg.go#L1120-L1131).

dkg/pedersen/reshare.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		kyberPubShares = append(kyberPubShares, kyberPubshare)
+	}
+
+	pubPoly, err := kshare.RecoverPubPoly(suite.G1(), kyberPubShares, threshold, len(keyShare.PublicShares))
+	if err != nil {
+		return nil, errors.Wrap(err, "recover pubpoly")
+	}
+
+	_, commits := pubPoly.Info()
```
aadomn: Could we use `restoreCommits` here to avoid code duplication?

dkg/pedersen/utils.go
```diff
@@ -0,0 +1,62 @@
...
+
+	pubKey, err := tbls.SecretToPublicKey(privKey)
+	if err != nil {
+		return tbls.PrivateKey{}, tbls.PublicKey{}, errors.Wrap(err, "derive pubkey from privkey")
+	}
+
+	return privKey, pubKey, nil
+}
+
+func keyShareToValidatorPubKey(result *kdkg.DistKeyShare, suite kdkg.Suite) (tbls.PublicKey, error) {
```
aadomn: Picky comment on the function name: I find `keyShareToValidatorPubKey` a bit confusing as we recover `ValidatorPublicKey` from commits instead of key shares. I think `commitsToValidatorPubKey` or `distKeyShareToValidatorPubKey` would be clearer.

cmd/addvalidators.go
```diff
@@ -53,13 +55,17 @@ func newAddValidatorsCmd(runFunc func(context.Context, addValidatorsConfig) erro
...
 
+			libp2plog.SetPrimaryCore(log.LoggerCore()) // Set libp2p logger to use charon logger
+
 			return runFunc(cmd.Context(), config)
 		},
 	}
 
 	// Bind `add-validator` flags.
 	cmd.Flags().IntVar(&config.NumValidators, "num-validators", 1, "The number of new validators to generate and add to the existing cluster.")
-	cmd.Flags().StringVar(&config.DataDir, "data-dir", ".charon", "The source charon folder with existing cluster data (lock, validator_keys, etc.).")
```
KaloyanTanev, line 62: Shall we deprecate it first? If`DataDir` is specified we can throw a warning and bind `PrivateKeyPath`, `LockFilePath`, `ValidatorKeysDir` to the values that we expect them to be inside the `DataDir`.

cmd/edit.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func newEditCmd(cmds ...*cobra.Command) *cobra.Command {
+	root := &cobra.Command{
+		Use:   "edit",
+		Short: "Subcommands provide functionality to modify existing cluster configurations",
+		Long:  "Subcommands allow users to modify existing distributed validator cluster configurations, such as adding and removing operators.",
+	}
+
+	root.AddCommand(cmds...)
```
KaloyanTanev, line 16: A nitpick, not really related to this line - can we prepend `edit` in the files that are used for this command? I feel it's a bit easier to find them in the explorer as there are a lot of files already in the cmd dir. Similar to how we have it for the test command and exit command. I see there is no consensus if we should use underscore or not between category and subcategory.

cmd/removeoperators.go
```diff
@@ -0,0 +1,158 @@
...
+			return slices.Contains(config.ParticipatingENRs, o.ENR)
+		})
+		if !ok {
+			return errors.New("participating-operator-enrs contains a non-existing operator")
+		}
+	}
+
+	f := len(lock.Operators) - lock.Threshold
+	if len(config.RemovingENRs) > f && len(config.ParticipatingENRs) == 0 {
+		return errors.New("participating-operator-enrs is required when removing more than F operators")
```
KaloyanTanev: Not sure if people will understand what `F operators` is?

```suggestion
		return errors.New("participating-operator-enrs is required when after the removal, the remaining amount of operators is below the current threshold")
```

dkg/pedersen/board.go
```diff
@@ -0,0 +1,357 @@
...
+
+// BroadcastValidatorPubKeyShare broadcasts a public key and collects the public keys of all peers.
+func (b *Board) BroadcastValidatorPubKeyShare(ctx context.Context, share []byte) error {
+	msg := &pb.ValidatorPubKeyShareMessage{
+		SessionId:      b.config.SessionID,
+		PublicKeyShare: share,
+	}
+
+	if err := b.broadcastP2P(ctx, valPubKeyShareMsg, msg); err != nil {
+		log.Error(b.logCtx, "Failed to broadcast val pubkey share", err)
```
KaloyanTanev, line 161: Why we don't return error here?

dkg/pedersen/board.go
```diff
@@ -0,0 +1,357 @@
...
+		PubKey: protoMsg.GetPublicKey(),
+	}
+	if protoMsg.GetShares() != nil {
+		ppk.PubKeyShares = protoMsg.GetShares().GetPublicKeyShares()
+	}
+
+	select {
+	case <-ctx.Done():
+		log.Error(ctx, "Dropping node pubkey, context done", nil, z.Str("from", peerID.String()))
+	default:
```
KaloyanTanev: Why do we need a default here?

dkg/pedersen/board.go
```diff
@@ -0,0 +1,357 @@
...
+
+	vpks := ValidatorPubKeyShare{
+		PeerID:          peerID,
+		ValidatorPubKey: protoMsg.GetPublicKeyShare(),
+	}
+
+	select {
+	case b.valPubKeySharesCh <- vpks:
+	case <-ctx.Done():
+		log.Error(b.logCtx, "Dropping validator pubkey share, context done", nil, z.Str("from", peerID.String()))
```
KaloyanTanev, line 282: If context is done, do we still have to return `nil, true, nil`?

dkg/pedersen/config.go
```diff
@@ -0,0 +1,70 @@
...
+	ThisPeerID    peer.ID
+	PeerMap       map[peer.ID]cluster.NodeIdx
+	Threshold     int
+	SessionID     []byte
+	PhaseDuration time.Duration
+	Reshare       *ReshareConfig
+}
+
+type ReshareConfig struct {
+	TotalShares  int
```
KaloyanTanev, line 30: I'm a bit confused what this one is? I thought it's `len(NewPeers) + len(OldPeers)` but doesn't seem like it?

dkg/pedersen/dkg.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/obolnetwork/charon/dkg/share"
+	"github.com/obolnetwork/charon/tbls"
+)
+
+// High-level protocol overview:
+// 1. Each node generates an ephemeral BLS key pair and broadcasts the public key, see board.BroadcastNodePubKey.
+// 2. Each node receives all other nodes' public keys before anybody starts the DKG protocol.
+// 3. Each node runs the Pedersen DKG protocol, which may run a few exchanges: Deals, Responses, and Justifications.
+//    - A single Pedersen DKG run produces a single shared secret key among all nodes.
+//    - To run the protocol for multiple validator keys, the protocol is run multiple times (there is room for optimization here).
```
KaloyanTanev, line 25: Is there still room for optimization?

dkg/pedersen/reshare.go
```diff
@@ -0,0 +1,284 @@
...
+		OldNodes:     oldNodes,
+		Threshold:    config.Reshare.NewThreshold,
+		OldThreshold: config.Threshold,
+		FastSync:     true,
+		Auth:         drandbls.NewSchemeOnG2(kbls.NewBLS12381Suite()),
+		Log:          newLogger(log.WithTopic(ctx, "pedersen")),
+	}
+
+	log.Info(ctx, "Starting pedersen reshare...",
+		z.Int("oldN", len(oldNodes)), z.Int("newN", len(newNodes)),
```
KaloyanTanev: I think abbreviations here might only cause confusions later on.
```suggestion
		z.Int("oldNodes", len(oldNodes)), z.Int("newNodes", len(newNodes)),
```

dkg/pedersen/reshare.go
```diff
@@ -0,0 +1,284 @@
...
+		z.Int("oldT", config.Threshold), z.Int("newT", config.Reshare.NewThreshold),
```
KaloyanTanev:
```suggestion
		z.Int("oldThreshold", config.Threshold), z.Int("newThreshold", config.Reshare.NewThreshold),
```

dkg/pedersen/dkg.go
pinebit: The problem here is that the nonce shall be the same for all operators in this ceremony. If we generate a random nonce, then it will require an exchange among all participants upfront, which adds friction. Under the hood, the nonce is needed to distinguish network packets in case more than one session is running at a time.
In the present code, `config.SessionID` is just the full cluster hash, because it does not make sense to run multiple dkg commands for the same cluster. It is know to all participants upfront which is convenient.

cmd/addoperators.go
pinebit: > the old operator set have shares that allow reconstruction

Correct, but the newly generated shares may be set for new `t`.
You are probably referring to the fact that the old set can persist the old shares and use them to double sign, etc.
From this perspective - yes, what you are saying makes total sense.

cmd/addoperators.go
pinebit: Removed

cmd/addoperators.go
pinebit: Removed the flag and the related logic. The threshold will remain the same for this command.

dkg/pedersen/dkg.go
aadomn: I see. Anyway after a successful execution the cluster hash will change so it's mainly a concern for successive failed executions I guess.

dkg/dkg.go
```diff
@@ -267,6 +269,11 @@ func Run(ctx context.Context, conf Config) (err error) {
 	// register bcast callbacks: node signatures and public shares
 	nodeSigCaster := newNodeSigBcast(peers, nodeIdx, caster)
 
+	// register pedersen protocol messages
```
KaloyanTanev, line 272: Shall we do those even if it's not Pedersen, but Frost?

dkg/nodesigs.go
```diff
@@ -117,21 +124,34 @@ func (n *nodeSigBcast) broadcastCallback(ctx context.Context, _ peer.ID, _ strin
...
-		return errors.Wrap(err, "get peer public key")
+		return errors.Wrap(err, "lock hash wait")
 	}
 
-	lockHash, err := n.lockHash(ctx)
+	if bytes.Equal(lockHash, noneData) {
+		n.setSig(noneData, msgPeerIdx)
+
+		return nil
+	}
```
KaloyanTanev, line 150: Why do we set the signature to none if there is no lock hash?

dkg/nodesigs.go
```diff
@@ -117,21 +124,34 @@ func (n *nodeSigBcast) broadcastCallback(ctx context.Context, _ peer.ID, _ strin
 		return errors.New("invalid node sig type")
 	}
 
-	sig := nodeSig.GetSignature()
 	msgPeerIdx := int(nodeSig.GetPeerIndex())
 
+	sig := nodeSig.GetSignature()
+	if bytes.Equal(sig, noneData) {
+		// For certain protocols we allow exchanging nil signatures.
+		n.setSig(sig, msgPeerIdx)
```
KaloyanTanev: nitpick, just for readability
```suggestion
		n.setSig(noneData, msgPeerIdx)
```

dkg/pedersen/dkg.go
pinebit: As discussed internally, the new approach has been introduced.

dkg/protocol.go
```diff
@@ -0,0 +1,290 @@
...
+	var lock cluster.Lock
+	if err := json.Unmarshal(b, &lock); err != nil {
+		return nil, errors.Wrap(err, "unmarshal cluster-lock.json", z.Str("path", lockFilePath))
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	eth1Cl := eth1wrap.NewDefaultEthClientRunner(executionEngineAddr)
+	go eth1Cl.Run(ctx)
```
KaloyanTanev, line 204: Can we make sure we stop this one after we finish with this function? Or alternatively, `LoadAndVerifyClusterLock` with the protocol context as an argument where we start eth1 client as well.

cmd/addvalidators.go
pinebit: Since it is under alpha, we have privilege to make breaking changes. We have already moved this to `edit` space, which is a breaking change.
Also it's always better to avoid any "implicit behavior".. imo. Let's me know if you have a strong opinion here then we can discuss it 1:1.

dkg/protocol.go
pinebit: This stops when `ctx` gets canceled. And in the commands it gets canceled when the command is done.

cmd/addvalidators.go
KaloyanTanev: Right, good point.

dkg/nodesigs.go
pinebit: Because we need to indicate a response is received for this slot... I did not want to refactor the entire component, and therefore introduced this workaround. The whole DKG package deserves a huge overhaul..

dkg/dkg.go
pinebit: It does not hurt, but makes less conditions => better readability.

dkg/pedersen/board.go
pinebit: This is a weird leftover, removed.

dkg/pedersen/board.go
pinebit: Inspired by the existing bcast code using forkjoin. A protocol is in charge of handing logical errors on its own. So either that or returning an error shall not impact the high level flow. With this approach we can quickly see all peers to which we failed to send.

cmd/edit.go
pinebit: Good idea, done as suggested.

dkg/pedersen/dkg.go
pinebit: Yes, we can send batches in parallel like Frost does. But I don't think it worth the effort as of now.

dkg/pedersen/config.go
pinebit: Total shares means total validators, not peers.

Issue #3978, opened 2025-09-16
Update minimum version warning log for fusaka fork


KaloyanTanev:
# 🎯 Problem to be solved
Previously we've logged when the CL is not electra-ready, which was quite useful. Would be nice if we change pre-fusaka to have the same functionality.

# 🛠️ Proposed solution
- [ ] Update the expected CL versions for Fusaka, once they are available

PR #4000, opened 2025-09-29
app/eth2wrap: bump min version

DiogoSantoss:
Bump CL min version to be fusaka-ready

category: feature 
ticket: #3978

### DIFF
--- a/app/eth2wrap/version.go
+++ b/app/eth2wrap/version.go
@@ -12,12 +12,12 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


KaloyanTanev (approved):
I'm not sure how parsing rcs works with our checks?

DiogoSantoss:
Parsing works fine but we use `Compare` from `app/version` which does not compare rc version, only major/minor/patch

Issue #3992, opened 2025-09-25
`charon deposit sign` && `charon deposit fetch`


OisinKyne:
# 🎯 Problem to be solved

We want to sign deposits for new validators, but `add validators` is high friction

# 🛠️ Proposed solution

Like we have `charon exit sign` and `charon exit fetch`. Lets make `charon deposit sign` and `charon deposit fetch`. 

It should only work if the cluster-lock withdrawal address for the pubkey is 0x0 (we can debate this, but i think it keeps things safer)

You should be able to specify `--public-keys` from the cluster-lock, (and maybe `--validator-number` to indicate the nth validator in a lock file? tbd)

You should be able to specify --compounding, which makes the withdrawal address 0x02, otherwise 0x01

You must specify one withdrawal address or as many withdrawal addresses as your `--public-keys` array

## To be figured out 

You may specify a fee recipient address, and the program also signs a builder-registration, we can do this in a second ticket, as it may complicate execution, and would require a cluster-lock modification. For now the user would have to keep the existing fee-recipient or maybe do the add-validator flow instead. 

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

# 👐 Additional acceptance criteria

The Obol API needs to support this behavior

# ❌ Out of Scope

Modifying fee-recipients of normal validators.

PR #4032, opened 2025-10-15
cmd: add API partial deposit flow

KaloyanTanev:
1. `charon deposit sign`: sign partial deposit -> publish to the api
2. `charon deposit fetch`: fetch deposits from the api -> aggregate signatures -> save full deposit locally

category: feature
ticket: #3992

### DIFF
--- a/cmd/depositsign.go
+++ b/cmd/depositsign.go
@@ -49,6 +49,7 @@ func newDepositSignCmd(runFunc func(context.Context, depositSignConfig) error) *
 	wrapPreRunE(cmd, func(cmd *cobra.Command, _ []string) error {
 		mustMarkFlagRequired(cmd, "withdrawal-addresses")
 		mustMarkFlagRequired(cmd, "validator-public-keys")
+
 		return nil
 	})
 
--- a/app/obolapi/deposit_test.go
+++ b/app/obolapi/deposit_test.go
@@ -1,0 +1,90 @@
+// Copyright © 2022-2025 Obol Labs Inc. Licensed under the terms of a Business Source License 1.1
+
+package obolapi_test
+
+import (
+	"context"
+	"encoding/hex"
+	"math/rand"
+	"net/http/httptest"
+	"testing"
+
+	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/stretchr/testify/require"
+
+	"github.com/obolnetwork/charon/app/obolapi"
+	"github.com/obolnetwork/charon/cluster"
+	"github.com/obolnetwork/charon/eth2util"
+	"github.com/obolnetwork/charon/eth2util/deposit"
+	"github.com/obolnetwork/charon/tbls"
+	"github.com/obolnetwork/charon/testutil/beaconmock"
+	"github.com/obolnetwork/charon/testutil/obolapimock"
+)
+
+func TestAPIDeposit(t *testing.T) {
+	kn := 4
+
+	beaconMock, err := beaconmock.New(t.Context())
+	require.NoError(t, err)
+
+	defer func() {
+		require.NoError(t, beaconMock.Close())
+	}()
+
+	mockEth2Cl := eth2Client(t, context.Background(), beaconMock.Address())
+
+	handler, addLockFiles := obolapimock.MockServer(false, mockEth2Cl)
+	srv := httptest.NewServer(handler)
+
+	defer srv.Close()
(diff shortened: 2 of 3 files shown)


app/obolapi/deposit.go
```diff
@@ -0,0 +1,158 @@
...
+	defer respBody.Close()
+
+	var dr FullDepositResponse
+	if err := json.NewDecoder(respBody).Decode(&dr); err != nil {
+		return []eth2p0.DepositData{}, errors.Wrap(err, "json unmarshal error")
+	}
+
+	withdrawalCredentialsBytes, err := hex.DecodeString(strings.TrimPrefix(dr.WithdrawalCredentials, "0x"))
+	if err != nil {
+		return []eth2p0.DepositData{}, errors.Wrap(err, "validator pubkey to bytes")
```
pinebit: Should this message say "withdrawal credentials" and not "validator pubkey"?

app/obolapi/deposit.go
```diff
@@ -0,0 +1,158 @@
...
+		return errors.Wrap(err, "http Obol API POST request")
+	}
+
+	return nil
+}
+
+// GetFullDeposit gets the full deposit message for a given validator public key, lock hash and share index.
+// It respects the timeout specified in the Client instance.
+func (c Client) GetFullDeposit(ctx context.Context, valPubkey string, lockHash []byte) ([]eth2p0.DepositData, error) {
+	valPubkeyBytes, err := from0x(valPubkey, 48) // public key is 48 bytes long
```
pinebit: Maybe it can be `len(eth2p0.BLSPubKey)` instead of magic number?

cmd/depositfetch.go
```diff
@@ -0,0 +1,110 @@
...
+
+const defaultDepositDataDir = ".charon/deposit-data-<TIMESTAMP>"
+
+func newDepositFetchCmd(runFunc func(context.Context, depositFetchConfig) error) *cobra.Command {
+	var config depositFetchConfig
+
+	cmd := &cobra.Command{
+		Use:   "fetch",
+		Short: "Fetch a full deposit.",
+		Long:  "Fetch a full validator deposit messages using a remote API.",
```
pinebit: Not sure this wording is correct or not.. Oisin can verify if you share the --help output internally.

KaloyanTanev: Yeah, I will ask him to go over the CLI once the rest looks OK

app/obolapi/deposit.go
```diff
@@ -0,0 +1,172 @@
...
+		return []eth2p0.DepositData{}, errors.Wrap(err, "json unmarshal error")
+	}
+
+	withdrawalCredentialsBytes, err := hex.DecodeString(strings.TrimPrefix(dr.WithdrawalCredentials, "0x"))
+	if err != nil {
+		return []eth2p0.DepositData{}, errors.Wrap(err, "withdrawal credentials to bytes")
+	}
+
+	// do aggregation
+	fullDeposits := []eth2p0.DepositData{}
```
DiogoSantoss, line 116: Can we do `make([]eth2p0.DepositData,dr.Amounts)`? and then later `fullDeposits[i]=...` ?

cmd/depositsign.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return errors.Wrap(err, "load keystore")
+	}
+
+	shares, err := keystore.KeysharesToValidatorPubkey(cl, valKeys)
+	if err != nil {
+		return errors.Wrap(err, "match local validator key shares with their counterparty in cluster lock")
+	}
+
+	pubkeys := []eth2p0.BLSPubKey{}
```
DiogoSantoss, line 108: `make([]eth2p0.BLSPubKey,config.ValidatorPublicKeys)` and then `pubkeys[i]=blsPK`

app/obolapi/deposit.go
KaloyanTanev: What do we gain from filling the slice with empty objects which we later overwrite over creating an empty one to which we later on amend new objects?

I personally never liked empty objects in slices as it's really error prone and useful only in specific cases.

DiogoSantoss: It should be faster since you are allocating all memory upfront and remove slice growth logic at runtime. I believe it also helps GC not having to deal with multiple arrays i.e. when the slice reaches its capacity a new one is allocated

Issue #3882, opened 2025-08-05
Errors revamp


KaloyanTanev:
# 🎯 Problem to be solved
Some errors in Charon can be a bit more descriptive. Even adding some tips on what might have been wrong.

# 🛠️ Proposed solution
Go over charon errors (lengthy task, might consider using some tool/AI to extract us all errors we wrap/create) and for those that is possible write something more descriptive.

DiogoSantoss:
We should add global CLI argument verification to catch cases where users accidentally pass flags as values to other flags.
E.g.: `charon run --p2p-external-hostname --p2p-tcp-address 0.0.0.0:3600`

pinebit:
Logging revamp has been done: https://github.com/ObolNetwork/charon/pull/4038
Next is to revamp `errors.New/Wrap` entries...

PR #4038, opened 2025-10-21
*: logging revamp

pinebit:
Charon logging revamp.

category: refactor
ticket: #3882

### DIFF
--- a/app/eth2wrap/synthproposer.go
+++ b/app/eth2wrap/synthproposer.go
@@ -220,7 +220,7 @@ func (h *synthWrapper) SubmitBlindedProposal(ctx context.Context, opts *eth2api.
 // SubmitProposal submits a beacon block or swallows it if marked as synthetic.
 func (h *synthWrapper) SubmitProposal(ctx context.Context, opts *eth2api.SubmitProposalOpts) error {
 	if IsSyntheticProposal(opts.Proposal) {
-		log.Debug(ctx, "Ignoring synthetic beacon block (already on-chain)")
+		log.Debug(ctx, "Ignoring synthetic beacon block proposal")
 		return nil
 	}
 
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -377,7 +377,7 @@ func (s *Scheduler) resolveAttDuties(ctx context.Context, slot core.Slot, vals v
 
 		pubkey, ok := vals.PubKeyFromIndex(attDuty.ValidatorIndex)
 		if !ok {
-			log.Warn(ctx, "Received attester duty for unknown validator. Ignoring", nil, z.U64("vidx", uint64(attDuty.ValidatorIndex)), z.U64("slot", slot.Slot))
+			log.Warn(ctx, "Received attester duty for unknown validator. The validator may not be part of this cluster. Ignoring. If edit command was recently executed, Charon and/or VC might have not been restarted or not read the new keys properly", nil, z.U64("vidx", uint64(attDuty.ValidatorIndex)), z.U64("slot", slot.Slot))
 			continue
 		}
 


app/eth2wrap/synthproposer.go
```diff
@@ -210,7 +210,7 @@ func fraction(transactions []bellatrix.Transaction) []bellatrix.Transaction {
 // SubmitBlindedProposal submits a blinded beacon block proposal or swallows it if marked as synthetic.
 func (h *synthWrapper) SubmitBlindedProposal(ctx context.Context, opts *eth2api.SubmitBlindedProposalOpts) error {
 	if IsSyntheticBlindedBlock(opts.Proposal) {
-		log.Debug(ctx, "Synthetic blinded beacon proposal swallowed")
+		log.Debug(ctx, "Ignoring synthetic blinded beacon proposal (already on-chain)")
```
KaloyanTanev: Probably this `(already on-chain)` shouldn't be there? It's not really on-chain, because it's a "fake" block

app/peerinfo/peerinfo.go
```diff
@@ -231,7 +231,7 @@ func (p *PeerInfo) sendOnce(ctx context.Context, now time.Time) {
 				peerCompatibleGauge.WithLabelValues(name).Set(0) // Set to false
 
 				// Log as error since user action required
-				log.Error(ctx, "Invalid peer version", err,
+				log.Error(ctx, "Peer is running an incompatible Charon version. Please coordinate with the operator to upgrade or downgrade to a compatible version", err,
```
KaloyanTanev: Not sure if we should use "operator" in Charon itself. I know we call entities that run Charon node operators, but I see that more as a business term, rather than code. IDK... Probably we can stick with peer?

```suggestion
				log.Error(ctx, "Peer is running an incompatible Charon version. Please coordinate with the peer to upgrade or downgrade to a compatible version", err,
```

core/scheduler/scheduler.go
```diff
@@ -377,7 +377,7 @@ func (s *Scheduler) resolveAttDuties(ctx context.Context, slot core.Slot, vals v
 
 		pubkey, ok := vals.PubKeyFromIndex(attDuty.ValidatorIndex)
 		if !ok {
-			log.Warn(ctx, "Ignoring unexpected attester duty", nil, z.U64("vidx", uint64(attDuty.ValidatorIndex)), z.U64("slot", slot.Slot))
+			log.Warn(ctx, "Received attester duty for unknown validator. The validator may not be part of this cluster. Ignoring", nil, z.U64("vidx", uint64(attDuty.ValidatorIndex)), z.U64("slot", slot.Slot))
```
KaloyanTanev: Could it also be some mismatch after an edit command?

core/tracking.go
```diff
@@ -65,7 +65,7 @@ func WithTracking(tracker Tracker, inclusion InclusionChecker) WireOption {
 		w.BroadcasterBroadcast = func(ctx context.Context, duty Duty, set SignedDataSet) error {
 			// Check inclusion even if we fail to broadcast, since peers may succeed.
 			if err := inclusion.Submitted(duty, set); err != nil {
-				log.Error(ctx, "Bug: failed to submit duty to inclusion checker", err)
+				log.Error(ctx, "Bug: Failed to submit duty to inclusion checker. This is an internal error that should not occur. Please report this issue with logs", err)
```
KaloyanTanev: Shall we stick to the new rhetoric of `Internal error`?


```suggestion
				log.Error(ctx, "Internal error: Failed to submit duty to inclusion checker. This indicates a tracking bug that should be reported", err)
```

p2p/ping.go
```diff
@@ -203,14 +203,14 @@ func newPingLogger(p2pNode host.Host, p peer.ID) func(context.Context, ping.Resu
...
-		log.Warn(ctx, "Peer not connected", nil, opts...)
+		log.Warn(ctx, "Peer not connected. Check network connectivity, firewall rules, and ensure peer addresses are reachable", nil, opts...)
 	}
 }
 
 func resolveBackoffMsgs(ctx context.Context, p2pNode host.Host, p peer.ID) map[string]string {
 	net, ok := p2pNode.Network().(*swarm.Swarm)
 	if !ok {
-		log.Error(ctx, "Not a swarm network", nil)
+		log.Error(ctx, "Internal error: Failed to access libp2p swarm network. This indicates a critical p2p subsystem failure", nil)
```
KaloyanTanev:
```suggestion
		log.Error(ctx, "Internal error: Failed to access libp2p swarm network. This indicates a critical p2p subsystem failure that should be reported", nil)
```

p2p/ping.go
```diff
@@ -224,7 +224,7 @@ func resolveBackoffMsgs(ctx context.Context, p2pNode host.Host, p peer.ID) map[s
 
 	msgs, ok := dialErrMsgs(err)
 	if !ok { // Some other error
-		log.Warn(ctx, "Peer dial failing", nil, z.Str("peer", PeerName(p)), z.Str("error", err.Error()))
+		log.Warn(ctx, "Peer dial failed. Verify peer is online and network path is clear", nil, z.Str("peer", PeerName(p)), z.Str("error", err.Error()))
```
KaloyanTanev: What does "network path is clear" mean 😅?

p2p/sender.go
```diff
@@ -103,7 +103,7 @@ func (s *Sender) addResult(ctx context.Context, peerID peer.ID, err error) {
 	if val, ok := s.states.Load(peerID); ok {
 		state, ok = val.(*peerState)
 		if !ok {
-			log.Warn(ctx, "Type assertion peer state failing", err, z.Str("peer", PeerName(peerID)))
+			log.Warn(ctx, "Internal error: Type assertion failed for peer state. This indicates a bug in peer state management", err, z.Str("peer", PeerName(peerID)))
```
KaloyanTanev:
```suggestion
			log.Warn(ctx, "Internal error: Type assertion failed for peer state. This indicates a bug in peer state management and should be reported", err, z.Str("peer", PeerName(peerID)))
```

p2p/ping.go
pinebit: I think it's meant that the patch has no extra characters, etc. But I fixed this.

app/eth2wrap/synthproposer.go
```diff
@@ -220,7 +220,7 @@ func (h *synthWrapper) SubmitBlindedProposal(ctx context.Context, opts *eth2api.
 // SubmitProposal submits a beacon block or swallows it if marked as synthetic.
 func (h *synthWrapper) SubmitProposal(ctx context.Context, opts *eth2api.SubmitProposalOpts) error {
 	if IsSyntheticProposal(opts.Proposal) {
-		log.Debug(ctx, "Synthetic beacon block swallowed")
+		log.Debug(ctx, "Ignoring synthetic beacon block (already on-chain)")
```
KaloyanTanev:
```suggestion
		log.Debug(ctx, "Ignoring synthetic beacon block")
```

core/scheduler/scheduler.go
KaloyanTanev: If so, then we can have something along the lines of

```
Received attester duty for unknown validator. The validator may not be part of this cluster. Ignoring. If edit command was recently executed, Charon and/or VC might have not been restarted or not read the new keys properly
```

pinebit: Sounds great! Added.

app/eth2wrap/synthproposer.go
pinebit: Made it `Ignoring synthetic beacon block proposal`

PR #4039, opened 2025-10-21
*: various optimizations

pinebit:
## Summary

Optimize codebase with Go 1.25 features and fix resource leaks across p2p, core, and app packages.

**Key improvements:**
- Fixed 6+ timer leaks by replacing `time.After` with `time.NewTimer/NewTicker` and proper cleanup
- Eliminated 3 busy-wait patterns (CPU spinning loops) with efficient channel-based notifications
- Reduced memory allocations using `clear()` builtin for map reuse in hot paths
- Simplified initialization with `sync.OnceFunc` replacing `sync.Once`
- Used builtin `min()`/`max()` functions for cleaner code
- Improved goroutine scheduling with buffered channels
- Migrated eth1 client to lifecycle management for proper startup/shutdown

**Packages affected:**
- `p2p/`: Timer leaks in bootnode ENR resolution and relay router
- `core/scheduler`: Replaced busy-wait loop with channel notifications for epoch resolution
- `core/aggsigdb`: Eliminated `runtime.Gosched()` busy-wait with buffered channel
- `core/parsigex`: Fixed timer leak in signature exchange await logic
- `app/retry`: Proper timer cleanup on context cancellation
- `app/eth2wrap`: Removed `time.Sleep` busy-wait, added `clear()` for map reuse
- `app/forkjoin`: Buffered results channel for better concurrency

All changes maintain test compatibility and pass race detection.

category: refactor
ticket: none

### DIFF
--- a/app/forkjoin/forkjoin.go
+++ b/app/forkjoin/forkjoin.go
@@ -162,7 +162,7 @@ func New[I, O any](rootCtx context.Context, work Work[I, O], opts ...Option) (Fo
 		wg         sync.WaitGroup
 		zero       O
 		input      = make(chan I, options.inputBuf)
-		results    = make(chan Result[I, O], options.workers)
+		results    = make(chan Result[I, O])
 		dropOutput = make(chan struct{})
 		done       = make(chan struct{})
 	)


app/eth2wrap/httpwrap.go
```diff
@@ -96,22 +96,11 @@ func (h *httpAdapter) Validators(ctx context.Context, opts *api.ValidatorsOpts)
 	*api.Response[map[eth2p0.ValidatorIndex]*apiv1.Validator],
 	error,
 ) {
-	var cancel func()
-
-	reqCtx := ctx
-
 	maxValAmt := max(len(opts.PubKeys), len(opts.Indices))
+	timeout := time.Duration(max(200, maxValAmt)*50) * time.Millisecond
```
KaloyanTanev: Is this the same as before?
AFAICT before we did stick with default timeout if `maxValAmt > 200`.
Now we change the timeout irregardless of the amount. Was the default timeout 10s (the result you get now if maxValAmt <= 200)?

app/forkjoin/forkjoin.go
```diff
@@ -162,7 +162,7 @@ func New[I, O any](rootCtx context.Context, work Work[I, O], opts ...Option) (Fo
 		wg         sync.WaitGroup
 		zero       O
 		input      = make(chan I, options.inputBuf)
-		results    = make(chan Result[I, O])
+		results    = make(chan Result[I, O], options.workers)
```
KaloyanTanev: Why do we make this channel buffered? Are we sure it won't break some things down the line, as it changes its behaviour

app/app.go
```diff
@@ -175,7 +175,7 @@ func Run(ctx context.Context, conf Config) (err error) {
 	}
 
 	eth1Cl := eth1wrap.NewDefaultEthClientRunner(conf.ExecutionEngineAddr)
-	go eth1Cl.Run(ctx)
+	life.RegisterStart(lifecycle.AsyncAppCtx, lifecycle.StartEth1Client, lifecycle.HookFuncCtx(eth1Cl.Run))
```
KaloyanTanev, line 178: nice :)

core/aggsigdb/memory_v2.go
```diff
@@ -4,7 +4,6 @@ package aggsigdb
```
KaloyanTanev, line 4: TBH we can revisit this memorydbV2.
Now that we have metrics of the types of feature flags operators have turned on, we can see how many have this V2 DB. I personally do not have the perspective how much it improves Charon.
I can do it as part of looking into the aggregation failures we see recently, as they are related to the DB.

app/forkjoin/forkjoin.go
pinebit: As it was explained by Claude, there was extreme goroutines contention here and this buffer resolves it. We have good test coverage for this object, so I am sure it did not break anything. However, it's hard to estimate the effect.

app/eth2wrap/httpwrap.go
pinebit: You are very right! Fixed.

KaloyanTanev (approved):

app/forkjoin/forkjoin.go
pinebit: As discussed 1:1, reverted this change.

PR #4042, opened 2025-10-21
core/validatorapi: add eth2wrap logic to proxy

DiogoSantoss:
Proxied requests from the VC to the BN didn't have our `eth2wrap` logic (such as sending to multiple BNs or retrying to fallback nodes) neither did they send our additional headers from `CHARON_BEACON_NODE_HEADERS`.
This changes our proxy to use the `eth2wrap` client to proxy to (possibly) multiple addresses instead of directly proxing to a single address.

category: feature
ticket: none

### DIFF
--- a/app/eth2wrap/genwrap/genwrap.go
+++ b/app/eth2wrap/genwrap/genwrap.go
@@ -244,9 +244,8 @@ func run(_ context.Context) error {
 	pkgs, err := packages.Load(
 		&packages.Config{
 			Mode: packages.NeedSyntax | packages.NeedTypesInfo | packages.NeedFiles | packages.NeedCompiledGoFiles | packages.NeedTypes,
-			Dir:  "/home/diogo/dev/work/go-eth2-client",
 		},
-		".",
+		"github.com/attestantio/go-eth2-client",
 	)
 	if err != nil {
 		return errors.Wrap(err, "load package")
--- a/core/validatorapi/router_internal_test.go
+++ b/core/validatorapi/router_internal_test.go
@@ -1463,7 +1463,7 @@ func TestBeaconCommitteeSelections(t *testing.T) {
 	handler.BeaconCommitteeSelectionsFunc = func(ctx context.Context, opts *eth2api.BeaconCommitteeSelectionsOpts) (*eth2api.Response[[]*eth2v1.BeaconCommitteeSelection], error) {
 		return wrapResponse(opts.Selections), nil
 	}
-	// Use beacon node testing handler to handled unmocked endpoints called by eth2http client
+	// Use beacon node testing handler to handle unmocked endpoints called by eth2http client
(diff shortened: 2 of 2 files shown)


app/eth2wrap/eth2wrap.go
```diff
@@ -228,6 +230,42 @@ func isSyncingError(err error) bool {
 	return strings.Contains(msg, "syncing")
 }
 
+// isBadGateway returns true when the error indicates a connectivity or upstream gateway issue.
+func isBadGateway(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	for current := err; current != nil; current = errors.Unwrap(current) {
```
KaloyanTanev: Good! A bit less readable as you pointed out, but better than matching random strings.

app/eth2wrap/httpwrap.go
```diff
@@ -127,3 +135,136 @@ func (h *httpAdapter) Domain(ctx context.Context, domainType eth2p0.DomainType,
...
+	rp.ErrorLog = stdlog.New(io.Discard, "", 0)
+
+	// Apply custom director for auth and headers
+	defaultDirector := rp.Director
+	rp.Director = func(outReq *http.Request) {
+		defaultDirector(outReq)
+
+		// Basic auth if present in target URL
+		if targetURL.User != nil {
+			password, _ := targetURL.User.Password()
```
KaloyanTanev: Do we skip error check here?

DiogoSantoss:
> Looks like the build is broken

Need to create a release on `go-eth2-client` to update this `go.mod`

app/eth2wrap/eth2wrap.go
```diff
@@ -205,8 +207,8 @@ func provide[O any](ctx context.Context, clients []Client, fallbacks []Client,
 	}
 	output, err := runForkJoin(clients)
 
-	// Call fallback nodes when request to beacon node timeout or if it's syncing
-	if err != nil && len(fallbacks) != 0 && (isTimeoutError(err) || isSyncingError(err)) {
+	// Call fallback nodes when request to beacon node timeout or if it's syncing or beacon node is unreachable
+	if err != nil && len(fallbacks) != 0 && (isTimeoutError(err) || isSyncingError(err) || isBadGateway(err)) {
```
KaloyanTanev, line 211: This is not related to the actual proxying logic, right?

KaloyanTanev (approved):

app/eth2wrap/eth2wrap.go
DiogoSantoss: It's not. While testing with offline primaries and relying only on fallback BNs I noticed that the errors where neither syncing nor timeout so it wouldn't trigger a retry. Those cover lower level "fail to connect" errors

PR #4043, opened 2025-10-22
*: replace Split in loops with more efficient SplitSeq

wyrapeseed:
strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.

category: refactor
ticket: none


pinebit:
Thank you @wyrapeseed for the contribution!

wyrapeseed:
Thanks for your review. 

If there's anything need I to do, please feel free to let me know.

PR #4044, opened 2025-10-22
app: errors revamp

pinebit:
app package errors revamp with claude.

category: refactor
ticket: #3882

### DIFF
--- a/app/obolapi/exit.go
+++ b/app/obolapi/exit.go
@@ -110,7 +110,7 @@ func (c Client) PostPartialExits(ctx context.Context, lockHash []byte, shareInde
 		Signature:                  signature,
 	})
 	if err != nil {
-		return errors.Wrap(err, "marshal json")
+		return errors.Wrap(err, "marshal PartialExitRequest to JSON")
 	}
 
 	ctx, cancel := context.WithTimeout(ctx, c.reqTimeout)
@@ -170,7 +170,7 @@ func (c Client) GetFullExit(ctx context.Context, valPubkey string, lockHash []by
 
 	var er FullExitResponse
 	if err := json.NewDecoder(respBody).Decode(&er); err != nil {
-		return ExitBlob{}, errors.Wrap(err, "unmarshal json")
+		return ExitBlob{}, errors.Wrap(err, "unmarshal FullExitResponse from JSON")
 	}
 
 	// do aggregation
--- a/app/utils.go
+++ b/app/utils.go
@@ -223,7 +223,7 @@ func CompareDirectories(originalDir, extractedDir string) error {
 			// Check if directory exists in extracted content
 			extractedInfo, err := os.Stat(extractedPath)
 			if err != nil {
-				return errors.Wrap(err, "stat extracted directory", z.Str("path", relPath))
+				return errors.Wrap(err, "directory should exist in extracted content", z.Str("path", relPath))
 			}
 
 			if !extractedInfo.IsDir() {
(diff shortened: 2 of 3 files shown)


app/eth2wrap/valcache.go
DiogoSantoss, line 1: nit: here I actually prefer the "cannot be" instead of "is", similar to "nickname cannot be..."

KaloyanTanev: I agree. If we change it here, we have to do in the other PRs as well. There was one with a lot of those changes.

app/obolapi/exit.go
```diff
@@ -110,7 +110,7 @@ func (c Client) PostPartialExits(ctx context.Context, lockHash []byte, shareInde
 		Signature:                  signature,
 	})
 	if err != nil {
-		return errors.Wrap(err, "json marshal error")
+		return errors.Wrap(err, "marshal json")
```
KaloyanTanev:
```suggestion
		return errors.Wrap(err, "marshal PartialExitRequest to JSON")
```

app/obolapi/exit.go
```diff
@@ -170,7 +170,7 @@ func (c Client) GetFullExit(ctx context.Context, valPubkey string, lockHash []by
 
 	var er FullExitResponse
 	if err := json.NewDecoder(respBody).Decode(&er); err != nil {
-		return ExitBlob{}, errors.Wrap(err, "json unmarshal error")
+		return ExitBlob{}, errors.Wrap(err, "unmarshal json")
```
KaloyanTanev:
```suggestion
		return ExitBlob{}, errors.Wrap(err, "unmarshal FullExitResponse from JSON")
```

app/utils.go
```diff
@@ -223,11 +223,11 @@ func CompareDirectories(originalDir, extractedDir string) error {
 			// Check if directory exists in extracted content
 			extractedInfo, err := os.Stat(extractedPath)
 			if err != nil {
-				return errors.Wrap(err, "directory should exist in extracted content", z.Str("path", relPath))
+				return errors.Wrap(err, "stat extracted directory", z.Str("path", relPath))
```
KaloyanTanev: What we had before was better IMO, this change is less understandable.
```suggestion
				return errors.Wrap(err, "directory should exist in extracted content", z.Str("path", relPath))
```

app/utils.go
```diff
@@ -236,11 +236,11 @@ func CompareDirectories(originalDir, extractedDir string) error {
 		// Compare file contents
 		extractedInfo, err := os.Stat(extractedPath)
 		if err != nil {
-			return errors.Wrap(err, "file should exist in extracted content", z.Str("path", relPath))
+			return errors.Wrap(err, "stat extracted file", z.Str("path", relPath))
```
KaloyanTanev: What we had before was better IMO, this change is less understandable.
```suggestion
			return errors.Wrap(err, "file should exist in extracted content", z.Str("path", relPath))
```

app/utils_test.go
```diff
@@ -132,7 +132,7 @@ func TestCompareDirectories(t *testing.T) {
 				return dir1, dir2
 			},
 			expectError: true,
-			errorMsg:    "file should exist in extracted content",
+			errorMsg:    "stat extracted file",
```
KaloyanTanev:
```suggestion
			errorMsg:    "file should exist in extracted content",
```

app/utils_test.go
```diff
@@ -175,7 +175,7 @@ func TestCompareDirectories(t *testing.T) {
 				return dir1, dir2
 			},
 			expectError: true,
-			errorMsg:    "directory should exist in extracted content",
+			errorMsg:    "stat extracted directory",
```
KaloyanTanev:
```suggestion
			errorMsg:    "directory should exist in extracted content",
```

app/eth2wrap/valcache.go
pinebit: Claude did that because there's no universal consensus, but the trend is to describe what IS wrong (state), not what action failed. It mentioned many sources like "Effective Go", but I totally agree we can set our own rules here. And since Kalo is the boss, I am modifying as suggested.

PR #4046, opened 2025-10-22
core: errors revamp

pinebit:
core package errors revamp with claude.

category: refactor
ticket: #3882

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -459,7 +459,7 @@ func compare[I any, V comparable, C any](ctx context.Context, d Definition[I, V,
 			return inputValueSource, nil
 		case inputValueSource = <-compareValue:
 		case <-timerChan:
-			log.Warn(ctx, "", errors.New("timeout waiting for comparison data"))
+			log.Warn(ctx, "", errors.New("timeout waiting for local data, used for comparing with leader's proposed data"))
 			return inputValueSource, errTimeout
 		}
 	}
--- a/core/tracker/inclusion.go
+++ b/core/tracker/inclusion.go
@@ -170,7 +170,7 @@ func (i *inclusionCore) Submitted(duty core.Duty, pubkey core.PubKey, data core.
 
 		defer func() {
 			if r := recover(); r != nil {
-				err = errors.New("determine if proposal was synthetic",
+				err = errors.New("could not determine if proposal was synthetic or not",
 					z.Str("proposal", fmt.Sprintf("%+v", block)),
 					z.Bool("blinded", block.Blinded),
 				)
--- a/core/tracker/tracker.go
+++ b/core/tracker/tracker.go
@@ -287,7 +287,7 @@ func analyseDutyFailed(duty core.Duty, allEvents map[core.Duty][]event, msgRootC
 	case zero:
 		failedErr = errors.New("no events for duty") // This should never happen.
 	default:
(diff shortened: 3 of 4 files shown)


core/qbft/qbft.go
```diff
@@ -459,7 +459,7 @@ func compare[I any, V comparable, C any](ctx context.Context, d Definition[I, V,
 			return inputValueSource, nil
 		case inputValueSource = <-compareValue:
 		case <-timerChan:
-			log.Warn(ctx, "", errors.New("timeout on waiting for data used for comparing local and leader proposed data"))
+			log.Warn(ctx, "", errors.New("timeout waiting for comparison data"))
```
KaloyanTanev: Given the "novelty" and that it's not as straightforward, I'd rather be a bit more explicit here
```suggestion
			log.Warn(ctx, "", errors.New("timeout waiting for local data, used for comparing with leader's proposed data"))
```

core/tracker/inclusion.go
```diff
@@ -170,7 +170,7 @@ func (i *inclusionCore) Submitted(duty core.Duty, pubkey core.PubKey, data core.
 
 		defer func() {
 			if r := recover(); r != nil {
-				err = errors.New("could not determine if proposal was synthetic or not",
+				err = errors.New("determine if proposal was synthetic",
```
KaloyanTanev: I'd keep it as it is. I usually try to follow what we have in the docs, which is to use negative words (cannot/failed/etc.) in new errors and not to use them in wrapping.
```suggestion
				err = errors.New("could not determine if proposal was synthetic or not",
```

core/tracker/tracker.go
```diff
@@ -287,7 +287,7 @@ func analyseDutyFailed(duty core.Duty, allEvents map[core.Duty][]event, msgRootC
 	case zero:
 		failedErr = errors.New("no events for duty") // This should never happen.
 	default:
-		failedErr = errors.New("duty failed at step", z.Int("step", int(failedStep))) // This should never happen.
+		failedErr = errors.New("duty at step", z.Int("step", int(failedStep))) // This should never happen.
```
KaloyanTanev:
```suggestion
		failedErr = errors.New("duty failed at step", z.Int("step", int(failedStep))) // This should never happen.
```

core/tracker/tracker_internal_test.go
```diff
@@ -1393,7 +1393,7 @@ func TestSubmittedProposals(t *testing.T) {
 				Version: eth2spec.DataVersionDeneb,
 			},
 		}, 1*time.Millisecond)
-		require.ErrorContains(t, err, "could not determine if proposal was synthetic or not")
+		require.ErrorContains(t, err, "determine if proposal was synthetic")
```
KaloyanTanev: To be approved only if the prod code change is approved.
```suggestion
		require.ErrorContains(t, err, "could not determine if proposal was synthetic or not")
```

core/tracker/tracker_internal_test.go
```diff
@@ -1402,6 +1402,6 @@ func TestSubmittedProposals(t *testing.T) {
...
```
KaloyanTanev: To be approved only if the prod code change is approved.
```suggestion
		require.ErrorContains(t, err, "could not determine if proposal was synthetic or not")
```

PR #4047, opened 2025-10-22
*: errors revamp

pinebit:
eth2util, p2p, tbls packages errors revamp with claude.

category: refactor
ticket: #3882

### DIFF
--- a/p2p/config.go
+++ b/p2p/config.go
@@ -129,7 +129,7 @@ func resolveListenUDPAddr(addr string) (*net.UDPAddr, error) {
 // multiAddrFromIPUDPPort returns a multiaddr composed of the provided ip (v4 or v6) and udp port.
 func multiAddrFromIPUDPPort(ip net.IP, port int) (ma.Multiaddr, error) {
 	if ip.To4() == nil && ip.To16() == nil {
-		return nil, errors.New("invalid IP address format")
+		return nil, errors.New("invalid IP address format for UDP")
 	}
 
 	var typ string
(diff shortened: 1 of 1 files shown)


DiogoSantoss (approved):

p2p/config.go
```diff
@@ -129,7 +129,7 @@ func resolveListenUDPAddr(addr string) (*net.UDPAddr, error) {
 // multiAddrFromIPUDPPort returns a multiaddr composed of the provided ip (v4 or v6) and udp port.
 func multiAddrFromIPUDPPort(ip net.IP, port int) (ma.Multiaddr, error) {
 	if ip.To4() == nil && ip.To16() == nil {
-		return nil, errors.New("invalid UDP ip address")
+		return nil, errors.New("invalid IP address format")
```
KaloyanTanev: I'd rather be explicit it's regarding UDP
```suggestion
		return nil, errors.New("UDP invalid IP address format")
```

p2p/config.go
```diff
@@ -150,7 +150,7 @@ func multiAddrFromIPUDPPort(ip net.IP, port int) (ma.Multiaddr, error) {
 // multiAddrFromIPTCPPort returns a multiaddr composed of the provided ip (v4 or v6) and tcp port.
 func multiAddrFromIPTCPPort(ip net.IP, port int) (ma.Multiaddr, error) {
 	if ip.To4() == nil && ip.To16() == nil {
-		return nil, errors.New("invalid TCP ip address")
+		return nil, errors.New("invalid IP address format")
```
KaloyanTanev: I'd rather be explicit it's regarding TCP
```suggestion
		return nil, errors.New("TCP invalid IP address format")
```

p2p/config.go
pinebit: Okay, but errors must start with lowercase, I will change as suggested

PR #4051, opened 2025-10-24
core/parsigdb: various fixes

pinebit:
Various fixes for parsigdb logic:

- One-time threshold notifications - notified map prevents duplicate notifications
- Batched lock optimization - Collects all threshold-reached keys before single lock
- Deterministic signature selection - Sorts by ShareIdx (lowest first)
- Returns exactly threshold signatures - Not all matching signatures
- Context propagation - Both StoreInternal and StoreExternal add duty to context
- Memory leak prevention - Keys only added once to keysByDuty

The key insight is that once a duty+pubkey reaches threshold, it should NEVER be notified again for that same duty+pubkey.

category: refactor
ticket: none

### DIFF
--- a/core/parsigdb/memory_internal_test.go
+++ b/core/parsigdb/memory_internal_test.go
@@ -12,6 +12,7 @@ import (
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/stretchr/testify/require"
 
+	"github.com/obolnetwork/charon/app/errors"
 	"github.com/obolnetwork/charon/cluster"
 	"github.com/obolnetwork/charon/core"
 	"github.com/obolnetwork/charon/testutil"
@@ -164,6 +165,27 @@ type testDeadliner struct {
 	ch    chan core.Duty
 }
 
+func (t *testDeadliner) Expire() bool {
+	for _, d := range t.added {
+		t.ch <- d
+	}
+
+	t.ch <- core.Duty{} // Dummy duty to ensure all piped duties above were processed.
+
+	t.added = nil
+
+	return true
+}
+
+func (t *testDeadliner) Add(duty core.Duty) bool {
+	t.added = append(t.added, duty)
+	return true
+}
+
+func (t *testDeadliner) C() <-chan core.Duty {
+	return t.ch
+}
+
 // TestConcurrentSubscribeAndStore tests that there are no race conditions
 // when subscribers are being added while Store operations are happening.
 // This verifies the fix for race conditions when reading subscriber slices.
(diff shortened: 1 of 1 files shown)


core/parsigdb/memory.go
```diff
@@ -205,7 +257,8 @@ func getThresholdMatching(typ core.DutyType, sigs []core.ParSignedData, threshol
 
 	if typ == core.DutySignature {
 		// Signatures do not support message roots.
-		return sigs, len(sigs) == threshold, nil
+		// Return exactly threshold number of signatures.
```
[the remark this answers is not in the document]
pinebit: Added clarifying comment.

core/parsigdb/memory.go
```diff
@@ -63,8 +65,14 @@ func (db *MemDB) StoreInternal(ctx context.Context, duty core.Duty, signedSet co
 		return err
 	}
 
+	// Copy subscribers under lock to avoid race conditions.
```
KaloyanTanev, line 68: To be fair, probably we will never hit race condition for subscribers as there are writes to it only at Charon start. There are no reads to this field before the writes and vice versa. I'm not sure if having this extra memory allocation is worth, given the above. However, I acknowledge the above flow isn't enforced anywhere atm.

For those subs fields (`internalSubs` and `threshSubs`) `RWMutex` is probably more suitable than `Mutex`, but not as suitable for the rest of the `MemDB`.

core/parsigdb/memory.go
```diff
@@ -102,17 +119,46 @@ func (db *MemDB) StoreExternal(ctx context.Context, duty core.Duty, signedSet co
...
+		return nil
+	}
+
+	// Single lock to check and update all notifications at once
+	db.mu.Lock()
+
+	output := make(map[core.PubKey][]core.ParSignedData)
+
+	for _, r := range reached {
+		if !db.notified[r.key] {
```
KaloyanTanev, line 135: Is the idea of this `notified` map to filter out duplicates?
If so... practically there can't be duplicates as a single pubkey cannot have duplicate duty for a single slot.

I think having `var reached []thresholdReached` and then recreating the same map structure we had before `output := make(map[core.PubKey][]core.ParSignedData)` from this slice isn't really helpful. We are creating more structures in memory + having more iteration through them.

On top it says `Collect keys that reached threshold (batch processing to reduce lock contention)`, but I can't see where this batch processing is that is supposed to help us?

core/parsigdb/memory.go
```diff
@@ -167,7 +214,12 @@ func (db *MemDB) store(k key, value core.ParSignedData) ([]core.ParSignedData, b
 	}
 
 	db.entries[k] = append(db.entries[k], clone)
-	db.keysByDuty[k.Duty] = append(db.keysByDuty[k.Duty], k)
+
+	// Only append key to keysByDuty if this is the first signature for this key.
+	// This prevents duplicate keys accumulating in keysByDuty.
+	if len(db.entries[k]) == 1 {
+		db.keysByDuty[k.Duty] = append(db.keysByDuty[k.Duty], k)
```
KaloyanTanev, line 221: To be fair, I can't see where this `keysByDuty` is read at all? I see only writes and deletes. Can we... get rid of it?

core/parsigdb/memory.go
```diff
@@ -205,7 +257,9 @@ func getThresholdMatching(typ core.DutyType, sigs []core.ParSignedData, threshol
 
 	if typ == core.DutySignature {
 		// Signatures do not support message roots.
-		return sigs, len(sigs) == threshold, nil
+		// We know len(sigs) >= threshold from the check above, so slicing is safe.
+		// Return exactly threshold number of signatures.
+		return sigs[:threshold], true, nil
```
KaloyanTanev, line 262: Not really, as above we check `<` and not `!=`. I don't think we can assume it's true here.

core/parsigdb/memory.go
```diff
@@ -219,10 +273,17 @@ func getThresholdMatching(typ core.DutyType, sigs []core.ParSignedData, threshol
...
-			return set, true, nil
+		if len(set) >= threshold {
+			// Sort by share index for deterministic, reproducible behavior
+			sort.Slice(set, func(i, j int) bool {
+				return set[i].ShareIdx < set[j].ShareIdx
+			})
+
+			return set[:threshold], true, nil
 		}
 	}
```
KaloyanTanev, line 288: What do we gain from that?

pinebit:
As discussed internally, this does not make significant improvements due to how we use this component. Going further this would increase costs for testings it, etc.

PR #4060, opened 2025-10-30
*: builder reg redesign

pinebit:
New builder registration mechanism. Check CoPilot's summary for details.

category: refactor
ticket: none

### DIFF
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -256,11 +256,10 @@ func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
 		}
 	}
 
-	if s.builderEnabled {
-		if s.getResolvedEpoch() != slot.Epoch() || s.submittedRegistrationEpoch != slot.Epoch() {
-			if err := s.submitValidatorRegistrations(ctx); err == nil {
-				s.submittedRegistrationEpoch = slot.Epoch()
-			}
+	// Safe to access s.submittedRegistrationEpoch without lock because scheduleSlot is called by the same goroutine.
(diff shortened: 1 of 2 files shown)


KaloyanTanev (approved):

core/scheduler/scheduler.go
```diff
@@ -690,7 +690,18 @@ func (s *Scheduler) submitValidatorRegistrations(ctx context.Context) error {
...
 	}
 
-	return s.eth2Cl.SubmitValidatorRegistrations(ctx, registrations)
+	submitRegistrationCounter.Add(1)
+
+	err := s.eth2Cl.SubmitValidatorRegistrations(ctx, registrations)
+	if err != nil {
+		submitRegistrationErrors.Add(1)
+
+		log.Error(ctx, "Failed to submit validator registrations", err)
```
KaloyanTanev, line 703: BTW, something interesting to keep in mind and we have no influence on:
MEV-boost will report success on registrations to the BN (and hence to us) if at least 1 succeeds. This is weird behaviour IMO as it should result in failure if at least 1 fails.

KaloyanTanev (approved):

core/scheduler/scheduler.go
pinebit: We should start talking to MEV directly...

KaloyanTanev: It won't solve our issue here though. As it's the aggregator (MEV-boost) that is returning an OK response.

And I really don't fancy signing up for doing MEV-boost's job ourselves and talking directly to the relays.

DiogoSantoss (approved):

Issue #4027, opened 2025-10-13
Fetch attester data on processed block


KaloyanTanev:
# 🎯 Problem to be solved
Currently we fetch attestation data on T=1/3 (= 4th second for Ethereum mainnet). This is fine in the general case, but in situations with beacon nodes under too much load blocks can be processed a bit later. This, rightfully, gives stale heads for attestation data at T=1/3.

# 🛠️ Proposed solution
Since recently we have added SSE events subscriptions of Charon nodes to BN events. We can leverage that to trigger request for attestation data up until some timeout.

i.e.: Ask for attestation data upon BN successfully processing new block (can be even before T=1/3) or if not received by T=1/3 + 300ms (time is debatable), ask anyway and get a stale head.

This feature should be under a feature flag.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #4095, opened 2025-11-13
core/scheduler: fetch attestation data on sse block event

DiogoSantoss:
Fetch attestation data on receiving SSE block event or after `1/3 + 300ms` as a fallback, possibly fetching a stale head.
Includes a small fix to `featureset/config.go` which had a data race on clean up after a test. The feature flags are `fetch_att_on_block` and `fetch_att_on_block_with_delay`.

category: feature
ticket: #4027
feature_flag: fetch_att_on_block

### DIFF
--- a/core/scheduler/scheduler.go
+++ b/core/scheduler/scheduler.go
@@ -335,7 +335,6 @@ func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
 					log.Error(dutyCtx, "Failed to trigger duty subscriber", err, z.U64("slot", slot.Slot))
 				}
 			}
-
 		}(duty, defSet)
 
 		if slot.LastInEpoch() {


core/scheduler/scheduler.go
```diff
@@ -116,6 +117,8 @@ type Scheduler struct {
 	builderEnabled             bool
 	schedSlotFunc              schedSlotFunc
 	epochResolved              map[uint64]chan struct{} // Notification channels for epoch resolution
+	eventTriggeredAttestations map[uint64]bool          // Track attestation duties triggered via sse block event
```
pinebit: This might be the perfect case for the new `sync.Map`

DiogoSantoss: Good idea

core/scheduler/scheduler.go
```diff
@@ -185,6 +186,57 @@ func (s *Scheduler) HandleChainReorgEvent(ctx context.Context, epoch eth2p0.Epoc
...
+	defSet, ok := s.getDutyDefinitionSet(duty)
+	if !ok {
+		// Nothing for this duty.
+		return
+	}
+
+	log.Debug(ctx, "Early attestation data fetch triggered by SSE block event", z.U64("slot", uint64(slot)))
+
+	// Trigger duty immediately (early fetch)
+	go s.triggerDuty(ctx, duty, defSet)
```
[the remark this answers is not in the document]
DiogoSantoss: This context is not short-lived nor tied to the HTTP request lifecycle

core/scheduler/scheduler.go
```diff
@@ -276,33 +328,22 @@ func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
...
+				if !s.waitForBlockEventOrTimeout(ctx, slot) {
+					return // context cancelled
 				}
-
-				if err := sub(dutyCtx, duty, clone); err != nil {
-					log.Error(dutyCtx, "Failed to trigger duty subscriber", err, z.U64("slot", slot.Slot))
+				_, alreadyTriggered := s.eventTriggeredAttestations.LoadOrStore(slot.Slot, true)
+				if alreadyTriggered {
+					return // already triggered via block event
 				}
```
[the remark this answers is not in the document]
DiogoSantoss: Since `LoadOrStore` is atomic, once step 1. sets the slot in `eventTriggeredAttestations`, step 3. will return `alreadyTriggered` set to `true` and return. In the reverse case, we could reach step 3. first but then step 1. wouldn't trigger any goroutine since `alreadyTriggered` would be `true`

core/scheduler/scheduler.go
```diff
@@ -333,6 +374,28 @@ func delaySlotOffset(ctx context.Context, slot core.Slot, duty core.Duty, delayF
 	}
 }
 
+// waitForBlockEventOrTimeout waits for attestation duty with timeout fallback.
+// Returns immediately if the duty was already triggered via block event.
+// Otherwise waits until T=1/3 + 300ms (fallback timeout).
```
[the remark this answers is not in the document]
DiogoSantoss: Fixed

core/scheduler/scheduler.go
```diff
@@ -116,6 +116,7 @@ type Scheduler struct {
 	builderEnabled             bool
 	schedSlotFunc              schedSlotFunc
 	epochResolved              map[uint64]chan struct{} // Notification channels for epoch resolution
+	eventTriggeredAttestations sync.Map                 // Track attestation duties triggered via sse block event (map[uint64]bool)
```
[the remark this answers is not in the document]
DiogoSantoss: Added trimming

core/scheduler/scheduler_test.go
```diff
@@ -506,6 +506,87 @@ func TestHandleChainReorgEvent(t *testing.T) {
...
+			for _, d := range triggeredDuties {
+				if d.Type == core.DutyAttester && d.Slot == 1 {
+					hasAttester1 = true
+					break
+				}
+			}
+			dutyMux.Unlock()
+
+			require.True(t, hasAttester1, "Attester duty for slot 1 should be triggered early by block event for slot 1")
+
```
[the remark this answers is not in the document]
DiogoSantoss: Fixed with require eventually

core/scheduler/scheduler_test.go
```diff
@@ -506,6 +506,87 @@ func TestHandleChainReorgEvent(t *testing.T) {
...
+			sched.Stop()
+		}
+
+		clock.Resume()
+	}
+
+	require.NoError(t, <-doneCh)
+}
```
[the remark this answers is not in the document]
DiogoSantoss: Added more tests

core/scheduler/scheduler.go
```diff
@@ -185,6 +186,57 @@ func (s *Scheduler) HandleChainReorgEvent(ctx context.Context, epoch eth2p0.Epoc
...
 }
 
+// triggerDuty triggers all duty subscribers with the provided duty and definition set.
+func (s *Scheduler) triggerDuty(ctx context.Context, duty core.Duty, defSet core.DutyDefinitionSet) {
+	instrumentDuty(duty, defSet)
+
+	dutyCtx := log.WithCtx(ctx, z.Any("duty", duty))
+	if duty.Type == core.DutyProposer {
+		var span trace.Span
+		dutyCtx, span = core.StartDutyTrace(dutyCtx, duty, "core/scheduler.scheduleSlot")
```
pinebit: Not sure the trace name is right `.scheduleSlot`..

core/scheduler/scheduler.go
```diff
@@ -690,6 +752,33 @@ func (s *Scheduler) trimDuties(epoch uint64) {
...
 	delete(s.dutiesByEpoch, epoch)
+
+	if featureset.Enabled(featureset.FetchAttOnBlock) {
+		s.trimEventTriggeredAttestations(epoch)
+	}
+}
+
+// trimEventTriggeredAttestations removes old slot entries from eventTriggeredAttestations.
+func (s *Scheduler) trimEventTriggeredAttestations(epoch uint64) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
pinebit: Why 5s timeout? I see other `FetchSlotsConfig` calls don't use any timeouts.
If this makes sense for this case though, let's move all magic timings to global constants (5s, 300ms, etc)

DiogoSantoss: On a second look, it's unnecessary since the `eth2wrap` already has a client with a configured timeout

core/scheduler/scheduler.go
DiogoSantoss: I didn't modify this name, it was already like that in `scheduleSlot` https://github.com/ObolNetwork/charon/blob/99bc19424cc7e1a9afe1ac29bb44a7ff08101a4d/core/scheduler/scheduler.go#L290 but I extracted that logic to a separate function to reuse in `HandleBlockEvent`
Are you suggesting changing to `.triggerDuty` ?

KaloyanTanev:
Let's rework this one after Christmas as we've discussed in the past week - separate the logic into 2 flags:
- `fetch_att_on_block`
- `fetch_att_on_block_with_delay`

If both are or only the second is enabled, delay will be used. If only the first is enabled, no delay will be added.

To be fair, the first one does not make a huge amount of sense. We can achieve only triggering early fetching, which will save us some time, but submitting will still be in 1/3 > x > 2/3 of slot time. There are 2 blockers for the attestation to be actually submitted early:
1. Consensus starts at 1/3 slot time. This is hard and scary to change. Does not really make sense to me to put that much effort in it for such a small reward.
2. VC asks for data at 1/3 slot time. This can be changed relatively easy, as most VCs have an option to use the SSE endpoint to get notified when there is a new block and ask for attestation data (what we do). As we have SSE proxying, it should work.

core/scheduler/scheduler.go
pinebit: No suggestion, whatever you think will be useful when browsing trace data in Grafana.

core/fetcher/fetcher.go
```diff
@@ -70,9 +90,24 @@ func (f *Fetcher) Fetch(ctx context.Context, duty core.Duty, defSet core.DutyDef
...
-		if err != nil {
-			return errors.Wrap(err, "fetch attester data")
+		// Check if attestation data was already fetched early and cached
+		if cached, ok := f.attDataCache.Load(duty.Slot); ok {
+			f.attDataCache.Delete(duty.Slot)
+			if data, valid := cached.(core.UnsignedDataSet); valid {
+				unsignedSet = data
+				log.Debug(ctx, "Using early-fetched attestation data from cache", z.U64("slot", duty.Slot))
+			} else {
+				// Type assertion failed, re-fetch
```
KaloyanTanev, line 101: Can we add a warning log here? It's not expected behaviour, so it'd be good to have it recognised.

core/scheduler/scheduler.go
```diff
@@ -157,6 +165,46 @@ func (s *Scheduler) HandleChainReorgEvent(ctx context.Context, epoch eth2p0.Epoc
 	}
 }
 
+// HandleBlockEvent handles SSE "block" events (block imported to fork choice) and triggers early attestation data fetching.
+func (s *Scheduler) HandleBlockEvent(ctx context.Context, slot eth2p0.Slot, bnAddr string) {
+	if !featureset.Enabled(featureset.FetchAttOnBlock) || s.fetcherFetchOnly == nil {
```
KaloyanTanev: Isn't `s.fetcherFetchOnly == nil` an unexpected scenario?

If so, can we put it in separate if statement after this one, log a warning and return?

core/scheduler/scheduler.go
```diff
@@ -306,6 +360,31 @@ func delaySlotOffset(ctx context.Context, slot core.Slot, duty core.Duty, delayF
...
+	case <-ctx.Done():
+		return false
+	case <-s.clock.After(time.Until(fallbackDeadline)):
+		// Check if block event triggered early fetch
+		if _, triggered := s.eventTriggeredAttestations.Load(slot.Slot); !triggered {
+			log.Debug(ctx, "Proceeding with attestation at T=1/3+300ms (no early block event)",
+				z.U64("slot", slot.Slot))
+		}
+		return true
+	}
```
KaloyanTanev, line 401: This would block always for T=1/3+300ms, right? Don't we want to unblock if we fetched a block using the SSE trigger?

core/scheduler/scheduler.go
```diff
@@ -275,14 +329,14 @@ func (s *Scheduler) scheduleSlot(ctx context.Context, slot core.Slot) {
...
-		if err != nil {
-			log.Warn(ctx, "Resolving duties error (retrying next slot)", err, z.U64("slot", slot.Slot))
-		}
+			if slot.LastInEpoch() {
+				err := s.resolveDuties(ctx, slot.Next())
+				if err != nil {
+					log.Warn(ctx, "Resolving duties error (retrying next slot)", err, z.U64("slot", slot.Slot))
+				}
+			}
+		}(duty, defSet)
```
KaloyanTanev: Why is this inside the go routine now?

core/scheduler/scheduler.go
```diff
@@ -663,6 +742,32 @@ func (s *Scheduler) trimDuties(epoch uint64) {
...
+	s.eventTriggeredAttestations.Range(func(key, _ any) bool {
+		slot, ok := key.(uint64)
+		if !ok {
+			return true // continue iteration
+		}
+		if slot < minSlotToKeep {
+			s.eventTriggeredAttestations.Delete(slot)
+		}
+		return true // continue iteration
+	})
```
KaloyanTanev, line 786: We never return false from this function? Do we even need a return value?

core/scheduler/scheduler.go
DiogoSantoss: Yes it will block for either 1/3 or 1/3+300 but that's what we want right ? To avoid triggering consensus too soon.

core/scheduler/scheduler.go
DiogoSantoss: My bad, issue when merging with main

core/scheduler/scheduler.go
DiogoSantoss: The `Range` function requires returning true, otherwise it stop iterating.

Issue #4059, opened 2025-10-28
`--nickname` flag in `charon dkg`


OisinKyne:
# 🎯 Problem to be solved

Figuring out who is taking part in a dkg is hard. It only shows peer names, and its hard to figure out who that is before we have prometheus running. The only way to figure it out is the startup log line. 

# 🛠️ Proposed solution

Add --nickname support to DKG including the log lines

# 🧪 Tests

# 👐 Additional acceptance criteria

On the launchpad side, if there is an ens name setup for the connected wallet, we should auto-populate that in the command we proffer people to run. 

On the DV-pod, we should reuse the same `.Values.charon.nickname` for this feature once it ships. 

# ❌ Out of Scope
If there is anything to highlight as out of scope for this issue, please outline it here.

OisinKyne:
Merged in #4105

PR #4105, opened 2025-11-18
dkg: add `nickname` flag

DiogoSantoss:
Add `--nickname` flag to DKG.
Peers will send their nickname via `sync` protocol which will be logged upon connecting for the first time:
```
17:38:03.733 INFO dkg        Connected to peer 1 of 3                 {"peer": "affectionate-kitchen", "nickname": "Alice"}
17:38:03.742 INFO dkg        Connected to peer 2 of 3                 {"peer": "amiable-mirror", "nickname": "Bob"}
```

category: feature
ticket: #4059

### DIFF
--- a/cmd/dkg.go
+++ b/cmd/dkg.go
@@ -25,7 +25,7 @@ func newDKGCmd(runFunc func(context.Context, dkg.Config) error) *cobra.Command {
 distributed validator key shares and a final cluster lock configuration. Note that all other cluster operators should run
 this command at the same time.`,
 		Args: cobra.NoArgs,
-		PreRunE: func(cmd *cobra.Command, args []string) error {
+		PreRunE: func(_ *cobra.Command, _ []string) error {
 			if len(config.Nickname) > 32 {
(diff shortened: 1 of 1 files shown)


dkg/dkgpb/v1/sync.pb.go
```diff
@@ -29,6 +29,7 @@ type MsgSync struct {
 	Shutdown      bool                   `protobuf:"varint,3,opt,name=shutdown,proto3" json:"shutdown,omitempty"`
 	Version       string                 `protobuf:"bytes,4,opt,name=version,proto3" json:"version,omitempty"`
 	Step          int64                  `protobuf:"varint,5,opt,name=step,proto3" json:"step,omitempty"`
+	Nickname      string                 `protobuf:"bytes,6,opt,name=nickname,proto3" json:"nickname,omitempty"`
```
KaloyanTanev, line 32: Good to note that this will make it incompatible with previous versions.
Either way we allow DKG only with the same MAJOR.MINOR, so it's not something new.

Issue #4107, opened 2025-11-19
Catch HTTP 409 Duplicate Request gracefully in `charon dkg --publish`


OisinKyne:
# 🎯 Problem to be solved

When we complete a DKG and publish it. One node succeeds, the rest fail, and get an error that looks concerning. However its only the HTTP409 duplicate request because they're not the first node to publish. 

# 🛠️ Proposed solution

Lets just catch the 409 as expected and info log something like "Cluster lock is already published.". We should still throw any other 4XX or 5XX

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #4121, opened 2025-11-24
*: handing 409 for obol api post

pinebit:
Handling 409 status code for Obol API's HTTP POST requests, e.g. when DKG is publishing the lock file.

category: refactor
ticket: #4107


OisinKyne (approved):
I think `Cluster lock is already published.` would be cleaner than:  `Duplicate request ignored by Obol API. Likely another node submitted the same data already.`

PR #4132, opened 2025-11-28
*: omit unnecessary reassignment

rifeplight:
The new version of Go has been optimized, and variables do not need to be reassigned.

For more info: https://tip.golang.org/wiki/LoopvarExperiment#does-this-mean-i-dont-have-to-write-x--x-in-my-loops-anymore

category: refactor
ticket: none


rifeplight:
Reverted according to the CI and copilot suggesstion.

KaloyanTanev:
Those changes are not what loopvar change in go v1.22 is about. The change in v1.22 is regarding the variables created from the for loop line. Those 3 variables changed are not related to the loop.

PR #4169, opened 2025-12-12
*: improved tracing

pinebit:
Fixed and improving of tracing data production.

category: refactor
ticket: none

### DIFF
--- a/core/validatorapi/validatorapi.go
+++ b/core/validatorapi/validatorapi.go
@@ -18,11 +18,13 @@ import (
 	"github.com/attestantio/go-eth2-client/spec/altair"
 	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
 	ssz "github.com/ferranbt/fastssz"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/obolnetwork/charon/app/errors"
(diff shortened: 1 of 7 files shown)


core/validatorapi/validatorapi.go
```diff
@@ -261,6 +262,14 @@ func (c *Component) Subscribe(fn func(context.Context, core.Duty, core.ParSigned
 
 // AttestationData implements the eth2client.AttesterDutiesProvider for the router.
 func (c Component) AttestationData(ctx context.Context, opts *eth2api.AttestationDataOpts) (*eth2api.Response[*eth2p0.AttestationData], error) {
+	var span trace.Span
+
+	duty := core.NewAttesterDuty(uint64(opts.Slot))
+	ctx, span = core.StartDutyTrace(ctx, duty, "core/validatorapi.AttestationData")
+
+	span.SetAttributes(attribute.Int64("committee_index", int64(opts.CommitteeIndex)))
```
KaloyanTanev, line 270: This here we expect it to be always 0 if `fetch_only_commidx_0` is turned on which is the majority now. It will also be made default in the upcoming months. Not sure if we need that attribute.

pinebit: It shall not hurt, especially if it is not zero, suddenly :) this way we can always confirm it is zero

PR #4205, opened 2026-01-06
cmd: skip inactive validators in exits

KaloyanTanev:
This PR solves 2 issues:
1. Upon signing exits for all validators, Charon does not check if all validators are active. We are changing that to sign only exits for validators that are either queued for activation or active.
2. Upon broadcasting exits for all validators, Charon does not check if all validators are active. We are changing that to broadcast only exits for validators that are active.

category: bug
ticket: none

### DIFF
--- a/cmd/exit_broadcast.go
+++ b/cmd/exit_broadcast.go
@@ -9,9 +9,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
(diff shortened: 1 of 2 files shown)


DiogoSantoss:
How bad is signing exits for validators which are not active yet ? Asking because I remember we had issues with `head` not having the most up to date state and returning stale validator state. Although this was only occurred when asking close to the activation epoch so maybe not relevant.

OisinKyne:
> How bad is signing exits for validators which are not active yet ?

You cannot sign a valid exit for a validator until it has an index assigned (meaning its in the activation queue at least). You cannot broadcast a valid exit until the validator has been active for at least 27 hours.

Issue #4200, opened 2025-12-18
Edit ready private-key-lock-file


pinebit:
# 🎯 Problem to be solved

Most of the `edit` commands replace the existing key shares. The process of cluster editing involves shutting down the VC and Charon instances and then restart with a new set of key shares. The new key shares appear as all new keys to VC and therefore its anti-slashing database will not work as expected (updating this DB is another concern that is to be managed separately from this ticket). To prevent double signing reliably, it is recommended to enforce a pause (grace period of two complete epochs) between shutting down and restarting of those components, and Charon shall enforce this.

# 🛠️ Proposed solution

- [ ] Approved proposal - read below.
- [ ] Core team consensus on the proposed solution

A proposal is to extend lease-lock pattern already embedded in the codebase as follow:
- Charon will update the lease lock with the most recent timestamp of the last signing event along with the cluster hash.
- Every edit command will see this lock file and update it by adding two epochs - that will become the deadline for the enforced grace period; two epoch is the recommended by Ethereum grace period.
- On a new `charon run` it will verify the current time against the lease lock timestamp as well as cluster hash match/mismatch and makes Charon waiting for the remaining grace period, after which it simply resumes booting all Charon components.
- If no edit command was executed in between, the lease lock timestamp will always be in past.

The counterpart work that needs to be mentioned here: https://github.com/ObolNetwork/charon-distributed-validator-node/pull/395 - the script that will update anti-slashing DB with the new keys.

Both measures shall reliably mitigate slashing risks.

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #4226, opened 2026-01-15
*: lock supporting grace period

pinebit:
## Add cluster lock hash validation to privkey lock file

Enhances the private key lock mechanism to prevent race conditions during cluster configuration changes.

### Changes
- Track `cluster_lock_hash` in the privkey lock file metadata
- Introduce 2-epoch grace period (768s) after cluster edits before allowing restart with new configuration
- Update `privkeylock.New()` signature to accept cluster lock file path
- Handle missing cluster lock files gracefully (for DKG scenarios)
- Skip grace period check for migrations from old format (empty hash)
- Edit commands copying the lock file, if present.

### Why
After cluster edits (adding/removing operators), validators shouldn't immediately start with the new configuration. The grace period ensures:
- Old configuration finishes duties gracefully
- Network has time to finalize pending attestations
- Prevents double-signing between old/new configurations

category: feature
ticket: #4200

### DIFF
--- a/app/privkeylock/privkeylock.go
+++ b/app/privkeylock/privkeylock.go
@@ -57,7 +57,7 @@ func New(privKeyFilePath, clusterLockFilePath, command string) (Service, error) 
 			elapsedPeriod := time.Since(meta.Timestamp)
 			if elapsedPeriod < gracePeriod {
 				waitTime := gracePeriod - elapsedPeriod
-				errText := fmt.Sprintf("existing private key lock file found with different cluster lock hash, you must wait for %v before starting charon with the new cluster hash", waitTime)
+				errText := fmt.Sprintf("an existing private key lock file is present with a different cluster lock hash, for safety reasons, you must wait for %v before starting charon with a modified cluster", waitTime)
 
 				return Service{}, errors.New(
 					errText,


app/privkeylock/privkeylock.go
```diff
@@ -17,50 +18,81 @@ var (
...
+
+	// gracePeriod is the duration after which a new cluster (new lock hash) can be run after an edit command.
+	// (we can't use chain spec at this time, so we use fixed duration of 768 seconds)
+	gracePeriod = 2 * 32 * 12 * time.Second
 )
 
 // New returns new private key locking service. It errors if a recently-updated private key lock file exists.
-func New(path, command string) (Service, error) {
-	content, err := os.ReadFile(path)
+func New(privKeyFilePath, clusterLockFilePath, command string) (Service, error) {
```
KaloyanTanev, line 28: nitpick: can we rename that to `privKeyFileLockPath`? Reading it as `privKeyFilePath` I've assumed it's the actual privKey

pinebit: I used to follow "filepath" convention as a common widely adopted term. For instance, the standard Golang's package "path/filepath" is an example.

app/privkeylock/privkeylock.go
```diff
@@ -17,50 +18,81 @@ var (
...
 				z.Str("command", meta.Command),
+				z.Str("cluster_lock_hash", meta.ClusterLockHash),
 			)
 		}
+
+		if meta.ClusterLockHash != "" && clusterLockHash != meta.ClusterLockHash {
+			elapsedPeriod := time.Since(meta.Timestamp)
+			if elapsedPeriod < gracePeriod {
+				waitTime := gracePeriod - elapsedPeriod
+				errText := fmt.Sprintf("existing private key lock file found with different cluster lock hash, you must wait for %v before starting charon with the new cluster hash", waitTime)
```
OisinKyne:
```suggestion
				errText := fmt.Sprintf("An existing private key lock file is present with a different cluster lock hash, for safety reasons, you must wait for %v before starting charon with a modified cluster", waitTime)
```

Idk will

pinebit: Changed as suggested

Issue #4009, opened 2025-10-01
Cache VC schedule duties requests


KaloyanTanev:
# 🎯 Problem to be solved

In our scheduler we ask the BN for duties for the validators we manage.
Underlying VC also does that. For those request we always ask the BN again for duties.

This results in duplicate requests, which puts unnecessary load on the BN. Sometimes duties fail because the VC is taking too long to respond.

# 🛠️ Proposed solution
- [ ] Cache the results from scheduler and VC in Charon
- [ ] Serve the VC call or the scheduler the data from the cache, if it's missing
- [ ] Potentially do not send requests twice if one request is already sent and we are waiting on its response

This solution is similar to what we do for active validators where we also maintain cache.

Some data points from our canary clusters

Attester duties for Prysm BN
<img width="1300" height="744" alt="Image" src="https://github.com/user-attachments/assets/20539415-8800-4d7c-bc30-0704640cb0d4" />

Proposer duties for Teku BN
<img width="2778" height="1578" alt="Image" src="https://github.com/user-attachments/assets/f1b6c765-d7dc-436a-9541-b7df597db667" />

PR #4227, opened 2026-01-16
*: add duties cache

KaloyanTanev:
Previously we've recognised a pattern in certain validator clients - making too many obsolete requests to Charon for checking active validators. This meant that Charon was sending the same amount of requests to the beacon node for such a heavy request. The assumption that the beacon node will have this data cached and serve us quickly was not always true. Because of that we have added active validators cache.

With a similar logic, some validator clients are sending requests too often for next slot's duties. This creates a lot of requests to the beacon node which can be heavy in cases of 1000+ validator keys. Moreover, Charon is also sending such requests to the beacon node, so the amount of duties requests is at least doubled each epoch. We can prevent that with a similar caching strategy. It has been implemented for attester, proposer and sync committee duties.

Another small change has been made - removed Teku VC testing from integration tests. Back in the days it was included there as some sanity check with a real and not a mock VC. However, nowadays, we are doing much more complicated and closer to real scenarios of actual VCs in Kurtosis.

category: feature
ticket: #4009

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -22,8 +22,8 @@ when storing metrics from multiple nodes or clusters in one Prometheus instance.
 | `app_beacon_node_sse_head_slot` | Gauge | Current beacon node head slot, supplied by beacon node`s SSE endpoint | `addr` |
 | `app_beacon_node_version` | Gauge | Constant gauge with label set to the node version of the upstream beacon node | `version` |
 | `app_cache_hits_total` | Counter | Total number of times the cache was used | `endpoint` |
-| `app_cache_misses_total` | Counter | Total number of times the cache was missed | `endpoint` |
 | `app_cache_invalidated_reorg_total` | Counter | Total number of times the cache was invalidated due to a chain reorg | `endpoint` |
+| `app_cache_misses_total` | Counter | Total number of times the cache was missed | `endpoint` |
 | `app_eth2_errors_total` | Counter | Total number of errors returned by eth2 beacon node requests | `endpoint` |
 | `app_eth2_latency_seconds` | Histogram | Latency in seconds for eth2 beacon node requests | `endpoint` |
 | `app_eth2_requests_total` | Counter | Total number of requests sent to eth2 beacon node | `endpoint` |


app/eth2wrap/cache.go
```diff
@@ -0,0 +1,471 @@
...
+	"sync"
+
+	eth2api "github.com/attestantio/go-eth2-client/api"
+	eth2v1 "github.com/attestantio/go-eth2-client/api/v1"
+	eth2p0 "github.com/attestantio/go-eth2-client/spec/phase0"
+
+	"github.com/obolnetwork/charon/app/errors"
+)
+
+const dutiesCacheTrimThreshold = 6
```
pinebit: Please add a comment explaining what is this for and why 6.

app/app.go
```diff
@@ -518,14 +524,21 @@ func wireCoreWorkflow(ctx context.Context, life *lifecycle.Manager, conf Config,
...
 		valCache.Trim()
 
-		_, _, refresh, err := valCache.GetBySlot(ctx, slotToFetch)
+		activeValidators, _, refresh, err := valCache.GetBySlot(ctx, slotToFetch)
 		if err != nil {
 			log.Error(ctx, "Failed to refresh validator cache", err)
 			return err
 		}
 
+		vIdxs := []eth2p0.ValidatorIndex{}
```
DiogoSantoss: A fancy new way of collecting map keys in Go 1.23:
```
vIdxs := slices.Collect(maps.Keys(activeValidators))
```
however setting the capacity (or size) is probably faster:
```
vIdxs := make([]eth2p0.ValidatorIndex, 0, len(activeValidators))
```
to avoid reallocations with append.

KaloyanTanev: Ah yes, I missed the `slices.Collect`.
I agree with you on the speed of `vIdxs := make([]eth2p0.ValidatorIndex, 0, len(activeValidators))`, but this is executed once on startup so performance isn't really crucial IMO. I'd rather stick with your first suggestion of `vIdxs := slices.Collect(maps.Keys(activeValidators))`

PR #4240, opened 2026-01-21
core: log error when bcast proposal

pinebit:
Added logging error if bcast failed to submit proposal.

category: refactor
ticket: none

### DIFF
--- a/core/bcast/bcast.go
+++ b/core/bcast/bcast.go
@@ -205,6 +205,13 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, set core.Sig
 				z.Any("pubkey", pubkey),
 				z.Bool("blinded", block.Blinded),
 			)
+		} else {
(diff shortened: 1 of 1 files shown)


core/bcast/bcast.go
```diff
@@ -205,6 +205,13 @@ func (b Broadcaster) Broadcast(ctx context.Context, duty core.Duty, set core.Sig
 				z.Any("pubkey", pubkey),
 				z.Bool("blinded", block.Blinded),
 			)
+		} else {
+			log.Error(ctx, "Failed to submit block proposal to beacon node",
```
KaloyanTanev, line 209: We will log a failed BN call if there is an error, as we will return error from this endpoint

pinebit: Yes but we won't log details such as "blinded"

KaloyanTanev: Hmmm okay. This will become obsolete in Gloas fork, but we can make some use of it until then.

PR #4246, opened 2026-01-22
*: dkg and pk ux improvements

pinebit:
After recent issues investigation, we would like to introduce these several improvement to UX, for dkg and privkey lock specifically. The main goal is to print better error messages.

category: refactor
ticket: none

### DIFF
--- a/app/privkeylock/privkeylock.go
+++ b/app/privkeylock/privkeylock.go
@@ -34,7 +34,7 @@ func New(privKeyFilePath, clusterLockFilePath, command string) (Service, error) 
 			return Service{}, errors.New("private key file does not exist, please check the path", z.Str("path", privKeyFilePath))
 		}
 
-		return Service{}, errors.Wrap(err, "stat private key file", z.Str("path", privKeyFilePath))
+		return Service{}, errors.Wrap(err, "fetch file info for private key file", z.Str("path", privKeyFilePath))
 	}
 
 	if info.IsDir() {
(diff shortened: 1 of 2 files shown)


app/privkeylock/privkeylock.go
```diff
@@ -26,28 +27,56 @@ var (
...
 // New returns new private key locking service. It errors if a recently-updated private key lock file exists.
 func New(privKeyFilePath, clusterLockFilePath, command string) (Service, error) {
+	// Sanity check: private key file exists
+	info, err := os.Stat(privKeyFilePath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return Service{}, errors.New("private key file does not exist, please check the path", z.Str("path", privKeyFilePath))
+		}
+
+		return Service{}, errors.Wrap(err, "stat private key file", z.Str("path", privKeyFilePath))
```
KaloyanTanev: I'm not sure `stat private key file` will mean anything significant to the user?

```suggestion
		return Service{}, errors.Wrap(err, "fetch file info for private key file", z.Str("path", privKeyFilePath))
```

app/privkeylock/privkeylock.go
```diff
@@ -26,28 +27,56 @@ var (
...
+		return Service{}, errors.New("private key file path is a directory, please check the path", z.Str("path", privKeyFilePath))
+	}
+
+	privKeyFileLockPath := privKeyFilePath + ".lock"
+
+	// Sanity check: we can write to the directory
+	parentDir := filepath.Dir(privKeyFileLockPath)
+	if info, err := os.Stat(parentDir); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return Service{}, errors.New("parent directory does not exist", z.Str("dir", parentDir), z.Str("lock_path", privKeyFileLockPath))
```
KaloyanTanev: Just for clarity
```suggestion
			return Service{}, errors.New("parent directory, fetched from private key lock file path, does not exist", z.Str("parent_directory", parentDir), z.Str("lock_path", privKeyFileLockPath))
```

dkg/dkg.go
```diff
@@ -284,6 +284,8 @@ func Run(ctx context.Context, conf Config) (err error) {
 
 	nextStepSync, stopSync, err := startSyncProtocol(ctx, p2pNode, key, def.DefinitionHash, peerIDs, cancel, conf.TestConfig, conf.Nickname)
 	if err != nil {
+		log.Error(ctx, "Failed to start sync protocol, please make sure you don't have any other charon or dkg processes running", err)
```
KaloyanTanev: As essentially DKG is also Charon, might confuse some users.
```suggestion
		log.Error(ctx, "Failed to start sync protocol, please make sure you don't have any other long-running charon or dkg processes running", err)
```

Issue #4102, opened 2025-11-15
Add amount of pending/successful attestations metrics


KaloyanTanev:
# 🎯 Problem to be solved
Currently we can track successful attester duties and successful submitted attestations.

However, data would like to know how many attestations we have also submitted (remember, the submission to the BN is a collection of signatures). This is because it's much easier to compare to external sources which track per single attestation.

# 🛠️ Proposed solution

Add two more metrics:
1. Counter of expected attestations for the slot. Currently data are using pending attester duty, we need something for all attestations expected in the slot.
2. Counter of successful attestations for the slot. Currently data are using successful attester duty, we need something for all attestations successful in the slot. Potentially after successful duty we increase a counter for the amount of attestations?

PR #4248, opened 2026-01-22
tracker: add attestation counter metric

DiogoSantoss:
Add two new metrics `core_tracker_attestation_expect_total` and `core_tracker_attestation_success_total` which track the count of individual attestations instead of attestation duties.

image

category: feature
ticket: #4102

### DIFF
--- a/core/tracker/tracker.go
+++ b/core/tracker/tracker.go
@@ -219,22 +219,22 @@ func lastStep(dutyType core.DutyType) step {
 }
 
 // countAttestations returns the number of attestations in the events for an attester duty.
-// For attester duties, it counts the unique pubkeys which represents the number of attestations.
+// For attester duties, it counts the fetcher events which represents the number of attestations.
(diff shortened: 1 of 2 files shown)


core/tracker/tracker.go
```diff
@@ -216,6 +218,25 @@ func lastStep(dutyType core.DutyType) step {
...
+		return 0
+	}
+
+	// Count unique pubkeys in fetcher events (which represents the expected attestations)
+	uniquePubkeys := make(map[core.PubKey]bool)
+	for _, e := range events {
+		if e.step == fetcher && e.pubkey != "" {
+			uniquePubkeys[e.pubkey] = true
+		}
+	}
```
KaloyanTanev: Do we need that? In what scenario we can see multiple attestations of the same public key in the same slot?

DiogoSantoss: Fair, we can just user a counter

PR #4261, opened 2026-01-29
dkg: fixing processkey

pinebit:
This fixes a bug in `restoreKey` routine that affects "remove operators" edit command.

category: bug
ticket: none


pinebit:
> Would it be possible to have a test case to catch this scenario?

I was thinking about this, but let's wait for the full report and I suspect we may need to rework a lot more, including more of the  tests...

aadomn (approved):

PR #4268, opened 2026-02-02
dkg: threshold validation

pinebit:
Added threshold validation for `RunDKG` and `RunReshareDKG`. Previous behavior delegated the check down the stack to allow underlying library (kyber) to validate the threshold. Given that Kyber has weak validation that can lead to weird late errors, it is more convenient to verify the threshold upfront.

Additionally added verification such that users cannot remove all nodes with "remove operators" command.

category: refactor
ticket: none

### DIFF
--- a/dkg/protocol_test.go
+++ b/dkg/protocol_test.go
@@ -173,7 +173,7 @@ func TestRemoveOperatorsProtocol_AllNodes(t *testing.T) {
 	ctx, cancel := context.WithCancel(t.Context())
 	defer cancel()
 
-	var errorOccurred atomic.Bool
+	var errorReported atomic.Bool
 
 	runProtocol(t, numNodes, func(relayAddr string, n int) error {
 		dkgConfig := createDKGConfig(t, relayAddr)
@@ -194,10 +194,9 @@ func TestRemoveOperatorsProtocol_AllNodes(t *testing.T) {
 
 		err := dkg.RunRemoveOperatorsProtocol(ctx, removeConfig, dkgConfig)
 		if err != nil {
-			errorOccurred.Store(true)
-
-			require.ErrorContains(t, err, "remove operation would remove all nodes from original cluster")
-			cancel()
+			if strings.Contains(err.Error(), "remove operation would remove all nodes from original cluster") {
+				errorReported.Store(true)
+			}
 
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


dkg/pedersen/dkg.go
```diff
@@ -51,6 +53,15 @@ func RunDKG(ctx context.Context, config *Config, board *Board, numVals int) ([]s
 		return int(a.Index) - int(b.Index)
 	})
 
+	threshold := config.Threshold
+	if threshold == 0 {
+		threshold = cluster.Threshold(len(nodes))
+	}
```
KaloyanTanev, line 61: Is this a plausible scenario if this is called from the CLI? If the default is 0, can we add a debug log here?

Can a user set a negative number? It might be worth it making it a uint?

dkg/pedersen/reshare.go
```diff
@@ -173,6 +174,48 @@ func RunReshareDKG(ctx context.Context, config *Config, board *Board, shares []s
...
+				z.Int("old_nodes", len(oldNodes)),
+				z.Int("removed_peers", len(config.Reshare.RemovedPeers)),
+			)
+		}
+	}
+
+	oldThreshold := config.Threshold
+	if oldThreshold == 0 {
+		oldThreshold = cluster.Threshold(len(oldNodes))
+	}
```
KaloyanTanev: Why are we calculating that? Shouldn't we rather enforce it to be set in the config, as it's something usually obtained from the pre-existing cluster lock?

dkg/pedersen/reshare.go
```diff
@@ -173,6 +174,48 @@ func RunReshareDKG(ctx context.Context, config *Config, board *Board, shares []s
...
+
+	newThreshold := config.Reshare.NewThreshold
+	if newThreshold == 0 {
+		newThreshold = cluster.Threshold(len(newNodes))
+	}
```
KaloyanTanev: Probably an info log here as well?

dkg/pedersen/reshare.go
```diff
@@ -173,6 +174,48 @@ func RunReshareDKG(ctx context.Context, config *Config, board *Board, shares []s
...
+
+	// Validate old threshold against current cluster size
+	if err := validateThreshold(len(oldNodes), oldThreshold); err != nil {
+		return nil, errors.Wrap(err, "invalid old threshold")
+	}
```
KaloyanTanev: Might not be worth the hassle validating something that was already running in production?

dkg/pedersen/dkg.go
pinebit: Improved this in the new iteration. Thank you for the feedback!

Issue #4247, opened 2026-01-22
Log the signed head in sync message submission


KaloyanTanev:
# 🎯 Problem to be solved
Given that VCs use different mechanisms to acquire the head to be signed for the sync message, we cannot deterministically choose an endpoint to gatekeep with a QBFT consensus between the nodes.

As a first step it would be nice to at least track who is providing signature of what - log the `beacon_block_root` Charon received for each `sync_message` partial signature.

PR #4269, opened 2026-02-02
core/validatorapi: log beacon block root

DiogoSantoss:
Log `beacon_block_root` of each sync committee message.

category: feature
ticket: #4247


core/validatorapi/validatorapi.go
```diff
@@ -950,6 +950,10 @@ func (c Component) SubmitSyncCommitteeMessages(ctx context.Context, messages []*
 			return err
 		}
 
+		log.Debug(ctx, "Sync committee message received from validator client",
+			z.U64("slot", uint64(msg.Slot)),
+			z.Any("beacon_block_root", msg.BeaconBlockRoot))
```
KaloyanTanev:
```suggestion
			z.Str("beacon_block_root", msg.BeaconBlockRoot.String()))
```
Otherwise we will see a byte slice, can't do much with it.

PR #4280, opened 2026-02-04
dkg: cap p2p msg size

pinebit:
Added dkg p2p wire message size validation.

category: refactor
ticket: none

### DIFF
--- a/dkg/sync/server.go
+++ b/dkg/sync/server.go
@@ -30,7 +30,7 @@ import (
 const (
 	protocolID = "/charon/dkg/sync/1.0.0/"
 
-	maxMessageSize = 10 * 1024 * 1024 // 10 MB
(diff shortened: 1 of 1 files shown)


dkg/sync/server.go
```diff
@@ -27,7 +27,11 @@ import (
 	"github.com/obolnetwork/charon/p2p"
 )
 
-const protocolID = "/charon/dkg/sync/1.0.0/"
+const (
+	protocolID = "/charon/dkg/sync/1.0.0/"
+
+	maxMessageSize = 10 * 1024 * 1024 // 10 MB
```
KaloyanTanev: I'm cautious this might bite us back at some point when Ethereum blocks get bigger and we forget about this cap.

pinebit: It's possible. Increased to 32Mb for now. I rely on meaningful error we will get in case of reaching this limit.

PR #4282, opened 2026-02-04
dkg: better nonce

pinebit:
Better nonce generation and variation for dkg.

category: refactor
ticket: none

### DIFF
--- a/dkg/pedersen/dkg.go
+++ b/dkg/pedersen/dkg.go
@@ -78,7 +78,7 @@ func RunDKG(ctx context.Context, config *Config, board *Board, numVals int) ([]s
 
 	shares := make([]share.Share, 0, numVals)
 
-	for i := 0; i < numVals; i++ {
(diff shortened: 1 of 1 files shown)


dkg/pedersen/dkg.go
```diff
@@ -71,7 +65,14 @@ func RunDKG(ctx context.Context, config *Config, board *Board, numVals int) ([]s
 
 	shares := make([]share.Share, 0, numVals)
 
-	for range numVals {
+	for i := 0; i < numVals; i++ {
```
KaloyanTanev: For simplicity
```suggestion
	for i := range numVals {
```

KaloyanTanev (approved):

dkg/pedersen/dkg.go
pinebit: Fixed, thank you

Issue #4166, opened 2025-12-10
Allow URL for lock-file CLI flag


pinebit:
# 🎯 Problem to be solved

The idea is suggested by Oisin here: https://github.com/ObolNetwork/obol-gitbook/pull/76#discussion_r2586637947

# 🛠️ Proposed solution
- [ ] Add support for https URLs for the `--lock-file` CLI flags, in particulate for all `edit` commands

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests

KaloyanTanev:
We might as well add it to other charon commands (not run though) that require cluster lock or definition? Not as important though.

PR #4285, opened 2026-02-05
*: fetching lock via http

pinebit:
Cluster lock file can be fetched via http is a URL is provided.

category: feature
ticket: #4166


pinebit:
I think a better approach would be not integrating this into charon, but supporting this in CDVN scripts.

Issue #4265, opened 2026-01-29
DKG is failing with error " Fatal error: cluster definition hashes verification failed."


HananINouman:
# 🐞 Bug Report

### Description

I created a definition on mainnet with a safe smart contract operator, but dkg only work with --no-verify

### Has this worked before in a previous version?
Not sure, but I remember i tested when we supported Safe in charon and it was working, not 100% sure though.

## 🔬 Minimal Reproduction
You can use this
https://api.obol.tech/v1/definition/0xfecca10d19152d372e1de7c14121268418925b0d8e86ed73bc4b7e371ac1e380

## 🔥 Error

" Fatal error: cluster definition hashes verification failed. Run with --no-verify to bypass verification at own risk: definition hash: bytes too long {"n": 65, "l": 130}"

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>

  </code>
</pre>

**What version of Charon are you running? (Which release)**

<pre>
  <code>

  </code>
</pre>

**Anything else relevant (validator index / public key)?**

PR #4291, opened 2026-02-06
*: add cluster definition v1.11

DiogoSantoss:
New cluster definition `v1.11` to support variable-sized signatures.

category: feature
ticket: #4265

### DIFF
--- a/cluster/definition.go
+++ b/cluster/definition.go
@@ -232,6 +232,10 @@ func (d Definition) VerifySignatures(eth1 eth1wrap.EthClientRunner) error {
 		return errors.New("older version signatures not supported")
 	}
 
+	if err := d.validateCreatorSignatureLength(); err != nil {
+		return err
+	}
+
 	// Check valid operator config signature for each operator.
 	operatorConfigHashDigest, err := digestEIP712(getOperatorEIP712Type(d.Version), d, Operator{})
 	if err != nil {
@@ -255,8 +259,7 @@ func (d Definition) VerifySignatures(eth1 eth1wrap.EthClientRunner) error {
 			return errors.New("empty operator config signature", z.Any("operator_address", o.Address))
 		}
 
-		// Validate signature lengths for v1.11.0+ (Safe multisig support)
-		if err := d.validateSignatureLengths(); err != nil {
+		if err := d.validateOperatorSignatureLengths(o); err != nil {
 			return err
 		}
 
@@ -325,24 +328,30 @@ func (d Definition) VerifySignatures(eth1 eth1wrap.EthClientRunner) error {
 	return nil
 }
 
-// validateSignatureLengths validates that all signatures in v1.11.0+ are multiples of 65 bytes.
-// This ensures compatibility with Safe multisig wallets where signature length = threshold × 65 bytes.
-func (d Definition) validateSignatureLengths() error {
+// validateCreatorSignatureLength validates creator signature length for v1.11.0+.
+func (d Definition) validateCreatorSignatureLength() error {
 	if !isAnyVersion(d.Version, v1_11) {
 		return nil // Skip validation for older versions
 	}
 
-	for i, op := range d.Operators {
-		if err := validateSignatureLength(op.ConfigSignature, "operator config signature"); err != nil {
-			return errors.Wrap(err, "invalid signature length", z.Int("operator_index", i), z.Str("address", op.Address))
-		}
-		if err := validateSignatureLength(op.ENRSignature, "operator enr signature"); err != nil {
-			return errors.Wrap(err, "invalid signature length", z.Int("operator_index", i), z.Str("address", op.Address))
-		}
-	}
-
 	if err := validateSignatureLength(d.Creator.ConfigSignature, "creator config signature"); err != nil {
 		return errors.Wrap(err, "invalid signature length", z.Str("address", d.Creator.Address))
+	}
+
+	return nil
+}
+
+// validateOperatorSignatureLengths validates a single operator's signature lengths for v1.11.0+.
+func (d Definition) validateOperatorSignatureLengths(op Operator) error {
+	if !isAnyVersion(d.Version, v1_11) {
+		return nil // Skip validation for older versions
+	}
+
+	if err := validateSignatureLength(op.ConfigSignature, "operator config signature"); err != nil {
+		return errors.Wrap(err, "invalid signature length", z.Str("address", op.Address))
+	}
+	if err := validateSignatureLength(op.ENRSignature, "operator enr signature"); err != nil {
+		return errors.Wrap(err, "invalid signature length", z.Str("address", op.Address))
 	}
 
 	return nil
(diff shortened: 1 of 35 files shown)


cluster/ssz.go
```diff
@@ -19,6 +19,7 @@ const (
 	sszMaxOperators      = 256
 	sszMaxValidators     = 65536
 	sszMaxDepositAmounts = 256
+	sszMaxSignature      = 384
```
KaloyanTanev: Where is this one coming from?

cluster/definition.go
```diff
@@ -970,6 +1062,44 @@ func unmarshalDefinitionV1x10(data []byte) (def Definition, err error) {
 	}, nil
 }
 
+func unmarshalDefinitionV1x11(data []byte) (def Definition, err error) {
```
KaloyanTanev: I couldn't spot where this function is different than the `unmarshalDefinitionV1x10`? If it's not, can we reuse `unmarshalDefinitionV1x10` and rename it to `unmarshalDefinitionV1x10to11`?

We do that for others like `unmarshalDefinitionV1x5to7`

cluster/definition.go
```diff
@@ -1128,6 +1258,27 @@ type definitionJSONv1x10 struct {
 	DefinitionHash     ethHex                    `json:"definition_hash"`
 }
 
+// definitionJSONv1x11 is the json formatter of Definition for version v1.11 or later.
+type definitionJSONv1x11 struct {
```
KaloyanTanev: Same as above, if nothing changed probably we can rename `definitionJSONv1x10` struct to `definitionJSONv1x10or11` struct?

cluster/definition.go
KaloyanTanev: Not something caused by your PR, but might use the occasion to fix that and make it

```suggestion
type definitionJSONv1x5or6or7 struct {
```

cluster/operator.go
```diff
@@ -21,10 +21,10 @@ type Operator struct {
 	ENR string `config_hash:"-" definition_hash:"1" json:"enr" ssz:"ByteList[1024]"`
 
 	// ConfigSignature is an EIP712 signature of the config_hash using privkey corresponding to operator Ethereum Address.
-	ConfigSignature []byte `json:"config_signature,0xhex" ssz:"Bytes65" config_hash:"-" definition_hash:"2"`
+	ConfigSignature []byte `json:"config_signature,0xhex" ssz:"Bytes65|ByteList[384]" config_hash:"-" definition_hash:"2"`
```
KaloyanTanev: Does this work properly? I'm a bit sceptical of the ambiguity it adds. I'd rather we stick to `ssz:"ByteList[384]"`

cluster/operator.go
```diff
@@ -21,10 +21,10 @@ type Operator struct {
...
 
 	// ENRSignature is a EIP712 signature of the ENR by the Address, authorising the charon node to act on behalf of the operator in the cluster.
-	ENRSignature []byte `json:"enr_signature,0xhex" ssz:"Bytes65" config_hash:"-" definition_hash:"3"`
+	ENRSignature []byte `json:"enr_signature,0xhex" ssz:"Bytes65|ByteList[384]" config_hash:"-" definition_hash:"3"`
```
KaloyanTanev: Same as above.

cluster/ssz.go
```diff
@@ -491,12 +494,160 @@ func hashDefinitionV1x10(d Definition, hh ssz.HashWalker, configOnly bool) error
 	})
 }
 
+// hashDefinitionV1x11 hashes the new definition.
+func hashDefinitionV1x11(d Definition, hh ssz.HashWalker, configOnly bool) error {
```
KaloyanTanev, line 498: Why do we need this whole thing? Can't we do something similar as we did to `hashDefinitionV1x10`, but even shorter, given there are no new fields?

cluster/ssz.go
DiogoSantoss: We need to provide a parameter for `ByteList` length, so this is 65*5 (5 signatures). We can change it to some other higher value.

cluster/ssz.go
DiogoSantoss: Here we need a new function because this new `v1.11` changes the signature encoding from `putBytesN` to `putByteList` and won't work with that composable pattern that other functions use.

cluster/operator.go
DiogoSantoss: I was to differentiate between "regular" signatures and safe signatures but agree that just a single type works fine

cluster/ssz.go
KaloyanTanev: Hmmm yeah, I see. I would say we should align with product/business team then and check what's the highest they expect. My hunch is we can go as high as 10 (for 10 operators in a 7/10 cluster).

cluster/helpers.go
```diff
@@ -127,6 +128,33 @@ func verifySig(expectedAddr string, digest []byte, sig []byte) (bool, error) {
...
+	if eth1 == nil {
+		return false, eoaErr
+	}
+
+	ok, err := eth1.VerifySmartContractBasedSignature(addr, [32]byte(digest), sig)
+	if errors.Is(err, eth1wrap.ErrNoExecutionEngineAddr) {
+		return false, eoaErr
+	}
+
+	return ok, err
```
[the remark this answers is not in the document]
KaloyanTanev: Irrelevant for go 1.26

cluster/helpers.go
```diff
@@ -127,6 +128,33 @@ func verifySig(expectedAddr string, digest []byte, sig []byte) (bool, error) {
...
+	}
+
+	if eth1 == nil {
+		return false, eoaErr
+	}
+
+	ok, err := eth1.VerifySmartContractBasedSignature(addr, [32]byte(digest), sig)
+	if errors.Is(err, eth1wrap.ErrNoExecutionEngineAddr) {
+		return false, eoaErr
+	}
```
pinebit, line 153: When a v1.11 Safe signature is longer than 65 bytes, EOA verification is skipped and this path requires an ERC-1271 client; with nil or the no-endpoint client it returns not verified. Existing callers such as `cmd/createcluster.loadDefinition` and the `LoadClusterLockAndVerify` utility commands still verify with no execution client, so Safe-signed definitions/locks are rejected before any configured endpoint can be used.

cluster/definition.go
```diff
@@ -255,16 +259,15 @@ func (d Definition) VerifySignatures(eth1 eth1wrap.EthClientRunner) error {
...
 		}
 
-		// Check that we have a valid config signature for each operator.
-		if ok, err := verifySig(o.Address, operatorConfigHashDigest, o.ConfigSignature); err != nil {
+		if err := d.validateOperatorSignatureLengths(o); err != nil {
+			return err
+		}
+
+		// Check that we have a valid config signature for each operator (EOA or ERC-1271).
+		if ok, err := verifySigOrERC1271(eth1, o.Address, operatorConfigHashDigest, o.ConfigSignature); err != nil {
```
pinebit, line 267: For v1.5-v1.10 definitions, `validateOperatorSignatureLengths` is a no-op, so this call now sends any non-65-byte signature to ERC-1271 instead of rejecting it. Those versions still hash signatures through the fixed Bytes65 path, so a contract that accepts a short signature can make a pre-v1.11 definition verify even though variable signatures are only represented by the new v1.11 schema.

pinebit (commented):
Added couple of comments observed with GPT

pinebit (approved):

Issue #4351, opened 2026-03-04
Sigagg metric


aadomn:
# 🎯 Problem to be solved
There are currently no dedicated Prometheus metrics to measure the time spent in partial sig verifications and sig aggregations. While negligible for small clusters, the CPU usage scales with the threshold parameter and the number of validators. It would be useful to have such metrics in order to assess if optimization at this level would be worth.

# 🛠️ Proposed solution
Add additional metrics for these specific operations:

- `core_parsigex_set_verification_seconds` measures the time to verify all partial sigs in one peer's P2P message for one duty
- `core_sigagg_slot_aggregation_seconds` measures the time to aggregate all partial sigs for one duty in one slot

PR #4357, opened 2026-03-05
*: add sigagg and parsigex metrics

aadomn:
Add new metrics to benchmark signature aggregations and partial signature verifications.
Address issue #4351

category: feature
ticket: none

### DIFF
--- a/core/sigagg/sigagg.go
+++ b/core/sigagg/sigagg.go
@@ -29,7 +29,7 @@ var slotAggregationDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
 	Subsystem: "sigagg",
 	Name:      "slot_aggregation_seconds",
 	Help:      "Total duration to aggregate all validators for a duty in a slot, in seconds",
-	Buckets:   []float64{.01, .05, .1, .25, .5, 1, 2.5, 5, 10, 25},
+	Buckets:   []float64{.001, .005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5},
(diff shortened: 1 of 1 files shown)


pinebit:
You need to run this to update `docs/metrics.md`

```
//go:generate go test . -run=TestMetricReference -update-markdown

func TestMetricReference(t *testing.T) {
...
```

pinebit:
To test pre-commit before committing/pushing, run `pre-commit` tool locally inside the repo (your changes must be staged).

core/parsigex/metrics.go
pinebit, line 1: I think it should not be that hard. The existing TestMetricsReference scans the entire codebase and finds all metrics automatically. You don't need metrics.go/metrics_test.go

pinebit: Everything works fine if you just delete the four files:
```
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    core/parsigex/metrics.go
        deleted:    core/parsigex/metrics_test.go
        deleted:    core/sigagg/metrics.go
        deleted:    core/sigagg/metrics_test.go

no changes added to commit (use "git add" and/or "git commit -a")
```

aadomn:
> Both metrics have different buckets. If this is on purpose, then Approved.

`sigagg` was initially using larger bucket values as I (wrongly) assumed it would take more time but it actually makes sense to use the same values as for `parsigex`.
Thanks for spotting this.

Issue #4211, opened 2026-01-08
VC missing validators false positive


KaloyanTanev:
# 🐞 Bug Report

### Description

Charon health check false positively triggers "VC missing validators". Usually happens when not all keys in a cluster are active.

KaloyanTanev:
We can as well change the metric to be useful for the edit commands - currently we check if the amount of keys matches, however, we can also check if they are the correct keys. Once operators start rotating keys, they might end up with the old set of keys in the VC, or both sets, or some mix of those.

It would be nice if this command not only checks if the amount of on-chain active keys are queried by the VC, but also if the partial keys the VC ask for match a full pubkey in the lock file.

PR #4359, opened 2026-03-05
*: remove VC Missing validators ready check

KaloyanTanev:
The check was flaky and there is no consistent way of checking which partial keys the VC is querying.

category: bug
ticket: #4211


pinebit (approved):
LGTM except for indices (communicated internally)

PR #4378, opened 2026-03-10
Fix race condition

OisinKyne:
category: bug
feature_flag: ?

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -459,8 +459,22 @@ func compare[I any, V comparable, C any](ctx context.Context, d Definition[I, V,
 			return inputValueSource, nil
 		case inputValueSource = <-compareValue:
 		case <-timerChan:
-			log.Warn(ctx, "", errors.New("timeout waiting for local data, used for comparing with leader's proposed data"))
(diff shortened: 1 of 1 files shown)


core/qbft/qbft.go
```diff
@@ -390,9 +435,53 @@ func Run[I any, V comparable](ctx context.Context, d Definition[I, V], t Transpo
...
+		case <-timerChan:
+			// When the eager timer fires at an absolute deadline, d.Compare may have
+			// already completed (e.g. instantly when compareAttestations is disabled).
+			// Go's select picks randomly among ready channels, so check compareErr
+			// before declaring a timeout to avoid spurious round changes.
+			select {
+			case err := <-compareErr:
+				if err != nil {
+					log.Warn(ctx, errCompare.Error(), err)
+					return inputValueSource, errCompare
```
pinebit, line 470: maybe to preserve errors chain: `return inputValueSource, errors.Wrap(err, errCompare.Error())`.

pinebit:
I suggest to convert this into a task, then we will plan for this test/fix in another planning meeting.

Issue #4293, opened 2026-02-06
Mutable `fee-recipient-addresses`


OisinKyne:
# 🎯 Problem to be solved

We want to modify fee recipient addresses of unallocated keys. Ideally we don't want to restart our cluster or experience any down time. 

# 🛠️ Proposed solution

Describe the solution to be implemented. TBD on an internal doc. 

(Ideas:
- Persist updates on disk in `.charon` if they supercede the lockfile, just load them in and start using them. 
- Submit them to a running charon's vAPI (from another `charon fee-recipient update` command, making our first client/daemon type UX
- Support some sort of distribution of this ask? the ux of an edit command?

# 🧪 Tests
- [ ] Tested by new automated unit/integration/smoke tests
- [ ] Manually tested on core team/canary/test clusters
- [ ] Manually tested on local compose simnet

PR #4407, opened 2026-03-19
*: feerecipient auto fetching

pinebit:
Add automatic background fetching of builder registrations from the Obol API to the `builderRegistrationService`. Previously operators had to manually run `charon feerecipient fetch` to pull fully signed registrations. Now the service periodically fetches from the API and merges results with file overrides, preferring the entry with the highest timestamp per validator.

Fetching schedule:
- Immediate fetch on charon start
- Every 1h if any validators have incomplete (partially signed) registrations
- Every 24h if all registrations are fully signed
- File override changes also trigger an immediate API fetch

The `Run` function now works even when no overrides file path is configured, as long as the Obol API address is set.

category: feature
ticket: #4293

### DIFF
--- a/app/builderregistration.go
+++ b/app/builderregistration.go
@@ -300,8 +300,11 @@ func (s *builderRegistrationService) Run(ctx context.Context) {
 			log.Warn(ctx, "File watcher error for builder registration overrides", err)
 
 		case <-fetchCh:
-			hasIncomplete := s.fetchFromAPI(ctx)
-			if hasIncomplete {
+			hasIncomplete, err := s.fetchFromAPI(ctx)
+			if err != nil {
+				log.Warn(ctx, "Builder registration API fetch failed", err)
+				fetchTimer.Reset(fetchIntervalIncomplete)
+			} else if hasIncomplete {
 				fetchTimer.Reset(fetchIntervalIncomplete)
 			} else {
 				fetchTimer.Reset(fetchIntervalComplete)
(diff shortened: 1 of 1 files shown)


app/builderregistration.go
```diff
@@ -88,26 +210,48 @@ func (s *builderRegistrationService) FeeRecipient(pubkey core.PubKey) string {
 	return s.feeRecipients[pubkey]
 }
 
-// Run watches the overrides file for changes and reloads when modified.
-// It blocks until ctx is cancelled.
+// Run watches the overrides file for changes, periodically fetches from the Obol API,
+// and reloads when either source is updated. It blocks until ctx is cancelled.
 func (s *builderRegistrationService) Run(ctx context.Context) {
-	if s.path == "" {
+	if s.path == "" && s.obolClient == nil {
```
KaloyanTanev, line 216: This will practically trigger only if the user runs Charon with
```
--overrides-file=""
--publish-address=""
```
, right?

What are the odds of having that? And if we want this to be disabled, shouldn't we have some better functionality rather than user having to specify the above 2 in such way?

This also relates to multiple decision trees further down in this function.

app/builderregistration.go
```diff
@@ -129,40 +274,107 @@ func (s *builderRegistrationService) Run(ctx context.Context) {
...
-		s.registrations = s.baseRegistrations
-		s.feeRecipients = feeRecipients
+	return nil
+}
 
-		return nil
+// fetchFromAPI calls the Obol API, processes the response, stores API overrides,
+// and calls recompute. Returns true if any validators have incomplete registrations
+// or if an error occurred (to retry sooner).
+func (s *builderRegistrationService) fetchFromAPI(ctx context.Context) (hasIncomplete bool) {
```
KaloyanTanev: Can we rename `hasIncomplete` to `retryEarly`?
Or even make it an error and retry on error?
Or change the function signature to `(hasIncomplete bool, err error)`?

Right now we are returning `hasIncomplete = true` on error and it feels a bit incorrect.

app/builderregistration.go
pinebit: Right, if both overrides and remote api are disabled, then the service simply provides cluster-lock data (builder registrations) as the single source of truth.

app/builderregistration.go
pinebit: Good point, fixed.

PR #4498, opened 2026-04-27
fix: prevent duty loss and fix misleading comment in deadliner

enlightened88:
Remove the default branch from the timer case in deadliner.run to ensure deadlined duties are never silently dropped when the output channel is full. Also fix an incorrect comment in getCurrDuty that said latest instead of earliest


KaloyanTanev:
@enlightened88 did you experience any bug or degraded performance because of the current logic? Currently the buffer is 10, which is more than enough to support all duties before they time out. For gloas we will likely increase it slightly though, because of PTC.

KaloyanTanev:
@enlightened88 I will be closing this PR by tomorrow if no update.

PR #4553, opened 2026-06-08
core/validatorapi: fix dynamic path proxy metric

KaloyanTanev:
Prometheus metrics were growing their cardinality uncontrollably on proxied requests that had slot/epoch/pubkey/val_idx in them.

category: bug
ticket: none

### DIFF
--- a/core/validatorapi/metrics.go
+++ b/core/validatorapi/metrics.go
@@ -98,11 +98,7 @@ func isNumeric(s string) bool {
 		return false
 	}
 
-	for _, r := range s {
-		if r < '0' || r > '9' {
-			return false
-		}
-	}
-
-	return true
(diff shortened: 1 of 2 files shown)


core/validatorapi/metrics.go
```diff
@@ -64,9 +64,45 @@ func observeAPILatency(endpoint string) func() {
...
+	return strings.Join(segments, "_")
+}
+
+// isNumeric returns true if s is non-empty and contains only ASCII digits.
+func isNumeric(s string) bool {
+	if s == "" {
+		return false
+	}
+
+	for _, r := range s {
```
pinebit: to keep it simple
```
return !strings.ContainsFunc(s, func(r rune) bool {
	return r < '0' || r > '9'
})
```

pinebit (approved):

PR #4576, opened 2026-07-03
*: allow incomplete validator_keys store when parsing

KaloyanTanev:
Previously Charon failed commands requiring the partial keys when only a subset of the keys is available in `validator_keys` directory:

```
13:54:58.491 ERRO cmd        Application failed to start: match local validator key shares with their counterparty in cluster lock: public key share from provided private key share not found in provided lock
        eth2util/keystore/keystore.go:288 .KeysharesToValidatorPubkey
        cmd/exit_sign.go:126 .runSignPartialExit
        cmd/exit_sign.go:44 .func1
        cmd/cmd.go:111 .func1
        main.go:19 .main
```

This is actually incorrect as the said key did exist, however, not the full set of keys in the lock were available in the directory. Scenarios in which allowing this is useful:
1. Operator has Charon in one machine and VC in separate machine;
2. Operator does not hold the validator keys in the Charon machine;
3. Operator wants to do some action with Charon's CLI (e.g.: sign exit, sign new deposit, etc.);
4. Operator moves part of the keys for which the operation will be done inside the Charon machine;
5. Operator runs the command for the said keys.

category: bug
ticket: none

### DIFF
--- a/eth2util/keystore/keystore.go
+++ b/eth2util/keystore/keystore.go
@@ -242,13 +242,13 @@ func checkDir(dir string) error {
 	return nil
 }
 
-// KeysharesToValidatorPubkey maps each share in cl to the associated validator private key.
-// It returns an error if a keyshare does not appear in cl, or if there's a validator public key associated to no
-// keyshare.
+// KeysharesToValidatorPubkey maps each provided private key share to the associated validator
+// public key in the cluster lock. It returns an error if a provided private key share does not belong to any validator
+// in the lock.
 func KeysharesToValidatorPubkey(lock cluster.Lock, shares []tbls.PrivateKey) (ValidatorShares, error) {
 	ret := make(map[core.PubKey]IndexedKeyShare)
 
-	var pubShares []tbls.PublicKey
+	pubShares := make([]tbls.PublicKey, 0, len(shares))
 
 	for _, share := range shares {
 		ps, err := tbls.SecretToPublicKey(share)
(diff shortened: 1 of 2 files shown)


pinebit (approved):

pinebit:
A few findings by Fable for your to consider:
```
  1. cmd/exit_sign.go:220 — charon exit sign --all now silently signs exits for only the validators whose keystores are on disk. An operator with a partially restored validator_keys dir (e.g.  keystore-0..4 of 10 — contiguous, so SequencedKeys passes) previously got a hard error; now the command succeeds, posts partial exits for 5 validators, and nothing in the output mentions the 5 skipped lock validators — the active_validators/inactive_validators log counts are both computed from the on-disk subset. If enough operators share the gap, those exits never reach threshold. The --all path should compare len(shares) against len(cl.Validators) and at least warn.

  2. eth2util/keystore/keystore.go:285 — the break after the first match makes a second share of the same validator fail with a factually wrong error. With shares = [sA1, sA2] both belonging to validator A (consolidated keystore dirs, or a duplicated keystore file), the inner loop matches sA1 and breaks, so matched for sA2 stays false and the function returns "public key share from provided private key share not found in provided lock" — even though it is in the lock. This input was accepted on main (the extra share was silently ignored), so it's a behavioral regression, and the message misdiagnoses the cause.

  3. eth2util/keystore/load.go:47 — the PR's motivating scenario still fails one layer upstream. All three callers reach the relaxed function via SequencedKeys, which rejects any keystore set whose file indexes aren't exactly 0..N-1. Copying just keystore-5.json (the PR body's use case) dies with "out of sequence keystore index" before keysharesToValidatorPubkey runs — only contiguous-prefix subsets, or renamed files, benefit. The fix likely needs to reach the load layer too, since matching is by derived pubkey and the filename index carries no meaning for these commands.
```

PR #4577, opened 2026-07-03
cmd: improve split keys behavior

pinebit:
Improve `charon create cluster --split-existing-keys` behavior:

- Load keystores in strict file index order when distinct fee recipient or withdrawal addresses are provided (via flags or a definition file), so `keystore-N.json` is paired with the Nth per-validator addresses. Previously only multiple `--withdrawal-addresses` triggered ordered loading; multiple fee recipients or a definition file with distinct addresses paired keys with addresses nondeterministically.
- Correct the `--split-existing-keys` help text: deposit data files are re-created in split mode using the provided withdrawal addresses, and must not be submitted for validators that are already active. The flag previously claimed deposit data is not re-created.
- Use American English spelling in the split keys warning output ("finalized").

Documentation is updated accordingly in ObolNetwork/obol-gitbook#164.

category: bug
ticket: none

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cmd/createcluster.go
```diff
@@ -198,7 +198,14 @@ func runCreateCluster(ctx context.Context, w io.Writer, conf clusterConfig) erro
...
 	if conf.SplitKeys {
-		useSequencedKeys := len(conf.WithdrawalAddrs) > 1
+		// Load keys in strict file index order when per-validator addresses may differ,
+		// so that keystore-N.json is paired with the Nth withdrawal and fee recipient address.
+		useSequencedKeys := len(conf.WithdrawalAddrs) > 1 || len(conf.FeeRecipientAddrs) > 1
+		if conf.DefFile != "" {
+			// A definition file replicates a single provided address for each validator,
+			// so only distinct addresses require strict key file ordering.
+			useSequencedKeys = hasDistinctAddrs(def.WithdrawalAddresses()) || hasDistinctAddrs(def.FeeRecipientAddresses())
+		}
```
KaloyanTanev, line 208: CMIIW, but this `useSequencedKeys` is used for optimisation further down the line? And now we are making the work here, setting this flag to true/false, so we skip doing the work down the line? It seems to me it's just moving the work from one place to another when using this `hasDistinctAddrs`.

KaloyanTanev (approved):
Approved, but honestly the code changes didn't seem

cmd/createcluster.go
pinebit: Not quite: `useSequencedKeys` isn't an optimization flag, it selects between two loading modes with different user-facing contracts, and `hasDistinctAddrs` is the condition for which contract we must impose:

- `true`: `LoadFilesUnordered().SequencedKeys()`: requires strict keystore-N.json naming (contiguous indices 0..N-1, no gaps or duplicates, password file adjacent) and returns keys in file-index order.
- `false`: `LoadFilesRecursively().Keys()`: the convenience mode documented since v1.8.0 — recursive directory search, arbitrary keystore file names, best-effort password matching - but the resulting key order is nondeterministic (concurrent decryption, completion order).

PR #4581, opened 2026-07-04
*: fix metrics high cardinality risks

pinebit:
Fixes high-cardinality risks found while reviewing all Prometheus metrics against the code that populates them:

- `cmd/relay`: delete all `relay_p2p_*` series of a peer when it disconnects. The relay accepts connections from arbitrary peers and clusters (open gater) and never deleted series, so a long-running relay accumulated `(peer, peer_cluster)` series indefinitely.
- `app/health`: the high-cardinality check compared the number of label keys per series (~10) against `100 * numValidators`, so it never fired. It now compares the series count per metric family.
- `core/validatorapi`: truncate the untrusted User-Agent header to 128 bytes and sanitize it to valid UTF-8 before using it as a label value.
- `app/promauto`: `ResetGaugeVec` no longer panics on label values containing `|` and `Reset` now matches exact leading label values instead of substrings anywhere, which could delete series of an unrelated peer whose labels contained the given value as a substring.

category: bug
ticket: none

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cmd/relay/p2p.go
```diff
@@ -163,12 +163,9 @@ func monitorConnections(ctx context.Context, p2pNode host.Host, bwTuples <-chan
...
 			for p, state := range peers {
 				if state.Active == 0 {
-					// No active connections, remove peer from state.
+					// No active connections, remove peer from state and delete its metric series.
 					delete(peers, p)
-
-					if state.ClusterHash != "" {
-						activeConnsCounter.WithLabelValues(state.Name, state.ClusterHash).Set(0)
-					}
+					deletePeerMetrics(state.Name)
```
KaloyanTanev, line 168: Wouldn't that delete historical metrics for a peer if it goes offline for a minute?

core/scheduler/scheduler.go
```diff
@@ -454,6 +458,9 @@ func (s *Scheduler) resolveDuties(ctx context.Context, slot core.Slot) error {
 		return err
 	}
 
+	// Delete metrics of validators no longer reported by the beacon node.
+	s.metricSweeper()
```
KaloyanTanev: I don't understand why that is needed? Do we protect ourselves against a bug in the very core of Ethereum consensus layer? Similarly as the previous question, would it delete historical metrics?

cmd/relay/p2p.go
pinebit: Historical data is safe, deleting a series from the client process only removes it from the live `/metrics` output. Everything Prometheus already scraped stays in its TSDB, so dashboards querying past ranges are unaffected.

core/scheduler/scheduler.go
pinebit: On historical concern - same answer as above, but here it's even better: these are gauges, not counters.

As for why, I agree this is a minor benefit. I decided just to revert this change here.

KaloyanTanev (approved):

PR #4610, opened 2026-07-28
core/deadline: increase sync message deadline

KaloyanTanev:
Increase sync message deadline. This is done as after the deadline hits, the aggregated sync message signatures are cleaned from the DB (previously at 8s + 1s margin). However, the sync contribution does need them as part of the data, so sometimes, under specific network conditions or VC types, the contribution might take a bit more than 9s and the sync message is missing by then.

category: bug
ticket: none


pinebit (approved):
Assuming we tested this, approved

PR #4620, opened 2026-08-05
cluster/*: reject lock with val count mismatch

KaloyanTanev:
Reject lock that has mismatch of number of validators with the definition. Essentially happens never, but a safety precaution.

category: misc
ticket: none


OisinKyne (approved):
I assume this doesn't fire when `--no-verify` is present

PR #4621, opened 2026-08-05
*: remove legacy un-versioned attestations

KaloyanTanev:
Pre-electra we have used attestation objects, but post-electra we have moved to versioned attestations. Remove the unused logic now.

category: misc
ticket: none

### DIFF
--- a/core/eth2signeddata.go
+++ b/core/eth2signeddata.go
@@ -16,7 +16,6 @@ import (
 
 var (
 	_ Eth2SignedData = VersionedSignedProposal{}
-	_ Eth2SignedData = Attestation{}
 	_ Eth2SignedData = VersionedAttestation{}
 	_ Eth2SignedData = SignedVoluntaryExit{}
 	_ Eth2SignedData = VersionedSignedValidatorRegistration{}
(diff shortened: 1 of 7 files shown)


OisinKyne (approved):
Some part of me worries i'll regret us removing the ability for charon to run on older forks, but i don't think this is the first occasion of us doing that so ticking approve

KaloyanTanev:
@OisinKyne This one doesn't prevent us running older forks! The difference is that previously we have used:

```go
// Attestation is the Ethereum 2 attestation structure.
type Attestation struct {
	AggregationBits bitfield.Bitlist `dynssz-max:"MAX_VALIDATORS_PER_COMMITTEE" ssz-max:"2048"`
	Data            *AttestationData
	Signature       BLSSignature `ssz-size:"96"`
}
```

While now we do use

```go
// VersionedAttestation contains a versioned attestation.
type VersionedAttestation struct {
	Version        DataVersion
	ValidatorIndex *phase0.ValidatorIndex
	Phase0         *phase0.Attestation
	Altair         *phase0.Attestation
	Bellatrix      *phase0.Attestation
	Capella        *phase0.Attestation
	Deneb          *phase0.Attestation
	Electra        *electra.Attestation
	Fulu           *electra.Attestation
}
```

Which has either the phase0 attestation from the first one, or the electra one.

PR #4628, opened 2026-08-06
core/qbft: preserve local value on comparison failure

KaloyanTanev:
Fixes the compare flow losing the already-fetched local value when the comparator reads the value and errors back-to-back: both buffered channel sends can complete before the select polls, which then picks the error case at random and drops the value. Subsequent rounds then block forever on the already-consumed input value source channel, so the node never prepares again, as seen in `TestChainSplit` flakes reaching max round. Extracts the select loop into `awaitCompare` and drains the pending value on the error and timeout paths.

category: bug
ticket: none

### DIFF
--- a/core/qbft/qbft.go
+++ b/core/qbft/qbft.go
@@ -490,9 +490,31 @@ func compare[I any, V comparable, C any](ctx context.Context, d Definition[I, V,
 	// If comparison or any other unexpected error occurs, the error is returned on compareErr channel.
 	go d.Compare(ctxCompare, msg, inputValueSourceCh, inputValueSource, compareErr, compareValue)
 
+	return awaitCompare(ctx, compareErr, compareValue, timerChan, inputValueSource)
+}
+
+// awaitCompare waits for the comparator's verdict and returns the latest local value.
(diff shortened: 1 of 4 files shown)


core/qbft/qbft.go
```diff
@@ -488,9 +488,31 @@ func compare[I any, V comparable, C any](ctx context.Context, d Definition[I, V,
 	// If comparison or any other unexpected error occurs, the error is returned on compareErr channel.
 	go d.Compare(ctxCompare, msg, inputValueSourceCh, inputValueSource, compareErr, compareValue)
 
+	return awaitCompare(ctx, compareErr, compareValue, timerChan, inputValueSource)
+}
+
+// awaitCompare waits for the comparator's verdict and returns the latest local value.
+func awaitCompare[C any](ctx context.Context, compareErr <-chan error, compareValue <-chan C, timerChan <-chan time.Time, inputValueSource C) (C, error) {
```
pinebit, line 497: Love generics

pinebit (approved):

core/qbft/qbft.go
KaloyanTanev: Controversial topic in golang... haha

PR #4642, opened 2026-08-12
p2p: bound relay address query responses

pinebit:
Relay address resolution read the HTTP response body with `io.ReadAll` and no size limit, using a zero-value `http.Client` with no timeout, in a loop that runs for the process lifetime. A configured relay returning an endless response body could therefore grow the heap until the node was OOM killed. Reported via the bug bounty program.

Limits the response to 64KB, which is well above a valid ENR string or multiaddr array, and rejects anything larger instead of parsing a truncated body. Adds a 10s per-attempt client timeout, since the request was previously bound only to the long-lived application context. Closes the response body on the non-2xx retry path, where it was leaked along with its connection on every retry.

Oversized responses warn and retry, matching how the existing non-2xx, JSON and multiaddr failures are handled; returning an error instead would make `resolveRelay` abandon the relay for the rest of the process lifetime.

Adds `p2p/bootnode_internal_test.go`, which had no coverage. `TestQueryRelayAddrsBoundsResponse` counts the bytes a streaming server manages to write and asserts a hard bound, so it exercises the size limit independently of the timeout. With the limit removed but the timeout in place, that server writes 2.8GB within the 10s window, so the timeout alone is not sufficient mitigation. The oversized case uses valid JSON padded past the limit so the size check rejects it rather than the parser, and a case at exactly the limit covers the boundary.

category: bug
ticket: none

### DIFF
--- a/p2p/bootnode.go
+++ b/p2p/bootnode.go
@@ -140,6 +140,15 @@ func resolveRelay(ctx context.Context, rawURL, lockHashHex, uuid string, callbac
 	}
 }
 
+const (
+	// maxRelayResponseSize is the maximum accepted relay query response size. Valid responses are
+	// either an ENR string or a small json array of multiaddrs, so this is a generous upper bound.
+	maxRelayResponseSize = 64 << 10 // 64KB
+
+	// relayQueryTimeout bounds a single relay query attempt, including reading the response body.
+	relayQueryTimeout = 10 * time.Second
+)
+
 // queryRelayAddrs returns the relay multiaddrs via a http GET query to the url.
 //
 // This supports resolving relay addrs from known http URLs which is handy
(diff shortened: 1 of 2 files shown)


KaloyanTanev:
Do we have metrics from the size of the relay data that was exchanged? If so, is it the case that those limits would suffice? With both QUIC and TCP. I generally think this PR has higher ROI than the other transport-related PRs, as it is relay related.

pinebit:
> Do we have metrics from the size of the relay data that was exchanged? If so, is it the case that those limits would suffice? With both QUIC and TCP. I generally think this PR has higher ROI than the other transport-related PRs, as it is relay related.

No, we don't have historical metrics.

I queried the production relays directly to get a real number:
- https://0.relay.obol.tech = 183 bytes
- https://1.relay.obol.tech = 179 bytes
Each returns exactly two multiaddrs: one TCP, one QUIC.

The 64KB limit is ~350× the actual production response (~180 bytes) and would still fit ~600 multiaddrs, so it's safe for both transports and any plausible relay configuration. Note that the limit applies only to the HTTP address-discovery query; the relayed p2p traffic itself (QUIC or TCP) is unaffected.

KaloyanTanev (approved):

Issue #4692, opened 2026-09-09
app: distribute cluster proposer config (fee recipient / gas limit) to validator clients


KaloyanTanev:
# 🎯 Problem to be solved

Charon currently **ignores** the VC's proposer config: `prepare_beacon_proposer` is a no-op (`core/validatorapi/router.go`, `submitProposalPreparations`) and `register_validator` is ignored (`core/validatorapi/validatorapi.go`, `SubmitValidatorRegistrations`). Charon injects the lock's fee recipient itself. As a result the VC-side fee recipient / gas limit are unmanaged and irrelevant today — CDVN VCs set none, and Teku even hardcodes `--validators-proposer-default-fee-recipient 0x0000…0000`.

Post-gloas the VC **signs** `fee_recipient` and `target_gas_limit` into `SignedProposerPreferences` (#4691), so its local config becomes load-bearing and must match the lock cluster-wide or the partial signatures won't aggregate. This issue: give Charon a way to push the cluster-canonical fee recipient + gas limit (lock `FeeRecipientAddresses` / `TargetGasLimit`) into the VCs.

This has **independent, pre-gloas value** — it makes VCs hold the correct fee recipient rather than relying on Charon's override — so it can land before the gloas duty.

Sub-task of the gloas epic #4324.

# 🛠️ Proposed solution

## Findings

- **No standard config file exists.** The ecosystem is standardizing an *API* instead — keymanager-APIs#87 (`/eth/v1/validator/config`, closed/contentious, not ratified), explicitly to kill the "different config dialect per client" burden it names for dappnode/eth-docker.
- **Per-client config surfaces:**
  - **Prysm / Teku / Lodestar** — one central `proposer_config` JSON: `--proposer-settings-file` / `--validators-proposer-config` / `--proposerSettingsFile`.
  - **Nimbus** — per-validator files `<validators-dir>/<pubkey>/suggested_fee_recipient.hex` + `suggested_gas_limit.json`.
  - **Lighthouse** (not in CDVN) — `validator_definitions.yml` (keystore-entangled; standalone fee-recipient file removed) or keymanager API.
- **Dynamic reload** (matters for coordinated `gas_limit` bumps, which are applied staggered across operators):
  - **Keymanager API is dynamic on all clients** (and the client persists the change to its own config file).
[...]

PR #4700, opened 2026-09-15
app: generate proposer config file for validator clients

KaloyanTanev:
Generates a `proposer-config.json` file in a `vc-config` directory next to the cluster lock on startup: created if missing, never modified afterwards. The file uses a charon-defined canonical schema (`version: 1`, modelled on the proposer settings format minus its pre-gloas legacy fields): `default_config` carries the per-field majority `fee_recipient` and `gas_limit` across this node's validator shares (from charon's effective builder registrations, including operator overrides, falling back to the lock), and `proposer_config` only lists validators diverging from it, carrying only the diverging fields. Per validator client adapters (the CDVN wrapper scripts) render each client's native format from it.

From the gloas fork, validator clients sign the fee recipient and gas limit into `SignedProposerPreferences`, so VCs must be configured with the cluster values instead of relying on charon overriding them (see #4692 for the full design). Generation is best-effort: failures are logged and never prevent duties. Also adds a `GasLimit` accessor to `BuilderRegistrationService` and rejects builder registration overrides with a zero gas limit.

category: feature
ticket: #4692


OisinKyne:
what's to be gained in the `vc-config` directory? if we just are making one file imo leaving .charon/ as the output minimises the chances it gets lost etc. i.e. `.charon/proposer_config.json` rather than anywhere else. (or i think in pluto have we been trying to make `distributed_validator` happen?

OisinKyne (approved):

Issue #4324, opened 2026-02-16
Gloas hardfork


KaloyanTanev:
# 🎯 Problem to be solved

Prepare Charon for gloas hardfork

# 🛠️ Proposed solution

- [ ] Fetch gloas hardfork and based on it:
  - [ ] Change attestation timing
  - [ ] Add PTC duty
- [ ] Add changes from devnet 0 [spec](https://github.com/ethereum/consensus-specs/tree/v1.7.0-alpha.2) and [Beacon API](https://github.com/ethereum/beacon-APIs/commits/v5.0.0-alpha.0/)

PR #4704, opened 2026-09-17
app: generate builder config for validator clients

KaloyanTanev:
Adds gloas builder configuration to the generated VC proposer config file. New `charon run` flags (`--builder-urls`, `--builder-min-bid`, `--builder-boost-factor`, `--builder-max-execution-payment`) emit a `builder` section on the file's `default_config`: `min_bid`, `builder_boost_factor` and `max_execution_payment` as defaults plus one `builders[]` entry per URL, with the schema reserving optional per-entry and per-validator overrides for future inputs. Since the values must be identical cluster-wide (they feed signed builder duties post-gloas), each node exposes its config via the `app_builder_config_*` metrics and gossips a hash of it via peerinfo, warning and setting `app_peerinfo_builder_config_mismatch{peer}` when a peer diverges.

category: feature
ticket: #4324


OisinKyne:
This is probably where i'd default in an obol url as mentioned on slack. 

Rn with this, #4700 and #4707 we use the words `proposer` `builder` and `feerecipient`, i think we should consider if we can pull these closer together in convention (or maybe they do need to be different, open to seeing what the vcs are using for these different pieces). 

Posting my other idea comments on those prs now, and then i think we get on a call and ensure i'm fully on the same page with the plan before merging

OisinKyne (approved):

PR #4707, opened 2026-09-21
cmd: forbid feerecipient fetch from the gloas fork

KaloyanTanev:
Rejects `charon alpha feerecipient fetch` on networks where the gloas hard fork is active: from the fork, builder registrations are superseded by signed proposer preferences and the fetched overrides no longer reach the validator client — a new command modifying the generated proposer configuration file will replace this flow. Adds `GloasHardForkTimestamp` to `eth2util.Network` (zero = not scheduled, so nothing changes until networks schedule the fork) and a `GloasActive` helper.

category: feature
ticket: #4692


OisinKyne:
lets plan out the ux of this. idk what naming convention we replace it with, `proposer` instead of `feerecipient`? we could just update this command (it is in alpha) rather than deprecating it for something else

OisinKyne (approved):
might be a little overkill in case idk gloas gets delayed or something but its only for a niche feature so probably okay

Issue #4691, opened 2026-09-09
core: proposer preferences signed duty (gloas ePBS)


KaloyanTanev:
# 🎯 Problem to be solved

From the Gloas fork onwards, `prepare_beacon_proposer` and `register_validator` are **superseded** by validator-signed `SignedProposerPreferences`, gossiped on the `proposer_preferences` topic, as the source of proposer `fee_recipient` and `target_gas_limit`. For a distributed validator this becomes a new **ahead-of-time signed duty**: aggregate the VCs' partial signatures and gossip the reconstructed signature *before* the proposal slot.

Sub-task of the gloas epic #4324 (Block production / ePBS). Gated upstream on attestantio/go-eth2-client#316 (submit proposer preferences).

# 🛠️ Proposed solution

Treat proposer preferences like **sync committee messages**: ungated (no consensus round), accept the VC's partial signature, aggregate on threshold, broadcast — returning `2XX` to the VC on submission without blocking for threshold.

## Key design facts (spec + Lighthouse impl)

- **Signed message**: `ProposerPreferences{dependent_root, proposal_slot, validator_index, fee_recipient, target_gas_limit}`, domain `DOMAIN_PROPOSER_PREFERENCES` at `proposal_epoch`. There is **no wall-clock timestamp** — unlike `ValidatorRegistrationV1.timestamp` (the field that forces us to pre-sign registrations and ignore VC ones), so the VCs' partial signatures **are aggregatable**.
- **Only path is aggregating VC partials at runtime.** Charon can't self-sign (no key material in the running process) and can't pre-sign offline (`proposal_slot` + `dependent_root` are live consensus values).
- **`dependent_root` = shuffling anchor**, `compute_shuffling_dependent_slot(E) = start_slot(E-1)-1` = last block of epoch `E-2` (`MIN_SEED_LOOKAHEAD=1`). It points ~1–2 epochs into the *past*, so it's stable and agreed cluster-wide except on a pathological deep reorg. Charon **serves** it via the v2 proposer-duties response, so the cluster converges on identical messages by construction.
- **Ahead-of-time, not at-trigger.** Broadcast window is the current epoch up to `MIN_SEED_LOOKAHEAD` epochs ahead; gossip validation `[IGNORE] The proposal slot has not started yet`. Missing it means the validator "will not accept any trustless bids for that slot" → self-build → lost MEV.
[...]

Issue #332, opened 2022-03-31
Add support for external/advertised p2p addresses


corverroos:
## Problem to be solved

Charon currently advertises the same IP it binds to. That means that it cannot bind to any other IP than publicly accessible IPs. This is not great and makes config harder. Example: in docker-compose, we cannot just bind to 0.0.0.0 and use the container/service name as advertised host, we have to bind to the container IP which is only known at runtime. 

## Proposed solution

Add support for external/advertised ips and hostnames. See prysm [cmd/flags.go](https://github.com/prysmaticlabs/prysm/blob/develop/cmd/flags.go#L128-L139) and [p2p/discovery.go](https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/p2p/discovery.go#L173-L194) for example.

PR #4708, opened 2026-09-21
core/validatorapi: serve v2 proposer duties dependent root

KaloyanTanev:
The existing `/eth/v2/validator/duties/proposer/{epoch}` route was wired to the v1 handler, serving v1's `E-1` `dependent_root` on the v2 path — from gloas, VCs copy this root verbatim into `SignedProposerPreferences`, so serving the wrong anchor would make the whole cluster sign gossip-invalid preferences. Adds an `eth2wrap.ProposerDutiesV2Provider` that fetches the beacon node's v2 endpoint directly (go-eth2-client doesn't support it yet, `TODO` to swap once attestantio/go-eth2-client#332 merges) and a validatorapi component method applying the usual pubshare swap, with the `E-2` root passing through untouched.

category: bug
ticket: #4691


OisinKyne (approved):
Three non-blocking nits

1. new(http.Client).Do(req) in httpwrap.go:212 â a fresh client per call means no connection pooling, so every VC duties poll pays a new TCP+TLS handshake to the BN. A package-level shared http.Client is a one-line fix. Given the whole thing is a stopgap until go-eth2-client #332 merges (TODO is properly marked), this is tolerable, but it's the cheapest improvement available.
2. No caching for v2 â v1 duties go through the duties cache; v2 hits the BN on every VC request. Fine as a temporary state, but if gloas VCs poll aggressively across many validators, this is where you'd feel it. Worth remembering when the go-eth2-client swap happens.
3. The license-header commit is a band-aid. The check-licence-header pre-commit hook demands the header, but neither the genwrap template (genwrap.go's tpl still starts at package eth2wrap) nor the mockery config emits it â so the next go generate/mockery run strips the headers again and someone repeats this commit. Suggest fixing at the source: add the header line to the genwrap template and a boilerplate file to the mockery config (or extend the hook's exclude beyond .pb.go to cover generated files).

KaloyanTanev:
Addressed (2) only: both `TODO(gloas)`s now note the #332 swap should also route v2 through the duties cache. (1) and (3) intentionally not taken — `new(http.Client).Do` matches the existing `app/obolapi` idiom, and the headers stay as-committed.

